PR #8, opened 2024-03-06
Feat : Trace_transaction RPC call implementation

Tbelleng:
# Pull Request type

- Feature ✅

## What is the new behavior?

Add trace_transaction RPC call and modify trace_block_transactions ✅
In this new implementation, we are going to separate transactions from a block into 2 vectors,
the first one contains transactions that we dont want to trace but we still need to re-execute them in order to
get the correct state update. 

The second vector contains transactions that we want to re-execute and trace.

⚠️ So for trace_block_transactions method, we are gonna re-execute and trace all transactions within a block, but concerning trace_transactions method, we will only target and put the transactions needed into the second vector to get its trace.

No ❌


crates/client/rpc/src/lib.rs
jbcaron, line 1: is it really necessary to use an intermediate variable?

crates/pallets/starknet/src/tests/re_execute_transactions.rs
jbcaron, line 1: is it really necessary to use an intermediate variable?

crates/pallets/starknet/src/simulations.rs
```diff
@@ -229,4 +248,45 @@ impl<T: Config> Pallet<T> {
...
+                    Self::execute_user_transaction(tx.clone(), chain_id, block_context, execution_config)
+                }
+                UserOrL1HandlerTransaction::L1Handler(tx, _fee) => {
+                    Self::execute_message(tx.clone(), chain_id, block_context, execution_config)
+                }
+            })
+            .collect();
+
+        let mut execution_infos = Vec::with_capacity(exec_transactions.len());
+        for result in exec_transactions {
```
jbcaron: is it possible to use a `.map` directly?

jbcaron (approved):

crates/client/rpc/src/lib.rs
Tbelleng: Here the function changed in order to take 2 args as params, the first vector its tx that we dont want to trace and the second one is the ones we want to trace, so i added an empty vector for your function in order to fit the new function definiton

crates/pallets/starknet/src/simulations.rs
Tbelleng: Good suggestion, im gonna try to 🤝

crates/pallets/starknet/src/tests/re_execute_transactions.rs
Tbelleng: Same comment as above, its now mandatory to use 2 vectors of transactions in order to get a proper re-execution

antiyro (approved):

PR #10, opened 2024-03-07
refactor(substrate_hash): :recycle: Substrate hash no longer retrieved via rpc call

Trantorian1:
# Pull Request type

Substrate block hash in `l2.rs` is retrieved via an internal RPC call to the running Deoxys node.

## What is the new behavior?

Substrate hash is now retrieved through the use of the rpc client initialized in `service.rs`, which has been passed to `l2.rs`.

No


crates/client/sync/src/l2.rs
```diff
@@ -28,7 +29,7 @@ use tokio::sync::mpsc::Sender;
 use tokio::task::JoinSet;
 
 use crate::commitments::lib::{build_commitment_state_diff, update_state_root};
-use crate::utility::get_block_hash_by_number;
+use crate::utility::update_config;
```
jbcaron, line 32: remove this

crates/client/sync/src/l2.rs
```diff
@@ -87,44 +88,21 @@ pub struct SenderConfig {
...
     start_at: u64,
-    rpc_port: u16,
     backend: Arc<mc_db::Backend<B>>,
-) {
+    client: Arc<C>,
+) where
+    B: BlockT,
+    C: HeaderBackend<B>,
+{
+    update_config(&config);
```
jbcaron, line 102: remove update_config()
update_config is now in `fn run_node()`

antiyro (approved):

PR #12, opened 2024-03-11
fix(root): :bug: Got state root to work

Trantorian1:
# Pull Request type

- Bugfix

Deoxys computes an invalid state root for blocks > 0.

## What is the new behavior?

State root is now correctly computed up to block ~800. This is because the class root is not yet taken into consideration, but that should be only a minor change.

Yes. Changes had to be made to the way in which contract storage was represented in db. Nodes with a existing db will need to re-sync starting from block 0.

## Other information

> :warning: The above changes have introduced significant performance overheads. While this is a working solution, yet more effort will have to be dedicated towards optimization.


.gitignore
```diff
@@ -53,8 +53,7 @@ starknet-e2e-test/contracts/build
 # vscode settings
 .vscode/settings.json
 
-# script files
-output_deoxys.json
-output_pathfinder.json
+# rpc output
+output*
```
jbcaron, line 57: maybe add the directory

jbcaron:
I think we are reaching a point where we should start tracking down our more or less provisional 🚧 forks of the crates

crates/client/mapping-sync/src/sync_blocks.rs
```diff
@@ -40,6 +40,7 @@ where
 
                         // Success, we write the Starknet to Substate hashes mapping to db
                         let mapping_commitment = mc_db::MappingCommitment {
+                            block_number: digest_starknet_block.header().block_number,
                             block_hash: substrate_block_hash,
                             starknet_block_hash: digest_starknet_block_hash.into(),
                             starknet_transaction_hashes: digest_starknet_block
```
jbcaron, line 46: why do we recover all the digest?

crates/client/rpc/src/lib.rs
```diff
@@ -215,6 +215,18 @@ where
         })
     }
 
+    /// Return the block hash of a given block number.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_number` - The number of the block you'd like to retrieve the hash (starknet block).
+    fn get_cached_block_hash_by_number(&self, block_number: u64) -> Option<StarkHash> {
```
jbcaron: Is it efficient in case of restart?

crates/client/storage/src/overrides/mod.rs
```diff
@@ -127,6 +141,36 @@ where
...
 
+    /// Get the storage keys and values from a specific contract address
+    ///
+    /// # Arguments
+    ///
+    /// * `address` - The contract address to fetch the storage from
+    ///
+    /// # Returns
+    /// * `Vec<(storage_key, storage_value)>` - The storage from the provided contract address
+    fn get_storage_from(
```
jbcaron, line 152: why not handle the error?

crates/client/sync/src/commitments/contracts.rs
```diff
@@ -30,50 +39,50 @@ pub struct ContractLeafParams {
...
-        for (storage_key, storage_value) in updates {
-            let key = BitVec::from_vec(Felt252Wrapper::from(storage_key.0.0).0.to_bytes_be()[..31].to_vec());
-            let value = Felt252Wrapper::from(*storage_value);
-            bonsai_storage.insert(key.as_bitslice(), &value.into()).expect("Failed to insert storage update into trie");
+    // Inserts old storage changes
+    let db = overrides.for_schema_version(&Undefined);
+    if let Some(hash) = maybe_block_hash {
+        if let Some(storage) = db.get_storage_from(hash, *contract_address) {
+            storage.into_iter().map(convert_storage).for_each(|(key, value)| {
+                if value != Felt252Wrapper::ZERO {
```
jbcaron, line 60: will the case of a zero value be fixed in the bonsai crate? in this case add a todo

crates/client/sync/src/commitments/lib.rs
```diff
@@ -146,109 +162,130 @@ where
...
+fn contract_state_leaf_hash(
+    csd: &CommitmentStateDiff,
+    overrides: &OverrideHandle<Block<Header<u32, BlakeTwo256>, OpaqueExtrinsic>>,
+    contract_address: &ContractAddress,
+    storage_updates: &IndexMap<StorageKey, StarkFelt>,
+    maybe_block_hash: Option<H256>,
+) -> Result<Felt252Wrapper, BonsaiStorageError<BonsaiDbError>> {
+    let storage_root = update_storage_trie(overrides, contract_address, storage_updates, maybe_block_hash)?;
+
+    let nonce =
```
jbcaron, line 225: Is an error transformed into default?

crates/client/sync/src/commitments/lib.rs
```diff
@@ -146,109 +162,130 @@ where
...
-
-    Ok(state_root)
+    overrides: Arc<OverrideHandle<Block<Header<u32, BlakeTwo256>, OpaqueExtrinsic>>>,
+    bonsai_contract: Arc<Mutex<BonsaiStorage<BasicId, BonsaiDb<B>, Pedersen>>>,
+    bonsai_class: Arc<Mutex<BonsaiStorage<BasicId, BonsaiDb<B>, Poseidon>>>,
+    block_number: u64,
+    substrate_block_hash: Option<H256>,
+) -> Felt252Wrapper {
+    // Update contract and its storage tries
+    let contract_trie_root =
```
jbcaron, line 174: pass the mutex as a parameter in order to lock for the shortest possible time in the child function

crates/client/sync/src/commitments/lib.rs
```diff
@@ -146,109 +162,130 @@ where
...
+        contract_trie_root(&csd, overrides, bonsai_contract.lock().unwrap(), block_number, substrate_block_hash)
```
jbcaron, line 175: pass the mutex as a parameter in order to lock for the shortest possible time in the child function

jbcaron (commented):
check the comments
I wonder if the error handling is correct at the query level in the DB

antiyro (approved):

.gitignore
Trantorian1: json output from `rpc_rmp` is only stored at the root, so a directory is not really needed. Unless you meant saving `output*` in a separate directory (I don't think that would be necessary as these are essentially just test files made to be deleted at a later date).

crates/client/mapping-sync/src/sync_blocks.rs
Trantorian1: This isn't my code sry, couldn't tell you. You should see with @antiyro

crates/client/storage/src/overrides/mod.rs
Trantorian1: Same thing here, see with @antiyro (this is code from a merge)

crates/client/sync/src/commitments/contracts.rs
Trantorian1: Adding a TODO and sending the info over to the Bonsai Lib team :+1:

crates/client/sync/src/commitments/lib.rs
Trantorian1: What do you mean by that?

If you are wondering whether conversion is performed automatically on errors caught with `?`, it is.

crates/client/rpc/src/lib.rs
Trantorian1: This code has been deleted as it was no longer in use and causing clippy errors.

crates/client/sync/src/commitments/lib.rs
Trantorian1: This really isn't necessary since:

1. each mutex is **only** used in the context of the function it's lock is being passed to.
2. `verify_l2` is never called asynchronously, only the functions inside are to be parallelized.

The use of mutex here is just to allow the madara backend to access the bonsai dbs in a thread safe-context, such as during RPC calls where they might need to be accessed to retrieve hashes.

crates/client/sync/src/commitments/lib.rs
Trantorian1: same as above, this is not necessary in this case since the mutex is not used anywhere else (except maybe in RPC calls), and we definitely do not want any mutation to occur while we are computing the state root.

.gitignore
jbcaron: in this case change this in `/output*`

crates/client/sync/src/commitments/lib.rs
jbcaron: it's about `.unwrap_or(..)`

jbcaron: it's about `.unwrap_or(..)`

crates/client/sync/src/commitments/lib.rs
jbcaron: I hadn't seen the for loop in the contract_trie_root function which would force lock and unlock the mutex at each iteration

.gitignore
Trantorian1: All the files we ignore at the root do not have a folder specified

crates/client/sync/src/commitments/lib.rs
Trantorian1: That's some old code that was refactored. Waiting for v13 compatibility to be able to test removing it.

jbcaron (approved):

PR #15, opened 2024-03-14
L2 sync: Parallel fetching of blocks, classes, state updates

cchudant:

- Blocks are fetched from the sequencer sequentially.

## What is the new behavior?

- L2 sync fetching of blocks, classes, state updates is now parallel. 

N/A (project is in alpha)

## Other information

- Usage of futures stream `buffered()` for parallel fetching,
- Error handling is still a mess, but I've started moving to `thiserror`-based error handling. I am not opposed to use `anyhow` instead, but I feel like `thiserror` is more appropriate here.
- `verify_l2` is sync and takes a long time, so I have put it in a tokio blocking task (see <https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code>)
- Rate limits could be handled better

### DIFF
--- a/crates/client/sync/src/l2.rs
+++ b/crates/client/sync/src/l2.rs
@@ -119,7 +119,7 @@ where
     const MAX_RETRY: usize = 15;
     for _ in 0..MAX_RETRY {
         log::debug!("fetch_block_and_updates {}", block_n);
-        let block = fetch_block(&provider, block_n);
+        let block = fetch_block(provider, block_n);
         let state_update = fetch_state_and_class_update(provider, block_n, overrides, client);
         let (block, state_update) = tokio::join!(block, state_update);
         log::debug!("fetch_block_and_updates: done {block_n}");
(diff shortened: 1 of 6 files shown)


antiyro (approved):

Trantorian1:
Don't forget to update `CHANGELOG.md` with a short description of the changes made, as well as linting the project (_you can use `./scripts/deoxys lint` for that purpose_)

crates/client/sync/src/l2.rs
```diff
@@ -111,105 +157,104 @@ pub async fn sync<B, C>(
...
-    block_sender.send(block_conv).await.map_err(|e| format!("failed to dispatch block: {e}"))?;
+async fn fetch_block(client: &SequencerGatewayProvider, block_number: u64) -> Result<p::Block, L2SyncError> {
+    let block = client.get_block(BlockId::Number(block_number)).await?;
 
-    Ok(())
+    Ok(block)
 }
 
-pub async fn fetch_genesis_block(config: FetchConfig) -> Result<mp_block::Block, String> {
+// FIXME: I don't know why this is required by the CLI,, this is not used within the sync
```
Trantorian1: This is an artefact of an older version of the code when this was used to retrieve the head of the chain during initialization, but is since no longer used. You can remove it.

Trantorian1 (approved):

PR #17, opened 2024-03-16
Deoxys-TUI: an optional TUI dashboard

noamoulin:
## What is the new behavior ?

When running the node with the `--tui` command-line argument (after building it with the `tui` feature), a dashboard shows some node's metrics like CPU usage, RAM usage, storage usage, network traffic, and  classified L1 and L2 logs.

Not really, just a cool feature

## Other information

The dashboard is made with ratatui, implemented as a crate and launched in a thread from `node/commands/run.rs` in `run_node` if the `--tui` argument is provided.

To be able to redirect substrate logs to the TUI, it is necessary to define a writer for tracing (used by substrate). However, in its current state, substrate doesn't allow us to define a custom subscriber, so I had to write a script to directly modify the sources of the version of substrate cloned by cargo. This script is automatically executed in Deoxys if built with the `tui` flag, so the user doesn't have to worry about it. This modification will not affect the default behavior of substrate and therefore of Deoxys (it ensures that if an environment variable `SUBSCRIBER_DEFINED` specifying that a subscriber has already been defined, substrate does not redefine it and does not cause a panic).

****Coming soon:**** Fully modular widgets and tabs, more data, p2p map...


Trantorian1:
:musical_note:  _It just works_ :notes:

antiyro (approved):

PR #20, opened 2024-03-19
Fix/unused deps

antiyro:
removed unsued dependencies


jbcaron (approved):
STARKNET_HIGHEST_BLOCK_HASH_AND_NUMBER was not 'public'  to enforce the use of the getter and the updater to access this variable.

Tbelleng (approved):
lfgggg seems legit to me Boss 🙇

PR #23, opened 2024-03-22
fix(sync): Cleaned mc-sync isolating fetch process + added shared SyncStatus

antiyro:
### fix(sync): Cleaned mc-sync isolating fetch process + added shared SyncStatus


Tbelleng (approved):
Seems good to me 🚀

PR #26, opened 2024-03-25
L2 commitments: hash commitments & l2 sync parallelism

cchudant:

- Some hash computations could be done in parallel.

## What is the new behavior?

- Some hash computations are done in parallel

N/A (project is in alpha)

## Other information

- rayon

### DIFF
--- a/crates/client/sync/src/commitments/events.rs
+++ b/crates/client/sync/src/commitments/events.rs
@@ -5,11 +5,11 @@ use bonsai_trie::{BonsaiStorage, BonsaiStorageConfig};
 use mp_felt::Felt252Wrapper;
 use mp_hashers::pedersen::PedersenHasher;
 use mp_hashers::HasherT;
+use rayon::prelude::*;
 use starknet_api::transaction::Event;
 use starknet_ff::FieldElement;
 use starknet_types_core::felt::Felt;
 use starknet_types_core::hash::Pedersen;
-use rayon::prelude::*;
 
 use crate::utils::constant::bonsai_identifier;
 
@@ -67,9 +67,7 @@ pub fn memory_event_commitment(events: &[Event]) -> Result<Felt252Wrapper, Strin
     let identifier = bonsai_identifier::EVENT;
 
     // event hashes are computed in parallel
-    let events = events.par_iter().map(|event| {
-        calculate_event_hash::<PedersenHasher>(event)
-    }).collect::<Vec<_>>();
+    let events = events.par_iter().map(|event| calculate_event_hash::<PedersenHasher>(event)).collect::<Vec<_>>();
 
     // once event hashes have finished computing, they are inserted into the local Bonsai db
     for (i, event_hash) in events.into_iter().enumerate() {
--- a/crates/client/sync/src/commitments/events.rs
+++ b/crates/client/sync/src/commitments/events.rs
@@ -67,7 +67,7 @@ pub fn memory_event_commitment(events: &[Event]) -> Result<Felt252Wrapper, Strin
     let identifier = bonsai_identifier::EVENT;
 
     // event hashes are computed in parallel
-    let events = events.par_iter().map(|event| calculate_event_hash::<PedersenHasher>(event)).collect::<Vec<_>>();
+    let events = events.par_iter().map(calculate_event_hash::<PedersenHasher>).collect::<Vec<_>>();
 
     // once event hashes have finished computing, they are inserted into the local Bonsai db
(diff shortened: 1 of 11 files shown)


cchudant:
Reworked things around l2 sync too, and a lot more things around commitments
Removed as a whole mutexes which were needed as BonsaiStorage is not thread safe

antiyro (approved):

crates/client/sync/src/commitments/events.rs
```diff
@@ -89,12 +85,8 @@ pub async fn memory_event_commitment(events: &[Event]) -> Result<Felt252Wrapper,
     let id = id_builder.new_id();
 
     // run in a blocking-safe thread to avoid starving the thread pool
-    let root_hash = spawn_blocking(move || {
-        bonsai_storage.commit(id).expect("Failed to commit to bonsai storage");
-        bonsai_storage.root_hash(identifier).expect("Failed to get root hash")
-    })
-    .await
-    .map_err(|e| format!("Failed to computed event root hash: {e}"))?;
+    bonsai_storage.commit(id).expect("Failed to commit to bonsai storage");
```
Trantorian1, line 88: Doesn't this risk starving the current thread? Why remove `spawn_blocking`

crates/client/sync/src/commitments/lib.rs
```diff
@@ -205,34 +213,52 @@ pub fn update_state_root(
...
-    bonsai_contract_storage: Arc<Mutex<BonsaiStorage<BasicId, BonsaiDb, Pedersen>>>,
+    bonsai_contract: &mut BonsaiStorage<BasicId, BonsaiDb, Pedersen>,
+    bonsai_contract_storage: &mut BonsaiStorage<BasicId, BonsaiDb, Pedersen>,
     block_number: u64,
     maybe_block_hash: Option<H256>,
 ) -> Result<Felt252Wrapper, BonsaiStorageError<BonsaiDbError>> {
     let identifier = bonsai_identifier::CONTRACT;
     bonsai_contract.init_tree(identifier)?;
 
+    let start1 = std::time::Instant::now();
```
Trantorian1, line 224: Is it really a good idea to keep timers in the codebase? Just wondering, open to feedback :)

crates/client/sync/src/commitments/lib.rs
```diff
@@ -205,34 +213,52 @@ pub fn update_state_root(
...
+        .storage_updates
+        .iter()
+        .par_bridge()
+        .map(|contract_address| {
+            contract_state_leaf_hash(csd, &overrides, contract_address.0, maybe_block_hash, bonsai_contract_storage)
+        })
+        .collect::<Result<Vec<_>, BonsaiStorageError<BonsaiDbError>>>()?;
+    log::debug!("contract_trie_root updates: {:?}", std::time::Instant::now() - start);
 
+    let start = std::time::Instant::now();
```
Trantorian1, line 251: Same here

crates/client/sync/src/commitments/transactions.rs
```diff
@@ -97,13 +94,8 @@ pub async fn memory_transaction_commitment(
...
     let id = id_builder.new_id();
 
-    // run in a blocking-safe thread to avoid starving the thread pool
-    let root_hash = spawn_blocking(move || {
-        bonsai_storage.commit(id).expect("Failed to commit to bonsai storage");
-        bonsai_storage.root_hash(identifier).expect("Failed to get root hash")
-    })
-    .await
-    .map_err(|e| format!("Failed to computed transaction root hash: {e}"))?;
+    bonsai_storage.commit(id).expect("Failed to commit to bonsai storage");
```
Trantorian1, line 97: Again, why not run this in the `spawn_blocking` ?

crates/client/sync/src/commitments/lib.rs
cchudant: i think they should go once i get perf and flamegraph working again
I left them here because adding/removing them between every commit was bothering me

crates/client/sync/src/commitments/events.rs
cchudant: This function is called within a spawn blocking task (called by crate::convert::block, it's spaghetti)

crates/client/sync/src/commitments/transactions.rs
cchudant: Same as above, this function is now called within a rayon context and not tokio

Trantorian1 (approved):

PR #28, opened 2024-03-27
feat(storage): :sparkles: Bonsai storage refactor and rpc call support

Trantorian1:
# Pull Request type

- Feature

Bonsai storage is handled separately from Substrate storage and does not interact with the runtime. Moreover, the manual use of bonsai storage in cases such as `commitments/lib.rs` is not type-safe and is error prone, especially when the amount of type-juggling and manual, often repetitive hash computation.

Other issues further arise when considering a situation with many parallel accesses to the db, which is typical in our use case. This PR aims to introduce some of the boilerplate necessary for minimizing storage locks during block sync by working of BonsaiDb transactional state. Since this is still a work in progress in upstream dependencies, parts of the code relevant to this issue have been commented out, with a more naive 'lock-on-use' approach being currently used.

## What is the new behavior?

- Created a new `StoraHandler` abstraction for handling bonsai storage in a type-safe way by providing a single source of mutation for the bonsai storage and handling type conversions and hashing for interfacing with the bonsai lib.
- Set up the boilerplate for writing changes though BonsaiDb transactional state and state merges.
- Updated rpc methods which rely on bonsai storage to use `StorageHandler`.
- Merged duplicate types in `crates/runtime/src/types.rs` and `crates/primitives/types`.

No.

## Other information

Looking for feedback on the implementation of `StorageHandler`.


antiyro (approved):

antiyro (approved):

Tbelleng (approved):
Seems good to me, Lfgg well done 🤝

PR #30, opened 2024-03-28
feat(api_key): Added support for --gateway-api to avoid rate limit from the gateway

antiyro:
## feat(api_key): Added support for --gateway-api to avoid rate limit from the gateway

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ git # Deoxys Changelog
 
 ## Next release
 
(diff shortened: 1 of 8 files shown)


Tbelleng (approved):

jbcaron (approved):
everything seems correct, an interesting feature would be to retrieve the key via an environment variable with `std::env`

antiyro:
> everything seems correct, an interesting feature would be to retrieve the key via an environment variable with `std::env`

Agree we should add a config file in general for parameters

PR #32, opened 2024-03-29
db: refactor the crate

cchudant:
# Pull Request type

- Refactoring

Resolves: #NA

## What is the new behavior?

Got rid of kvdb and kvdb-rocksdb.

N/A

## Other information

All locks are removed, accesses to bonsai storages are mediated through the BonsaiStorageAccess wrapper, which you can use to do read accesses, or get a owned BonsaiStorage to make writes.
PersistentBonsaiDatabase is now implemented properly
All Backend db accesses are now using rocksdb directly


cchudant:
How should we proceed with #28 and this one? @Trantorian1

antiyro (approved):
I 100% approve those changes, great work!

Trantorian1:
The idea would be be to merge #32 first. I can then go over it and migrate most of our Substrate storage into the new db. I'll be integrating that as part of #28 @cchudant

crates/client/db/src/bonsai_db.rs
```diff
@@ -1,265 +1,404 @@
...
+        // if let Some(max_number_snapshot) = self.config.max_saved_snapshots {
+        //     while self.snapshots.len() > max_number_snapshot {
+        //         self.snapshots.pop_first();
+        //     }
+        // }
+    }
+
+    fn transaction(&self, id: BasicId) -> Option<Self::Transaction> {
+        log::trace!("Generating RocksDB transaction");
+        if let Some(snapshot) = self.snapshots.get(&id) {
```
Trantorian1, line 385: If I got this right, this will only be able to retrieve database transactions if we have stored the _exact_ snapshot for the given `id`?

crates/client/db/src/lib.rs
```diff
@@ -175,33 +280,47 @@ impl DeoxysBackend {
                         paritydb_path: starknet_database_dir(db_config_dir, "paritydb"),
                         cache_size: 0,
                     },
-                    _ => return Err("Supported db sources: `rocksdb` | `paritydb` | `auto`".to_string()),
+                    _ => bail!("Supported db sources: `rocksdb` | `paritydb` | `auto`"),
```
Trantorian1, line 283: This is more out of curiosity: when do you consider it to be a better fit to use `thiserror` vs `anyhow`?

crates/client/db/src/lib.rs
```diff
@@ -230,15 +349,15 @@ impl DeoxysBackend {
         BACKEND_SINGLETON.get().map(|backend| &backend.sierra_classes).expect("Backend not initialized")
     }
 
-    pub fn bonsai_contract() -> &'static Arc<Mutex<BonsaiStorage<BasicId, BonsaiDb, Pedersen>>> {
+    pub fn bonsai_contract() -> &'static BonsaiStorageAccess<Pedersen> {
```
Trantorian1, line 352: Once we've merged this I'll start working on integrating this into #28: do you think it would be a good idea to delegate mutability control into the `StorageHandler` struct (see #28)? ie:

- `StorageHandler` would contain a single static readable reference for each bonsai trie.
- When mutable access is required, a clone of the backing bonsai trie woudl be created.
- Merging back into the backing trie would be done by `StorageHandler`. This can be done in a safer way by giving a unique type to writable tries so that class trie updates cannot be merged back to the contract trie for example.

What is your opinion on this?

crates/client/db/src/mapping_db.rs
```diff
@@ -45,30 +46,31 @@ impl MappingDb {
...
+        let block_mapping_col = self.db.get_column(Column::BlockMapping);
+
+        match self.db.get_cf(&block_mapping_col, starknet_block_hash.encode())? {
             Some(raw) => Ok(Some(Vec::<DHashT>::decode(&mut &raw[..])?)),
             None => Ok(None),
         }
     }
 
     /// Register that a Substrate block has been seen, without it containing a Starknet one
     pub fn write_none(&self, block_hash: DHashT) -> Result<(), DbError> {
```
Trantorian1, line 58: Do we know what this is used for? The name kinda screams 'delete me' from my point of view haha. Maybe in a later pr, just wondering.

Trantorian1 (commented):
All seems good to me. Please just fix the state root and check the comments below. I'm wondering if errors in the state root might not have to do with how our writable Bonsai dbs are handled vs read-only ones.

crates/client/db/src/bonsai_db.rs
cchudant: yes, this is taken from https://github.com/keep-starknet-strange/bonsai-trie/blob/oss/src/databases/rocks_db.rs
I don't fully understand snapshots, but from what I understand they have a lifetime that must be less than the DB lifetime: they don't persist across restarts

crates/client/db/src/lib.rs
cchudant: My thinking was: thiserror for everything, anyhow for stuff that sets up the startup.
However, I think it's kinda bad for user experience as anyhow's messages are more developer-oriented and show stacktraces vs errors that are user errors
maybe we should just put everything on thiserror at some point

crates/client/db/src/lib.rs
cchudant: If i understand correctly, `BonsaiStorageAccess` is already pretty much your `StorageHandler` in spirit without all the trie-specific utility functions, is that correct?

crates/client/db/src/mapping_db.rs
cchudant: i really have no clue :)

crates/client/db/src/lib.rs
Trantorian1: Yep, the idea is just to provide some extra type safety over raw usage of the backend database, which in my experience can be error-prone (so for example we can only add contracts to the contract bonsai trie, and we don't have to remember to remove the first 5 bits of any key). This would quickly broaden though, as the plan is to have this handle most of the deoxys storage.

Trantorian1 (approved):

PR #34, opened 2024-04-02
Feat : ✨ Adding estimate_fee for v0.7.0

Tbelleng:
# Pull Request type

- Feature

Modifying Estimate_fee rpc call in order to be aligned with starknet 0.7.0 spec-version

## What is the new behavior?

New EstimateFee Rpc call introduce a more detailed version of Fee, adding da gas consumption and new units (FRI and WEI)

No

## Other information


Tbelleng:
Just have to change the runtime declaration and it should be good

jbcaron (commented):
is this struct really necessary `SimulationFlagForEstimateFee` ?

antiyro (approved):

PR #58, opened 2024-04-18
fix(compute): 🐛 Fixed prepare_data_availability_modes computation

antiyro:


Tbelleng (approved):
Lfg, seems legit to me boss, good work 👍

jbcaron (approved):
looks ok

PR #60, opened 2024-04-18
fix(class): Fix Sierra classes conversion (missing abis)

antiyro:


Tbelleng (approved):
Seems good to me

jbcaron (approved):
Good job Boss

PR #64, opened 2024-04-19
fix: :bug: fix StateReader for get_storage_at()

jbcaron:
# Retrieve the value of a key from a given block in the bonsai lib

# Pull Request type

- Bugfix

- `get_storage_at` does not use cache of 'BlockifierStateAdapter' to retrieve keys
- `get_storage_at` retrieve the value of the keys on the last block

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ git # Deoxys Changelog
 
 ## Next release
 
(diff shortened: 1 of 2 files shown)


antiyro (approved):

Trantorian1 (approved):

Tbelleng (approved):
Better code than the Boss Himself, respect to you sir 🫡

PR #70, opened 2024-04-22
Added boilerplate for Bonsai storage migration

Trantorian1:
# Pull Request type

- Feature

Currently, Deoxys relies heavily on the Substrate runtime storage. This poses issues as it was designed for sparse usage, and is not pertinent for the storage of data which is not consensus-critical. Furthermore, use of the Substrate runtime storage and the associated `StorageOverride`s built on top has lead to a convoluted use of storage which is hard to reason about.

> Another issue worth noting is that we currently use rocksdb / substrate for _sequential_ insertions: this can be hevilly parallelized.

## What is the new behavior?

- Added new storage primitives in `crates/client/src/storage_handler` for accessing storage statically and _outside_ of the Substrate runtime.
- Started migrating Substrate runtime storage to the Bonsai lib `BonsaiStorage` and `RevertibleStorage` via `storage_handler`. This inculdes the following mappings:
    - `block_number` -> `block_hash`
    - `block_hash` -> `block_number`
    - `contract_address` -> `class_hash` & `nonce`
    - `class_hash` -> `contract_abi` & `contract_class`
    - `class_hash` -> `compiled_class_hash`
- Insertion and retrieval from any of these new storage primitives can now be done in parallel _lock-free_.
- All storage apart from new blocks have been moved out of the runtime: as a result `on_initialized` is no longer used for storing data. Any related logic, such as serializing data to communicate it to the runtime for storage, have been removed. Storage updates are now handled as part of `l2.rs`.
- Storage insertions have been heavily parallelized where possible.
- Removed the use of `StorageOverride` from the code base.

Yes. Any node already running will have to be re-run from the block 0 to update the new databases.

## Other information

Note that this is PR mostly introduces the boilerplate necessary for a full migration to Bonsai lib storage. Further improvements remain to be made:

- currently, there remains an issue where the wrong state root is being calculated. This is probably due to me inserting default `Nonces` in the db in some special cases.
- significant improvements in storage access using transactional state are possible after **[pr #28](https://github.com/keep-starknet-strange/bonsai-trie/pull/28)** on the bonsai-lib

Please keep these points in mind as you review this PR.

> Note that multiple changes to the bonsai-lib are also essential to this pr. These are present on my fork of the repo and are yet to be merged.


jbcaron (approved):

crates/client/db/src/storage_handler/contract_data.rs
```diff
@@ -0,0 +1,146 @@
...
+        let tree: BTreeMap<u64, StorageContractData> = match db
+            .get_cf(&column, contract_address.encode())
+            .map_err(|_| DeoxysStorageError::StorageRetrievalError(StorageType::ContractData))?
+        {
+            Some(bytes) => BTreeMap::decode(&mut &bytes[..])
+                .map_err(|_| DeoxysStorageError::StorageDecodeError(StorageType::ContractData))?,
+            None => BTreeMap::new(),
+        };
+
+        Ok(tree.get(&block_number).cloned())
```
jbcaron: you must retrieve the value for the lower or equal index closest to the block number
```rust
 tree.get.range(..=index).rev().next().map(|(_, value)| value)
```

antiyro (approved):

crates/client/rpc/src/methods/trace/utils.rs
```diff
@@ -451,3 +418,14 @@ where
...
 }
+
+pub fn block_number_by_id(id: BlockId) -> u64 {
+    match id {
+        BlockId::Number(number) => number,
+        BlockId::Hash(block_hash) => match storage_handler::block_number().get(&Felt252Wrapper(block_hash)) {
+            Ok(Some(block_number)) => block_number,
+            _ => get_highest_block_hash_and_number().1,
+        },
+        BlockId::Tag(_) => get_highest_block_hash_and_number().1,
```
jbcaron, line 429: tag pending not supported

jbcaron (approved):

PR #75, opened 2024-04-25
feat(storage): Updated storage to store state diff

Trantorian1:
# Pull Request type

- Feature

Deoxys does not store state diffs between blocks which make it especially hard to retrieve them during blockifier execution.

## What is the new behavior?

Starknet core state diffs are now stored in database and are accessible at any point in the codebase to reconstruct the state update for a given block. This has also reduced reliance on `StateUpdateWrapper`, which will be removed from the codebase along with `ContractClassWrapper` in a future pr.

Yes. Due to the changes to storage, any running database will have to re-start synchronization from block 0.


crates/client/db/src/storage_handler/block_state_diff.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::{Column, DatabaseExt, DeoxysBackend};
+
+pub struct BlockStateDiffView;
+
+impl BlockStateDiffView {
+    pub fn insert(&mut self, block_number: u64, state_diff: StateDiff) -> Result<(), DeoxysStorageError> {
+        let db = DeoxysBackend::expose_db();
+        let column = db.get_column(Column::BlockStateDiff);
+
+        let bytes = serde_json::to_vec(&state_diff)
```
jbcaron, line 14: you should use crate bincode

jbcaron (approved):

antiyro (approved):

crates/client/db/src/storage_handler/block_state_diff.rs
Trantorian1: Since the starknet core `StateDiff` already implemented `Serialized` and `Deserialize`, I dind't much see the need to add another crate to do the same job :man_shrugging:.

Are there any potential advantages to using `bincode` I might have missed?

PR #82, opened 2024-04-28
feat(storage): :sparkles: `BonsaiStorage` logs are no longer stored in db

Trantorian1:
# Pull Request type

- Bugfix

Bonsai lib logs currently take up a cumulative 93.1% of the database space, with the bonsai lib as a whole being responsible for 98.8% of database usage.

image

## What is the new behavior?

Removed reliance on Bonsai lib: `BonsaiStorage` is no longer used to store contract storage in a revertible way. Instead, we use our own custom solution to handle that (kudos to @jbcaron for comming up with this). This has resulted in a 13.3x reduction in storage space. Note that this still introduces a temporary duplication of storage, with contract key/value being stored both in the Bonsai lib and our custom storage. 

image

> :information_source: Note that the Bonsai lib is still responsible for 51.6% of the storage in Deoxys.

Yes. Due to changes in the database, any node currently running will have to re-start synchronizing from the block 0.

## Other information

Some of the code for inserting contract storage into the db might seem messy. Keep in mind I will be making another pr tomorrow rewriting all rocksdb storage access to make better use write batches and multiget.


crates/client/db/src/lib.rs
```diff
@@ -206,6 +208,7 @@ impl Column {
             Column::ContractClassData => "contract_class_data",
             Column::ContractData => "contract_data",
             Column::ContractClassHashes => "contract_class_hashes",
+            Column::ContractStorage => "contrac_storage",
```
antiyro, line 211: "contract"

antiyro (approved):

crates/client/db/src/storage_handler/contract_storage.rs
```diff
@@ -0,0 +1,294 @@
...
+                        Some(bytes) => bincode::deserialize(&bytes)
+                            .map_err(|_| DeoxysStorageError::StorageDecodeError(StorageType::ContractStorage)),
+                        None => Ok(History::default()),
+                    },
+                )
+            })
+            .collect();
+
+        let mut batch = WriteBatchWithTransaction::<true>::default();
+        for (key, mut history, value) in izip!(keys, histories?, values) {
```
jbcaron, line 253: return the error before
`histories?`

jbcaron (approved):
Cf comment

PR #86, opened 2024-04-29
perf(rocksdb): :zap: Better use of Rocksdb multiget and batch writes

Trantorian1:
# Pull Request type

- Feature

Our current usage of rocksdb heavily parallelises database read and write operations when commiting. This is not necessarily more efficient, as rocsdkb has it's own capabilities for retrieving and writing multiple values from and to the database.

## What is the new behavior?

Made better use of rocksdb multiget and batch writes. Storage handler `commit` operations are now synchronous. Note that each commit is still run inside a separate task which is joined as this still provides better performance, but the commit operation itself is synchronous. See `storage_update.rs` for more information.

Changes were also made to the data serialization model, with `bincode` now being the primary method of serialization over `parity-scale-codec`. Changes have also been made to `StorageContractData` to avoid the usage of binary trees to store class hash and nonce history, preferring instead our own revertible `History<T>` implementation. This should also have the effect of reducing storage in this area. New accessors were also added to simply retrieving the nonce and class hash as compared to before.

Work was also done refactoring primitive types `StorageContractData` and `StorageContractClassData`. These have been moved from `carates/primitives` to a new `client/db/src/storage_handler/primitives` folder. Please place any struct used as a storage wrapper here in the future.

> :zap:  **All in all this has resulted in a 10% improvement in sync time.**

Yes. Due to change in storage, nodes already running will have to restart synchronizing from block 0.


crates/client/db/src/storage_handler/block_hash.rs
```diff
@@ -10,17 +9,17 @@ impl BlockHashView {
     pub fn insert(&mut self, block_number: u64, block_hash: &Felt252Wrapper) -> Result<(), DeoxysStorageError> {
         let db = DeoxysBackend::expose_db();
         let column = db.get_column(Column::BlockNumberToHash);
-        db.put_cf(&column, block_number.encode(), block_hash.encode())
+        db.put_cf(&column, bincode::serialize(&block_number).unwrap(), bincode::serialize(&block_hash).unwrap())
```
jbcaron, line 12: maybe create new Storage error for serialize and deserialize

crates/client/db/src/storage_handler/contract_data.rs
```diff
@@ -24,132 +20,153 @@ impl StorageView for ContractDataView {
...
             false => Ok(false),
         }
     }
 }
 
 impl ContractDataView {
-    pub fn get_at(
+    pub fn get_nonce(self, contract_address: &ContractAddress) -> Result<Option<Nonce>, DeoxysStorageError> {
+        self.get(contract_address)
+            .map(|option| option.map(|contract_data| contract_data.nonce.get().cloned().unwrap_or_default()))
```
jbcaron, line 49: use `as_ref()` instead of `cloned()`

jbcaron (approved):

antiyro (approved):

PR #88, opened 2024-04-30
fix: db, add test, add codec traits

jbcaron:
# fix: db, add test, add codec traits

## Pull Request type

- Bugfix
- Feature

## What is the new behavior?

- add codec traits on Starkfelt and History<T>
- use `get_storage_at` instead of using bonsai
- add unit tests

No


Tbelleng (approved):
Seems good to me, well done sir 🤝

antiyro (approved):
Let's see how the db compaction operates

PR #93, opened 2024-05-02
Fix/genesis

jbcaron:
# Fix: genesis

## Pull Request type

- Bugfix

- state root does not match at block 255

## What is the new behavior?

-
-
-

## Other information

block_hash is not stored for block 0


antiyro (approved):
approved thanks

cchudant (approved):

PR #94, opened 2024-05-02
Allow compiling without libm + Fix state corruption during sync when doing CTRL+c

cchudant:
# Pull Request type

- Bugfix

Resolves: #NA

## What is the new behavior?
- Allow compiling without libm
- Fix state corruption during sync when doing CTRL+c
No

## Other information
Also, made the create_block do stuff in parallel wrt. state/class updates


cchudant:
Compiling without sound is done using `cargo run --no-default-features`

jbcaron (approved):

antiyro (approved):
looks fine

PR #97, opened 2024-05-03
feat: store key/value in `--disble-root` mode

jbcaron:
# feat: store key/value in `--disble-root` mode

## Pull Request type

- Feature

- key/value are not stored in `--disable-root` mode (#96)

## What is the new behavior?

- key/value are stored in `--disable-root` mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ git # Deoxys Changelog
 
 ## Next release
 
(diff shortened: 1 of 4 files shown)


antiyro (approved):
Nice, better to store keys in disable

cchudant (approved):

antiyro (approved):
Looks better now thanks

PR #99, opened 2024-05-03
fix: class update

jbcaron:
# fix: class update

## Pull Request type

- Bugfix

## What is the new behavior?

- fetch `old_declared_contracts`


antiyro (approved):
Seems good to me

cchudant (approved):
looks good

PR #100, opened 2024-05-06
fix: remove `unwrap` in storage

jbcaron:
# fix: remove `unwrap` in storage

## Pull Request type

- Bugfix

- if the block sync stops with the storage updates already in db but not the block, when restarting the program crashes

## What is the new behavior?

- remove unwrap in the storage


antiyro (approved):

cchudant (approved):
yay

PR #106, opened 2024-05-08
perf: convert blocks in parallel

cchudant:
# Pull Request type

- Perf

sequencial

## What is the new behavior?

blocks are converted and commitments computed in parallel, up to 10 blocks in the future

No


antiyro (approved):
approving speed tested

jbcaron (approved):

PR #110, opened 2024-05-09
fix: graceful shutdown of rocksdb on ctrl+c

cchudant:
# Pull Request type

- Bugfix

bad basically

## What is the new behavior?

- fix: graceful shutdown of rocksdb on ctrl+c
- fix: better error handling around l1 and l2 sync
- perf: compile with target_cpu=skylake by default

No

## Other information

there's so much more to do


antiyro (approved):
Nice work

PR #111, opened 2024-05-09
performance improvement for data storage key/value history

jbcaron:
# performance improvement for data storage key/value history

## Pull Request type

## What is the new behavior?

- use encode on `Starkfelt`
- `add_to_history_encoded` enhances performance by allowing new values to be added to the encoded history without requiring the decoding of the entire dataset, thereby reducing processing operations and associated costs.

DB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- perf: storage key with encode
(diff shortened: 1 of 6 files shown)


cchudant (approved):
nice

antiyro (approved):

PR #114, opened 2024-05-13
fix: block context, fetch class, simulate flags

jbcaron:
# Fix: RPC

## Pull Request type

- Bugfix

- we recover classes already stored in db if it is deployed on a contract

## What is the new behavior?

- consistent creation of simulation flags
- creation of the block context from the corresponding block header
- fix is_missing_class: only classes not in DB are fetch


antiyro (approved):
Thanks for spotting missing classes

PR #115, opened 2024-05-13
feat: backups

cchudant:
# Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?

3 new CLI options:
```sh
--backup-every-n-blocks 30
--backup-dir ./rocksdb_backups
--restore-from-latest-backup
```
Rocksdb backups are incremental, meaning having 3 backups actually doesnt store 3 whole copies of the db, so it's way smaller.
Substrate DB is **NOT** backed up in this PR. so be careful

No

## Other information

Future work: it is possible to add metadata to backups, and then choose a backup to load based on that metadata
Meaning I kind of want to make a `--restore-from-block <block_n>` arg to restore from a specific block instead of the latest backup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- feat(db): backups
 - fix: state root for nonce
 - fix: store the first history in storage ket
 - perf: improved perfs with parallelized iteration over tx hashes cache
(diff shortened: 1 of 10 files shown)


cchudant:
Also in this PR: makes the node crash on state root mismatch

jbcaron (approved):

antiyro (approved):
One of the greatest features Deoxys ever had for dev purposes

PR #116, opened 2024-05-14
fix: state root - replaced_classes

jbcaron:
# fix: state root - replaced_classes

# Pull Request type

- Bugfix

## What is the new behavior?

-


antiyro (approved):
fixing state root correctly thanks

PR #120, opened 2024-05-15
feat: fetch block and state update in only one request

jbcaron:
# feat: fetch block and state update in only one request

# Pull Request type

- Feature

- if the fetch of a single class fails in a block, the entire fetch of the classes is retry

## What is the new behavior?

- fetch block and state update in only one request
- retry fetch for each class independently

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- feat: fetch block and state update in only one request
 - fix: creation of the block context
 - fix: is_missing_class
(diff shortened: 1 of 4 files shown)


jbcaron:
# 🚧 TODO:
- [x] `is_missing_class`: improve db error handling
- [ ] `fetch_class` : remove expect
- [x] `retry`: check the relevance of L2SyncError
- [ ] ~fetch: use anyhow to report errors~

cchudant (approved):

antiyro (approved):

PR #124, opened 2024-05-21
Feat/meta

jbcaron:
## Pull Request type

## What is the new behavior?

- using OnceCell instead of lazy_static for static Configuration
- using  const instead of lazy_static thanks to the `FieldElement::from_mont` constructor which does not return a result
- remove crate `mp-storage`

No


antiyro (approved):
approving changes

PR #129, opened 2024-05-23
db: batch and parallelize history multiget/insert

cchudant:
# Pull Request type

- Bugfix

Resolves: #NA

## What is the new behavior?

- DB path has been changed
- Refactor a bit around db initialization
- ContractStorage history has been changed

old dbs will need to resync from scratch


antiyro (approved):
tested

PR #131, opened 2024-05-26
History using rocksdb keys

cchudant:
# Pull Request type

- Bugfix

Contract keys as (contract, storage key) => list of [block_n, felt]
Contract class hashes and nonces as contract => list of [block_n, (class_hash, nonce)]

## What is the new behavior?

Let prefix = (contract, storage key)

Put a new value at block_n:
we put_cf at key=(prefix || block_n) the new value

Get at block_n:
create a rocksdb prefix iterator for prefix, reverse mode until prefix start, seek to prefix||block_n
get next item from the iterator

ContractData has been split to ContractToClassHash and ContractToNonce.

Yes, need to resync

## Other information

Need review around the RPCs @jbcaron


antiyro (approved):
approving changes

jbcaron (approved):

PR #180, opened 2024-06-26
fix typo

apoorvsadana:


antiyro:
Thanks, this will be fixed in @cchudant next PR.

PR #196, opened 2024-07-06
fix(metrics): removed influx and added l2_state_size data

antiyro:
# Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-


jbcaron (approved):
The get_directory_size function looks efficient overall, but here are a few points to consider to ensure it doesn’t introduce unnecessary overhead:
- **Error Handling:** The function currently ignores errors when reading directory entries or fetching metadata.
- **Concurrency:** If the directory contains a large number of files or deeply nested structures, processing them sequentially could become a bottleneck. Consider using a concurrent approach to traverse the directory entries in parallel.
- **Metrics Frequency Consideration:** Since this function is used for metrics, consider how frequently these metrics need to be collected. If the function is called frequently, ensure that it does not introduce significant performance overhead.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix(metrics): removed influx and added l2_state_size endpoint
```
cchudant: nit: its not really an endpoint

crates/client/sync/src/utils/utility.rs
```diff
@@ -125,3 +125,24 @@ pub fn trim_hash(hash: &Felt) -> String {
 
     format!("{}...{}", prefix, suffix)
 }
+
+pub fn get_directory_size(path: &Path) -> u64 {
```
cchudant: I would prefer if we were using https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.get_column_family_metadata_cf
which returns, for a given column, its size as
```rs
pub struct ColumnFamilyMetaData {
    // The size of this column family in bytes, which is equal to the sum of
    // the file size of its "levels".
    pub size: u64,
    // The name of the column family.
    pub name: String,
    // The number of files in this column family.
    pub file_count: usize,
}
```
Plus, we can do a `sum` gauge and a individual gauge for each column

CHANGELOG.md
antiyro: sorry i forgot to change this

crates/client/sync/src/l2.rs
```diff
@@ -381,6 +383,12 @@ async fn update_sync_metrics(
     block_metrics.l1_gas_price_wei.set(f64::from_u128(block_header.l1_gas_price.eth_l1_gas_price).unwrap_or(0f64));
     block_metrics.l1_gas_price_strk.set(f64::from_u128(block_header.l1_gas_price.strk_l1_gas_price).unwrap_or(0f64));
 
+    if block_number % 1000 == 0 {
```
0xChqrles: avoid magic number and replace `1000` by a constant eg. `STORAGE_SIZE_UPDATE_STEP`

0xChqrles (changes requested):

crates/client/sync/src/metrics/block_metrics.rs
```diff
@@ -7,6 +7,7 @@ pub struct BlockMetrics {
     pub l2_sync_time: Gauge<F64>,
     pub l2_avg_sync_time: Gauge<F64>,
     pub l2_latest_sync_time: Gauge<F64>,
+    pub l2_state_size: Gauge<F64>,
```
cchudant, line 10: I'd also like a HistogramVec to have a gauge for every column if possible :)

crates/client/sync/src/l2.rs
antiyro: Please go back to your organization!

crates/client/sync/src/metrics/block_metrics.rs
antiyro: doing it rn

cchudant (approved):
Perfect :)

Issue #189, opened 2024-06-28
feat(sequencer): Add support for a transaction pool to achieve sequencing


antiyro:
### Is there an existing issue?

- [X] I have searched the existing issues

### Motivation

In the context of adding support for the `--authority` mode regarding block creation as mentioned in #188, we need to implement a transaction pool that adheres to the current rules of the mainnet sequencer.

### Request

This issue serves as a reference for implementing the transaction pool, which will be used to receive, re-execute, and order incoming transactions via the following methods:

- [`starknet_addInvokeTransaction`](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L11)
- [`starknet_addDeclareTransaction`](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L67)
- [`starknet_addDeployAccountTransaction`](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L144)

The transaction pool will follow the system outlined below:

1. Creation of an empty block and start of the block time.
2. Reception of a transaction via the methods (add_invoke, declare, deploy_account).
3. Execution of this transaction:
   a. Creation of a block context with block information, including fees.
   b. Creation of the state adapter to have a snapshot of the chain at the moment of the block with all the methods to access the state for blockification.
   c. Creation of a temporary cache from the state adapter.
   d. Execution of the transaction with a caching system.
   e. Retrieval of the execution trace, creation of the statediff using the cache.
   f. Committing the temporary cache to the StateAdapter cache.
4. Ordering of the executed transactions.
5. End of block time, creation of the Block Header.
6. Storage of the block.

Note that during the entire process from block creation to storage, the block must be available in the pending state.

### Solution

This is an overview for a primitive implementation of the transaction pool adhering to the current Starknet specifications. Changes may be made as more information about the protocol itself becomes available.

### Are you willing to help with this request?

Yes!

antiyro:
Addionatl UX feature mentioned by @tdelabro: add support for an `add_deploy_account` transaction if a wallet does an invoke but has not been deployed yet. This is used by argent and braavos on mainnet.

PR #198, opened 2024-07-08
Migrate to blockifier v0.8 and Sequencing / mempool part 1

cchudant:
The aim of this PR is to add basic mempool and block production to match how the starkware sequencer works.

Currently, the 3 write RPCs (`add_invoke_transaction`, `add_deploy_account_transaction`, `add_declare_transaction`) are handled by directly calling the real starkware sequencer. To support sequencing, we may replace that with a trait, with two implementations: the first one, will fallback to the starkware sequencer (or a configurable one), and the second one will pass that to our new mempool structure for block production.

### Mempool logic

Note: When we are talking about `tip` here, we mean that in the sense of the `tip` field in broadcasted transactions, meaning the "to tip $20" sense of the verb - each transaction has a tip, and for conflicting transactions (to the same account and nonce) we need to prioritize the highest tip.

We also want to support accepting txs in any order - meaning if we receive a nonce that is too high (up to a point, probably) we should accept it anyway in case a future transaction will fill up the "nonce gap".

The mempool will be a tip heap, with one item per contract address. That item will be called a "NonceSequence" for the sake of our mempool - and it is just a nonce heap, where we can insert nonces in any order for the given contract.

When receiving a transaction, we need to do validation of nonces, max fees, fees, cairo `__validate__` etc. This is already implemented by blockifier using StatefulValidator -- https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/src/blockifier/stateful_validator.rs
We just need to call that.
Then, we put it in the mempool, replacing any transaction with the same contract_addr, nonce pair if the tip is higher.

The mempool will need to implement other stuff such as:
- mempool limits: we should evict transactions when the mempool is full
- transaction eviction based on how old they are (see how geth handle limits, for exemple - and maybe ask people from starkware how their sequencer handle that)
- we need to think hard about how one could DoS us.

### Block production

Block production will be run on a separate thread, outside of the current tokio runtime, and will probably have a higher thread priority to make sure RPC requests and the future p2p service cannot DoS us.

Blockifier already implements what they call the "bouncer", which returns and allow us to stop adding transactions to a block when it is full.

We only need to `pop` nonce sequences from the heap, giving us the highest tip txs first, put them into a block until the block is full or no further transaction can be added. At this stage, if the execution of a transaction fails, we (implemented in blockifier, actually) marks this transaction as rejected and fees are processed accordingly. This invalidates the whole nonce sequence.

The revalidation (including reexecution of `__validate__`), tx execution, revert-handling, etc. is already done in blockifier: https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/src/blockifier/transaction_executor.rs#L177
It also implements concurrent (optimistic) transaction execution, which we may want to try out.

The current starknet sequencer has the concept of a "pending block". I am not 100% sure how this works on starknet. **(TODO: pending block writeup)**

### Further reading / useful resources

https://github.com/starkware-libs/mempool is a crate by starkware that aims to handle the mempool entirely, as a separate service. It is very barebone for now, and doesn't support actual validation or actual transactions. It also doesn't support having multiple transactions with different nonces in the mempool for a single contract address, which we want to support ASAP.

I have read some of the geth code for inspiration: transaction picking for block production https://github.com/ethereum/go-ethereum/blob/master/miner/ordering.go#L91 - tx pool https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go
A lot of good things to pick from there: the basic logic of a price heap with an item per unique contract, and in that item a nonce heap is the same as what I intend to implement.
Something like their tx pool metrics will need to be implemented in our node too.
Furthermore, they have a system where if the txpool is full; they find the "spammer nodes" and start evicting their transaction first to make room. We will need a logic to limit mempool size and policies for eviction like that too.

### Future

Next starknet versions intend to add the concept of a paymaster for transaction fees - afaik this is not yet implemented ; but the starkware sequencer implements paying transactions in STRK, by getting the STRK/ETH quote from a oracle. That is already implemented in blockifier it seems, but we need to provide the STRK/ETH quote.
Volition - and L2 DA will need to be implemented when it will be available (i am not aware any of it is currently available or wanted in madara yet)

Leader election - Running deoxys/madara on a k8s cluster - if the sequencer goes down, we want another one of the nodes to become the sequencer. This can be done by simply doing k8s-style leader election, meaning acquiring a Lock object and it will be auto released if the process goes down, letting another deoxys/madara node take the lock and continue sequencing where the other one left off.


apoorvsadana:
> and for conflicting transactions (to the same account and nonce) we need to prioritize the highest tip

I am not sure if need this right now. There's good value in being FCFS for the start. Arbitrum, Starknet and I believe a lot of sequencers today are FCFS. It removes/ MEV complications and makes design simpler.

> we mean that in the sense of the tip field in broadcasted transactions

Is the `tip` field currently being used on mainnet? I think this is to be used once we decentralise. Right now, the resource bounds are used for V3 and max_fee for V2 afaik.

> if we receive a nonce that is too high (up to a point, probably) we should accept it anyway in case a future transaction will fill up the "nonce gap"

Good point! An edge case we faced with Madara over here was
1. We need to validate every txn before adding it to the mempool to avoid a DDoS (even if it is of a future nonce)
2. When using accounts like Argent, they deploy the account with the first txn. So they send two txs immediately i.e. deploy account + invoke (whatever the first tx was)
3. If you try to validate this it fails because for the invoke the account isn't deployed yet
4. So we had to add an edge case if a txn fails because of a no contract exists error and the nonce is 0x1, you just skip validation
5. Also, all txs are revaliated before actual execution

Here's a reference of the [code](https://github.com/keep-starknet-strange/madara/blob/7b405924b5859fdfa24ec33f152e56a97a047e31/crates/pallets/starknet/src/lib.rs#L779)

> We will need a logic to limit mempool size and policies for eviction like that too.

A FCFS will make this simpler, just keep removing oldest txs. We could do optimisations but I feel this works pretty week for a decently large queue (substrate had 8k by default afaik).

cchudant:
@apoorvsadana 

Oh, I see. So, no tip heap, just first come first serve. I got the impression the tip field and MEV was a thing already because the mempool from starkware-libs is being designed with that in mind.

> Good point! An edge case we faced with Madara over here was

I've come accross that in blockifier too, ahah: https://github.com/starkware-libs/blockifier/blob/main/crates/native_blockifier/src/py_validator.rs#L83
It seems they implement in on the python bindings (not sure why?) - we'll have to implement it in deoxys/madara too

apoorvsadana:
> I got the impression the tip field and MEV was a thing already because the mempool from starkware-libs is being designed with that in mind

I think they're building this for when Starknet will be decentralised but there's no spec for that yet and I am honestly not sure when that happens. My opinion is to start with something simple like FCFS and make changes incrementally as and when it makes sense.

> I've come accross that in blockifier too

Oh wow, didn't know that.

> we'll have to implement it in deoxys/madara too

You mean implement Python bindings? I am not sure if I understood.

cchudant:
@apoorvsadana 
> You mean implement Python bindings? I am not sure if I understood.

The logic for that specific case is in the python bindings, which we don't use -- i am not sure why it isnt on the main crate, but that means that we have to copy that in madara

apoorvsadana:
Ya that's what we did previously as well. Copy parts from the blockifier when we can't directly call them. One issue we faced last time was that it wasn't very clear what parts were copied and what parts need to be updated everytime we did a rebase (@tdelabro mostly had context as he did the rebase). It would be nice to have some file structures/comments/README in place to check whenever we're copying code from the blockifier :)

cchudant:
It took more time than I thought because I had to update blockifier.
As you can see from the number of changed files, updating blockifier is a pain :)

I have started a ChainConfig primitive that will be relevant for both starknet and appchains.
Mempool insertion is done, I'm working on block production
Mempool Inner is the struct behind a global lock, and as such it will become a choke point and I decided on a design that will make it as fast as possible.
It has a ton of invariants, though but it is self contained and will have a proptest.
https://github.com/cchudant/deoxys/blob/sequencing/crates/client/mempool/src/inner.rs

crates/client/db/src/block_db.rs
```diff
@@ -63,37 +62,37 @@ impl DeoxysBackend {
...
+        let block_n = bincode::deserialize(&res)?;
         Ok(Some(block_n))
     }
 
     fn block_hash_to_block_n(&self, block_hash: &Felt) -> Result<Option<u64>> {
         let col = self.db.get_column(Column::BlockHashToBlockN);
         let res = self.db.get_cf(&col, bincode::serialize(block_hash)?)?;
         let Some(res) = res else { return Ok(None) };
-        let block_n = codec::Decode::decode(&res)?;
+        let block_n = bincode::deserialize(&res)?;
```
apoorvsadana, line 77: what's the advantage of bincode over the previous implementation?

crates/client/exec/src/block_context.rs
```diff
@@ -1,33 +1,46 @@
...
+            self.block_context.clone(),
+            // No concurrency yet.
+            TransactionExecutorConfig { concurrency_config: Default::default() },
+        )
+    }
+
+    pub fn tx_validator(&self) -> StatefulValidator<BlockifierStateAdapter> {
+        // When deploying an account and invoking a contract at the same time, we want to skip the validation step for the invoke tx.
+        // This number is the maximum nonce the invoke tx can have to qualify for the validation skip.
+        // TODO: is the number 2 good here?
```
apoorvsadana: Since nonce start from 0, `1` should also work here. Setting a higher number would be a tradeoff in UX and security. Do you think it makes sense to move this to an env variable or is that an overkill?

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -163,62 +126,62 @@ impl<'a> StateReader for BlockifierStateAdapter<'a> {
             return Err(StateError::UndeclaredClassHash(class_hash));
         };
 
-        Ok(CompiledClassHash(class_info.compiled_class_hash.to_stark_felt()))
+        Ok(CompiledClassHash(class_info.compiled_class_hash))
     }
 }
 
-impl<'a> State for BlockifierStateAdapter<'a> {
```
apoorvsadana, line 170: Is this pending?

crates/client/exec/src/block_context.rs
```diff
@@ -36,10 +49,10 @@ impl<'a> ExecutionContext<'a> {
...
 
-        CachedState::new(BlockifierStateAdapter::new(self.backend, on_top_of), GlobalContractCache::new(16))
+        CachedState::new(BlockifierStateAdapter::new(Arc::clone(&self.backend), on_top_of))
     }
 
-    pub fn new(backend: &'a DeoxysBackend, block_info: &DeoxysMaybePendingBlockInfo) -> Result<Self, Error> {
+    pub fn new(backend: Arc<DeoxysBackend>, block_info: &DeoxysMaybePendingBlockInfo) -> Result<Self, Error> {
         let (db_id, protocol_version, block_number, block_timestamp, sequencer_address, l1_gas_price, l1_da_mode) =
             match block_info {
                 DeoxysMaybePendingBlockInfo::Pending(block) => (
```
apoorvsadana, line 61: in case of pending `block_number`, should we throw an error instead of return default `0`

crates/client/exec/src/block_context.rs
```diff
@@ -62,39 +75,32 @@ impl<'a> ExecutionContext<'a> {
...
-        let block_info = blockifier::block::BlockInfo {
+        let block_info = blockifier::blockifier::block::BlockInfo {
             block_number: BlockNumber(block_number),
             block_timestamp: BlockTimestamp(block_timestamp),
-            sequencer_address: sequencer_address.to_stark_felt().try_into().unwrap(),
+            sequencer_address: sequencer_address
+                .try_into()
+                .map_err(|_| Error::InvalidSequencerAddress(sequencer_address))?,
             gas_prices: (&l1_gas_price).into(),
             // TODO: Verify if this is correct
```
apoorvsadana: looks correct

crates/primitives/block/src/chain_config.rs
```diff
@@ -0,0 +1,213 @@
...
+                        pedersen: usize::MAX,
+                        poseidon: usize::MAX,
+                        range_check: usize::MAX,
+                        range_check96: usize::MAX,
+                    },
+                    gas: 5_000_000,
+                    n_steps: 40_000_000,
+                    message_segment_length: usize::MAX,
+                    n_events: usize::MAX,
+                    state_diff_size: usize::MAX,
```
apoorvsadana: The max state diff size is limited by the blob size on the L1 (131072 bytes currently). Should we use that limit here?

crates/primitives/block/src/chain_config.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+        Err(UnsupportedProtocolVersion(version))
+    }
+
+    pub fn starknet_mainnet() -> Self {
+        // Sources:
+        // - https://docs.starknet.io/tools/important-addresses
+        // - https://docs.starknet.io/tools/limits-and-triggers (bouncer & block times)
+        // - pending_block_update_time: educated guess
+        // - bouncer builtin_count, message_segment_length, n_events, state_diff_size are probably wrong
```
apoorvsadana, line 86: We can ask this on the `madara-core` or `madara-starkware` channel

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+            self.executor.bouncer.bouncer_config.block_max_capacity = config_bouncer.clone();
+        } else {
+            let current_pending_tick = self.current_pending_tick;
+            self.current_pending_tick += 1;
+
+            let n_pending_ticks_per_block = self.backend.chain_config().block_time.as_millis()
+                / self.backend.chain_config().pending_block_update_time.as_millis();
+            let n_pending_ticks_per_block = n_pending_ticks_per_block as usize;
+
+            if current_pending_tick == 0 || current_pending_tick >= n_pending_ticks_per_block {
```
apoorvsadana: why do we need to ignore the first tick?

crates/primitives/block/src/chain_config.rs
```diff
@@ -0,0 +1,213 @@
...
+            versioned_constants: [
+                (StarknetVersion::STARKNET_VERSION_0_13_0, BLOCKIFIER_VERSIONED_CONSTANTS_0_13_0.deref().clone()),
+                (StarknetVersion::STARKNET_VERSION_0_13_1, BLOCKIFIER_VERSIONED_CONSTANTS_0_13_1.deref().clone()),
+                (StarknetVersion::STARKNET_VERSION_0_13_1_1, BLOCKIFIER_VERSIONED_CONSTANTS_0_13_1_1.deref().clone()),
+                (StarknetVersion::STARKNET_VERSION_0_13_2, VersionedConstants::latest_constants().clone()),
+            ]
+            .into(),
+            latest_protocol_version: StarknetVersion::STARKNET_VERSION_0_13_2,
+            block_time: Duration::from_secs(6 * 60),
+            pending_block_update_time: Duration::from_secs(2),
```
apoorvsadana, line 112: is my understanding correct that `block_time` should always be divisible by `pending_block_update_time` to make sense? if yes, should we add an assert during initialisation?

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+                // first tick is ignored.
+                // out of range ticks are also ignored.
+                return Ok(());
+            }
+
+            // Reduced bouncer capacity for the current pending tick
+
+            // (current_val / n_ticks_per_block) * current_tick_in_block
+            let frac = n_pending_ticks_per_block * current_pending_tick;
```
apoorvsadana: Does this match the formula in the comment?

Also, what's the reasoning behind this math exactly? In Starknet, the block time is actually dynamic. If we can fill the blocks earlier then the block is sealed in much less than 6 minutes. You can see the variation here - https://www.starknet-monitor.com/

crates/client/mempool/src/inner.rs
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    pub fn pop_next_chunk(&mut self, dest: &mut Vec<MempoolTransaction>, n: usize) {
+        for _ in 0..n {
+            let Some(tx) = self.pop_next() else { break };
+            dest.push(tx);
+        }
+    }
+
+    pub fn readd_txs(&mut self, txs: Vec<MempoolTransaction>) {
```
apoorvsadana: do you mean `read`?

crates/client/exec/src/execution.rs
```diff
@@ -37,37 +39,60 @@ impl<'a> ExecutionContext<'a> {
...
-                    Transaction::L1HandlerTransaction(_) => None,
+                    Transaction::L1HandlerTransaction(_) => None, // There is no minimal_l1_gas field for L1 handler transactions.
                 };
-                let mut state = CachedState::<_>::create_transactional(&mut cached_state);
+
+                let make_reexec_error =
+                    |err| TxReexecError { block_n: self.db_id, hash, index: executed_prev + index, err };
+
+                let mut transactional_state = TransactionalState::create_transactional(&mut cached_state);
+                let execution_flags = ExecutionFlags { charge_fee, validate, concurrency_mode: false };
```
apoorvsadana, line 62: Is this always false? We also use concurrency in `tx_executor`. If these two values will always be the same maybe we should move this to a CLI flag?

crates/client/exec/src/execution.rs
```diff
@@ -37,37 +39,60 @@ impl<'a> ExecutionContext<'a> {
...
-                    .execute(&mut state, &self.block_context, charge_fee, validate)
-                    .and_then(|mut tx_info| {
-                        if tx_info.actual_fee.0 == 0 {
-                            tx_info.actual_fee = blockifier::fee::fee_utils::calculate_tx_fee(
-                                &tx_info.actual_resources,
-                                &self.block_context,
-                                &fee_type,
+                    .execute_raw(&mut transactional_state, &self.block_context, execution_flags)
+                    .and_then(|mut tx_info: TransactionExecutionInfo| {
+                        // TODO: why was this here again?
```
apoorvsadana, line 67: replying to `TODO`, I believe there's a case where fee is not calculated by default for L1 handler transactions and if max_fee is zero. We did this in Madara but we were also on an older version.

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn produce_block(&mut self) -> Result<(), Error> {
+        // Complete the block with full bouncer capacity.
+        let finish_block = true;
+        self.update_pending_block(finish_block)?;
+
+        // Convert the pending block to a closed block and save to db.
+
+        // Prepare for next block.
+        let mut pending_block = self.mempool.get_or_create_pending_block()?;
+        let bouncer_config = self.backend.chain_config().bouncer_config.clone();
```
apoorvsadana: `bouncer_config` isn't being used

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        Ok(Self { backend, mempool, executor, current_pending_tick: 0 })
+    }
+
+    fn update_pending_block(&mut self, finish_block: bool) -> Result<(), Error> {
+        let config_bouncer = &self.backend.chain_config().bouncer_config.block_max_capacity;
+
+        if finish_block {
+            // Full bouncer capacity
+            self.executor.bouncer.bouncer_config.block_max_capacity = config_bouncer.clone();
```
apoorvsadana: Is it possible to return here? Or does execute txs modify the block in someway (even though it doesn't seem to execute any txs in this case)

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+        // Add back the unexecuted transactions to the mempool.
+        self.mempool.readd_txs(rest_txs_to_process);
+
+        Ok(())
+    }
+
+    fn produce_block(&mut self) -> Result<(), Error> {
+        // Complete the block with full bouncer capacity.
+        let finish_block = true;
+        self.update_pending_block(finish_block)?;
```
apoorvsadana: Should we call `store_block` somewhere here? I can't find where the block is stored after being completed

crates/client/mempool/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        match self.backend.get_block(&DbBlockId::Pending)? {
+            Some(block) => Ok(block),
+            None => {
+                // No pending block: we create one :)
+
+                let block_info =
+                    self.backend.get_block_info(&BlockId::Tag(BlockTag::Latest))?.ok_or(Error::NoGenesis)?;
+                let block_info = block_info.as_nonpending().ok_or(Error::Internal("Latest block is pending".into()))?;
+
+                Ok(DeoxysMaybePendingBlock {
```
apoorvsadana: Is this being stored in the DB?

crates/client/mempool/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        // Get pending block
+        let pending_block_info = self.get_or_create_pending_block()?;
+
+        // If the contract has been deployed for the same block is is invoked, we need to skip validations.
+        // NB: the lock is NOT taken the entire time the tx is being validated. As such, the deploy tx
+        //  may appear during that time - but it is not a problem.
+        let deploy_account_tx_hash = if let AccountTransaction::Invoke(tx) = &tx {
+            let mempool = self.inner.read().expect("Poisoned lock");
+            if mempool.has_deployed_contract(&tx.tx.sender_address()) {
+                Some(tx.tx_hash) // we return the wrong tx hash here but it's ok because the actual hash is unused by blockifier
```
apoorvsadana, line 83: It's strange that the blockifier takes the hash but doesn't use it. It's nice that this works right now but I think we should add a TODO to send the actual hash. If the blockifier updates later to use this hash then our code will silently fail without us noticing.

crates/client/mempool/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        // The timestamp *does not* take the transaction validation time into account.
+        let arrived_at = ArrivedAtTimestamp::now();
+
+        // Get pending block
+        let pending_block_info = self.get_or_create_pending_block()?;
+
+        // If the contract has been deployed for the same block is is invoked, we need to skip validations.
+        // NB: the lock is NOT taken the entire time the tx is being validated. As such, the deploy tx
+        //  may appear during that time - but it is not a problem.
+        let deploy_account_tx_hash = if let AccountTransaction::Invoke(tx) = &tx {
```
apoorvsadana, line 80: I think a Declare is also possible here.

crates/client/mempool/src/inner.rs
```diff
@@ -0,0 +1,536 @@
...
+                            panic!("Force add should never error")
+                        }
+                        // Rollback the prior mutation.
+                        if let Some(contract_address) = &deployed_contract_address {
+                            if !self.deployed_contracts.remove(contract_address) {
+                                return Err(TxInsersionError::AccountAlreadyDeployed);
+                            }
+                        }
+
+                        return Err(TxInsersionError::NonceConflict);
```
apoorvsadana, line 259: should we return the same error thrown by `insert`?

crates/client/db/src/block_db.rs
cchudant: `codec` has been removed, we are now only using bincode, but a previous refactor left this in as it was not a priority.
I removed it here since removing it was faster than converting it to not use StarkFelts.

crates/client/exec/src/block_context.rs
cchudant: I'll move that to the chain config I think

crates/client/exec/src/blockifier_state_adapter.rs
cchudant: So, it turns out blockifier doesn't use the State implmementation for adapters, we have to take the state diff and apply it ourserlves - or make the CachedState write reads to the underlying State, which would need this impl
It is pending yes

crates/client/exec/src/block_context.rs
cchudant: So the thinking here is: This part was made to work when the actual genesis block is in a pending state (meaning, we are executing on a pending block and there are no latest block in db)
This is the reason this unwrap_or(0) exists: in this case we would be executing genesis

Tracing the execution using the trace rpcs require reexecuting old blocks, so this is useful for them

crates/client/mempool/src/lib.rs
cchudant: I think the relevant is gone in the next release anyway...

crates/primitives/block/src/chain_config.rs
cchudant: I think we should yes

crates/client/mempool/src/lib.rs
cchudant: > I think a Declare is also possible here.

I don't think the starkware sequencer supports using classes before they are declared?
It's just for deploy account and invoke I think

cchudant:
I am missing the commitment computation, and a bit of plumbery and testing
#213 is touching the commitments and I will probably rebase on that when it gets merged

crates/client/exec/src/blockifier_state_adapter.rs
apoorvsadana: Oh, you mean the blockifier just creates a state diff on the `CachedState` and then we need to manually take the state diff and write it to our DB?

crates/client/mempool/src/lib.rs
apoorvsadana: I didn't understand, will this code not be used in the next release?

crates/client/mempool/src/lib.rs
apoorvsadana: So I was actually saying that it's possible that 

Nonce 0 - Deploy account
Nonce 1 - Declare some random class (we should ignore validation)

And the above check ignores validation only if Nonce 1 is an invoke if I understand it correctly, so this case will be missed right?

Cargo.toml
```diff
@@ -61,7 +62,8 @@ rocksdb = { version = "0.22", features = [
 ] }
 
 # Bonsai trie dependencies
-bonsai-trie = { default-features = false, git = "https://github.com/cchudant/bonsai-trie.git", branch = "fix_inserts_remove_leaks", features = [
+# bonsai-trie = { default-features = false, git = "https://github.com/cchudant/bonsai-trie.git", branch = "fix_inserts_remove_leaks", features = [
+bonsai-trie = { default-features = false, path = "../bonsai-trie", features = [
```
tdelabro: revert to github import

Cargo.toml
```diff
@@ -137,16 +139,22 @@ serde_json = { version = "1.0", default-features = false, features = ["std"] }
...
+proptest-derive = "0.5.0"
+env_logger = "0.11.3"
 
 [patch.crates-io]
 starknet-core = { git = "https://github.com/kasarlabs/starknet-rs.git", branch = "fork" }
 starknet-providers = { git = "https://github.com/kasarlabs/starknet-rs.git", branch = "fork" }
 
 # wait for PR https://github.com/starknet-io/types-rs/pull/76 to be merged
-starknet-types-core = { git = "https://github.com/jbcaron/types-rs.git", rev = "fd0e062" }
+# starknet-types-core = { path = "../types-rs" }
```
tdelabro: remove commented code?

crates/client/db/src/block_db.rs
```diff
@@ -30,29 +32,26 @@ pub struct TxIndex(pub u64);
...
 // inconsistent state.
 impl DeoxysBackend {
-    pub fn chain_info(&self) -> Result<ChainInfo> {
-        let col = self.db.get_column(Column::BlockStorageMeta);
-        let res = self.db.get_pinned_cf(&col, ROW_CHAIN_INFO)?.ok_or(DeoxysStorageError::MissingChainInfo)?;
-        Ok(bincode::deserialize(res.as_ref())?)
-    }
-
-    pub(crate) fn assert_chain_info(&self, expected: &ChainInfo) -> anyhow::Result<()> {
+    pub(crate) fn assert_chain(&self, expected: &ChainConfig) -> anyhow::Result<()> {
```
tdelabro: It's named `assert_chain` but I see an insertion being made in db if res.chain_id == expected.chain_id
You need a more explicit name as well as a bit of an explanation of what the purpose is.
Also `asset` in Rust context means that the method panic if the assertion is false (here if the expected chainId is not the correct one). I don't see it being the case here

crates/client/db/src/bonsai_db.rs
```diff
@@ -138,6 +138,9 @@ impl BonsaiDatabase for BonsaiDb<'_> {
     fn write_batch(&mut self, batch: Self::Batch) -> Result<(), Self::DatabaseError> {
         Ok(self.db.write_opt(batch, &self.write_opt)?)
     }
+
+    #[cfg(test)]
+    fn dump_database(&self) { unimplemented!() }
```
tdelabro: Is it forgotten?

crates/client/db/src/lib.rs
```diff
@@ -355,19 +356,28 @@ impl Drop for DeoxysBackend {
...
 
-        let backend = Arc::new(Self { backup_handle, db, last_flush_time: Default::default() });
-        backend.assert_chain_info(chain_info)?;
+        let backend = Arc::new(Self {
+            backup_handle,
+            db,
+            last_flush_time: Default::default(),
+            chain_config: Arc::clone(&chain_config),
+        });
+        backend.assert_chain(&chain_config)?;
```
tdelabro: Why would `backend.chain_config` be any different from the one you just initialized it with?
Why do you feel the need to check again?

crates/client/exec/src/block_context.rs
```diff
@@ -1,33 +1,46 @@
...
+        // This number is the maximum nonce the invoke tx can have to qualify for the validation skip.
+        // TODO: is the number 2 good here?
+        let max_nonce_for_validation_skip = Nonce(Felt::from(2u64));
+        StatefulValidator::create(self.init_cached_state(), self.block_context.clone(), max_nonce_for_validation_skip)
+    }
+
+    pub fn init_cached_state(&self) -> CachedState<BlockifierStateAdapter> {
         let on_top_of = match self.db_id {
             DbBlockId::Pending => Some(DbBlockId::Pending),
             DbBlockId::BlockN(block_n) => {
```
tdelabro, line 45: `Number` rather than `BlockN`makes it  `DbBlockId::Number` rather than ``DbBlockId::BlockN` where block is repeated two times

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -1,54 +1,34 @@
...
 use starknet_api::state::StorageKey;
 use starknet_core::types::Felt;
+use std::sync::Arc;
 
 /// `BlockifierStateAdapter` is only use to re-executing or simulate transactions.
-/// None of the setters should therefore change the storage persistently,
-/// all changes are temporary stored in the struct and are discarded after the execution
-pub struct BlockifierStateAdapter<'a> {
-    backend: &'a DeoxysBackend,
+/// All of the setters change the storage permanently. Use blockifier's `CachedState` for a cache.
```
tdelabro: They change the storage permanently now? So they are not suited for simulation and re-execution as it is being explained

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -1,54 +1,34 @@
...
 }
 
-impl<'a> StateReader for BlockifierStateAdapter<'a> {
-    fn get_storage_at(&mut self, contract_address: ContractAddress, key: StorageKey) -> StateResult<StarkFelt> {
-        if *contract_address.key() == StarkFelt::ONE {
-            let block_number = (*key.0.key()).try_into().map_err(|_| StateError::OldBlockHashNotProvided)?;
+impl StateReader for BlockifierStateAdapter {
+    fn get_storage_at(&self, contract_address: ContractAddress, key: StorageKey) -> StateResult<Felt> {
+        if *contract_address.key() == Felt::ONE {
+            let block_number = felt_to_u64(key.0.key()).map_err(|_| StateError::OldBlockHashNotProvided)?;
```
tdelabro: Add some comments and link to the specs here: https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-state/#address_0x1

Explain why it is treated differently

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -59,11 +39,10 @@ impl<'a> StateReader for BlockifierStateAdapter<'a> {
                         format!("Failed to retrieve block hash for block number {block_number}",),
                     )
                 })?
-                .ok_or(StateError::OldBlockHashNotProvided)?
-                .to_stark_felt());
+                .ok_or(StateError::OldBlockHashNotProvided)?);
         }
 
-        let Some(on_top_of_block_id) = self.on_top_of_block_id else { return Ok(StarkFelt::ZERO) };
+        let Some(on_top_of_block_id) = self.on_top_of_block_id else { return Ok(Felt::ZERO) };
```
tdelabro, line 57: This is an optimization around the assumption that genesis state is always empty
Make it explicit.

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -1,54 +1,34 @@
...
-    fn get_storage_at(&mut self, contract_address: ContractAddress, key: StorageKey) -> StateResult<StarkFelt> {
-        if *contract_address.key() == StarkFelt::ONE {
-            let block_number = (*key.0.key()).try_into().map_err(|_| StateError::OldBlockHashNotProvided)?;
+impl StateReader for BlockifierStateAdapter {
+    fn get_storage_at(&self, contract_address: ContractAddress, key: StorageKey) -> StateResult<Felt> {
+        if *contract_address.key() == Felt::ONE {
+            let block_number = felt_to_u64(key.0.key()).map_err(|_| StateError::OldBlockHashNotProvided)?;
 
             return Ok(self
                 .backend
```
tdelabro, line 46: With this implementation, doesn't it means that I can query block hashes from future blocks?
Let's say my chain tip is at block 100.
My node knows the block hash of block 80.

I create a blockifier state adapter `on_top_of_block_id: Some(50)` and try to address 0x1 read key 80.
`get_block_hash(&BlockId::Number(80))` will return the hash of block 80, because the node is aware of it, even if the block is in the future regarding `on_top_of_block_id`

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -76,15 +55,11 @@ impl<'a> StateReader for BlockifierStateAdapter<'a> {
...
+            .unwrap_or(Felt::ZERO))
     }
 
-    fn get_nonce_at(&mut self, contract_address: ContractAddress) -> StateResult<Nonce> {
+    fn get_nonce_at(&self, contract_address: ContractAddress) -> StateResult<Nonce> {
         log::debug!("get_nonce_at for {:#?}", contract_address);
-        if let Some(nonce) = self.nonce_update.get(&contract_address) {
-            return Ok(*nonce);
-        }
         let Some(on_top_of_block_id) = self.on_top_of_block_id else { return Ok(Nonce::default()) };
```
tdelabro, line 75: Same assumption of an empty genesis state.

Document it on top of the trait implementation block

crates/client/exec/src/blockifier_state_adapter.rs
tdelabro: Part of our requirements for blockifer was to expose the cached state as a Trait, with abort and commit methods to dump it onto the "actual" storage.

crates/client/exec/src/call.rs
```diff
@@ -50,11 +53,21 @@ impl<'a> ExecutionContext<'a> {
 
         let mut cached_state = self.init_cached_state();
 
+        let class_hash = cached_state
+            .get_class_hash_at(storage_address)
+            .map_err(TransactionExecutionError::StateError)
+            .map_err(make_err)?;
```
tdelabro, line 59: Is there a way to only compute this in case the execution failed? 
Al it is only used in the error

crates/client/exec/src/execution.rs
```diff
@@ -1,17 +1,19 @@
...
+use blockifier::transaction::transactions::{ExecutableTransaction, ExecutionFlags};
 use starknet_api::transaction::TransactionHash;
 
 use crate::{Error, ExecutionContext, ExecutionResult, TxFeeEstimationError, TxReexecError};
 
-impl<'a> ExecutionContext<'a> {
+impl ExecutionContext {
+    /// Execute transactions. The returned `ExecutionResult`s are the results of the `transactions_to_trace`. The results of `transactions_before` are discarded.
+    /// This function is useful for tracing trasaction execution, by reexecuting the block.
     pub fn execute_transactions(
```
tdelabro: rename to `re_execute_transaction` to make it more explicit that the changes are not persisted

crates/client/exec/src/execution.rs
```diff
@@ -37,37 +39,60 @@ impl<'a> ExecutionContext<'a> {
...
 
+                // We need to estimate gas too.
                 let minimal_l1_gas = match &tx {
                     Transaction::AccountTransaction(tx) => Some(
                         estimate_minimal_gas_vector(&self.block_context, tx)
                             .map_err(TransactionExecutionError::TransactionPreValidationError)
                             .map_err(|err| TxFeeEstimationError { block_n: self.db_id, index, err })?,
                     ),
-                    Transaction::L1HandlerTransaction(_) => None,
+                    Transaction::L1HandlerTransaction(_) => None, // There is no minimal_l1_gas field for L1 handler transactions.
```
tdelabro, line 55: Link to the official spec/doc

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
+//! This should probably be moved into another crate.
```
tdelabro: Yup. Create a block_producer crate

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+use dc_db::DeoxysBackend;
+use dc_exec::{ExecutionContext, BlockifierStateAdapter};
+use dp_utils::graceful_shutdown;
+use std::sync::Arc;
+
+use crate::{clone_account_tx, Error, Mempool};
+
+pub struct BlockProductionTask {
+    backend: Arc<DeoxysBackend>,
+    mempool: Arc<Mempool>,
```
tdelabro: If moved to another crate, it should be generic over different mempool implementations.
This requires creating a Mempool trait under primitives.
Unlock dependency injection for unitests with an really simple in-memory mempool

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+    executor: TransactionExecutor<BlockifierStateAdapter>,
+    current_pending_tick: usize,
+}
+
+const TX_CHUNK_SIZE: usize = 128;
```
tdelabro: What is that? I don't understand by reading the name of the const. 
It needs some comment to explain.
Or move it closer to where it is used

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+use dp_utils::graceful_shutdown;
+use std::sync::Arc;
+
+use crate::{clone_account_tx, Error, Mempool};
+
+pub struct BlockProductionTask {
+    backend: Arc<DeoxysBackend>,
+    mempool: Arc<Mempool>,
+    executor: TransactionExecutor<BlockifierStateAdapter>,
+    current_pending_tick: usize,
```
tdelabro: I need a comment explaining this field to me

crates/client/mempool/src/block_production.rs
tdelabro: I don't understand what is going on here.
Why isn't `self.executor.bouncer.bouncer_config.block_max_capacity` set to `self.backend.chain_config().bouncer_config.block_max_capacity` from the getgo

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    fn update_pending_block(&mut self, finish_block: bool) -> Result<(), Error> {
+        let config_bouncer = &self.backend.chain_config().bouncer_config.block_max_capacity;
+
+        if finish_block {
+            // Full bouncer capacity
+            self.executor.bouncer.bouncer_config.block_max_capacity = config_bouncer.clone();
+        } else {
+            let current_pending_tick = self.current_pending_tick;
```
tdelabro: I don't understand at all what is going on with this whole tick logic.

What I was expecting is:
- when you create a `BlockProductionTask` you register it's creation time, and it's `block_max_capacity`
- then every time you add tx to the block, you make sure that `current_time < start_time + block_time_limit` and you check for the bouncer return.


And that's it.

tdelabro: What would be the issue with such a design?

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+                n_steps: config_bouncer.n_steps / frac,
+                state_diff_size: config_bouncer.state_diff_size / frac,
+            };
+        }
+
+        let mut txs_to_process = Vec::with_capacity(TX_CHUNK_SIZE);
+        self.mempool.take_txs_chunk(&mut txs_to_process, TX_CHUNK_SIZE);
+
+        let blockifier_txs: Vec<_> =
+            txs_to_process.iter().map(|tx| Transaction::AccountTransaction(clone_account_tx(&tx.tx))).collect();
```
tdelabro, line 206: Is there a way to avoid this useless iteration?
`take_txs_chunk` could directly do the conversion while it is iterating

tdelabro: especially because we are not using the `arrived_at` in the current context

tdelabro: oh we are using in when we put it back into the mempool.

Still, we should aim to reduce iteration as much as possible.

Modifying blockifier `execute_txs` so that it takes an iterator could be a good solution but will take time

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+        // Prepare for next block.
+        let mut pending_block = self.mempool.get_or_create_pending_block()?;
+        let bouncer_config = self.backend.chain_config().bouncer_config.clone();
+        self.executor = ExecutionContext::new(Arc::clone(&self.backend), &pending_block.info)?.tx_executor();
+        self.current_pending_tick = 0;
+
+        Ok(())
+    }
+
+    pub async fn block_production_task(&mut self) -> Result<(), anyhow::Error> {
```
tdelabro, line 345: SHouldn't the signature be:
`pub async fn block_production_task(&mut self) -> !`

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,159 @@
...
+                _ = interval_block_time.tick() => {
+                    if let Err(err) = self.produce_block() {
+                        log::error!("Block production task has errored: {err:#}");
+                    }
+                },
+                _ = interval_pending_block_update.tick() => {
+                    if let Err(err) = self.update_pending_block(false) {
+                        log::error!("Pending block update task has errored: {err:#}");
+                    }
+                },
```
tdelabro: I think it is okay to panic and shut down the whole node if one of those task fail.

tdelabro (dismissed):
Branch is broken:

```shell
deoxys git:(sequencing) cargo check
    Updating crates.io index
error: failed to select a version for `bonsai-trie`.
    ... required by package `dc-db v0.1.0 (/Users/tdelabro/Documents/code/starkware/deoxys/crates/client/db)`
versions that meet the requirements `*` (locked to 0.1.0) are: 0.1.0

the package `dc-db` depends on `bonsai-trie`, with features: `std` but `bonsai-trie` does not have these features.


failed to select a version for `bonsai-trie` which could resolve this conflict
```

crates/client/db/src/block_db.rs
cchudant: This function checks a that the program was started on a db of the wrong chain (ie. main vs sepolia) and returns an anyhow error if it does.
You are right about assert meaning panic in usual rust code but I feel like this function name isn't that far of a stretch. I'll change it anyway

crates/client/db/src/bonsai_db.rs
cchudant: No, it's because I am seeing ghost in rust-analyzer, which sometimes ask me to implement it even though I shouldn't (it's a required method only in the `test` cfg in bonsai-trie)

crates/client/db/src/lib.rs
cchudant: Ah I see why you want a different name now. It's checking if the chain config in DB is the same as &chain_config given to us from CLI args.

crates/client/mempool/src/block_production.rs
cchudant: We want ticks every two seconds - that makes it easier for optimistic parallelism later, and also it allows us to do writes to DB only every two seconds. I've heard (unsure?) this is also how the current starkware sequencer operates (- or at least that's how it makes sense to me)
Each of those tick store the pending block at the end.

Each of those tick have a fraction of the total bouncer capacity for the whole block.
Once the time has passed for the full block we complete it to the maximum bouncer capacity, and convert the pending block to a full block (ie. do commitment computations)

crates/client/mempool/src/block_production.rs
cchudant: Okay, careful here - I really don't want a lock to be held while running a user supplied function in the mempool. So, mempool stuff returns / accept slices and vecs and not Iterator instances.
For blockifier yea, i've tried my best to remove most of those useless copies but once you get to actually interface with blockifier it becomes a mess and they always need new vecs..

crates/client/mempool/src/block_production.rs
cchudant: I think I don't bubble up anything here right now but I might want to in the future
Could be Result<!, anyhow::Error> true but that doesnt really matter

I think all my services have that `Result<(), anyhow::Error>` type in the codebase

crates/client/mempool/src/block_production.rs
cchudant: Yeah, i've been wondering about that. I'll see the kind of errors I actually see before choosing either way.

crates/client/exec/src/blockifier_state_adapter.rs
cchudant: The State trait is unused at the moment only the StateRead one is used in blockifier
> Oh, you mean the blockifier just creates a state diff on the CachedState and then we need to manually take the state diff and write it to our DB?

Yes but we already have the logic for that from the sync crate, that's the refactor/plumbering i've been doing

crates/client/exec/src/block_context.rs
cchudant: True

crates/client/exec/src/blockifier_state_adapter.rs
cchudant: CachedState is already doing that

crates/client/exec/src/blockifier_state_adapter.rs
cchudant: There is also an error here with us returning the 10 last blocks here

crates/client/exec/src/blockifier_state_adapter.rs
cchudant: ah, true

crates/client/mempool/src/block_production.rs
cchudant: This is not yet commented I think we shouldn't have put the PR in ready yet
I am moving things around quite a lot around here

apoorvsadana: I might be repeating points from my comments above but 

I think we can check for filling blocks instead of a block time tick (we can have a very large one if needed but I assume that we will reach block capacity before that limit. this stems from our discussion in Brussels that we need longer time blocks to reduce # of blocks to reduce proving costs). We would need our block times to be dynamic based on the traffic.

> Each of those tick have a fraction of the total bouncer capacity for the whole block.

Why do we need this? If there were a 100 txs in 2 seconds that will fill the block, we are fine to do that right?

cchudant: > Why do we need this? If there were a 100 txs in 2 seconds that will fill the block, we are fine to do that right?

Then the blockchain isn't bounded in size, correct?

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,332 @@
...
+        let config_bouncer = &self.backend.chain_config().bouncer_config.block_max_capacity;
+
+        if finish_block {
+            // Full bouncer capacity
+            self.executor.bouncer.bouncer_config.block_max_capacity = config_bouncer.clone();
+        } else {
+            let current_pending_tick = self.current_pending_tick;
+            self.current_pending_tick += 1;
+
+            let n_pending_ticks_per_block = self.backend.chain_config().block_time.as_millis()
```
EvolveArt: this shouldn't be computed here

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        // Execute the transactions.
+        let all_results = self.executor.execute_txs(&blockifier_txs);
+
+        // Split the `txs_to_process` vec into two iterators.
+        let mut to_process_iter = txs_to_process.into_iter();
+        // This iterator will consume the first part of `to_process_iter`.
+        let consumed_txs_to_process = to_process_iter.by_ref().take(all_results.len());
+
+        let excuted_txs: Vec<_> = consumed_txs_to_process.collect();
```
EvolveArt: typo ?

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,332 @@
...
+                .get_compiled_contract_class(*class_hash)?;
+            Ok((*class_hash, contract_class.get_visited_segments(class_visited_pcs)?))
+        })
+        .collect::<TransactionExecutorResult<_>>()?;
+
+    Ok(visited_segments)
+}
+
+fn finalize_execution_state<S: StateReader>(
+    executed_txs: &[MempoolTransaction],
```
EvolveArt: unused?

crates/client/mempool/src/block_production.rs
```diff
@@ -0,0 +1,332 @@
...
+                    }
+
+                    self.block.inner.transactions.push(mempool_tx.tx.into());
+                    self.block.inner.receipts.push(execution_info.transaction_receipt.into());
+
+                    Some(execution_info.transaction_receipt)
+                }
+                Err(err) => {
+                    log::error!("Unsuccessful execution of transaction {:?}: {err:#}", mempool_tx.tx_hash());
+                    None
```
EvolveArt: we should do stuff here, it's not because it failed that it shouldn't be stored, especially if it was an actual reverted tx and not a `StateError` or `BlockFull`

crates/client/mempool/src/inner.rs
EvolveArt: but it doesn't read, it actually *force* insert

crates/client/mempool/src/inner.rs
cchudant: sure, i wanted to be extra-verbose here in this file since it's more critical

crates/client/mempool/src/inner.rs
cchudant: ah, it's re_add not read_d
the block production pops transactions as chunks, so that we can support optimistic parallelization
but there is the case where the block gets full before the end of the chunk, and in this case you need to re-add transactions to the mempool, with a special `force` mode to handle nonce collision.

i'll change the function name and add a comment about that

crates/client/mempool/src/lib.rs
cchudant: Yes, I believe this is what blockifier expects us to do
The predicate for validation skipping in is really hard coded to be an invoke after a deploy_account in blockifier

crates/client/mempool/src/lib.rs
cchudant: I think the field is actually gone in newer blockifier or totally different

crates/client/exec/src/execution.rs
cchudant: Yes, it'll move there

EvolveArt (approved):

cchudant:
closes #189

apoorvsadana (approved):

Issue #205, opened 2024-07-16
feat(l1): Forward L1 gas price data to the block header


antiyro:
### Is there an existing issue?

- [X] I have searched the existing issues

### Motivation

We need to fetch the real time L1 gas price from the Layer 1 and forward it to the block header.

### Request

The request is to implement a new module within the Deoxys project that retrieves the L1 gas price using the new crate `l1` as defined in issue [#204](https://github.com/KasarLabs/deoxys/issues/204). The module should be located at `crates/client/{l1/eth}` and named `gas_price.rs`. This module must then be integrated into the block creation process as outlined in issue [#189](https://github.com/KasarLabs/deoxys/issues/189). Specifically, the L1 gas price needs to be forwarded to the block header during the sequencing phase.

### Solution

1. **Create the Module:**
   - Develop a new Rust module named `gas_price.rs` within the directory `crates/client/{l1/eth}`.
   - This module will utilize the `l1` client to fetch the current gas price from the L1 network (using alloy rs getter probably).

2. **Integrate with Block Creation Process:**
   - The best way is probably to keep track of it every x seconds and let the block production process call the value.
   - Ensure the gas price is correctly assigned to the `l1_gas_price` field in the `Header` struct, which is defined in the [header.rs file](https://github.com/KasarLabs/deoxys/blob/main/crates/primitives/block/src/header.rs).

The `GasPrices` struct should reflect the following fields (please adapt in consequence):

```rust
#[derive(Clone, Debug, Default, serde::Serialize, serde::Deserialize)]
     pub struct GasPrices {
         pub eth_l1_gas_price: u128,
         pub strk_l1_gas_price: u128,
         pub eth_l1_data_gas_price: u128,
         pub strk_l1_data_gas_price: u128,
     }
```

3. **Testing and Validation:**
   - Write unit and integration tests to ensure the module retrieves the correct gas prices.
   - Validate that the gas prices are accurately incorporated into the block headers during the sequencing process.

### Are you willing to help with this request?

Yes!

Issue #204, opened 2024-07-16
feat(l1): Implement l1<>l2 messaging


antiyro:
### Is there an existing issue?

- [X] I have searched the existing issues

### Motivation

We need to implement the reading of messages from L1 in Madara. For more context about L1-L2 messaging, I recommend visiting this [link](https://docs.starknet.io/architecture-and-concepts/network-architecture/messaging-mechanism/). This implementation is relatively straightforward and does not require significant changes, as it complements the L2 state verification mechanism via L1 already implemented in Deoxys. The principle is simple: for L1 verification, we continuously subscribe to the `logStateUpdate` event from the core contract to receive the latest verified block number, block hash, and state root. The L1 messaging works similarly by subscribing to the `logMessageToL2` event and aggregating the received data into an L1 handler transaction format to the transaction pool.

### Request

This implementation involves a slight refactor of how we initialize an Ethereum client in Deoxys. Currently, it is a simple module `l1.rs` in `client/sync` [link](https://github.com/KasarLabs/deoxys/blob/main/crates/client/sync/src/l1.rs). We would like to have a dedicated `l1` crate in `client/` [link](https://github.com/KasarLabs/deoxys/blob/main/crates/client/) to facilitate its multiple uses across the codebase. This crate will also be used to retrieve the L1 gas price. The L1<>L2 messaging implementation is already in Madara. The initial objective is to set up a proper L1 client on the client side and then implement the necessary modules (L1<>L2 messaging, gas price, `logStateUpdate`).

### Solution

1. **Set Up a Proper L1 Client:**
   - Create a dedicated `l1` crate in the `client/` directory.
   - Ensure this crate can be easily dispatched for multiple uses across the codebase.

2. **Implement Necessary Modules:**
   - L1<>L2 messaging module: Subscribe to the `logMessageToL2` event and aggregate data into the L1 handler transaction format.
   - Gas price module: Retrieve the L1 gas price (dedicated pr will be linked here).
   - `logStateUpdate` module: Subscribe to the `logStateUpdate` event to continuously receive the latest verified `block number`, `block hash`, and `state root`.

### Are you willing to help with this request?

Yes!

antiyro:
I would like to talk about the possibility of refactoring this using [alloy.rs](http://alloy.rs/) instead of [ether.rs](http://ether.rs/) to ensure an up to date library usage and avoid upcoming refactos.

antiyro:
[...]

PR #208, opened 2024-07-18
Added L1 crate

Mohiiit:
# Pull Request type

- Feature

Pre-req for: #204 #205 

## What is the new behavior?

- It creates the eth client from rpc url along with wallet using the private key provided

No


crates/client/sync/src/eth/src/client.rs
apoorvsadana, line 1: can we delete this?

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,143 @@
...
+/// anvil -b 5 --config-out $BUILD_DIR/anvil.json
+/// PRE_PRIVATE=$(jq -r '.private_keys[0]' $BUILD_DIR/anvil.json)
+pub const DEFAULT_PRIVATE_KEY: &str = "ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct EthereumClientConfig {
+    #[serde(default)]
+    pub provider: EthereumProviderConfig,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub wallet: Option<EthereumWalletConfig>,
```
apoorvsadana: we can remove the wallet right? we won't be sending any txs from the sequencer

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,143 @@
...
+//! in order to share the same configuration file.
+//!
+//! struct MyServiceConfig {
+//!     pub provider: EthereumProviderConfig,
+//!     pub wallet: Option<EthereumWalletConfig>,
+//!     pub contracts: Option<StarknetContracts>,
+//! }
+//!
+//! Default provider and wallet configurations are set for use with Anvil
+//! - local Ethereum environment.
```
apoorvsadana: I didn't really get this comment. I think it's from Madara. Do you know what this means or do you want to change it?

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,143 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HttpProviderConfig {
+    pub wallet: Option<EthereumWallet>,
+    #[serde(default = "default_rpc_endpoint")]
+    pub rpc_endpoint: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tx_poll_interval_ms: Option<u64>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gas_price_poll_ms: Option<u64>,
```
apoorvsadana: We can remove these fields right?

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct StarknetContracts {
+    pub core_contract: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HttpProviderConfig {
+    pub wallet: Option<EthereumWallet>,
```
apoorvsadana: we can remove this as well?

crates/client/sync/src/eth/src/error.rs
```diff
@@ -0,0 +1,18 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Failed to parse provider URL: {0}")]
+    ProviderUrlParse(#[source] url::ParseError),
```
apoorvsadana: after we remove unnecessary fields, we should remove this as well

cchudant:
@Mohiiit @apoorvsadana 
Hi - At a glance, I have a few things to say about the direction the PR is taking.

Do you want me to start a pre-review or should I wait until the PR is more ready?

apoorvsadana:
Hey @cchudant, I guess there was a file restructure that @antiyro pointed out but feel free to add any comments you've :)

crates/client/sync/src/eth/src/error.rs
```diff
@@ -0,0 +1,12 @@
...
```
cchudant: As a general rule, I'd prefer errors looking like
```rs
    #[error("Failed to parse provider URL: {0:#}")]
    ProviderUrlParse(#[from] url::ParseError),
```

#[from] implies source.

Cargo.toml
```diff
@@ -3,6 +3,7 @@ members = [
   "crates/client/db",
   "crates/client/exec",
   "crates/client/sync",
+  "crates/client/sync/src/eth",
```
cchudant: I'd like it to be in crates/client/eth directly, this is too much crate nesting

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,89 @@
+
```
cchudant: All of this config will need to move in the very near future to the `ChainConfig` struct I'm introducing in #198.
This is fine for now as #198 isnt merged yet but the change will need to be done if it gets merged before this one.

crates/client/sync/src/eth/src/config.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::error::Error;
+
+/// Default Anvil local endpoint
+pub const DEFAULT_RPC_ENDPOINT: &str = "http://127.0.0.1:8545";
+/// Default Anvil chain ID
+pub const DEFAULT_CHAIN_ID: u64 = 31337;
+/// Default private key derived from starting Anvil as follows:
+/// anvil -b 5 --config-out $BUILD_DIR/anvil.json
+/// PRE_PRIVATE=$(jq -r '.private_keys[0]' $BUILD_DIR/anvil.json)
+pub const DEFAULT_PRIVATE_KEY: &str = "ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
```
cchudant: what does that private key correspond to? a wallet?
why would there be a default one?

cchudant (changes requested):
- The l1.rs file from the sync crate needs to be rewritten and moved to a proper L1 watch standalone service.

- Rpc url needs to be a CLI arg with clap, as it already is right now.


- #198 needs ethereum prices - it gets them from a dependency inversion trait that looks like

```rs
/// This trait enables the block production task to fill in the L1 info.
/// Gas prices and DA mode 
pub trait L1DataProvider {
    /// Get L1 data gas prices. This needs an oracle for STRK prices.
    fn get_gas_prices(&self) -> GasPrices;
    /// Get the DA mode for L1 Ethereum. This can be either Blob or Calldata, whichever is cheaper at the moment.
    fn get_da_mode(&self) -> L1DataAvailabilityMode;

    // TODO
    // fn get_pending_l1_handler_txs(&self) -> Vec<L1HandlerTransaction>
}
```

I'd like you to define that trait in your eth crate for now, and implement it (no need to rebase on #198 -- keep the pr on main)
I think you can't really get STRK prices though, leave that blank for now?

cchudant:
I don't think we need a wallet - we only need to connect to an rpc endpoint and read stuff

Mohiiit:
A few changes that would be helpful and I can work on these:

- we can bring the state_update functions from l1.rs to the state_update.rs inside the eth crate along with refactoring it to alloy rs
- in the l1.rs we can keep the sync function which would be responsible for calling the state update worker, l1 gas worker and l1 messaging worker
- we can have a worker trait which can help us in future to add more workers seamlessly
- we can move EthClient from l1.rs to client.rs inside the eth crate and the sync function inside l1.rs can create this using the URL 

cc: @cchudant

cchudant:
@Mohiiit 
Hello :)

>    we can move EthClient from l1.rs to client.rs inside the eth crate and the sync function inside l1.rs can create this using the URL
>    we can bring the state_update functions from l1.rs to the state_update.rs inside the eth crate along with refactoring it to alloy rs

Yes, please do
I want the l1.rs file from the sync crate gone entirely: I want the sync crate to have nothing to do with L1.

For context, the sync crate is in the process of slowly being slimmed down - with sequencing, I am moving all of the commitment computation out of that crate, and with p2p the whole fetching process will need to be remade.
In the very near future the sync crate would only be reduced to L2 feeder gateway fetching (and will probably get renamed).

The L1 watcher should be a whole separate service from `SyncService`, with its own CLI arg struct.

Relevant parts of the code: 
- [node > cli > sync.rs](https://github.com/KasarLabs/deoxys/blob/main/crates/node/src/cli/sync.rs#L18-L24): L1 related CLI params
- [node > service > sync.rs](https://github.com/KasarLabs/deoxys/blob/main/crates/node/src/cli/sync.rs#L18-L24): The current sync service

`node` is the crate that parses CLI args and has the `fn main()` of the app -- it is responsible for starting the services for the whole app

You will need to create copies of these two files for the new L1 service.

- [node > main.rs](https://github.com/KasarLabs/deoxys/blob/main/crates/node/src/main.rs): Where all services are started

>    in the l1.rs we can keep the sync function which would be responsible for calling the state update worker, l1 gas worker and l1 messaging worker
>    we can have a worker trait which can help us in future to add more workers seamlessly

I think this would be too much boilerplate for now.
Let's start something with something very simple and watch for L1 updates in a loop - without any trait
Kind of like what l1.rs is doing right now

Currently, nothing would depend on the new eth crate: we currently store the L1 tip to a meta column in the database, and getting l1 transaction status just checks the validated L1 tip from db.
That design is debatable but let's keep it for now I think.

Once we have that, we can consider adding gas prices to the l1 watcher

crates/client/eth/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+authors.workspace = true
+description = "This crate is responsible to handle l1 communication"
+edition = "2021"
```
cchudant: Please use
```rs
authors.workspace = true
edition.workspace = true
license.workspace = true
repository.workspace = true
version.workspace = true
homepage.workspace = true
```

Can you also fix that in the `Cargo.toml` of the sync crate? seems we missed it :)

crates/client/eth/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+homepage = "https://github.com/KasarLabs/deoxys"
+license = "MIT"
+name = "dc-eth"
+publish = false
+repository = "https://github.com/KasarLabs/deoxys"
+version = "0.1.0"
+
+[features]
+default = []
+m = ["dep:rodio"]
```
cchudant: Unused feature

crates/client/eth/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
```
cchudant, line 14: I think most of those dependencies are not needed for that crate

cchudant (changes requested):

Mohiiit:
@cchudant 

if I understood it correctly what we want is:
```
crates
└── node
    └── src
        ├── cli
        │   └── l1.rs   <!-- which would have all the cli args for the l1 -->
        └── service
            └── l1.rs   <!-- which would for now just work for state_update -->
```

also given we want l1 out of sync crate, that means the logic of implementing state_update etc in future would inside the new eth crate?

and we will be moving the code from l1.rs to client.rs inside the eth crate as well for creating the eth client.

and the logic for state update would go inside the state_update.rs inside the eth crate and that would be used in the service >> l1.rs or shall we directly implement it inside the service >> l1.rs for now and then refactor it again?

cchudant:
> @cchudant
> 
> if I understood it correctly what we want is:
> 
> ```
> crates
> └── node
>     └── src
>         ├── cli
>         │   └── l1.rs   
>         └── service
>             └── l1.rs   
> ```
> 
> also given we want l1 out of sync crate, that means the logic of implementing state_update etc in future would inside the new eth crate?

Correct, but the state update function should be in the eth crate. No other logic in the `node` crate other than startup logic for the service.
You can also make the Service struct in the eth crate directly, as I did with DatabaseService for exemple

> and we will be moving the code from l1.rs to client.rs inside the eth crate as well for creating the eth client.

inside the eth crate, however you structure it
 
> and the logic for state update would go inside the state_update.rs inside the eth crate and that would be used in the service >> l1.rs or shall we directly implement it inside the service >> l1.rs for now and then refactor it again?

logic in eth crate

@Mohiiit

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- feat: added l1 crate
```
tdelabro: is `dc-eth` the l1 crate?

crates/client/eth/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    type Error = Error;
+
+    fn try_from(config: EthereumProviderConfig) -> Result<Self, Self::Error> {
+        match config {
+            EthereumProviderConfig::Http(config) => {
+                let rpc_endpoint =
+                    config.rpc_endpoint.parse().map_err(|e: url::ParseError| Error::ProviderUrlParse(e))?;
+                // let client =
+                //     RpcClient::new_http(rpc_endpoint);
+                // let provider = ProviderBuilder::<_, Ethereum>::new().on_client(client);
```
tdelabro: forgotten comment

crates/client/eth/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub mod config;
+pub mod error;
+
+use alloy::providers::{Provider, ProviderBuilder, ReqwestProvider};
+
+use crate::config::EthereumProviderConfig;
+use crate::error::Error;
+
+pub struct EthereumClient {
+    #[allow(dead_code)]
```
tdelabro: don't allow dead code

crates/client/eth/src/config.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub rpc_endpoint: String,
+}
+
+fn default_rpc_endpoint() -> String {
+    DEFAULT_RPC_ENDPOINT.into()
+}
+
+fn default_chain_id() -> u64 {
+    DEFAULT_CHAIN_ID
+}
```
tdelabro: This is never used

crates/client/eth/src/config.rs
```diff
@@ -0,0 +1,77 @@
...
+impl EthereumProviderConfig {
+    pub fn rpc_endpoint(&self) -> &String {
+        match self {
+            Self::Http(config) => &config.rpc_endpoint,
+        }
+    }
+}
+
+impl StarknetContracts {
+    pub fn core_contract(&self) -> Result<Address, Error> {
```
tdelabro: core_contracts, with a `s` to be coherent with the rest

crates/client/eth/src/config.rs
```diff
@@ -0,0 +1,77 @@
...
+pub const DEFAULT_RPC_ENDPOINT: &str = "http://127.0.0.1:8545";
+/// Default Anvil chain ID
+pub const DEFAULT_CHAIN_ID: u64 = 31337;
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct EthereumClientConfig {
+    #[serde(default)]
+    pub provider: EthereumProviderConfig,
+    #[serde(default)]
+    pub contracts: StarknetContracts,
```
tdelabro: I would replace this field with 
`pub core_contrac_address: Eth::ContractAddress`, and don't add `    #[serde(default)]`.


We don't want at all to use the default string `""` if this is not given by the config. The deserialization should fail instead

crates/client/eth/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+tokio = { workspace = true, features = [
+  "macros",
+  "parking_lot",
+  "test-util",
+  "signal",
+] }
+url = { workspace = true }
+
+[dev-dependencies]
+# test_utils = { path = "./test_utils" }
```
tdelabro: forgotten

crates/client/eth/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+authors.workspace = true
+description = "This crate is responsible to handle l1 communication"
+edition = "2021"
+homepage = "https://github.com/KasarLabs/deoxys"
+license = "MIT"
+name = "dc-eth"
```
tdelabro, line 3: rename it to something more meaningful `dc-eth-core-contract`

crates/client/eth/src/utils.rs
```diff
@@ -0,0 +1 @@
+
```
tdelabro: remove the file?

tdelabro (dismissed):

crates/client/eth/Cargo.toml
antiyro: I prefere `dc-eth` actually since it concerns the eth client itselfs and may bring together other eth acceses than the the core contract

crates/client/eth/src/config.rs
Mohiiit: here you mean `alloy::primitives::Address` or `starknet_core::types::EthAddress`?

CHANGELOG.md
Mohiiit: yes ser

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+pub struct EthereumClient {
+    provider: Arc<ReqwestProvider>,
+    l1_core_address: Address,
+}
+
+sol!(
+    #[allow(missing_docs)]
+    #[sol(rpc)]
+    StarknetCore,
+    "src/abis/starknet_core.json"
```
apoorvsadana, line 47: we should comment where we got this json from

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+use futures::StreamExt;
+use primitive_types::H256;
+use serde_json::Value;
+use starknet_api::hash::StarkFelt;
+use starknet_types_core::felt::Felt;
+use url::Url;
+
+pub struct EthereumClient {
+    provider: Arc<ReqwestProvider>,
+    l1_core_address: Address,
```
apoorvsadana:
```suggestion
    l1_core_contract_address: Address,
```

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+sol!(
+    #[allow(missing_docs)]
+    #[sol(rpc)]
+    StarknetCore,
+    "src/abis/starknet_core.json"
+);
+
+impl EthereumClient {
```
apoorvsadana, line 66: we should test all these functions with integration tests using `anvil`

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+        if let Some(last_log) = logs.last() {
+            let last_block: u64 = last_log.block_number.context("no block number in log")?;
+            Ok(last_block)
+        } else {
+            bail!("no event found")
+        }
+    }
+
+    /// Get the last Starknet block number verified on L1
+    pub async fn get_last_block_number(&self) -> anyhow::Result<u64> {
```
apoorvsadana:
```suggestion
    pub async fn get_last_verified_block_number(&self) -> anyhow::Result<u64> {
```

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+        let block_number = contract.stateBlockNumber().call().await.unwrap();
+        let last_block_number: u64 = (block_number._0).as_u64();
+        Ok(last_block_number)
+    }
+
+    /// Get the last Starknet state root verified on L1
+    pub async fn get_last_state_root(&self) -> anyhow::Result<StarkFelt> {
+        let client = self.provider.clone();
+        let address: Address = self.l1_core_address;
+        let contract = StarknetCore::new(address, client);
```
apoorvsadana: should we just store `StarknetCore` instead of the address inside `EthereumClient`

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+        let state_root = contract.stateRoot().call().await.unwrap();
```
apoorvsadana: we should avoid unwraps everywhere

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+        u256_to_starkfelt(state_root._0)
+    }
+
+    /// Get the last Starknet block hash verified on L1
+    pub async fn get_last_block_hash(&self) -> anyhow::Result<StarkFelt> {
```
apoorvsadana:
```suggestion
    pub async fn get_last_verified_block_hash(&self) -> anyhow::Result<StarkFelt> {
```

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        Ok(L1StateUpdate { global_root, block_number, block_hash })
+    }
+
+    /// Subscribes to the LogStateUpdate event from the Starknet core contract and store latest
+    /// verified state
+    pub async fn listen_and_update_state(
+        &self,
+        backend: &DeoxysBackend,
+        start_block: u64,
```
apoorvsadana: `start_block` isn't being used

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,172 @@
...
+        block_metrics: BlockMetrics,
+        chain_id: Felt,
+    ) -> anyhow::Result<()> {
+        let client = self.provider.clone();
+        let address: Address = self.l1_core_address;
+
+        let latest_block: u64 =
```
apoorvsadana: never used

apoorvsadana (commented):
I am not sure if the comments are in the scope of this PR as currently this only migrates ethers-rs to alloy rs. Will do another review when we've completed the other stuff.

crates/client/eth/src/client.rs
Mohiiit: good idea, we won't have to create one everytime

crates/client/eth/src/config.rs
Mohiiit: this config rs might get deleted only, because it's not been used so far

crates/client/eth/src/lib.rs
Mohiiit: fixed

crates/client/eth/Cargo.toml
Mohiiit: fixed

crates/client/eth/Cargo.toml
Mohiiit: fixed

crates/client/eth/Cargo.toml
Mohiiit: fixed

crates/client/eth/src/lib.rs
Mohiiit: not valid, as config won't be used

crates/client/eth/Cargo.toml
Mohiiit: fixed

crates/client/eth/src/utils.rs
Mohiiit: it's in use now

crates/client/sync/src/eth/src/error.rs
Mohiiit: fixed

Cargo.toml
Mohiiit: fixed

crates/client/sync/src/eth/src/config.rs
Mohiiit: we won't be using any private key/wallet

crates/client/eth/src/client.rs
Mohiiit: invalid, as we are using the contract instance directly now

crates/client/eth/src/client.rs
Mohiiit: tests added

crates/client/eth/src/client.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: done

crates/client/eth/src/client.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: fixed

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
 
 ## Next release
-
+- test: tests added for eth client and event subscription
+- refactor: removed l1.rs from the sync crate and refactored it with alloy inside eth crate
+- refactor: eth client (l1 crate) refactored with alloy
 - fix(metrics): removed influx and added l2_state_size data
 - fix: command to start the Madara client
 - refactor: database error unification
+- feat: added l1 crate
```
apoorvsadana: let's move this to the top as well

crates/client/eth/src/abis/simple_storage.json
apoorvsadana, line 1: If this is for tests then let's move this to a tests folder. Let's also add the `.sol` file for this

crates/client/eth/src/abis/starknet_core_new.json
apoorvsadana, line 1: let's replace this with `starknet_core` this as discussed

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+use anyhow::{bail, Context};
+use bitvec::macros::internal::funty::Fundamental;
+use starknet_api::hash::StarkFelt;
+use url::Url;
+
+use crate::client::StarknetCoreContract::StarknetCoreContractInstance;
+use crate::{config::L1StateUpdate, utils::u256_to_starkfelt};
+
+sol!(
+    #[allow(missing_docs)]
```
apoorvsadana: let's add the docs for this contract?

crates/client/sync/src/eth/src/config.rs
Mohiiit: yes removed

crates/client/sync/src/eth/src/config.rs
Mohiiit: yes not required, removed

crates/client/sync/src/eth/src/config.rs
Mohiiit: yes removed

crates/client/sync/src/eth/src/config.rs
Mohiiit: done

crates/client/sync/src/eth/src/error.rs
Mohiiit: this particular error type is required in one of the cases

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+        let block_number = self.provider.get_block_number().await?.as_u64();
+        Ok(block_number)
+    }
+
+    /// Get the block number of the last occurrence of a given event.
+    pub async fn get_last_event_block_number<T: SolEvent>(&self) -> anyhow::Result<u64> {
+        let latest_block: u64 = self.get_latest_block_number().await?;
+
+        let filter = Filter::new()
+            .from_block(latest_block - 6000)
```
apoorvsadana, line 88: there was a comment here before ig, about how we chose the number 6000

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+            .to_block(latest_block)
+            .address(*self.l1_core_contract.address());
+
+        let logs = self.provider.get_logs(&filter).await?;
+
+        let filtered_logs = logs.clone().into_iter().filter_map(|log| log.log_decode::<T>().ok()).collect::<Vec<_>>();
```
apoorvsadana: `clone` looks unnecessary here as we don't use `logs` anywhere after, can you try removing it

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub async fn get_last_verified_block_number(&self) -> anyhow::Result<u64> {
+        let block_number = self.l1_core_contract.stateBlockNumber().call().await?;
+        let last_block_number: u64 = (block_number._0).as_u64();
+        Ok(last_block_number)
+    }
+
+    /// Get the last Starknet state root verified on L1
+    pub async fn get_last_state_root(&self) -> anyhow::Result<StarkFelt> {
+        let state_root = self.l1_core_contract.stateRoot().call().await?;
+        println!("state root is: {:?}", state_root._0.clone());
```
apoorvsadana: we should avoid `println!`. if we need something to be logged for debugging purposes we can use `log::debug!` or `log::trace!`

crates/client/sync/src/eth/src/client.rs
Mohiiit: it's being used now

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    use tokio;
+
+    use once_cell::sync::Lazy;
+
+    static ANVIL: Lazy<AnvilInstance> =
+        Lazy::new(|| Anvil::new().fork("https://eth.merkle.io").fork_block_number(20395662).spawn());
+
+    #[fixture]
+    fn anvil() -> &'static AnvilInstance {
+        &ANVIL
```
apoorvsadana: we don't need to use `Lazy`, we can add `#[once]` on top of the `#[fixture]` afaik and it will be handled automatically

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    #[fixture]
+    fn eth_client(anvil: &'static AnvilInstance) -> EthereumClient {
```
apoorvsadana: by using `#[once]` we should be able to combine the two fixtures into just one

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    use super::*;
+    use alloy::node_bindings::{Anvil, AnvilInstance};
+    use alloy::primitives::{address, U256};
+    use rstest::*;
+    use tokio;
+
+    use once_cell::sync::Lazy;
+
+    static ANVIL: Lazy<AnvilInstance> =
+        Lazy::new(|| Anvil::new().fork("https://eth.merkle.io").fork_block_number(20395662).spawn());
```
apoorvsadana: let's make this a const inside the testing module

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        EthereumClient { provider: Arc::new(provider), l1_core_contract: contract.clone() }
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_get_latest_block_number(eth_client: EthereumClient) {
+        let block_number =
+            eth_client.provider.get_block_number().await.expect("issue while fetching the block number").as_u64();
+        assert_eq!(block_number, 20395662, "provider unable to get the correct block number");
```
apoorvsadana: we should use `20395662` from the constant as mentioned above

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn test_get_last_event_block_number(eth_client: EthereumClient) {
+        let block_number = eth_client
+            .get_last_event_block_number::<StarknetCoreContract::LogStateUpdate>()
+            .await
+            .expect("issue while getting the last block number with given event");
+        assert_eq!(block_number, 20395662, "block number with given event not matching");
```
apoorvsadana: it will be great if we can add etherscan link to the specific event

apoorvsadana: valid for all below test cases as well

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,174 @@
...
+    sol,
+    transports::http::{Client, Http},
+};
+use anyhow::{bail, Context};
+use bitvec::macros::internal::funty::Fundamental;
+use starknet_api::hash::StarkFelt;
+use url::Url;
+
+use crate::client::StarknetCoreContract::StarknetCoreContractInstance;
+use crate::{config::L1StateUpdate, utils::u256_to_starkfelt};
```
apoorvsadana: we should move `L1StateUpdate` to `state_update.rs` and with the `get_initial_state` function

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,150 @@
...
+            .fork("https://eth.merkle.io")
+            .fork_block_number(20395662)
+            .try_spawn()
+            .expect("issue while forking");
+        let rpc_url: Url = anvil.endpoint().parse().expect("issue while parsing");
+        let provider = ProviderBuilder::new().on_http(rpc_url.clone());
+
+        let address = anvil.addresses()[0];
+
+        let contract1 = SimpleStorage::deploy(provider.clone(), "initial value".to_string()).await.unwrap();
```
apoorvsadana:
```suggestion
        let contract = SimpleStorage::deploy(provider.clone(), "initial value".to_string()).await.unwrap();
```

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        let event1 = contract1.event_filter::<SimpleStorage::ValueChanged>();
```
apoorvsadana:
```suggestion
        let event = contract1.event_filter::<SimpleStorage::ValueChanged>();
```

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,150 @@
...
+        let mut stream1 = event1.watch().await.unwrap().into_stream();
```
apoorvsadana:
```suggestion
        let mut stream = event.watch().await.unwrap().into_stream();
```

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    sol!(
+        #[derive(Debug)]
+        #[sol(rpc)]
+        SimpleStorage,
+        "src/abis/simple_storage.json"
+    );
+
+    #[tokio::test]
+    async fn test_event_subscription() {
```
apoorvsadana: I thought we will test `listen_and_update_state` function over here by emitting log state update from the dummy contract. In the current setup, we're just testing for alloy functionality

crates/client/eth/src/utils.rs
```diff
@@ -0,0 +1,41 @@
+use crate::client::StarknetCoreContract;
+use crate::config::L1StateUpdate;
+use alloy::primitives::{I256, U256};
+use anyhow::bail;
+use starknet_api::hash::StarkFelt;
+use starknet_types_core::felt::Felt;
+
+pub const LOG_STATE_UPDTATE_TOPIC: &str = "0xd342ddf7a308dec111745b00315c14b7efb2bdae570a6856e088ed0c65a3576c";
```
apoorvsadana: is this needed now

crates/client/eth/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+pub fn convert_log_state_update(
```
apoorvsadana, line 7: we should add unit tests for these.

crates/client/eth/src/abis/starknet_core_new.json
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: docs means the source?

crates/client/eth/src/utils.rs
Mohiiit: added

crates/client/eth/src/utils.rs
Mohiiit: removed

crates/client/eth/src/state_update.rs
Mohiiit: fixed

crates/client/eth/src/state_update.rs
Mohiiit: fixed

crates/client/eth/src/state_update.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: refactored

crates/client/eth/src/client.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: done

crates/client/eth/src/client.rs
Mohiiit: done

crates/client/eth/src/client.rs
Mohiiit: refactored, actually we do need two `[#once]` because else the anvil instance won't be allocated static memory and it's goes out of stack after the eth_client function, so we need once for both the anvil instance and the eth_client

crates/client/eth/src/client.rs
Mohiiit: done

crates/client/eth/src/client.rs
Mohiiit: I missed it, fixed

crates/client/eth/src/client.rs
Mohiiit: fixed

crates/client/eth/src/client.rs
Mohiiit: added

crates/client/eth/src/state_update.rs
Mohiiit: added

crates/client/eth/src/client.rs
apoorvsadana: I don't what docs this is looking for, I thought it was docs about what this contract does

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,164 @@
...
+    use alloy::primitives::U256;
+    use rstest::*;
+    use tokio;
+
+    const ETH_URL: &str = "https://eth.merkle.io";
+    const L1_BLOCK_NUMBER: u64 = 20395662;
+    const CORE_CONTRACT_ADDRESS: &str = "0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4";
+    const L2_BLOCK_NUMBER: u64 = 662703;
+    const L2_BLOCK_HASH: &str = "563216050958639290223177746678863910249919294431961492885921903486585884664";
+    const L2_STATE_ROOT: &str = "1456190284387746219409791261254265303744585499659352223397867295223408682130";
```
apoorvsadana, line 138: can you link to the state update call in the block as well so it's easier to check the root, l2 block number and hash

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    Ok(L1StateUpdate { global_root, block_number, block_hash })
+}
+
+/// Subscribes to the LogStateUpdate event from the Starknet core contract and store latest
+/// verified state
+pub async fn listen_and_update_state(
+    eth_client: EthereumClient,
+    backend: &DeoxysBackend,
+    block_metrics: BlockMetrics,
```
apoorvsadana: I think we can use reference of `BlockMetrics` everywhere, it avoids unnecessary clones

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,261 @@
...
+    ///
+    /// This test performs the following steps:
+    /// 1. Sets up a mock Ethereum environment using Anvil
+    /// 2. Initializes necessary services (Database, Metrics)
+    /// 3. Deploys a dummy contract and sets up an Ethereum client
+    /// 4. Starts listening for state updates
+    /// 5. Fires an event from the dummy contract
+    /// 6. Waits for event processing and verifies the block number
+    #[tokio::test]
+    async fn test_event_subscription() {
```
apoorvsadana:
```suggestion
    async fn test_listen_and_update_state() {
```

apoorvsadana: in fact we decided on a naming convention in the orchestrator which i think we should reuse here. essentially, test name would be `<function_to_be_tested>_<condition>_<expected_outcome>`. examples from orchestrator

- process_job_with_small_blob_size_fails
- process_job_with_pending_block_fails
- process_job_works

do you think we can rename our tests based on this?

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,261 @@
...
+        // Set up chain info
+        let chain_info =
+            ChainInfo { chain_id: Felt::from_dec_str("11153111").unwrap(), chain_name: "testing".to_string() };
+
+        // Set up database paths
+        let current_dir = env::current_dir().expect("Failed to get current directory");
```
apoorvsadana: can you check the temp dir thing if it can be used here?

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,261 @@
...
+                    fire_event(&contract).await;
+                }
+            }
+        });
+
+        // Wait for event processing
+        tokio::time::sleep(Duration::from_secs(EVENT_PROCESSING_TIME)).await;
+
+        // Signal to stop event firing
+        stop_flag.store(true, Ordering::Relaxed);
```
apoorvsadana: I think we can synchronously wait for an event and just check the DB straight away

crates/client/eth/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+        // 0x1b48eb57e000 is the hex value of the given random
+        let starkfelt = StarkFelt::from(random);
+        let trimmed = trim_hash(&starkfelt.to_felt());
+        assert_eq!(trimmed, "0x1b48eb...57e000");
+
+        let random: u128 = 12345678123456789;
+        // 0x2bdc542f0f5915 is the hex value of the given random
+        let starkfelt = StarkFelt::from(random);
+        let trimmed = trim_hash(&starkfelt.to_felt());
+        assert_eq!(trimmed, "0x2bdc54...0f5915");
```
apoorvsadana: should be two different #[case]

crates/client/eth/src/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+        assert_eq!(expected, result, "log state update conversion not working");
+    }
+
+    #[test]
+    fn test_u256_to_starkfelt() {
+        let random: u128 = 12345678123456789;
+        let starkfelt = StarkFelt::from(random);
+
+        let random_u256: U256 = U256::from_str_radix(random.to_string().as_str(), 10).unwrap();
+        let result = u256_to_starkfelt(random_u256).unwrap();
```
apoorvsadana: i am not sure if this test is technically testing something because we are using sort of the same code that is used inside the function itself (`StarkFelt`)

apoorvsadana: maybe a better case would be to just hardcode a few responses?

crates/client/eth/src/client.rs
Mohiiit: added

crates/client/eth/src/state_update.rs
Mohiiit: done

crates/client/eth/src/state_update.rs
Mohiiit: refactored

crates/client/eth/src/state_update.rs
Mohiiit: added

crates/client/eth/src/state_update.rs
Mohiiit: yes removed this

crates/client/eth/src/utils.rs
Mohiiit: done

crates/client/eth/src/utils.rs
Mohiiit: refactored

crates/client/eth/src/abis/simple_storage.json
apoorvsadana: this is pending. do we need simple storage now?

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,212 @@
...
+            .try_spawn()
+            .expect("Failed to spawn Anvil instance");
+        let rpc_url: Url = anvil.endpoint().parse().expect("Failed to parse RPC URL");
+
+        // Set up Ethereum provider and contracts
+        let provider = ProviderBuilder::new().on_http(rpc_url.clone());
+        let contract = DummyContract::deploy(provider.clone()).await.unwrap();
+        let core_contract = StarknetCoreContract::new(contract.address().clone(), provider.clone());
+
+        let eth_client = EthereumClient { provider: Arc::new(provider), l1_core_contract: core_contract.clone() };
```
apoorvsadana: actually, can we not import the fixture from eth client directly?

crates/client/eth/src/abis/simple_storage.json
Mohiiit: nope not needed

crates/client/eth/src/state_update.rs
Mohiiit: anvil fixture added

crates/client/eth/src/client.rs
```diff
@@ -0,0 +1,164 @@
...
+pub mod eth_client_getter_test {
+    use super::*;
+    use alloy::node_bindings::{Anvil, AnvilInstance};
+    use alloy::primitives::U256;
+    use rstest::*;
+    use tokio;
+
+    // https://etherscan.io/tx/0xcadb202495cd8adba0d9b382caff907abf755cd42633d23c4988f875f2995d81#eventlog
+    // The txn we are referring to it is here ^
+    const ETH_URL: &str = "https://eth.merkle.io";
```
cchudant: I don't know what to feel about putting an rpc provider in our tests
An env var might be better what do you think?

crates/client/eth/src/state_update.rs
```diff
@@ -0,0 +1,207 @@
...
+pub fn update_l1(
+    backend: &DeoxysBackend,
+    state_update: L1StateUpdate,
+    block_metrics: &BlockMetrics,
+    chain_id: Felt,
+) -> anyhow::Result<()> {
+    // This is a provisory check to avoid updating the state with an L1StateUpdate that should not have been detected
+    //
+    // TODO: Remove this check when the L1StateUpdate is properly verified
+    if state_update.block_number > 500000u64 || chain_id == TEST_CHAIN_ID {
```
cchudant: could you replace this with `if state_update.block_number > 500000u64 && chain_id == MAIN_CHAIN_ID` so that this doesn't get triggered in app chains

crates/client/metrics/src/block_metrics.rs
```diff
@@ -1,4 +1,6 @@
-use dc_metrics::{Gauge, MetricsRegistry, PrometheusError, F64};
+// use dc_metrics::{Gauge, MetricsRegistry, PrometheusError, F64};
+
+use crate::{Gauge, MetricsRegistry, PrometheusError, F64};
 
 #[derive(Clone, Debug)]
 pub struct BlockMetrics {
```
cchudant: I would prefer if you kept BlockMetrics where it was in sync, but move the l1 stuff to a new struct L1Metrics inside of the eth crate

i think each service having its own metrics struct for the metrics it expose is better

Cargo.toml
```diff
@@ -99,6 +102,8 @@ starknet_api = "=0.10.0"
 cairo-lang-starknet-classes = "2.6.0"
 cairo-lang-utils = "2.6.0"
 
+alloy = { version = "0.2.0", features = ["full", "json"] }
```
cchudant: the "full" flag here is probably overkill if we only want to support just sending rpc requests

cchudant (changes requested):
Very cool :) looks almost perfect

crates/client/metrics/src/block_metrics.rs
Mohiiit: it was creating a circular dependency, `sync` crate is calling the `sync` function from the `eth/l1` crate and `eth` crate needed the `BlockMetrics`.

crates/client/eth/src/client.rs
Mohiiit: yeah sounds good, one more thing that needs to added is secret in the github for the workflow. Because tests depends on it.

Mohiiit: added `.env.test` for this

crates/client/eth/src/state_update.rs
Mohiiit: fixed

Cargo.toml
Mohiiit: resolved

crates/client/metrics/src/block_metrics.rs
Mohiiit: @cchudant added l1BlockMetrics and added it in the client.rs inside the eth crate as of now, it would also have the l1_block_price, for the price from the l1

crates/client/eth/src/client.rs
```diff
@@ -27,6 +50,10 @@ sol!(
 pub struct EthereumClient {
     pub provider: Arc<ReqwestProvider>,
     pub l1_core_contract: StarknetCoreContractInstance<Http<Client>, RootProvider<Http<Client>>>,
+    // l1 metric :
+    // l1 block number
+    // l1 gas price in wei
+    // l1 gas price in strk (from oracle wei <-> strk)
```
apoorvsadana: we need to `L1BlockMetrics` here right?

crates/client/sync/src/metrics/block_metrics.rs
```diff
@@ -30,8 +28,8 @@ impl BlockMetrics {
                 .register(Gauge::new("deoxys_l2_latest_sync_time", "Gauge for deoxys L2 latest sync time")?)?,
             l2_state_size: registry
                 .register(Gauge::new("deoxys_l2_state_size", "Gauge for node storage usage in GB")?)?,
-            l1_block_number: registry
-                .register(Gauge::new("deoxys_l1_block_number", "Gauge for deoxys L1 block number")?)?,
+            // l1_block_number: registry
+            //     .register(Gauge::new("deoxys_l1_block_number", "Gauge for deoxys L1 block number")?)?,
```
apoorvsadana: we can delete this

crates/client/sync/src/metrics/block_metrics.rs
```diff
@@ -13,7 +11,7 @@ pub struct BlockMetrics {
     pub transaction_count: Gauge<F64>,
     pub event_count: Gauge<F64>,
     // L1 network metrics
-    pub l1_block_number: Gauge<F64>,
+    // gas price is also define in sync/block_metrics.rs but this would be the gas used in the block and it's price
```
apoorvsadana: you mean l1 metrics here?

Mohiiit: yeah resolved

crates/client/eth/src/client.rs
Mohiiit: resolved

crates/client/sync/src/metrics/block_metrics.rs
Mohiiit: resolved

cchudant (approved):

apoorvsadana (approved):

apoorvsadana (approved):

PR #213, opened 2024-07-23
add primitive StateUpdate, implement commitment for v0.13.2

### LINKED ISSUE #191: feat(0.13.2): Add support for upcoming Starknet version 0.13.2

### Motivation

The purpose of this issue is to track all issues related to the migration from the current version [v0.13.1.1](https://docs.starknet.io/starknet-versions/version-notes/#version0.13.1.1) to version 0.13.2. Learn more about v0.13.2 with the pre-release notes defined [here](https://community.starknet.io/t/starknet-v0-13-2-pre-release-notes/114223).

This version focuses on performance improvements. In addition, we are making some necessary adjustments to prepare for the transition to a Starknet p2p network between nodes that will occur in a later version.

### Request

Among the changes in this version, we primarily (but not exclusively) note the following updates (check the associated issue if opened for more information):

- [ ] Implement optimisitic parallelization to the sequencer level ([more infos](https://community.starknet.io/t/optimistic-parallelization-revived/114121#optimistic-parallelization-revived-1))
- [x] Refactor [`LogStateUpdate`](https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/starknet/solidity/Starknet.sol#L36) to deserialize every sequence of blocks rather than for every block
- [x] Implement new block hash definition ([starknet-api](https://github.com/starkware-libs/starknet-api/tree/15219992b54b170dcf80fa5cceab56847ea2f7f0/src/block_hash) for reference)
- [ ] Assign all FGW request/response changes (new version, new fields)
- [x] Assign cairo vm changes 
- [ ] Assign blockifier changes 

⚠️ This version does not have any JSON-RPC changes. V0.8.0 will be released alongside Starknet v0.13.3, at which point all support for versions < 0.7.0 will stop.

### Solution

_No response_

Yes!

jbcaron:
# Pull Request type

- Feature

Resolves: #191 

## What is the new behavior?

- improvement of error handling in `convert_and_verify_block`
- new primitive StateUpdate
- management of different versions of commitment and hash
- update `cairo-lang-starknet` V2 for class compilation
- simplification and improvement of root state update
- remove csd conversion

YES: db

## Other information

- ✅ synchronization test of the entire network integration state


crates/client/db/src/lib.rs
```diff
@@ -467,8 +467,6 @@ impl DeoxysBackend {
 }
 
 pub mod bonsai_identifier {
-    pub const CONTRACT: &[u8] = "0xcontract".as_bytes();
-    pub const CLASS: &[u8] = "0xclass".as_bytes();
-    pub const TRANSACTION: &[u8] = "0xtransaction".as_bytes();
-    pub const EVENT: &[u8] = "0xevent".as_bytes();
+    pub const CONTRACT: &[u8] = b"0xcontract";
+    pub const CLASS: &[u8] = b"0xclass";
```
tdelabro, line 471: Why are those prefixed with `0x`?
It seems arbitrary and doesn't make much sense as what follows is not an hex string. If it is for some important reason can you add it as a comment

crates/primitives/transactions/src/compute_hash.rs
```diff
@@ -116,56 +132,66 @@ impl InvokeTransactionV3 {
...
-                calldata_hash,
-                Felt::ZERO, // Fees are set to zero on L1 Handler txs
-                chain_id,
-                nonce,
-            ])
+            self.compute_hash_inner(chain_id, offset_version)
         }
     }
+
+    pub fn compute_hash_pre_v0_7(&self, chain_id: Felt) -> Felt {
```
tdelabro, line 143: No offset for legacy?

crates/primitives/transactions/src/lib.rs
```diff
@@ -17,7 +18,7 @@ const SIMULATE_TX_VERSION_OFFSET: Felt =
 /// See `https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transactions/` for more details.
 
 pub const LEGACY_BLOCK_NUMBER: u64 = 1470;
-pub const LEGACY_L1_HANDLER_BLOCK: u64 = 854;
+pub const V0_7_BLOCK_NUMBER: u64 = 833;
 
 //  b"SN_MAIN" == 0x534e5f4d41494e
 pub const MAIN_CHAIN_ID: Felt = Felt::from_hex_unchecked("0x534e5f4d41494e");
```
tdelabro: what is the `pub const INTE_CHAIN_ID: Felt = Felt::from_hex_unchecked("0x0");` value for?
What does INTE even mean?

crates/primitives/state_update/src/lib.rs
```diff
@@ -0,0 +1,173 @@
+mod into_starknet_core;
```
tdelabro, line 1: Can you explain the reason we cannot use `starknet_core` types directly?
Why do we have to duplicate the whole thing?

crates/client/sync/src/commitments/mod.rs
```diff
@@ -144,33 +53,67 @@ pub fn calculate_state_root(contracts_trie_root: Felt, classes_trie_root: Felt)
...
-        },
+        || class_trie_root(backend, declared_classes, block_number).expect("Failed to compute class root"),
     );
+
     calculate_state_root(contract_trie_root, class_trie_root)
 }
 
+/// Compute the root hash of a list of values.
+// The `HashMapDb` can't fail, so we can safely unwrap the results.
+pub fn compute_root<H>(values: &[Felt]) -> Felt
```
tdelabro, line 80: `values` could be an iterator.
Would avoid needless collect from the callers

tdelabro (changes requested):

crates/client/db/src/lib.rs
jbcaron: @cchudant is there a reason?

crates/primitives/transactions/src/compute_hash.rs
jbcaron: There should not have been offset before v0.11.1 (legacy block)
In any case we cannot execute before v0.13 with blockifier, so no simulate

crates/primitives/transactions/src/lib.rs
jbcaron: this is an error, this is not used but is supposed to be the integration chain_id which should be b"SN_INTEGRATION_SEPOLIA"

crates/primitives/state_update/src/lib.rs
jbcaron: We use our own primitives for all data in the database because we use bincode to serialize the data. The serialization methods of external crates do not allow us to serialize/deserialize using bincode. Using our own primitives also enables us to implement all our methods without needing to create traits or conversion implementations.

crates/client/sync/src/commitments/mod.rs
jbcaron: You are right that using an iterator could avoid needless collections.
In most callers, we use `par_iter` from rayon to benefit from parallel processing.
However, rayon produces parallel iterators (ParallelIterator), which are not directly compatible with standard iterators (Iterator).
Additionally, methods like enumerate are not directly available on ParallelIterator, further complicating the direct use of iterators in this context.

antiyro (approved):

crates/primitives/block/src/header.rs
```diff
@@ -214,6 +211,169 @@ impl Header {
...
+                Felt::from_bytes_be_slice(b"STARKNET_BLOCK_HASH0"),
+                Felt::from(self.block_number),
+                self.global_state_root,
+                self.sequencer_address,
+                Felt::from(self.block_timestamp),
+                concat_counts(self.transaction_count, self.event_count, self.state_diff_length, self.l1_da_mode),
+                self.state_diff_commitment,
+                self.transaction_commitment,
+                self.event_commitment,
+                self.receip_commitment,
```
cchudant: receipt

crates/client/sync/src/commitments/mod.rs
```diff
@@ -144,33 +53,67 @@ pub fn calculate_state_root(contracts_trie_root: Felt, classes_trie_root: Felt)
...
+where
+    H: StarkHash + Send + Sync,
+{
+    // the identifier cannot be empty otherwise it can create a bug in the bonsai
```
cchudant: mark that as TODO(bonsai-trie) please

cchudant (changes requested):

PR #214, opened 2024-07-29
test: :white_check_mark: add tests crate `db`

jbcaron:
## Pull Request type

- Testing

## What is the current behavior?

Resolves: #NA

## What is the new behavior?

- Trait Implementations:
  - Implemented the Default trait on primitives to allow for the creation of default instances during testing.
  - Implemented Eq and PartialEq traits on primitives to enable equality checks within tests.
- Unit Tests:
  - Block Header Storage and Retrieval:
    - Added tests to ensure that block headers are correctly stored and can be accurately retrieved.
  - Transaction Storage and Retrieval:
    - Added tests to validate the storage of transactions and their subsequent retrieval, ensuring that the data remains consistent.
  - Receipts Storage and Retrieval:
    - Introduced tests for storing and retrieving receipts, confirming that all relevant details are preserved.
  - Transaction Search:
    - Implemented tests to verify that the transaction search functionality works as expected, returning the correct transaction data based on provided search criteria.
- Pending Block Management:
  - Added tests to cover the management of pending blocks, ensuring that pending block is handled properly in various scenarios, such as during storage and retrieval processes.

## Does this introduce a breaking change?

No

PR #215, opened 2024-08-05
Rpc tests

cchudant:
# Pull Request type

- Testing

No

## Other information

I will add test helpers along the way.


cchudant:
done:
- block_hash_and_number
- all the get block rpcs
- get_transaction_{by_block_id_and_index,by_hash,receipt,status}
- get_state_update
- get_class_hash_at / get_nonce / get_storage_at 

next:
- get_class / get_class_at
- syncing
- get_events
- call / estimate_fee / estimate_message_fee and the trace transaction rpcs (execution rpcs)

I also migrated node and rpcs to using the new ChainConfig from #198

ChainConfig will get its own crate soon

crates/client/db/src/lib.rs
```diff
@@ -66,33 +61,15 @@ pub(crate) async fn open_rocksdb(
 
     opts.set_env(&env);
 
-    let backup_hande = if let Some(backup_dir) = backup_dir {
-        let (restored_cb_sender, restored_cb_recv) = oneshot::channel();
-
-        let (sender, receiver) = mpsc::channel(1);
```
jbcaron, line 72: what is the mpsc channel for?

crates/client/rpc/src/methods/read/get_block_with_receipts.rs
```diff
@@ -63,3 +64,95 @@ pub fn get_block_with_receipts(
...
+    use starknet_core::types::{BlockTag, Felt, L1DataAvailabilityMode, ResourcePrice};
+
+    #[rstest]
+    fn test_get_block_with_receipts() {
+        let _ = env_logger::builder().is_test(true).try_init();
+        let (backend, rpc) = open_testing();
+        let SampleChain1 { block_hashes, expected_txs, expected_receipts, .. } = make_sample_chain_1(&backend);
+
+        // Block 0
+        let res = MaybePendingBlockWithReceipts::Block(BlockWithReceipts {
```
jbcaron, line 84: #214  add default traits on primitives (`..Default::default()`)

crates/node/src/cli/mod.rs
```diff
@@ -66,14 +73,43 @@ impl RunCmd {
         }
         self.name.as_ref().unwrap()
     }
+}
+
+/// Starknet network types.
+#[derive(Debug, Clone, Copy, clap::ValueEnum, PartialEq)]
+pub enum NetworkType {
```
jbcaron, line 80: you will need to add a custom network with a user-defined chainid and feeder url to be consistent with the ChainConfig type

jbcaron (approved):
- add error testing on `get_BLOCK...` methods
- add test for `get_storage`, uninitialized key on existing contract should return Felt::ZERO

crates/client/db/src/lib.rs
cchudant: backup restore task, i moved elsewhere it so that DeoxysBackend::open_for_testing is not async for convenience

when backups are enabled, a thread is spawned that owns the rocksdb BackupEngine (it is not thread safe) and it receives backup requests using a mpsc channel
There is also another oneshot channel involved: when restoring the db at startup, we want to wait for the backupengine to finish restoration before returning from `open()`

I will add a comment about that

crates/node/src/cli/mod.rs
cchudant: yepyep coming in another pr

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: pending contract storage not stored properly
```
apoorvsadana, line 6: I guess we can add more things here, wdyt?

crates/client/db/src/contract_db.rs
```diff
@@ -46,6 +48,8 @@ impl DeoxysBackend {
                 // Get pending or fallback to latest block_n
                 let col = self.db.get_column(pending_col);
                 // todo: smallint here to avoid alloc
+
+                // NB: pending has keys in bincode, not bytes
```
apoorvsadana: what's NB 😅?

crates/client/db/src/lib.rs
```diff
@@ -66,33 +61,15 @@ pub(crate) async fn open_rocksdb(
...
-        log::debug!("blocking on db restoration");
-        restored_cb_recv.await.context("Restoring database")?;
-        log::debug!("done blocking on db restoration");
-
-        Some(sender)
-    } else {
-        None
-    };
-
+    #[cfg(not(feature = "testing"))]
```
apoorvsadana: is this needed?

crates/client/db/src/lib.rs
```diff
@@ -355,6 +334,7 @@ struct BackupRequest {
 
 impl Drop for DeoxysBackend {
     fn drop(&mut self) {
+        #[cfg(not(feature = "testing"))]
```
apoorvsadana: is this needed?

crates/client/rpc/src/methods/read/block_hash_and_number.rs
```diff
@@ -20,3 +20,56 @@ pub fn block_hash_and_number(starknet: &Starknet) -> StarknetRpcResult<BlockHash
...
+mod tests {
+    use super::*;
+    use crate::test_utils::open_testing;
+    use dp_block::{DeoxysBlockInfo, DeoxysBlockInner, DeoxysMaybePendingBlock, DeoxysMaybePendingBlockInfo, Header};
+    use dp_state_update::StateDiff;
+    use rstest::rstest;
+    use starknet_core::types::Felt;
+
+    #[rstest]
+    fn test_block_hash_and_number() {
```
apoorvsadana: Should we also check that if a Pending block is present then the block returned by `block_hash_and_number` is the non pending one?

crates/client/rpc/src/methods/read/get_block_transaction_count.rs
```diff
@@ -28,3 +28,30 @@ pub fn get_block_transaction_count(starknet: &Starknet, block_id: BlockId) -> St
...
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_1, open_testing, SampleChain1};
+    use rstest::rstest;
+    use starknet_core::types::BlockTag;
+
+    #[rstest]
+    fn test_get_block_transaction_count() {
+        let (backend, rpc) = open_testing();
+        let SampleChain1 { block_hashes, .. } = make_sample_chain_1(&backend);
```
apoorvsadana: let's use more descriptive names instead of 1/2?

crates/client/rpc/src/methods/read/get_block_with_receipts.rs
```diff
@@ -63,3 +64,95 @@ pub fn get_block_with_receipts(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_1, open_testing, SampleChain1};
+    use rstest::rstest;
+    use starknet_core::types::{BlockTag, Felt, L1DataAvailabilityMode, ResourcePrice};
+
+    #[rstest]
+    fn test_get_block_with_receipts() {
```
apoorvsadana: so from what I understand, we store the data inside the DB with `make_sample_chain_1` and then we fetch that data with `get_block_with_receipts`. this seems to test the receipt functionality but in my the view, the ideal RPC test case would

- (as a sequencer), execute a txn using the RPC method and then check if the receipt was built correctly
- (as a full node), sync from the gateway a txn and check if the receipt was built correctly (or do we get receipts from the gateway as well in the full node, I am not sure)

crates/client/rpc/src/methods/read/get_block_with_tx_hashes.rs
```diff
@@ -63,3 +63,86 @@ pub fn get_block_with_tx_hashes(
...
+    use super::*;
+    use crate::test_utils::{make_sample_chain_1, open_testing, SampleChain1};
+    use rstest::rstest;
+    use starknet_core::types::{BlockTag, Felt, L1DataAvailabilityMode, ResourcePrice};
+
+    #[rstest]
+    fn test_get_block_with_tx_hashes() {
+        let _ = env_logger::builder().is_test(true).try_init();
+        let (backend, rpc) = open_testing();
+        let SampleChain1 { block_hashes, tx_hashes, .. } = make_sample_chain_1(&backend);
```
apoorvsadana: same as above. currently it looks like we're checking db storage and not the RPC logic itself or maybe I am missing something

crates/client/rpc/src/methods/read/get_block_with_txs.rs
```diff
@@ -67,3 +67,86 @@ pub fn get_block_with_txs(starknet: &Starknet, block_id: BlockId) -> RpcResult<M
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_1, open_testing, SampleChain1};
+    use rstest::rstest;
+    use starknet_core::types::{BlockTag, Felt, L1DataAvailabilityMode, ResourcePrice};
+
+    #[rstest]
+    fn test_get_block_with_txs() {
```
apoorvsadana: same as above

crates/client/rpc/src/methods/read/get_class_hash_at.rs
```diff
@@ -22,7 +22,46 @@ pub fn get_class_hash_at(starknet: &Starknet, block_id: BlockId, contract_addres
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_2, open_testing, SampleChain2};
+    use rstest::rstest;
+    use starknet_core::types::BlockTag;
+
+    #[rstest]
+    fn test_get_class_hash_at() {
```
apoorvsadana: same as above

crates/client/rpc/src/methods/read/get_state_update.rs
```diff
@@ -80,3 +80,55 @@ pub fn get_state_update(starknet: &Starknet, block_id: BlockId) -> StarknetRpcRe
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_2, open_testing, SampleChain2};
+    use rstest::rstest;
+
+    #[rstest]
+    fn test_get_state_update() {
```
apoorvsadana: same as above

crates/client/rpc/src/methods/read/get_storage_at.rs
```diff
@@ -55,3 +55,74 @@ pub fn get_storage_at(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{make_sample_chain_2, open_testing, SampleChain2};
+    use rstest::rstest;
+    use starknet_core::types::BlockTag;
+
+    #[rstest]
+    fn test_get_storage_at() {
```
apoorvsadana: same as above

apoorvsadana (commented):
So I have some questions on how are we looking at RPC test cases. In my view, the test cases can be divided into two categories

## Sequencer RPC tests

Here, I think we should send the transaction via one RPC and then check if we got the expected result. For example, if we want to check transaction hash caclculation, then, we should first do a dummy invoke transaction and check if we got the expected transaction hash. Or, for example, if we want to check transaction count, we should send 5 txs, manually seal a block (like in Substrate) and then call the transaction count method and see if it worked correctly. I might be biased on the approach because this is how we did it in Madara previously. But curious to understand more about your approach here.

## Full node RPC tests

Here, it would be similar to above but instead of sending transaction we would ideally mock the feeder gateway responses. I believe @cchudant said something like this on Slack.

Also, one suggestion would be to maybe split RPC tests into different PRs :)

crates/client/rpc/src/methods/read/get_storage_at.rs
```diff
@@ -55,3 +55,74 @@ pub fn get_storage_at(
...
+                        assert_eq!(
+                            get_storage_at(&rpc, contracts[contract_i], keys[key_i], block_n),
+                            Err(StarknetRpcApiError::ContractNotFound),
+                            "get storage at blockid {block_n:?}, contract #{contract_i}, key #{key_i} should not found"
+                        );
+                    }
+                }
+            }
+        };
+
```
Mohiiit, line 98: shall we be testing about `BLOCK_NOT_FOUND` as well

crates/client/rpc/src/methods/read/get_transaction_by_block_id_and_index.rs
```diff
@@ -34,3 +34,54 @@ pub fn get_transaction_by_block_id_and_index(
...
+    use starknet_core::types::BlockTag;
+
+    #[rstest]
+    fn test_get_transaction_by_block_id_and_index() {
+        let _ = env_logger::builder().is_test(true).try_init();
+        let (backend, rpc) = open_testing();
+        let SampleChain1 { block_hashes, expected_txs, .. } = make_sample_chain_1(&backend);
+
+        // Block 0
+        assert_eq!(get_transaction_by_block_id_and_index(&rpc, BlockId::Number(0), 0).unwrap(), expected_txs[0]);
```
Mohiiit, line 52: we can also add tests for `BLOCK_NOT_FOUND` and `INVALID_TXN_INDEX` as well

crates/client/rpc/src/methods/read/get_transaction_receipt.rs
```diff
@@ -65,3 +65,44 @@ pub async fn get_transaction_receipt(
...
+        let res = TransactionReceiptWithBlockInfo {
+            receipt: expected_receipts[2].clone(),
+            block: ReceiptBlock::Block { block_hash: block_hashes[2], block_number: 2 },
+        };
+        assert_eq!(get_transaction_receipt(&rpc, tx_hashes[2]).unwrap(), res);
+
+        // Pending
+        let res =
+            TransactionReceiptWithBlockInfo { receipt: expected_receipts[3].clone(), block: ReceiptBlock::Pending };
+        assert_eq!(get_transaction_receipt(&rpc, tx_hashes[3]).unwrap(), res);
```
Mohiiit, line 105: tests for `TxnHashNotFound` not present

crates/node/src/cli/mod.rs
Mohiiit: was thinking about it as well, given for app chains it would be much more simpler to have feeder url in the params

crates/client/db/src/contract_db.rs
cchudant: nota bene, which is like `Note:` here or `Pay attention:`
should I stop putting NBs all over the codebase? I feel like this is a common enough abbreviation but i'm french and biased

crates/client/db/src/lib.rs
cchudant: no :)
i just wanted to have cleaner logs in tests but I can remove it

crates/client/db/src/lib.rs
cchudant: same, i'll probably remove it

crates/client/rpc/src/methods/read/block_hash_and_number.rs
cchudant: sure why not

crates/client/rpc/src/methods/read/get_transaction_receipt.rs
cchudant: done

crates/client/rpc/src/methods/read/get_transaction_by_block_id_and_index.rs
cchudant: done

CHANGELOG.md
cchudant: done

antiyro (approved):

PR #216, opened 2024-08-05
feat(client): l1<>l2 messaging crate

### LINKED ISSUE #204: feat(l1): Implement l1<>l2 messaging

azurwastaken:
# Pull Request type

- Feature

Resolves: #204

## What is the new behavior?

- Added a crate with a worker that listen for LogMessageToL2 event on main contract and handle it as the old madara did.
- Added some entries to the db for messaging related data

No 

## Usefull information

🚨Require #208 🚨
This PR include a lot of code from #208 as we started from it to build with the feature from the eth crate.

A few notes on the current design:
- There is no guarantee that L1->L2 messages will be executed in a centralized context.
- There is no ordering guarantee, nonces are solely used 

### Overall Workflow
- [X] Create a stream of LogMessageToL2 events
- [X] Get last synced block from Messaging DB
- [X] Consume the stream and log event
- [ ] Process message 
    - [X] Parse tx fee
    - [X] Parse transaction from event
    - [X] Check if message has already been processed
    - [X] Build transaction
    - [Waiting for mempool #198 ] Submit tx to mempool
    - [X] Update Messaging DB
    
### Specific case to handle    
- [ ] Handle Message Cancellation (https://github.com/keep-starknet-strange/madara/issues/1638)

### Tests

- [ ] E2E test 1
    - Launch Anvil Node
    - Launch Worker
    - Send L1->L2 message
    - Assert that event is emitted on L1
    - Assert that even was caught by the worker with correct data
    - Assert the tx hash computed by the worker is correct
    - Assert that the tx has been included in the mempool
    - Assert that DB was correctly updated (last synced block & nonce)
    - Assert that the tx was correctly executed

- [ ]  E2E test 2
    - Should fail if we try to send multiple messages with same nonces

- [ ]  E2E test 3 
    - Message Cancellation


Mohiiit:
Hey @azurwastaken, we were thinking of a different approach about the structure of the code. So instead of creating l1-messaging a different crate, we can have it inside the eth-crate, just like update_state. And as of now we are joining l1_sync with l1_messages_sync and l2_sync, so I was thinking to join l1_sync and l2_sync and inside the l1_sync we will have l1_messages_sync, state_update_sync and l1_gas_price sync, wdyt?

so structure of the code we had in mind was:

- eth-crate
    |_____ state_update
    |_____ l1_messages
    |_____ l1_gas_price
    |_____ maybe a `sync.rs` with sync function which calls above 3
    
and the above `sync` function get called in the `sync` crate, inside the `starknet_sync_worker`, wdyt?

cc: @cchudant

azurwastaken:
> Hey @azurwastaken, we were thinking of a different approach about the structure of the code. So instead of creating l1-messaging a different crate, we can have it inside the eth-crate, just like update_state. And as of now we are joining l1_sync with l1_messages_sync and l2_sync, so I was thinking to join l1_sync and l2_sync and inside the l1_sync we will have l1_messages_sync, state_update_sync and l1_gas_price sync, wdyt?
> 
> so structure of the code we had in mind was:
> 
> * eth-crate
>   |_____ state_update
>   |_____ l1_messages
>   |_____ l1_gas_price
>   |_____ maybe a `sync.rs` with sync function which calls above 3
> 
> and the above `sync` function get called in the `sync` crate, inside the `starknet_sync_worker`, wdyt?
> 
> cc: @cchudant

Hey @Mohiiit , yes you are right, its porbably better if everything is together

cchudant:
> Hey @azurwastaken, we were thinking of a different approach about the structure of the code. So instead of creating l1-messaging a different crate, we can have it inside the eth-crate, just like update_state. And as of now we are joining l1_sync with l1_messages_sync and l2_sync, so I was thinking to join l1_sync and l2_sync and inside the l1_sync we will have l1_messages_sync, state_update_sync and l1_gas_price sync, wdyt?
> 
> so structure of the code we had in mind was:
> 
>     * eth-crate
>       |_____ state_update
>       |_____ l1_messages
>       |_____ l1_gas_price
>       |_____ maybe a `sync.rs` with sync function which calls above 3
> 
> 
> and the above `sync` function get called in the `sync` crate, inside the `starknet_sync_worker`, wdyt?
> 
> cc: @cchudant

yes that makes sense
however a small thing here: i dont want the eth crate to be called from sync, i'd like it to be a Service so that i can just activate it even when doing block production (not sync) mode

also in the future this eth crate will expose a common interface to the rest of the app that is swappable with other chains (e.g for starknet L3s) but i'm getting ahead of myself

azurwastaken:
Outdated, see #220

PR #219, opened 2024-08-07
feat(client): l1 gas price

### LINKED ISSUE #205: feat(l1): Forward L1 gas price data to the block header

Mohiiit:
# Pull Request type

- Feature

Resolves: #205 

## What is the new behavior?

- Added a worker within the eth crate in order to fetch the gas prices from the l1
- Added tests regarding the same

No

## Other Information

In the current scope of the PR, we are not aiming to add oracle to fetch the STRK <-> ETH price for the strk conversion of the fees.

## Current Progress

- [x] Creating L1 gas price worker
- [x] unit test for the worker
- [x] supplying gas price to the block production service


crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+    *LAST_UPDATE_TIMESTAMP.lock().await
+}
+
+pub async fn gas_price_worker(
+    eth_client: &EthereumClient,
+    l1_data_provider: Arc<dyn L1DataProvider>,
+    infinite_loop: bool,
+) -> anyhow::Result<()> {
+    let poll_time = eth_client.gas_price_poll_ms.unwrap_or(DEFAULT_GAS_PRICE_POLL_MS);
+    update_last_update_timestamp().await;
```
apoorvsadana: is this incorrect? i can see the timestamp is updated inside `update_gas_price` and it should only be there from what I understand

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+
+        if current_timestamp - last_update_timestamp > 10 * poll_time as u128 {
+            panic!(
+                "Gas prices have not been updated for {} ms. Last update was at {}",
+                current_timestamp - last_update_timestamp,
+                last_update_timestamp
+            );
+        }
+
+        if !infinite_loop {
```
apoorvsadana: the purpose of the `infinite_loop` variable was to call this function once at the start of the node without looping to set the initial gas price, I think we should do that as well here

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    // The RPC responds with 301 elements for some reason. It's also just safer to manually
+    // take the last 300. We choose 300 to get average gas caprice for last one hour (300 * 12 sec block
+    // time).
+    let (_, blob_fee_history_one_hour) =
+        fee_history.base_fee_per_blob_gas.split_at(fee_history.base_fee_per_blob_gas.len().max(300) - 300);
+
+    let avg_blob_base_fee = blob_fee_history_one_hour.iter().sum::<u128>() / blob_fee_history_one_hour.len() as u128;
+
+    let eth_gas_price = fee_history.base_fee_per_blob_gas.last().context("Setting l1 last confirmed block number")?;
```
apoorvsadana:
```suggestion
    let eth_gas_price = fee_history.base_fee_gas.last().context("Getting eth gas price")?;
```

apoorvsadana: anyway we can ensure this in test cases? I think anvil allows you to set a custom gas price when you start it

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    // Get the current gas price
+    let current_gas_price = l1_data_provider.get_gas_prices().await;
+    let eth_gas_price = current_gas_price.eth_l1_gas_price;
+
+    // Update the metrics
+    eth_client.l1_block_metrics.l1_block_number.set(latest_block_number as f64);
+    eth_client.l1_block_metrics.l1_gas_price_wei.set(eth_gas_price as f64);
+
+    // We're ignoring l1_gas_price_strk
```
apoorvsadana, line 86: can we import the strk code as well? i remember pragma did a PR. or do we want to do it in another issue

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+        create_ethereum_client(None)
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn gas_price_worker_works(eth_client: &'static EthereumClient) {
+        let l1_data_provider: Arc<dyn L1DataProvider> = Arc::new(GasPriceProvider::new());
+
+        // Run the worker for a short time
+        let worker_handle = tokio::spawn(gas_price_worker(eth_client, l1_data_provider.clone(), false));
```
apoorvsadana: - if `infinite_loop` is false, do we need to spawn tokio? we can just await here right?
- we should also write a case for infinite loop true
- how does the current code work without a delay?

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,235 @@
...
+        assert_eq!(updated_prices.eth_l1_gas_price, 1, "ETH L1 gas price should be 1 in test environment");
+
+        // Assert that the ETH L1 data gas price has been updated to 1 (as per test environment)
+        assert_eq!(updated_prices.eth_l1_data_gas_price, 1, "ETH L1 data gas price should be 1 in test environment");
+
+        // Verify that the last update timestamp is recent
+        let now = SystemTime::now().duration_since(UNIX_EPOCH).expect("Time went backwards").as_secs() as u128;
+
+        let last_update_timestamp = get_last_update_timestamp().await;
+        assert!(last_update_timestamp > now - 60, "Last update timestamp should be within the last minute");
```
apoorvsadana: what is the timestamp when the test starts?

crates/client/mempool/src/l1.rs
```diff
@@ -1,12 +1,73 @@
...
     /// Get the DA mode for L1 Ethereum. This can be either Blob or Calldata, whichever is cheaper at the moment.
     fn get_da_mode(&self) -> L1DataAvailabilityMode;
 
     // fn get_pending_l1_handler_txs
 }
+
+#[async_trait::async_trait]
+impl L1DataProvider for GasPriceProvider {
+    async fn get_gas_prices(&self) -> GasPrices {
+        (*self.gas_prices.lock().await).clone()
```
apoorvsadana: let's mention that it's 4 `u128` so it's ok to clone (unless we think it isn't?)

crates/node/src/cli/l1.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[derive(Clone, Debug, clap::Args)]
+pub struct L1SyncParams {
+    /// Disable L1 sync.
+    #[clap(long, alias = "no-l1-sync")]
+    pub sync_l1_disabled: bool,
+
+    /// The L1 rpc endpoint url for state verification.
+    #[clap(long, value_parser = parse_url, value_name = "ETHEREUM RPC URL")]
+    pub l1_endpoint: Option<Url>,
```
apoorvsadana, line 16: you can enforce in clippy that `sync_l1_disabled` false requires `l1_endpoint` afaik

crates/node/src/service/l1.rs
```diff
@@ -0,0 +1,81 @@
...
+            if let Some(l1_rpc_url) = &config.l1_endpoint {
+                Some(l1_rpc_url.clone())
+            } else {
+                return Err(anyhow::anyhow!(
+                    "❗ No L1 endpoint provided. You must provide one in order to verify the synced state."
+                ));
+            }
+        } else {
+            None
+        };
```
apoorvsadana:
```suggestion
       let l1_endpoint = config.sync_l1_disabled
            .not()
            .then_some(config.l1_endpoint.clone()
                .context("❗ No L1 endpoint provided. You must provide one in order to verify the synced state.")?
            );
```

crates/node/src/service/l1.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        let core_address = Address::from_slice(l1_core_address.as_bytes());
+        let eth_client = EthereumClient::new(l1_endpoint.unwrap(), core_address, metrics_handle)
```
apoorvsadana:
```suggestion
        let eth_client = EthereumClient::new(l1_endpoint?, core_address, metrics_handle)
```

crates/client/eth/src/l1_gas_price.rs
Mohiiit: this is for the first time, time stamp is defined in the lazy static, and when the `gas worker` is called, it will update it for the first time and then for the next time it will get updated in the `update_gas_price`

crates/client/eth/src/l1_gas_price.rs
Mohiiit: we would need an oracle for that right? I will check for the PR, it can be another PR with the scope of adding oracle

crates/client/eth/src/l1_gas_price.rs
Mohiiit: - yeah I think that should work, using the await only
- will do, we have that when we are testing for the fail case but will create for this as well
- there is a delay, it's in the eth_client

crates/client/eth/src/l1_gas_price.rs
Mohiiit: it get's updated in the gas worker

crates/client/eth/src/l1_gas_price.rs
Mohiiit: yeah that's true, I think we can achieve this using mocks right?

crates/client/eth/src/l1_gas_price.rs
Mohiiit: resolved

crates/client/eth/src/l1_gas_price.rs
Mohiiit: this is removed because we are calling the update_gas_price while creating the sync service

crates/client/eth/src/l1_gas_price.rs
Mohiiit: resolved

crates/node/src/cli/l1.rs
Mohiiit: resolved

crates/node/src/service/l1.rs
Mohiiit: resolved

crates/node/src/service/l1.rs
Mohiiit: resolved

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,276 @@
...
+use futures::lock::Mutex;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::sleep;
+
+use lazy_static;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+lazy_static::lazy_static! {
+    static ref LAST_UPDATE_TIMESTAMP: Arc<Mutex<u128>> = Arc::new(Mutex::new(0));
```
cchudant: can this be inside the service instead of global state?

crates/node/src/service/l1.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        let gas_price_sync_disabled = config.gas_price_sync_disabled;
+        let gas_price_poll_ms = config.gas_price_poll_ms;
+
+        if !gas_price_sync_disabled {
+            let eth_client = eth_client.clone().ok_or_else(|| {
+                anyhow::anyhow!("EthereumClient is required to start the l1 sync service but not provided.")
+            })?;
+            // running at-least once before the block production service
+            let _ = futures::executor::block_on(dc_eth::l1_gas_price::gas_price_worker(
```
cchudant: what is that block_on? this looks suspicious

crates/node/src/cli/l1.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[clap(long, alias = "no-l1-sync", conflicts_with = "l1_endpoint")]
+    pub sync_l1_disabled: bool,
+
+    /// The L1 rpc endpoint url for state verification.
+    #[clap(long, value_parser = parse_url, value_name = "ETHEREUM RPC URL", required_unless_present = "sync_l1_disabled")]
+    pub l1_endpoint: Option<Url>,
+
+    /// Disable the gas price sync service. The sync service is responsible to fetch the fee history from the ethereum.
+    #[clap(long, alias = "no-gas-price-sync")]
+    pub gas_price_sync_disabled: bool,
```
cchudant, line 20: i like it when everything is configurable thank you :)

crates/node/src/service/sync.rs
```diff
@@ -34,32 +31,13 @@ impl SyncService {
...
-                return Err(anyhow::anyhow!(
-                    "❗ No L1 endpoint provided. You must provide one in order to verify the synced state."
-                ));
-            }
-        } else {
-            None
-        };
-
-        let core_address = Address::from_slice(config.network.l1_core_address().as_bytes());
-        let eth_client = EthereumClient::new(l1_endpoint.unwrap(), core_address, metrics_handle)
```
cchudant: when making #221 i found out that you broke `sync_l1_disabled` on main in the last pr due to this unwrap  - i fixed it there for now but is it fixed in this pr?

crates/client/eth/src/sync.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub async fn l1_sync_worker(
+    backend: &DeoxysBackend,
+    eth_client: &EthereumClient,
+    chain_id: Felt,
+    l1_data_provider: Arc<dyn L1DataProvider>,
+    gas_price_sync_disabled: bool,
+    gas_price_poll_ms: u64,
+) -> anyhow::Result<()> {
+    let state_update_fut = async { state_update_worker(backend, eth_client, chain_id).await };
```
cchudant: this func can be simplified to
```rs
tokio::try_join!(
  state_update_worker(backend,eth_client, chain_id),
  async {
    if !gas_price_sync_disabled {
      gas_price_worker(eth_client, l1_data_provider, true, gas_price_poll_ms).await?;
    }
    Ok(())
  },
)?;
```

crates/client/mempool/src/l1.rs
```diff
@@ -1,12 +1,73 @@
...
+}
 
 /// This trait enables the block production task to fill in the L1 info.
 /// Gas prices and DA mode
+#[async_trait::async_trait]
 pub trait L1DataProvider: Send + Sync {
     /// Get L1 data gas prices. This needs an oracle for STRK prices.
-    fn get_gas_prices(&self) -> GasPrices;
+    async fn get_gas_prices(&self) -> GasPrices;
+    async fn set_gas_prices(&self, new_prices: GasPrices);
```
cchudant: can you remove those setters from the trait? this is an internal impl detail that other crate shouldn't have access

crates/client/mempool/src/lib.rs
```diff
@@ -52,7 +52,7 @@ impl Mempool {
         Mempool { backend, l1_data_provider, inner: Default::default() }
     }
 
-    pub fn accept_account_tx(
+    pub async fn accept_account_tx(
```
cchudant: mmh i don't like that being async i need to think about this

crates/client/eth/src/l1_gas_price.rs
cchudant: also use a SystemTime directly instead of an untyped integer unix epoch?

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        let last_update_timestamp = get_last_update_timestamp().await;
+        let current_timestamp = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("Failed to get current timestamp")
+            .as_millis();
+
+        if current_timestamp - last_update_timestamp > 10 * gas_price_poll_ms as u128 {
+            panic!(
```
cchudant: why panic here?

crates/client/eth/src/l1_gas_price.rs
cchudant: i think this function would look better if it was split into two:
the loop body is a function, which corresponds to when the current `infinite_loop` false when called directly, and a separate function for the infinite task which calls the first function

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -0,0 +1,276 @@
...
+                current_timestamp - last_update_timestamp,
+                last_update_timestamp
+            );
+        }
+
+        if !infinite_loop {
+            return Ok(());
+        }
+
+        sleep(Duration::from_millis(gas_price_poll_ms)).await;
```
cchudant: use 
```rs
  let mut interval = tokio::time::interval(pending_block_poll_interval);
  interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
  loop {
    // work
    interval.tick().await;
  }
  ```
instead of sleep for more accurate intervals and tick skipping behavior in case the node is overloaded

Comment (cchudant) [reply crates/client/eth/src/l1_gas_price.rs]: actually, infinite loops should look like this:
```rs
    let mut interval = tokio::time::interval(pending_block_poll_interval);
    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
    while wait_or_graceful_shutdown(interval.tick()).await.is_some() {
```
to handle graceful shutdown of the node

cchudant (changes requested):
I am not sure if this pull request:
- fetches gas prices everytime a transaction is added to the mempool / a block is produced -- the L1DataProvider functions are async
- or is an infinite task running in the background, watching for l1 gas prices and updating a mutex that is shared with the mempool through the L1DataProvider abstraction -- the L1DataProvider functions are sync

which of the two does this pr implement?

I don't like the first one as:
- the mempool needs L1 prices for mempool validation
- the block production task needs L1 prices to produce blocks
- i don't like the idea of l1 gas prices request being initiated by a add_transaction rpc
- i don't like having the block production be slowed down by l1, as it's potentially time critical

The second approach has a major drawback that i'm realizing now:
- we shouldnt accept any mempool transaction before the first gas prices arrive.
- we cannot produce any block before the first gas prices arrive otherwise we could have a huge problem

I think the easiest way to solve it is:
- have L1DataProvider look like this
```rs
pub struct NewBlockL1Info {
  pub data_availability_mode: L1DataAvailabilityMode, // blob or calldata
  pub gas_prices: GasPrices,
}

trait L1DataProvider {
  /// This returns immediately when GasPrices are available,
  /// or will wait until the L1 watcher has started up and got the first prices.
  async fn get_gas_prices(&self) -> NewBlockL1Info;
}

// tokio watch channel
struct ProvideL1Data(tokio::sync::watch::Receiver<Option<NewBlockL1Info>>);
impl L1DataProvider for ProvideL1Data {
  async fn get_gas_prices(&self) -> NewBlockL1Info {
    self.0.wait_for(Option::is_some).await.expect("recv error").clone()
  }
}

async fn l1_gas_prices_task(set_gas_prices: tokio::sync::watch::Sender<Option<NewBlockL1Info>>) -> anyhow::Result<()> {
  loop {
    get prices
    if set_gas_prices.send(Some(gas_prices)).is_err() {
      // channel is closed, meaning no one is listening to gas prices
      return Ok(());
    }
    interval sleep
  }
}
```

crates/node/src/service/l1.rs
Mohiiit: this is to fetch the gas price before returning the sync service, it make sure we have gas price

crates/node/src/service/sync.rs
Mohiiit: I checked that, and yeah it's resolved in this PR

crates/client/eth/src/l1_gas_price.rs
Mohiiit: in case we are not able to fetch the latest gas prices, it was present in the madara code as well. I believe it exist to make sure we don't have gas fees deviation. but we can have a flag for this as well

Mohiiit:
> I am not sure if this pull request:
> 
> * fetches gas prices everytime a transaction is added to the mempool / a block is produced -- the L1DataProvider functions are async
> * or is an infinite task running in the background, watching for l1 gas prices and updating a mutex that is shared with the mempool through the L1DataProvider abstraction -- the L1DataProvider functions are sync
> 
> which of the two does this pr implement?

it implements the second one, the functions are async because reading from mutex is async and hence all the functions inside the block production service which requires the gas fees needs to be async




> The second approach has a major drawback that i'm realizing now:
> 
> * we shouldnt accept any mempool transaction before the first gas prices arrive.

that's why we are getting the gas price in the `L1SyncService::new`

crates/client/mempool/src/l1.rs
Mohiiit: resolved

crates/client/mempool/src/lib.rs
Mohiiit: resolved

crates/client/eth/src/l1_gas_price.rs
Mohiiit: resolved

crates/client/eth/src/l1_gas_price.rs
Mohiiit: resolved

cchudant (approved):
looks good to me :)

apoorvsadana (approved):

PR #220, opened 2024-08-07
feat(client): l1<>l2 messaging crate

### LINKED ISSUE #204: feat(l1): Implement l1<>l2 messaging

azurwastaken:
# Pull Request type

- Feature

Resolves: #204

## What is the new behavior?

- Added a crate with a worker that listen for LogMessageToL2 event on main contract and handle it as the old madara did.
- Added some entries to the db for messaging related data

No 

## Usefull information

I reopened this pr because there was an issue with our old fork. This one is from a fresh one.

A few notes on the current design:
- There is no guarantee that L1->L2 messages will be executed in a centralized context.
- There is no ordering guarantee, nonces are solely used 

### Overall Workflow
- [X] Create a stream of LogMessageToL2 events
- [X] Get last synced block from Messaging DB
- [X] Consume the stream and log event
- [ ] Process message 
    - [X] Parse tx fee
    - [X] Parse transaction from event
    - [X] Check if message has already been processed
    - [X] Build transaction
    - [Waiting for update to accepte blockifier tx] Submit tx to mempool
    - [X] Update Messaging DB
    
### Specific case to handle    
- [X] Handle Message Cancellation (https://github.com/keep-starknet-strange/madara/issues/1638)

### Tests

- [x] E2E test 1
    - [X] Launch Anvil Node
    - [X] Launch Worker
    - [X] Send L1->L2 message
    - [X] Assert that event is emitted on L1
    - [x] Assert that even was caught by the worker with correct data
    - [x] Assert the tx hash computed by the worker is correct
    - [ ] Assert that the tx has been included in the mempool
    - [X] Assert that DB was correctly updated (last synced block & nonce)
    - [ ] Assert that the tx was correctly executed

- [X]  E2E test 2
    - [X] Should fail if we try to send multiple messages with same nonces

- [X]  E2E test 3 
    - [X] Message Cancellation


azurwastaken:
> Hey @azurwastaken, we were thinking of a different approach about the structure of the code. So instead of creating l1-messaging a different crate, we can have it inside the eth-crate, just like update_state. And as of now we are joining l1_sync with l1_messages_sync and l2_sync, so I was thinking to join l1_sync and l2_sync and inside the l1_sync we will have l1_messages_sync, state_update_sync and l1_gas_price sync, wdyt?
> 
> so structure of the code we had in mind was:
> 
> * eth-crate
>   |_____ state_update
>   |_____ l1_messages
>   |_____ l1_gas_price
>   |_____ maybe a `sync.rs` with sync function which calls above 3
> 
> and the above `sync` function get called in the `sync` crate, inside the `starknet_sync_worker`, wdyt?
> 

Copying back this comment from Mohiit

azurwastaken:
> yes that makes sense however a small thing here: i dont want the eth crate to be called from sync, i'd like it to be a Service so that i can just activate it even when doing block production (not sync) mode
> 
> also in the future this eth crate will expose a common interface to the rest of the app that is swappable with other chains (e.g for starknet L3s) but i'm getting ahead of myself

and this one from cchudant

crates/client/db/src/messaging_db.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+impl LastSyncedEventBlock {
+    pub fn new(block_number: u64, event_index: u64) -> Self {
+        LastSyncedEventBlock { block_number, event_index }
+    }
+}
+
+impl DeoxysBackend {
+    pub fn messaging_last_synced_l1_block_with_event(&self) -> Result<Option<LastSyncedEventBlock>> {
```
antiyro: Could you add some rust doc on top of your functions/structure. We aim to have a complete rust doc in the future

antiyro (approved):

crates/client/db/src/messaging_db.rs
antiyro: overall seems correct to me thanks! @azurwastaken @EvolveArt

azurwastaken: Added some, tell me if its ok @antiyro !

crates/client/db/src/messaging_db.rs
cchudant, line 1: messaging db is a vague name, l1_db maybe?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,494 @@
...
+                meta.block_number,
+                meta.transaction_hash,
+                meta.log_index,
+                event.fromAddress
+            );
+
+            // Check if cancellation was initiated
+            let event_hash = get_l1_to_l2_msg_hash(&event)?;
+            tracing::info!("⟠ Checking for cancelation, event hash : {:?}", event_hash);
+            let cancellation_timestamp = client.get_l1_to_l2_message_cancellations(event_hash).await?;
```
cchudant, line 77: oh cool you handle message cancellations!

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,494 @@
...
+async fn process_l1_message(
+    backend: &DeoxysBackend,
+    event: &LogMessageToL2,
+    l1_block_number: &Option<u64>,
+    event_index: &Option<u64>,
+    chain_id: &ChainId,
+) -> anyhow::Result<Option<TransactionHash>> {
+    let transaction = parse_handle_l1_message_transaction(event)?;
+    // Ensure that L1 message has not been executed
+    match backend.messaging_update_nonces_if_not_used(transaction.nonce) {
```
cchudant: is the nonce enough here to uniquely identify an l1 message?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,494 @@
...
+use starknet_api::transaction::{
+    Calldata, Fee, L1HandlerTransaction, Transaction, TransactionHash, TransactionVersion,
+};
+use starknet_api::transaction_hash::get_transaction_hash;
+use starknet_types_core::felt::Felt;
+
+pub async fn sync(backend: &DeoxysBackend, client: &EthereumClient, chain_id: &ChainId) -> anyhow::Result<()> {
+    tracing::info!("⟠ Starting L1 Messages Syncing...");
+
+    let last_synced_event_block = match backend.messaging_last_synced_l1_block_with_event() {
```
cchudant, line 45: I think you store an event_index in db that you never use here

crates/client/eth/src/client.rs
```diff
@@ -116,6 +118,13 @@ impl EthereumClient {
         let block_hash = self.l1_core_contract.stateBlockHash().call().await?;
         u256_to_felt(block_hash._0)
     }
+
+    /// Get the last Starknet block hash verified on L1
```
cchudant: wrong doc :)

crates/client/db/src/messaging_db.rs
```diff
@@ -0,0 +1,189 @@
...
+            writeopts.disable_wal(true);
+
+            let mut btree_set = BTreeSet::new();
+            btree_set.insert(nonce);
+
+            self.db.put_cf_opt(&messaging_column, NONCES_PROCESSED, bincode::serialize(&btree_set)?, &writeopts)?;
+            return Ok(true);
+        };
+
+        let mut nonce_tree: BTreeSet<Nonce> = bincode::deserialize(&res)?;
```
cchudant: this is gonna very get expensive once a lot of txs are processed

cchudant (changes requested):
this is cool! I didn't expect you to handle the niche message cancellation feature
I dont get the nonce part, why are you storing them in the db?

crates/client/eth/src/client.rs
azurwastaken: Oups :o) corrected, also renamed the variable because it was wrong too

crates/client/db/src/messaging_db.rs
azurwastaken: its very specific to messaging so l1_messaging_db maybe ?

crates/client/eth/src/l1_messaging.rs
azurwastaken: Was done like this in the old madara so i feel like it is

crates/client/eth/src/l1_messaging.rs
EvolveArt: event index could be used to uniquely identify that event, it's not used here but could be great to have for monitoring etc

crates/client/db/src/messaging_db.rs
cchudant: ah, i was thinking l1_db in the sense that there is everything L1 related there, so that we would probably move the l1_sync_tip from blocksdb to this one in the future and put other l1 stuff as needed there too

crates/client/eth/src/l1_messaging.rs
cchudant: But I think you risk reprocessing an event here, if the node shuts down when processing event 2 in a block, you start over from event 1
I thought this event index was there to avoid that

crates/client/eth/src/l1_messaging.rs
cchudant: ah
image

https://github.com/starkware-libs/cairo-lang/blob/4e233516f52477ad158bc81a86ec2760471c1b65/src/starkware/starknet/eth/StarknetMessaging.sol#L105

this nonce is very different from the nonce from account transactions, you're right

crates/client/db/src/l1_db.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// # Panics
+    ///
+    /// This function does not panic.
+    pub fn messaging_update_last_synced_l1_block_with_event(
+        &self,
+        last_synced_event_block: LastSyncedEventBlock,
+    ) -> Result<(), DbError> {
+        let messaging_column = self.db.get_column(Column::Messaging);
+        let mut writeopts = WriteOptions::default(); // todo move that in db
+        writeopts.disable_wal(true);
```
apoorvsadana, line 105: Do we disable WAL for all write operations or is there a logic to disable only for some?

crates/client/db/src/lib.rs
```diff
@@ -242,6 +248,8 @@ impl Column {
             ContractToNonces => "contract_to_nonces",
             ContractClassHashes => "contract_class_hashes",
             ContractStorage => "contract_storage",
+            Messaging => "messaging",
+            MessagingNonce => "messaging_nonce",
```
apoorvsadana:
```suggestion
            L1Messaging => "l1_messaging",
            L1MessagingNonce => "l1_messaging_nonce",
```

will this be more clear?

crates/client/eth/src/client.rs
```diff
@@ -116,6 +118,25 @@ impl EthereumClient {
...
+    ///
+    /// - msg_hash : Hash of L1 to L2 message
+    ///
+    /// # Return
+    ///
+    /// - A felt representing a timestamp :
+    ///     - 0 if the message has not been cancelled
+    ///     - timestamp of the cancellation if it has been cancelled
+    /// - An Error if the call fail
+    pub async fn get_l1_to_l2_message_cancellations(&self, msg_hash: FixedBytes<32>) -> anyhow::Result<Felt> {
```
apoorvsadana: this function seems to be specific to the `l1_messaging` worker, should we move it there?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+    event: &LogMessageToL2,
+    l1_block_number: &Option<u64>,
+    event_index: &Option<u64>,
+    chain_id: &ChainId,
+) -> anyhow::Result<Option<TransactionHash>> {
+    let transaction = parse_handle_l1_message_transaction(event)?;
+    let tx_nonce = transaction.nonce;
+
+    // Ensure that L1 message has not been executed
+    match backend.has_nonce(tx_nonce) {
```
apoorvsadana: should we rename to `has_l1_messaging_nonce` and `set_l1_messaging_nonce` for better readability?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+                meta.log_index,
+                event.fromAddress
+            );
+
+            // Check if cancellation was initiated
+            let event_hash = get_l1_to_l2_msg_hash(&event)?;
+            tracing::info!("⟠ Checking for cancelation, event hash : {:?}", event_hash);
+            let cancellation_timestamp = client.get_l1_to_l2_message_cancellations(event_hash).await?;
+            if cancellation_timestamp != Felt::ZERO {
+                tracing::info!("⟠ L1 Message was cancelled in block at timestamp : {:?}", cancellation_timestamp);
```
apoorvsadana, line 79: Should we set the nonce and last synced L1 block here to avoid re processing?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+        Ok(None) => {
+            unreachable!("Should never be None")
+        }
+        Err(e) => {
+            tracing::error!("⟠ Madara Messaging DB unavailable: {:?}", e);
+            return Err(e.into());
+        }
+    };
+    let event_filter = client.l1_core_contract.event_filter::<StarknetCoreContract::LogMessageToL2>();
+    let mut event_stream = event_filter
```
apoorvsadana, line 56: We should add here a check that the Ethereum block has been finalized (i.e. received 2/3rd attestations). this is how it works on public Starknet. It was missing in Madara before but might be a good chance to add it now.

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+            // Check if cancellation was initiated
+            let event_hash = get_l1_to_l2_msg_hash(&event)?;
+            tracing::info!("⟠ Checking for cancelation, event hash : {:?}", event_hash);
+            let cancellation_timestamp = client.get_l1_to_l2_message_cancellations(event_hash).await?;
+            if cancellation_timestamp != Felt::ZERO {
+                tracing::info!("⟠ L1 Message was cancelled in block at timestamp : {:?}", cancellation_timestamp);
+                continue;
+            }
+
+            match process_l1_message(backend, &event, &meta.block_number, &meta.log_index, chain_id).await {
```
apoorvsadana, line 95: if the event_index has already been processed we can skip here right?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+                uint256 toAddress = 3256441166037631918262930812410838598500200462657642943867372734773841898370;
+                uint256 selector = 774397379524139446221206168840917193112228400237242521560346153613428128537;
+                uint256[] memory payload = new uint256[](7);
+                payload[0] = 96;
+                payload[1] = 1659025;
+                payload[2] = 38575600093162;
+                payload[3] = 5;
+                payload[4] = 4543560;
+                payload[5] = 1082959358903034162641917759097118582889062097851;
+                payload[6] = 221696535382753200248526706088340988821219073423817576256483558730535647368;
```
apoorvsadana: I think this is the same as above, should we move it to a function?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+            format!("event hash : {:?}", contract.getL1ToL2MsgHash().call().await.expect("failed to get hash")._0)
+                .as_str()
+        ));
+
+        // TODO : Assert that the tx has been included in the mempool
+
+        // Assert that the event is well stored in db
+        let last_block =
+            db.backend().messaging_last_synced_l1_block_with_event().expect("failed to retrieve block").unwrap();
+        assert_ne!(last_block.block_number, 0);
```
apoorvsadana, line 418: we should check the event index and nonce as well?

apoorvsadana, line 418: Nitpick: should we check if the block_number is greater than what it was before the DB started?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+        tokio::time::sleep(Duration::from_secs(5)).await;
+        let last_block =
+            db.backend().messaging_last_synced_l1_block_with_event().expect("failed to retrieve block").unwrap();
+        assert_ne!(last_block.block_number, 0);
+
+        // Send the event a second time
+        let _ = contract.fireEvent().send().await.expect("Failed to fire event");
+        tokio::time::sleep(Duration::from_secs(5)).await;
+        // Assert that the last event in db is still the same as it is already processed (same nonce)
+        assert_eq!(
```
apoorvsadana, line 463: should we also check nonce?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,497 @@
...
+            tokio::spawn(async move { sync(db.backend(), &eth_client, &chain_config.chain_id).await })
+        };
+
+        // Mock cancelled message
+        let _ = contract.setIsCanceled(true).send().await;
+        let _ = contract.fireEvent().send().await.expect("Failed to fire event");
+        tokio::time::sleep(Duration::from_secs(5)).await;
+        let last_block =
+            db.backend().messaging_last_synced_l1_block_with_event().expect("failed to retrieve block").unwrap();
+        assert_eq!(last_block.block_number, 0);
```
apoorvsadana, line 503: same as above

crates/client/db/src/l1_db.rs
azurwastaken: as far as i saw we disable it everywhere

crates/client/db/src/lib.rs
azurwastaken: done

crates/client/eth/src/client.rs
azurwastaken: done

crates/client/eth/src/l1_messaging.rs
azurwastaken: good

crates/client/eth/src/l1_messaging.rs
azurwastaken: The db is empty at start as we create different temp dir for each test so it's always 0.
I can change it if you think its better

crates/client/eth/src/l1_messaging.rs
azurwastaken: done

crates/client/eth/src/l1_messaging.rs
azurwastaken: done

crates/client/eth/src/l1_messaging.rs
azurwastaken: done for nonce but event_index is 0 as its the first

crates/client/eth/src/l1_messaging.rs
azurwastaken: not sure to understand here, do you mean if we get two time the same event before the process end ?
because nonce is checked right after at the very beginning of process_l1_message and to me, last synced block should be push to db once all the processing went smooth but maybe i'm wrong. What do you think ?

crates/client/eth/src/l1_messaging.rs
azurwastaken: done

crates/client/db/src/l1_db.rs
apoorvsadana: got it, i think we should move this to some global config in that case (in a new PR probably), wdyt @azurwastaken @cchudant

crates/client/eth/src/l1_messaging.rs
```diff
@@ -0,0 +1,526 @@
...
+use dc_db::{l1_db::LastSyncedEventBlock, DeoxysBackend};
+use dp_utils::channel_wait_or_graceful_shutdown;
+use starknet_api::core::{ChainId, ContractAddress, EntryPointSelector, Nonce};
+use starknet_api::transaction::{
+    Calldata, Fee, L1HandlerTransaction, Transaction, TransactionHash, TransactionVersion,
+};
+use starknet_api::transaction_hash::get_transaction_hash;
+use starknet_types_core::felt::Felt;
+
+impl EthereumClient {
```
apoorvsadana, line 21: nitpick: i was thinking we can pass the `EthereumClient` as a ref to this function instead of implementing it inside it. i was thinking of the `EthereumClient` as a more generalised struct but that's just my view :)

crates/client/eth/src/l1_messaging.rs
apoorvsadana: So my view was that if we have a lot of cancelled messages on the L1 then our code will read the messages but do nothing. It might be a good idea to store in DB that we've now processed these messages (which is do nothing in the case of cancelled messages) so that we don't run them again if we restart

crates/client/eth/src/l1_messaging.rs
apoorvsadana: is this pending?

crates/client/db/src/l1_db.rs
azurwastaken: totally agree

crates/client/eth/src/l1_messaging.rs
azurwastaken: oh i see, i'll add that

azurwastaken: done

crates/client/eth/src/l1_messaging.rs
azurwastaken: I see your point. For me EthereumClient is more of an interface for interacting with L1 (and core contract). wdyt @cchudant ?

crates/client/eth/src/l1_messaging.rs
azurwastaken: no no its live. Maybe you had same bug as me where github doesn't show change from all commit in te "File changed" tab

crates/client/eth/src/l1_messaging.rs
apoorvsadana: should we set last synced also

azurwastaken: is it really "synced" if it has been cancelled ? how do we differentiate message that are cancelled or not ?

crates/client/eth/src/l1_messaging.rs
apoorvsadana: oh nvm, I guess I was checking at the wrong place, sorry

crates/client/eth/src/l1_messaging.rs
azurwastaken: done

crates/client/eth/src/l1_messaging.rs
azurwastaken: Not sure to understand here. its fine as long as we check nonce imo

crates/client/db/src/l1_db.rs
cchudant: I definitely need do that doc yes, we have a non-default rocksdb configuration and a lot of thought was put into these decisions

moved to #231 

the short answer is that we disable the WAL, and with atomic flushes activated we don't need the wal for the db to be fault tolerant
flushing is instead done between blocks (non-concurrently with writes) every five seconds during sync

crates/client/eth/src/l1_messaging.rs
cchudant: i don't have a preference, but i would think that anvil already provides us the generalized ethereum client, that struct is more of the interface to the core contract more specifically
maybe it should be renamed core contract client then

crates/client/eth/src/l1_messaging.rs
apoorvsadana: Well I would defined synced as "we've processed all event that were there to process till this block". So even if it's cancelled, we've technically synced till that chain as we've processed everything until there.

crates/client/eth/src/l1_messaging.rs
apoorvsadana: As in, what nonce check?

cchudant (approved):
looks good to me :partying_face:

crates/client/eth/src/l1_messaging.rs
azurwastaken: we check the LogMessagetoL2 nonce to ensure unicity of the event at the beginning of process_l1_message so i guess its enough to avoid reprocessing as its not the block nonce but the one of the event

crates/client/eth/src/l1_messaging.rs
EvolveArt: let's keep it like this for now I think, not sure it's really core contract specific

PR #224, opened 2024-08-09
Fix Docker Compose instructions

neelkamath:
# Pull Request type

The Docker Compose docs don't work.

## What is the new behavior?

- The `.env` file to be created is `.gitignore`d.
- There's no warning due to the obsolete `version` field in the Docker Compose file.
- The Docker Compose profile (`mainnet`/`testnet`) is documented so the command actually succeeds.

No


antiyro:
Could you add changelog to run CI?

antiyro:
Sorry for the late response here

antiyro:
Now you'll have to fix the rest of the CI too if you can, so I can merge it.

antiyro (approved):

antiyro:
Force merging this since it doesn't introduce anything critical and coverage is failing only because of external contribution.

Thanks @neelkamath

Issue #223, opened 2024-08-09
feat(v8): Node readiness overhaul


cchudant:
### Is there an existing issue?

- [X] I have searched the existing issues

### Motivation

The `/health` endpoint on the node is a health check endpoint designed to work with [kubernetes startup probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) so that the node does not appear as available in k8s (i.e. not ready to answer rpc requests) when it is starting up

Returning anything other than 200 on this endpoint marks the node as not ready.

This is not a priority, but this feature would be important for node operators running multiple madara node in a cluster, so that a madara node that is restarting / starting for the first time does not get registered in the load balancer and doesn't receive rpc calls while it is starting.

### Request

The current endpoint is always up at the same time rpcs go up

### Solution

Readiness should be defined as:
- in sync mode, show ready when the chain is fully synced by default (configurable via an arg)
- in block production mode, we should wait for the first gas prices (see comment on #219) to arrive before marking the node as ready

In any case, I'd like to have a generic struct like
```rs
pub type ServiceName = String;
struct ServiceReadiness {
  service: String,
  on_ready: oneshot::Receiver<()>,
}
pub struct ReadinessProbe(Arc<Mutex<Vec<Readiness>>>);
impl ReadinessProbe {
  fn register_service_readiness(service: ServiceName) -> oneshot::Sender<()> {}
  fn is_all_ready(&self) -> bool {}
}
```
this would be shared between the services, and the `start` function of each service would register their readiness condition.

### Are you willing to help with this request?

Yes!

heemankv:
Status update from codebase review on 2026-05-05.

Classification: Partially addressed but still open.

Current state:
The node now exposes `/health` and `/ready`, which is real progress, but the issue asked for a broader readiness model than the current endpoint-only implementation. Evidence: `madara/node/src/service/rpc/server.rs:139-155`.

Keeping this issue open based on the current tree.

PR #226, opened 2024-08-09
fix: l1_sync

### LINKED ISSUE #225: bug: `--sync-l1-disabled` option crash the node

### Description of bug

when launching Madara with the **--sync-l1-disabled** option without providing an L1 endpoint, the node crashes

```bash
thread 'main' panicked at crates/node/src/service/sync.rs:55:58:
called `Option::unwrap()` on a `None` value
```

### Steps to reproduce

_No response_

jbcaron:
## Pull Request type

- Bugfix

- the node crash with ***--sync-l1-disabled*** option
- double **l1_gas_price** metrics

Resolves: #225
Introduced by: #223

## What is the new behavior?

- eth client is not created with option ***--sync-l1-disabled*** 

No

## Other information

The metrics server must be updated accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: #225
 - feat: experimental block production and mempool
 - refactor: L1BlockMetric is intialized inside the EthereumClient new function
 - refactor: BlockMetrics divided in L1BlockMetrics and BlockMetrics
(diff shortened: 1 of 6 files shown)


antiyro (approved):
Perfect thanks

cchudant:
introduced by #225 not #223

PR #227, opened 2024-08-09
test: primitives

jbcaron:
The pull request  focuses on adding and refining unit tests across primitives and client/sync modules.
## Pull Request type

- Testing

## What is the new behavior?

- the sync process from FGW now directly transforms state updates into our primitive types instead of passing through starknet-cors’s intermediate types
- removing the ToStateUpdateCore trait by directly implementing the From trait instead
- adding unit tests across all crates of primitives:
  - block header
  - block inner
  - class
  - receipt
  - state update
  - transaction
- The conversion between starknet-api transaction and our own transaction type has been updated. Previously, we used a reference . However, the new implementation now takes ownership of the Transaction instead of borrowing it. This change was made because the starknet-api now uses Felt as the primitive type for transactions. By taking ownership, it avoids unnecessary copies of Felt and reduces the need for iterations over vectors during the conversion process.

No

## Other information

our unit tests are currently being conducted on inconsistent data, I’ve opened #230  to address the issue


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next release
 
+- test: add unitests primitives
+- fix: #225
```
EvolveArt: duplicate

crates/primitives/class/src/compile.rs
```diff
@@ -342,10 +355,28 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_compressed_sierra_contract_class_to_json() {
+    async fn test_compressed_sierra_contract_v_1_1_0_class_to_json() {
+        let provider = SequencerGatewayProvider::starknet_alpha_mainnet();
```
EvolveArt, line 359: so this test depends on the current state of mainnet and the gateway ? shouldn't we mock the provider here?

crates/primitives/convert/src/felt.rs
```diff
@@ -48,7 +48,7 @@ mod tests {
         assert_eq!(felt_to_u128(&Felt::THREE).unwrap(), 3);
         assert_eq!(felt_to_u128(&Felt::from(u64::MAX)).unwrap(), u64::MAX as u128);
         assert_eq!(felt_to_u128(&Felt::from(u128::MAX)).unwrap(), u128::MAX);
-        assert!(felt_to_u128(&(Felt::from(u128::MAX) + Felt::ONE)).is_err());
-        assert!(felt_to_u128(&Felt::MAX).is_err());
+        assert!(matches!(felt_to_u128(&(Felt::from(u128::MAX) + Felt::ONE)), Err(FeltToU128Error)));
+        assert!(matches!(felt_to_u128(&Felt::MAX), Err(FeltToU128Error)));
```
EvolveArt: we can use `assert_matches` here

crates/primitives/convert/src/to_felt.rs
```diff
@@ -92,3 +92,55 @@ impl_for_wrapper!(Nonce);
...
+            0x56, 0x78,
+        ];
+        let h160 = H160::from_slice(&value);
+        assert_eq!(h160.to_felt(), Felt::from_bytes_be_slice(&value));
+    }
+
+    #[test]
+    fn test_chain_id_to_felt() {
+        let main_chain_id = ChainId::Mainnet;
+        assert_eq!(main_chain_id.to_felt(), Felt::from_bytes_be_slice(b"SN_MAIN"));
```
EvolveArt: for the test to be relevant shouldn't we hardcode the Felt value instead of using the `from_bytes_be_slice` method?

crates/primitives/receipt/src/lib.rs
```diff
@@ -322,43 +351,159 @@ mod tests {
...
+    fn test_event_compute_hash_pedersen() {
+        let event = Event {
+            from_address: Felt::from(1),
+            keys: vec![Felt::from(2), Felt::from(3)],
+            data: vec![Felt::from(4), Felt::from(5)],
+        };
+
+        let hash = event.compute_hash_pedersen();
+
+        assert_eq!(hash, Felt::from_hex_unchecked("0x770591674368ef723f40ddef92ee7cf2fcf3e244afa15cd0a703fce83a415c1"));
```
EvolveArt: where does this value come from, add in comment

crates/primitives/transactions/src/into_starknet_api.rs
```diff
@@ -294,33 +535,70 @@ fn version(version: &Felt) -> starknet_api::transaction::TransactionVersion {
...
+
+        let tx: Transaction = dummy_tx_deploy_account_v1().into();
+        let api_tx = starknet_api::transaction::Transaction::try_from(&tx).unwrap();
+        let tx_back: Transaction = api_tx.into();
+        assert_eq!(tx, tx_back);
+
+        let tx: Transaction = dummy_tx_deploy_account_v3().into();
+        let api_tx = starknet_api::transaction::Transaction::try_from(&tx).unwrap();
+        let tx_back: Transaction = api_tx.into();
+        assert_eq!(tx, tx_back);
```
EvolveArt: would love an util function that could be used accross *ALL* tests that convert then convert back to ensure no data is lost during marshalling. 
Something like `assert_consistent_conversion<T1, T2>(a, b)` wdyt ? Would help a lot with clarity

EvolveArt (changes requested):

CHANGELOG.md
jbcaron: solve with merge main

crates/primitives/convert/src/felt.rs
jbcaron: 👍

jbcaron: you’re right that relying on the current state of the mainnet and the gateway isn’t ideal. However, this is a temporary solution. Storing the entire class within the project would be too heavy. For testing purposes, we need simple and lightweight class definitions from different versions. 

we can also open an issue to create dedicated test data (blocks, classes, etc.) that we can use in the future to further improve the reliability of our tests.

crates/primitives/class/src/compile.rs
jbcaron: you’re right that relying on the current state of the mainnet and the gateway isn’t ideal. However, this is a temporary solution. Storing the entire class within the project would be too heavy. For testing purposes, we need simple and lightweight class definitions from different versions.

we can also open an issue to create dedicated test data (blocks, classes, etc.) that we can use in the future to further improve the reliability of our tests.

crates/primitives/transactions/src/into_starknet_api.rs
jbcaron: like that?
```rust
pub fn assert_consistent_conversion<T1, T2>(a: T1)
where
	T1: Clone + PartialEq + std::fmt::Debug + From<T2>,
	T2: Clone + PartialEq + std::fmt::Debug + From<T1>,
{
	let b: T2 = a.clone().into();
	let c: T1 = b.clone().into();
	assert_eq!(a, c);
	let d: T2 = c.clone().into();
	assert_eq!(b, d);
}
```

crates/primitives/class/src/compile.rs
EvolveArt: let's open an issue yes

crates/primitives/transactions/src/into_starknet_api.rs
EvolveArt: yup should be good

EvolveArt: just change on the other places where it can be used

crates/primitives/class/src/compile.rs
jbcaron: #230

antiyro (approved):

EvolveArt (approved):

PR #229, opened 2024-08-12
fix(namings): Replace old naming with adapted namings

antiyro:
This pull request aims to replace the old namings with new namings to avoid confusions:

- deoxys/Deoxys/DEOXYS will be replaced by madara/Madara/MADARA
- dp/dc will be replace by mp/mc (Madara primitive/Madara client)


jbcaron:
can we close this PR now that #243 is merged?

PR #232, opened 2024-08-21
Block import crate

cchudant:
Closes #228  for now, as this takes most of the code from that pr except the pipeline.

- chain config is now in its own primitive crate, and the StarknetVersion struct was moved there
- class block ids are not stored in the class info anymore
- all of the commitment computation and block conversion is done in a new crate block-import, and mempool uses it to store blocks too
- dc_sync is now only the pipeline and the fetching

I have a few commits on top of this one, which add a genesis provider that also uses the block import crate - i'm opening the pr without these for now so that @jbcaron can work on top of this without having to deal with broken code


crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+        if expected != got {
+            return Err(BlockImportError::TransactionCommitment { got, expected });
+        }
+    }
+
+    Ok(got)
+}
+
+/// Compute the events commitment for a block.
+fn event_commitment(block: &UnverifiedFullBlock, _validation: &Validation) -> Result<Felt, BlockImportError> {
```
jbcaron, line 228: it would be necessary to verify that the receipt types and the tx_hash match the associated transactions, even if the commitment is calculated

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+/// Compute the root hash of a list of values.
+// The `HashMapDb` can't fail, so we can safely unwrap the results.
+//
+// perf: Note that committing changes still has the greatest performance hit
+// as this is where the root hash is calculated. Due to the Merkle structure
+// of Bonsai Tries, this results in a trie size that grows very rapidly with
+// each new insertion. It seems that the only vector of optimization here
+// would be to parallelize the tree traversal on insertion and optimize hash computation.
+// It seems lambdaclass' crypto lib does not do simd hashing, we may want to look into that.
+fn compute_merkle_root<H: StarkHash + Send + Sync>(values: &[Felt]) -> Felt {
```
jbcaron, line 304: Shouldn’t this function be in a dedicated crate?

crates/client/block_import/src/verify_apply.rs
```diff
@@ -0,0 +1,281 @@
...
+        let _exclusive = self.mutex.lock();
+
+        let backend = Arc::clone(&self.backend);
+        self.pool.spawn_rayon_task(move || verify_apply_pending_inner(&backend, block, validation)).await
+    }
+}
+
+/// This needs to be called sequentially, it will apply the state diff to the db, verify the state root and save the block.
+/// This runs on the [`rayon`] threadpool however as it uses parallelism inside.
+pub fn verify_apply_inner(
```
jbcaron, line 62: Can the block storage in the database and the trie not be done in parallel? Is it too much load for the database?

crates/client/block_import/src/pre_validate.rs
cchudant: I think not, because i don't see any other crate than block import using it

in fact the i want to make bonsai-trie an implementation detail that only block import and db care about

cchudant: dc-block-import could be considered part of the madara ""backend"" along with dc-db

dc-p2p dc-fgw-sync and everything else would be the ""frontend""

crates/client/block_import/src/pre_validate.rs
cchudant: I don't think I understand, it is checked in transaction_commitment I believe?

crates/client/block_import/src/pre_validate.rs
cchudant: using very loose quotation marks

crates/client/block_import/src/verify_apply.rs
cchudant: thats true we should do that

crates/client/block_import/src/pre_validate.rs
cchudant: ah i thought you were talking about the global tries

EvolveArt:
pls rebase @cchudant

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+    // collecting into a Result has.
+    // little known fact this uses the impl FromIterator for () from std, nice trick
+    let (mut commitments, mut converted_classes) = Default::default();
+    [
+        Box::new(|| {
+            commitments = block_commitments(&block, &validation)?;
+            Ok(())
+        }) as Box<dyn FnOnce() -> Result<(), BlockImportError> + Send>,
+        Box::new(|| {
+            converted_classes = convert_classes(&validation, classes)?;
```
EvolveArt: we should be consistent with arguments order

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+        .collect();
+
+    // Transaction commitment
+    let got = if starknet_version < StarknetVersion::STARKNET_VERSION_0_13_2 {
+        compute_merkle_root::<Pedersen>(&tx_hashes_with_signature)
+    } else {
+        compute_merkle_root::<Poseidon>(&tx_hashes_with_signature)
+    };
+
+    if let Some(expected) = block.commitments.transaction_commitment {
```
EvolveArt: this will only be called in full node mode right ? since in authority all commitments are set to None, maybe we can avoid that

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+// The `HashMapDb` can't fail, so we can safely unwrap the results.
+//
+// perf: Note that committing changes still has the greatest performance hit
+// as this is where the root hash is calculated. Due to the Merkle structure
+// of Bonsai Tries, this results in a trie size that grows very rapidly with
+// each new insertion. It seems that the only vector of optimization here
+// would be to parallelize the tree traversal on insertion and optimize hash computation.
+// It seems lambdaclass' crypto lib does not do simd hashing, we may want to look into that.
+fn compute_merkle_root<H: StarkHash + Send + Sync>(values: &[Felt]) -> Felt {
+    //TODO: replace the identifier by an empty slice when bonsai supports it
```
EvolveArt, line 305: create issue pls so we can track these

crates/client/block_import/src/pre_validate.rs
EvolveArt: agreed that it should probably live in its own crate and also with more unit tests

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+    validation: &Validation,
+    declared_classes: Vec<DeclaredClass>,
+) -> Result<Vec<ConvertedClass>, BlockImportError> {
+    declared_classes.into_par_iter().map(|class| class_conversion(validation, class)).collect()
+}
+
+fn class_conversion(_validation: &Validation, class: DeclaredClass) -> Result<ConvertedClass, BlockImportError> {
+    let DeclaredClass { class_hash, contract_class, compiled_class_hash } = class;
+
+    // TODO(class_hash): uncomment this when the class hashes are computed correctly accross the entire state
```
EvolveArt: add issue in todo here too

crates/client/block_import/src/pre_validate.rs
```diff
@@ -0,0 +1,336 @@
...
+) -> Result<Vec<Felt>, BlockImportError> {
+    if receipts.len() != transactions.len() {
+        return Err(BlockImportError::TransactionEqualReceiptCount {
+            receipts: receipts.len(),
+            transactions: transactions.len(),
+        });
+    }
+
+    // mismatched block hash is allowed for blocks 1466..=2242 on mainnet
+    let is_special_trusted_case =
```
EvolveArt, line 170: that should be noted that an authority node cannot be run with the chain id `SN_MAIN` then or this will be triggered

crates/client/block_import/src/verify_apply.rs
```diff
@@ -0,0 +1,281 @@
...
+    RayonPool, UnverifiedHeader, ValidatedCommitments, Validation,
+};
+
+mod classes;
+mod contracts;
+
+pub struct VerifyApply {
+    pool: Arc<RayonPool>,
+    backend: Arc<DeoxysBackend>,
+    // Only one thread at once can verify_apply.
```
EvolveArt: why is that?

crates/client/block_import/src/verify_apply.rs
```diff
@@ -0,0 +1,281 @@
...
+    let global_state_root = update_tries(backend, &block, block_number)?;
+
+    // Block hash
+    let (block_hash, header) = block_hash(&block, &validation, block_number, parent_block_hash, global_state_root)?;
+
+    // store block, also uses rayon heavily internally
+    backend
+        .store_block(
+            DeoxysMaybePendingBlock {
+                info: DeoxysMaybePendingBlockInfo::NotPending(DeoxysBlockInfo {
```
EvolveArt, line 81: `NotPending` might not be the best name, but rather `Closed` or `Build` wdyt ?

crates/client/mempool/src/block_production.rs
```diff
@@ -325,14 +336,20 @@ impl BlockProductionTask {
...
+            &self.importer,
+            block_to_close,
+            &new_state_diff,
+            self.backend.chain_config().chain_id.clone(),
+            block_n,
+        )
+        .await?;
+        self.block.info.header.parent_block_hash = import_result.block_hash; // fix temp parent block hash for new pending :)
+
+        // self.backend.store_block(closed_block.into(), new_state_diff, declared_classes)?;
```
EvolveArt: remove comment

crates/client/mempool/src/block_production.rs
```diff
@@ -308,7 +319,7 @@ impl BlockProductionTask {
         Ok(())
```
EvolveArt, line 319: so ignoring the first tick means 

 BLOCK N BLOCK N+1
|.......|. -> this one is ignored or just the | so the time block is finalized?

For this to work we have to make sure both clocks are synchronized tho

EvolveArt: Also `continue_block_production` is probably a better name

jbcaron (approved):

crates/client/block_import/src/pre_validate.rs
cchudant: [#35 on bonsai-trie](https://github.com/madara-alliance/bonsai-trie/issues/35)

crates/client/block_import/src/pre_validate.rs
cchudant: i don't know if adding unit tests for that would be *that* useful this is the very basic usage of bonsai-trie that is already tested inside the bonsai-trie crate

i don't know if making a crate is a good idea, i would actually prefer if we lowered the number of primitive crates we currently have
and also i don't see why any other crate would ever want to compute a transaction commitment except the crate creating and validating blocks

crates/client/block_import/src/verify_apply.rs
cchudant: we can't, because the db needs the block hash and global_state_root, and we don't have the global_state_root until we have computed the trie
we could overlap some of the db writes inside dc-db, like, everything except the actual block header which we could store last once we have the global state root, but this would make the interface between dc-db and block-import a mess i think
Each of these steps are parallelized internally, so I don't think we would see much of a perf change at all

I am adding a comment pointing to this potential improvement so that we can potentially investigate it deeper if we ever want to but I'm pretty sure this is not worth it

crates/client/block_import/src/verify_apply.rs
cchudant: so, this was the main motivation for separating conversion (now pre-validate) and verify_l2/store_bock (now verify apply) initially:
The global tries necessarily need their updates to be sequential and in order
If a block A changes key K1, the hash of every of its parent nodes need to be updated and recomputed in the global tree, and block B cannot be processed before A is finished because B needs to be aware of that key

The goal of the pre-validate step is to be parallel over blocks, we only compute stuff that donesnt rely on db there to make the verify_apply task lighter. The verify-apply task really is the choke-point then

these names suck by the way if you have better names i'm down to change

crates/client/block_import/src/verify_apply.rs
cchudant: i've been thinking about this for some time, in starknet-types-core they say MaybePendingBlockInfo::Block/Pending and i kind of don't like it
I really like Closed but I don't know if it's that much better

crates/client/mempool/src/block_production.rs
cchudant: > For this to work we have to make sure both clocks are synchronized tho

I think I need to add a a reset to the pending block interval yes, something like
```rs
instant = interval_block_time.tick() => {
    if let Err(err) = self.produce_block().await {
        log::error!("Block production task has errored: {err:#}");
    }
    // this sets the next deadline to ensure the pending block tick and block time always match up
    interval_pending_block_update.reset_at(instant + interval_pending_block_update.period());
},
```
i'd like to have tests around that part that use tokio::time::advance to skip block times, and see how well this works because i'm almost sure i missed other issues

> Also continue_block_production is probably a better name

there is already a function named "continue_block" :(

cchudant: we also talked about removing tokio entirely in the block production because of DoS issues, right now i'm leaning towards making another tokio thread pool with higher thread priorities instead but we may switch all of this to hand written non async timers
I think just having a dedicated tokio pool is fine, I don't see how this would have any impact on perf vs a dedicated thread even if we don't use it that much

crates/client/block_import/src/verify_apply.rs
EvolveArt: Definitely is

antiyro (approved):

PR #235, opened 2024-08-23
feat: class import

### LINKED ISSUE #233: handling missing old_declared_contracts on mainnet state updates before block 2700
#### Context:
On the mainnet, state updates up to block 2700 (excluded), it has been observed that newly declared Cairo `class_hashes` are not included in the `old_declared_contracts` field. This omission creates a challenge: these `class_hashes` cannot be retroactively added to the `old_declared_contracts` field, as doing so would alter the state root of already produced blocks.

#### Issue:
The current solution involves scanning all deployed contract addresses to retrieve the class_hashes during the import of these blocks, then verifying their existence in the database and fetching them if necessary.

#### Proposed Solution:
I propose to include a hardcoded mapping in the binary, specifically a table that associates each block number with the corresponding list of `class_hashes`.

#### Implementation:
- This mapping will be embedded directly in the binary, ensuring that when these blocks are imported, the `class_hashes` can be fetched directly from this table rather than through a secondary, error-prone process.
- The table will be used solely for blocks up to 2700 (excluded), where the `old_declared_contracts` field is missing.

#### Benefits:
- **Efficiency**: Directly fetching `class_hashes` from a predefined table will significantly reduce the overhead associated with importing these blocks.
- **Accuracy**: This solution eliminates the risk of missing or incorrectly fetching `class_hashes` since the data is predefined and associated with the correct block numbers.

#### Conclusion:
This issue aims to streamline the handling of state updates before block 2700 by providing a robust, binary-level solution to the issue of missing `old_declared_contracts`. I recommend this approach as it provides a clear, efficient, and secure method for importing these blocks.

Please provide feedback on the proposed solution.

jbcaron:
## Pull Request type

Resolves: #233

## What is the new behavior?

- computation hash management: now based on Starknet version instead of block numbers
- database changes: removed the legacy JSON class from the database
- legacy class decompression: decompress the legacy class and create a blockifier class just-in-time for execution
- sierra class compilation: sierra class is now compiled using our custom type
- class fetch optimization: optimized class fetching through FGW


crates/client/block_import/src/pre_validate.rs
```diff
@@ -166,9 +187,10 @@ fn transaction_hashes(
         });
     }
 
-    // mismatched block hash is allowed for blocks 1466..=2242 on mainnet
-    let is_special_trusted_case =
-        validation.chain_id == ChainId::Mainnet && block_n.is_some_and(|n| (1466..=2242).contains(&n));
+    // mismatched block hash is allowed for block 1469 on mainnet
+    // this block contains a part of transactions computed with the legacy hash function
+    // and the other part with the post-legacy hash function
+    let is_special_trusted_case = validation.chain_id == ChainId::Mainnet && starknet_version.is_tx_hash_inconsistent();
```
Trantorian1, line 193: Should we not be making sure the block number is 1469? This seems to leave room for invalid state in the future if we are fed incorrect data.

crates/client/block_import/src/pre_validate.rs
```diff
@@ -136,28 +138,47 @@ fn convert_classes(
...
+                });
+            }
+            Ok(ConvertedClass::Sierra(SierraConvertedClass {
+                class_hash: sierra.class_hash,
+                info: SierraClassInfo { contract_class: Arc::new(sierra.contract_class), compiled_class_hash },
+                compiled: Arc::new(compiled_class),
+            }))
+        }
+        DeclaredClass::Legacy(legacy) => {
+            // TODO: verify that the class hash is correct
```
Trantorian1, line 168: Are there any issues in currently checking the class hash for Legacy classes?

crates/client/db/src/class_db.rs
```diff
@@ -85,91 +83,81 @@ impl DeoxysBackend {
...
+            })
+            .collect::<Vec<_>>()
+            .par_chunks(DB_UPDATES_BATCH_SIZE)
+            .try_for_each_init(
+                || self.db.get_column(col_compiled),
+                |col, chunk| {
+                    let mut batch = WriteBatchWithTransaction::default();
+                    for (key, value) in chunk {
+                        let key_bin = bincode::serialize(key)?;
+                        // TODO: find a way to avoid this allocation
```
Trantorian1, line 152: Which allocation?

crates/client/db/src/class_db.rs
```diff
@@ -85,91 +83,81 @@ impl DeoxysBackend {
         Ok(self.get_class_info(id, class_hash)?.is_some())
     }
 
-    pub fn get_class(
+    pub fn get_sierra_compiled(
```
Trantorian1, line 86: This is great, much cleaner than before :fire:

crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -111,16 +111,35 @@ impl StateReader for BlockifierStateAdapter {
...
+                    .backend
+                    .get_sierra_compiled(&on_top_of_block_id, &info.compiled_class_hash)
+                    .map_err(|err| {
+                        log::warn!("Failed to retrieve sierra compiled class {class_hash:#}: {err:#}");
+                        StateError::StateReadError(format!("Failed to retrieve compiled class {class_hash:#}"))
+                    })?
+                    .ok_or(StateError::StateReadError(format!(
+                        "Inconsistent state: compiled sierra class {class_hash:#} not found"
+                    )))?;
+                // TODO: convert ClassCompilationError to StateError
```
Trantorian1, line 135: Hasn't this been implemented in the below `map_err`?

crates/client/sync/src/fetch/fetchers.rs
```diff
@@ -277,52 +269,58 @@ async fn fetch_class_updates(
...
-        },
-    ))
-    .await?;
+) -> anyhow::Result<Vec<ClassUpdate>> {
+    let chain_id: Felt = backend.chain_config().chain_id.to_felt();
+
+    // for blocks before 2597 on mainnet new classes are not declared in the state update
+    // https://github.com/madara-alliance/madara/issues/233
+    let legacy_classes: Vec<_> = if chain_id == MAIN_CHAIN_ID && block_id.block_n() < Some(2597) {
+        let block_number = block_id.block_n().unwrap(); // Safe to unwrap because of the condition above
```
Trantorian1, line 278: Wouldn't it still be possible for this to be `None`? I'm assuming `None < Some(2597) == true`. If that is the case we might want to consider unwrapping to the maximum integer value of a block number.

crates/client/sync/src/fetch/fetchers.rs
```diff
@@ -277,52 +269,58 @@ async fn fetch_class_updates(
...
+        MISSED_CLASS_HASHES.get(&block_number).cloned().unwrap_or_default()
```
Trantorian1, line 279: Accessing a non-garded hashmap from within an async function seems pretty risky. Consider adding a lock (unless this is the only place this is being used, which I might be missing).

crates/primitives/class/src/lib.rs
```diff
@@ -1,58 +1,125 @@
...
+pub struct SierraClassInfo {
+    pub contract_class: Arc<FlattenedSierraClass>,
     pub compiled_class_hash: Felt,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum ContractClass {
-    Sierra(FlattenedSierraClass),
-    Legacy(CompressedLegacyContractClass),
+    Sierra(Arc<FlattenedSierraClass>),
```
Trantorian1, line 85: Why the added `Arc`?

crates/primitives/class/resources/missed_classes.json
Trantorian1, line 1: Potentially a missed ignore?

Trantorian1 (commented):
Seems good to me. Please check the below comments. It might also be a good idea to have someone like @cchudant check out this PR if you feel like you need a deeper look at the underlying logic of this part of the code, since I feel a bit rusty in this regard.

crates/client/sync/src/fetch/fetchers.rs
jbcaron: None for pending

crates/primitives/class/resources/missed_classes.json
jbcaron: no it's the legacy classes which are not in the states diff of the first blocks of the mainnet

crates/client/sync/src/fetch/fetchers.rs
jbcaron: it's a static reference

Trantorian1: Still though, nothing prevents a write from occurring at the same time as a read to it.

crates/client/block_import/src/pre_validate.rs
cchudant: so - the protocol version field is absent from blocks before (idk which version) on mainnet.
We used to carry the block number around everywhere and decide the legacy special cases and hash compute based on that, along with the protocol version, and chain id as a trio

this kinda sucked, so now we decide when to apply the legacy special cases based on starknet version only, no block number. We decide the starknet version when it is absent from the block number and chain id.
However these blocks are so old that even people from starkware don't know what version they are supposed to be - so we came up with fake versions and filled in the gaps :)

@jbcaron I think we should remove the chain id check here, as these fake protocol version shouldnt be able to be used outside of mainnet

Can you also document what I just said in the StarknetVersion?

crates/client/db/src/class_db.rs
cchudant: bincode::serialize

cchudant: returns a Vec in this case

cchudant: avoiding this alloc is easy we just have to use smallvec tbh, but I never got to doing it

crates/client/sync/src/fetch/fetchers.rs
cchudant: pending blocks don't have a block number (for no reason in particular, actually)
can you use expect() here though?

crates/primitives/chain_config/src/starknet_version.rs
```diff
@@ -16,12 +22,45 @@ impl StarknetVersion {
         StarknetVersion([major, minor, patch, build])
     }
 
-    pub const STARKNET_VERSION_0_11_1: StarknetVersion = StarknetVersion([0, 11, 1, 0]);
-    pub const STARKNET_VERSION_0_13_0: StarknetVersion = StarknetVersion([0, 13, 0, 0]);
-    pub const STARKNET_VERSION_0_13_1: StarknetVersion = StarknetVersion([0, 13, 1, 0]);
-    pub const STARKNET_VERSION_0_13_1_1: StarknetVersion = StarknetVersion([0, 13, 1, 1]);
-    pub const STARKNET_VERSION_0_13_2: StarknetVersion = StarknetVersion([0, 13, 2, 0]);
-    pub const STARKNET_VERSION_LATEST: StarknetVersion = Self::STARKNET_VERSION_0_13_2;
+    pub const VERSION_NONE: StarknetVersion = StarknetVersion([0, 0, 0, 0]);
```
cchudant: what's version none?

crates/primitives/chain_config/src/starknet_version.rs
```diff
@@ -16,12 +22,45 @@ impl StarknetVersion {
...
+    pub const VERSION_0_7_0: StarknetVersion = StarknetVersion([0, 7, 0, 0]);
```
cchudant: I think StarknetVersion::V0_7_0 would be even better of a name

crates/primitives/class/src/class_hash.rs
```diff
@@ -1,300 +1,384 @@
...
-            assert_eq!(legacy.class_hash().unwrap(), class_hash);
-        } else {
-            panic!("Not a Lecacy contract");
-        }
+        println!("computed_class_hash in {:?}", start.elapsed());
+        assert_eq!(computed_class_hash, class_hash);
     }
 }
+
+// use anyhow::{Context, Error, Ok, Result};
```
cchudant, line 85: should we remove that code?

also i think we should move toward tdelabro's hash chain struct, it's pretty cool

cchudant (commented):
finally, class hashes!! :partying_face:

crates/client/block_import/src/pre_validate.rs
Trantorian1: Great, thanks for the clarification.

Trantorian1 (approved):

cchudant (approved):

PR #236, opened 2024-08-26
e2e tests: RPC read test cases

Mohiiit:
PR includes tests for the all rpc read calls, in all cases we will be syncing a few blocks from the sepolia testnet and will make read calls and verify the information present in madara is right or not and whether the rpc calls are working fine or not.

For source of truth, I am using juno's data.

# Pull Request type

- Testing

Resolves: #

## What is the new behavior?

- tests has been added in test directory under the rpc folder, plan is to add write test cases there as well in the future

 No 

## Progress

- [x] get block hash and block number
- [ ] call (ignored)
- [ ] estimate fee (ignored)
- [ ] estimate message fee (ignored)
- [x] get block transaction count
- [x] get block with receipt
- [x] get block with transaction hashes
- [x] get block with transactions
- [x] get class
- [x] get class at
- [x] get class hash at
- [x] get state update
- [x] get nonce
- [x] get events
- [x] get storage at
- [x] get transaction by block id and index
- [x] get transaction by hash
- [x] get transaction receipt
- [x] get transaction status


crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1038 @@
...
+    use starknet_providers::Provider;
+    use std::fmt::Write;
+    use std::fs::File;
+    use std::io::BufReader;
+    use std::io::Read;
+    use tokio::sync::{Mutex, OnceCell};
+
+    static MADARA: Lazy<OnceCell<Arc<Mutex<MadaraCmd>>>> = Lazy::new(OnceCell::new);
+
+    async fn setup_madara() -> Arc<Mutex<MadaraCmd>> {
```
apoorvsadana: this could be a fixture. do we need Arc or Mutex? does a #[once] fixture autmatically handle it?

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1038 @@
...
+            block_hash: Felt::from_hex_unchecked("0x7a906dfd1ff77a121b8048e6f750cda9e949d341c4487d4c6a449f183f0e61d"),
+            parent_hash: Felt::from_hex_unchecked("0x78b67b11f8c23850041e11fb0f3b39db0bcb2c99d756d5a81321d1b483d79f6"),
+            block_number: 2,
+            new_root: Felt::from_hex_unchecked("0xe005205a1327f3dff98074e528f7b96f30e0624a1dfcf571bdc81948d150a0"),
+            timestamp: 1700475581,
+            sequencer_address: Felt::from_hex_unchecked(
+                "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8",
+            ),
+            l1_gas_price: ResourcePrice {
+                price_in_fri: Felt::from_hex_unchecked("0x0"),
```
apoorvsadana: should we make `Felt::from_hex_unchecked` a macro?

crates/tests/src/rpc/read.rs
apoorvsadana, line 1: we should probably add a comment somewhere that these were taken from Juno

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1038 @@
...
+        };
+
+        assert_eq!(decompressed_program, original_program);
+        assert_eq!(contract_entry_points, expected_contract_entry_points);
+        assert_eq!(contract_abi, expected_contract_abi);
+    }
+
+    #[tokio::test]
+    async fn test_get_class_at_works() {
+        let madara = get_shared_state().await;
```
apoorvsadana, line 943: a lot of common code between this and above, should we make an `assert_contract_class` function?

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1038 @@
...
+               }
+           },
+           "id": 1
+       }'
+
+       {"jsonrpc":"2.0","result":[{"gas_consumed":"0xcfc","gas_price":"0x3b9ada0f","data_gas_consumed":"0x0","data_gas_price":"0x1","overall_fee":"0x305eea75ac4","unit":"WEI"}],"id":1}%
+    */
+    #[ignore]
+    #[tokio::test]
+    async fn test_estimate_fee_works() {
```
apoorvsadana, line 1005: wait so are these working?

Mohiiit: added ignore on these, the data might be helpful in case we test it again

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1209 @@
...
+               "block_id": "latest"
+           },
+           "id": 1
+       }'
+    */
+    #[tokio::test]
+    async fn test_get_block_txn_with_tx_works() {
+        let madara = get_shared_state().await;
+        let json_client = JsonRpcClient::new(HttpTransport::new(madara.rpc_url.clone()));
+        let block = { json_client.get_block_with_txs(BlockId::Number(2)).await.unwrap() };
```
cchudant: fmt here, for braces

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1209 @@
...
+
+    async fn setup_madara() -> MadaraCmd {
+        let mut madara = MadaraCmdBuilder::new()
+            .args(["--network", "sepolia", "--no-sync-polling", "--n-blocks-to-sync", "20", "--no-l1-sync"])
+            .run();
+
+        madara.wait_for_ready().await;
+        madara.wait_for_sync_to(19).await;
+
+        // The lock is released when guard goes out of scope
```
cchudant: i dont understand this comment

cchudant (approved):

crates/tests/src/rpc/read.rs
Mohiiit: resolved

crates/tests/src/rpc/read.rs
Mohiiit: done

crates/tests/src/rpc/read.rs
Mohiiit: done

crates/tests/src/rpc/read.rs
Mohiiit: my bad, removed

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1207 @@
...
+               ],
+               "simulation_flags": [ "SKIP_VALIDATE"],
+               "block_id": {
+                   "block_number": 1
+               }
+           },
+           "id": 1
+       }'
+
+       Note: The test is ignored as of now because madara doesn't support historical data retrieval.
```
antiyro: We support historical data retrieval. We dont support it trough our bonsai trie. But all RPC endpoints are supported on latest (v0.7.1)

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1207 @@
...
+                   "calldata": []
+               },
+               "block_id": {
+                   "block_number": 19
+               }
+           },
+           "id": 1
+       }'
+
+       Note: The test is ignored as of now because madara doesn't support historical data retrieval.
```
antiyro: same here

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1207 @@
...
+                   "payload": ["755459732925565255818828693099006464755347258332","0","0"]
+               },
+               "block_id": {
+                   "block_number": 8
+               }
+           },
+           "id": 1
+       }'
+
+       Note: The test is ignored as of now because madara doesn't support historical data retrieval.
```
antiyro: same here

crates/tests/src/rpc/read.rs
```diff
@@ -0,0 +1,1207 @@
...
+            l1_da_mode: L1DataAvailabilityMode::Calldata,
+            starknet_version: "0.12.3".to_string(),
+            transactions: vec![Felt::from_hex_unchecked(
+                "0x701d9adb9c60bc2fd837fe3989e15aeba4be1a6e72bb6f61ffe35a42866c772",
+            )],
+        });
+        assert_eq!(block, expected_block);
+    }
+
+    /*
```
antiyro: I would rather comment doc using `///` since we are doing it this way all over the codebase:

```rust
/// Sends a request to fetch the latest block with transactions.
/// 
/// Example curl command:
/// 
/// ```bash
/// curl --location 'https://free-rpc.nethermind.io/sepolia-juno/' \
/// --header 'Content-Type: application/json' \
/// --data '{
///     "jsonrpc": "2.0",
///     "method": "starknet_getBlockWithTxs",
///     "params": {
///         "block_id": "latest"
///     },
///     "id": 1
/// }'
/// ```
```

antiyro: Also I would avoid referencing `https://free-rpc.nethermind.io/sepolia-juno/` and replace it with a madara public rpc (we have this one here https://rpc.kasar.io/deoxys)

antiyro: actually it is on mainnet

antiyro: so lets rely on nethermind for now

crates/tests/src/rpc/read.rs
Mohiiit: sorry, my bad what I meant was the calls with starknet version <0.13.0, this returns an error.

```
if block.info.protocol_version() < &FALLBACK_TO_SEQUENCER_WHEN_VERSION_BELOW {
        return Err(StarknetRpcApiError::UnsupportedTxnVersion);
    }
```

crates/tests/src/rpc/read.rs
Mohiiit: done

crates/tests/src/rpc/read.rs
Mohiiit: updated the comment

crates/tests/src/rpc/read.rs
Mohiiit: documentation updated

apoorvsadana (approved):

PR #237, opened 2024-08-27
dev: Block import refactoring

akhercha:
# Pull Request type

Resolves: #NA

### Added
- `Validate` trait implementation for `UnverifiedPendingFullBlock` and `UnverifiedFullBlock` 
- `tx_batch_size` as CLI parameter

### Updated
- `types` of `block_import` moved to `primitives`,
- rayon pool moved to `primitives`

## Not so sure about
- new block primitives - if they'll be used only in the `block_import` crate, should they live there? If we're gonna use them elsewhere it should be fine in `primitives`.
- `compute_merkle_root` & `transaction_hashes` + `TransactionHashesError` are currently in `primitives/block/src/lib.rs`. Where should they live?

Any thoughts? @cchudant


README.md
```diff
@@ -195,6 +195,14 @@ Here are more advanced command-line options, organized by namespace, for running
 
 </details>
 
+<details>
+<summary>Block Production</summary>
+
+- **`--no-block-production`**: Disable the block production service,
```
cchudant: we need to enable the --authority flag to enable block production mode, and we dont want --no-block-production in the readme I think
running authority mode + no block production doesn't really make sense unless you're running some weird configuration for an end to end test or something like that

crates/primitives/validation/src/lib.rs
cchudant, line 1: I am not really convinced by the Validate trait and primitive crate, I think it's not really making anything cleaner and we end up with another primitive crate with a single lib.rs file
The trait expresses the ability to `validate` something with a context, but it's only implemented by 2 operations that are called differently anyway, and that trait is very specific to block import (see how there is a `spawn_validate` in the trait - that's very specific to block-import internals right?
I think the multiplicity of crates like that may actually hurt us more than help us too, not just compile speeds but also in cleanliness

@jbcaron @EvolveArt @apoorvsadana what do you think?

crates/primitives/block/src/deoxys_block.rs
cchudant, line 1: we really should rename DeoxysBlock at some point x)

crates/primitives/block/src/lib.rs
```diff
@@ -127,170 +81,388 @@ impl From<BlockId> for starknet_core::types::BlockId {
...
+    pub transactions: Vec<Transaction>,
+    pub state_diff: StateDiff,
+    pub receipts: Vec<TransactionReceipt>,
+    pub converted_classes: Vec<ConvertedClass>,
+}
 
-        assert_eq!(pending_as_maybe_pending.as_block_id(), BlockId::Tag(BlockTag::Pending));
-        assert_eq!(not_pending_as_maybe_pending.as_block_id(), BlockId::Number(0));
+// ===================
+// TODO(akhercha) : move those *somewhere* else... not sure yet
```
cchudant, line 386: is that a todo you want fix before merge?

crates/primitives/validation/src/lib.rs
cchudant: mmh i am starting to change my mind, maybe this isnt that bad

I think in the future we should really merge dp-validate, dp-block, dp-state-update, dp-transaction, dp-receipt into a single primitive crate

crates/primitives/validation/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+use dp_rayon_pool::RayonPool;
+
+#[async_trait::async_trait]
+pub trait Validate: Send + Sync + Sized + 'static {
+    type Output: Send + 'static;
+    type Context: Send + 'static;
+    type ValidationError: Send;
+
+    // TODO: context should be optional
```
cchudant: remove that todo imo, if we want it to be optional we'd just put `()` as the context, thats the cleanest way to do it right?

crates/primitives/block/src/lib.rs
```diff
@@ -127,170 +81,388 @@ impl From<BlockId> for starknet_core::types::BlockId {
...
+            event_commitment: event_c,
+            state_diff_length: self.state_diff.len() as _,
+            state_diff_commitment: state_diff_c,
+            receipt_commitment: receipt_c,
+        })
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct BlockValidationContext {
```
cchudant, line 269: i like that name thank you

crates/primitives/validation/src/lib.rs
cchudant: I think I would feel better about Validate trait if it wasnt as mixed between "very specific to block import" (spawn_validate and rayon-specific) and "very abstract" with Context being an associated type

But I still think we should just remove that trait entirely, it's only implemented by UnverifiedFullBlock and UnverifiedFullPendingBlock, there is not really a reason to have a trait

cchudant (changes requested):

crates/primitives/validation/src/lib.rs
akhercha: Ah yup truee, exactly what I was looking for, was only seeing the `Option<Context>` and I hated it

crates/primitives/block/src/lib.rs
akhercha: If possible yup - that's related to the note at the bottom of my PR description, I could probably move them to the `transaction` primitive crate, but I like that they're close to where they're actually *only* used.

Kinda relates to the whole block primitives stuff I did - maybe it's just better in `block_import` like you did (close to where it's only used).

It really just depends on the future - if they'll be used elsewhere the primitives stuff make sense, else I think I'll revert that.
Wdyt?

README.md
akhercha: Oh yeah make sense - removed that part from the README for now

crates/primitives/validation/src/lib.rs
EvolveArt: I feel like it still makes sense although instead of implementing it just for the Block structures it could be implemented by the underlying structs that's how I imagined it. So transaction hashes, receipts..
In that way it makes more sense, cause let's say tomorrow there is a new field in the block that needs validation on syncing we just need to implement the trait on this struct and we're good.

akhercha:
PR closed as the changes don't have unanimous approval. We'll keep the refactor for later and if the validate trait is necessary for tests, we'll re-implement it in the associated PRs.

PR #238, opened 2024-08-27
feat: devnet

cchudant:
# Pull Request type

- Feature

## What is the new behavior?

Make a genesis block and add a simple test

No


cchudant:
i ended up going back and forth a lot on the approach, and i'm still quite not happy with it

cchudant:
i think it's in a state where you could review it but my two tests are kinda verbose i'll fix that real quick

cchudant:
i also added the launching of anvil in the `scripts/e2e-coverage.sh`, if you didnt know that script existed it just launches all the tests and outputs coverage

cairo/target/dev/madara_contracts_AccountUpgradeable.contract_class.json
```diff
@@ -0,0 +1 @@
...m","variants":[{"name":"OwnerAdded","type":"openzeppelin_account::account::AccountComponent::OwnerAdded","kind":"nested"},{"name":"OwnerRemoved","type":"openzeppelin_account::account::AccountComponent::OwnerRemoved","kind":"nested"}]},{"type":"event","name":"openzeppelin_introspection::src5::SRC5Component::Event","kind":"enum","variants":[]},{"type":"event","name":"openzeppelin_upgrades::upgradeable::UpgradeableComponent::Upgraded","kind":"struct","members":[{"name":"class_hash","type":"core::starknet::class_hash::ClassHash","kind":"data"}]},{"type":"event","name":"openzeppelin_upgrades::upgradeable::UpgradeableComponent::Event","kind":"enum","variants":[{"name":"Upgraded","type":"openzeppelin_upgrades::upgradeable::UpgradeableComponent::Upgraded","kind":"nested"}]},{"type":"event","name":"madara_contracts::account::AccountUpgradeable::Event","kind":"enum","variants":[{"name":"AccountEvent","type":"openzeppelin_account::account::AccountComponent::Event","kind":"flat"},{"name":"SRC5Event","type":"openzeppelin_introspection::src5::SRC5Component::Event","kind":"flat"},{"name":"UpgradeableEvent","type":"openzeppelin_upgrades::upgradeable::UpgradeableComponent::Event","kind":"flat"}]}]}
```
EvolveArt, line 1: I don't think compilation artifacts should be commited.

crates/client/mempool/src/block_production.rs
```diff
@@ -250,8 +251,11 @@ impl BlockProductionTask {
                     self.block.inner.transactions.push(converted_tx.transaction);
                 }
                 Err(err) => {
-                    // TODO: revert handling
-                    log::error!("Unsuccessful execution of transaction {:?}: {err:#}", mempool_tx.tx_hash());
+                    // These are the transactions that have errored but we can't revert them. It can be because of an internal server error, but
+                    // errors during the execution of Declare and DeployAccount also appear here as they cannot be reverted.
+                    // We reject them.
+                    // Note that this is a big DoS vector.
```
EvolveArt, line 263: lets create an issue on this

crates/client/mempool/src/genesis/balances.rs
```diff
@@ -0,0 +1,48 @@
...
+            let erc20_contract = chain_config.parent_fee_token_address;
+            let kv = storage_diffs.contract_mut(erc20_contract);
+            kv.insert(low_key, bal.wei);
+            // kv.insert(high_key, Felt::ZERO);
+
+            // STRK erc20
+            let erc20_contract = chain_config.native_fee_token_address;
+            let kv = storage_diffs.contract_mut(erc20_contract);
+            kv.insert(low_key, bal.fri);
+            // kv.insert(high_key, Felt::ZERO);
```
EvolveArt: why do we ignore high value ?

crates/client/mempool/src/genesis/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    use rstest::{fixture, rstest};
+    use starknet_core::types::{
+        BroadcastedDeclareTransaction, BroadcastedDeployAccountTransaction, BroadcastedInvokeTransaction,
+        BroadcastedInvokeTransactionV3, BroadcastedTransaction, DeclareTransactionResult,
+        DeployAccountTransactionResult, InvokeTransactionResult, ResourceBounds, ResourceBoundsMapping,
+    };
+
+    use std::sync::Arc;
+
+    struct PreparedDevnet {
```
EvolveArt: I think that's pretty cool that this live in its own crate under a feature flag ?

crates/client/mempool/src/genesis/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+            })
+            .collect()
+    }
+}
+
+// We allow ourselves to lie about the contract_address. This is because we want the UDC and the two ERC20 contracts to have well known addresses on every chain.
+
+/// Universal Deployer Contract.
+const UDC_CLASS_DEFINITION: &[u8] =
+    include_bytes!("../../../../../cairo/target/dev/madara_contracts_UniversalDeployer.contract_class.json");
```
EvolveArt: let's define a proper `CAIRO_CONTRACTS_PATH` constant

crates/client/mempool/src/genesis/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+            .iter()
+            .map(|(contract, map)| ContractStorageDiffItem {
+                address: contract.to_felt(),
+                storage_entries: map.iter().map(|(key, &value)| StorageEntry { key: key.to_felt(), value }).collect(),
+            })
+            .collect()
+    }
+}
+
+// We allow ourselves to lie about the contract_address. This is because we want the UDC and the two ERC20 contracts to have well known addresses on every chain.
```
EvolveArt: should be documented in a readme tho

EvolveArt (commented):
I see there is a test using some kind of `PreparedDevnet`, is Genesis usable with an node in authority mode yet ?

cairo/target/dev/madara_contracts_AccountUpgradeable.contract_class.json
cchudant: i personally think it's fine and just way simpler for everyone, but that may be a controversial take
i've worked on projects in the past where those kind of unimportant compilation artifacts were committed and it was soo much simpler

Here I think I'll hook up scarb to a build.rs file and compile these contracts when doing `cargo build`

crates/client/mempool/src/block_production.rs
cchudant: this is how starknet is designed and we'll have to talk about it in full node discussions -- we can't fix that on our end and it reaaaally changes a lot of things

I believe the whole sierra to casm compiler would need to be rewritten entirely in cairo. so that class compilation is factored in the fees and the fees are verifiable in SNOS.

crates/client/mempool/src/genesis/balances.rs
cchudant: the erc20 spec requires support up to u256 but it doesn't fit into a Felt as a Felt is 252bit - that's why the cairo compiler will end up splitting u256's in two
however there are places where blockifier treats STRK and ETH amounts as u128, and so we don't need the high bits at all and STRK/ETH wouldnt need any of that, as do any appchain (for fee token contracts at least) I'm pretty sure

I'll have to recheck blockifier I think they semi-support it

crates/client/mempool/src/genesis/mod.rs
cchudant: this for tests only, you mean you want that for appchain developers?

crates/client/mempool/src/genesis/mod.rs
cchudant: i tried that and `concat!` to concat the path did not work with include_bytes

cchudant: that's devastatingly sad

crates/client/mempool/src/genesis/mod.rs
cchudant: I'm in the process of making an issue about that, but I need to understand the requirements from the snos/proving perspective. I think lying about these will need them to special case this on their end too, and that might be a pain.
I'll forward the slack message to you

cairo/target/dev/madara_contracts_AccountUpgradeable.contract_class.json
EvolveArt: If we do it in the build.rs file I'm fine with it

crates/client/mempool/src/genesis/mod.rs
EvolveArt: I mean if people want to use madara only as a devnet like they would use katana

cairo/target/dev/madara_contracts_AccountUpgradeable.contract_class.json
cchudant: done :)

antiyro (approved):

EvolveArt (approved):

PR #242, opened 2024-09-03
feat: RPC Versioning

akhercha:
# Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?

Introduction of versions for our RPC API.

Versions are defined in the `mc-rpc::versions` module, organized as follow:
```bash
versions
├── mod.rs
└── v0_7_1
    ├── mod.rs
    ├── api.rs               # contains the API definition of the current version
    └── methods              # contains all the methods implementation
         ├── mod.rs
         ├── read/
         ├── trace/
         └── write/
```

All the implementations are then merged in our `RpcModule` in the `versioned_rpc_api` function.

Methods will be version-prefixed to avoid collisions, for example:
* `starknet_V0_7_1_blockNumber`,
* `starknet_V0_8_0_blockNumber`
* etc...

The "routing" is done in the `VersionMiddlewareLayer` middleware, where depending on the version provided in the route by the request's user, we forward to the right method.

If no version is provided, the latest RPC version available is used, defined in `RpcVersion::RPC_VERSION_LATEST`.

No

## Possible improvement

I don't like the current process of adding a new version ; it is:

1. Create a new constant in the `RpcVersion` primitive, like: `pub const RPC_VERSION_0_7_1: RpcVersion = RpcVersion([0, 7, 1]);`
2. Add it in the `SUPPORTED_RPC_VERSIONS`  array,
3. Add the version API + implementation in the `mc-rpc::versions` module,
4. Add the module when merging the API methods:
```rust
pub fn versioned_rpc_api(starknet: &Starknet, read: bool, write: bool, trace: bool) -> anyhow::Result<RpcModule<()>> {
    let mut rpc_api = RpcModule::new(());

    merge_rpc_versions!(
        rpc_api, starknet, read, write, trace,
        v0_7_1, // We can add new versions by adding the version module below
        // , v0_8_0 (for example)
    );
// ...
```

Lot of dependency between objects, the middleware also uses the `SUPPORTED_RPC_VERSIONS` array for routing.

It is fine though because we don't have new versions everyday, that's an uncommon event.
But happy to hear your suggestions!


crates/client/rpc/src/test_utils.rs
```diff
@@ -13,11 +13,41 @@ use dp_state_update::{
...
 
-use crate::{providers::TestTransactionProvider, Starknet};
+use crate::{providers::AddTransactionProvider, Starknet};
+
+#[cfg(test)]
+pub struct TestTransactionProvider;
+
+#[cfg(test)]
+#[async_trait]
+impl AddTransactionProvider for TestTransactionProvider {
```
cchudant, line 31: i pulled in the automock crate in #238 for this kind of thing, what do you think?

crates/proc-macros/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+//!     #[method(name = "blockNumber")]
+//!     fn block_number(&self) -> anyhow::Result<u64>;
+//! }
+//! ```
+//!
+//! The macro will generate this code:
+//! ```rust,ignore
+//! #[rpc(server, namespace = "starknet")]
+//! pub trait JsonRpcV0_7_1 {
+//!     #[method(name = "V0_7_1_blockNumber")]
```
cchudant, line 26: this does not follow the spec unfortunately, we'd like to have a prefix like `/v0.7.1/` in the URL instead - otherwise our rpcs won't be compatible with the rest of the ecosystem
that's also what juno and pathfinder do iirc

cchudant:
i like the approach quite a lot though, it's very good :partying_face:

crates/client/rpc/src/test_utils.rs
akhercha: way better yup!

crates/proc-macros/src/lib.rs
akhercha: As discussed in DMs it's okey - it does not change the API since we have a middleware that proxies the requests depending on the version requested.

Though do we prefer to have endpoints prefixed with `rpc` like -> `http://localhost:9944/rpc/v0.7.0` or do we prefer `http://localhost:9944/v0.7.0`? @antiyro

crates/primitives/chain_config/src/rpc_version.rs
```diff
@@ -0,0 +1,153 @@
...
+    UnsupportedVersion,
+}
+
+impl RpcVersion {
+    pub const fn new(major: u8, minor: u8, patch: u8) -> Self {
+        RpcVersion([major, minor, patch])
+    }
+
+    pub fn from_request_path(path: &str) -> Result<Self, RpcVersionError> {
+        let parts: Vec<&str> = path.split('/').collect();
```
EvolveArt, line 34: can't we use a regex to simplify this function?

akhercha: I try to avoid regexes for performance & safety (like overly permissive regex) - @cchudant wdyt?

Imo it's fine like this but can do it

crates/proc-macros/src/lib.rs
akhercha: Just realized that I'm also not sure if it should be:
* `/rpc/v0.7.0`
* or `/rpc/v0_7_0`

Happy to change - took the dot notation for now since I saw that suggestion on slack.

antiyro: I think this would be cleaner: `http://localhost:9944/rpc/v0_7_0
`

akhercha: Gotcha, updated 🫡

EvolveArt (approved):

akhercha:
Closed in favor of:
https://github.com/madara-alliance/madara/pull/247

Because this branch is coming from our fork the CI doesn't work properly.

PR #243, opened 2024-09-03
fix: replaced old namings with adapted namings

antiyro:
just replaced:

- deoxys->madara
- Deoxys->Madara
- kasarlabs -> madara-alliance

and some details


apoorvsadana (approved):

scripts/deoxys
jbcaron, line 1: rename this file

crates/client/sync/Cargo.toml
```diff
@@ -2,12 +2,12 @@
...
-homepage = "https://github.com/KasarLabs/deoxys"
+homepage = "https://github.com/madara-alliance/madara"
 license = "MIT"
-name = "dc-sync"
+name = "mc-sync"
 publish = false
-repository = "https://github.com/KasarLabs/deoxys"
-version = "0.1.0"
+repository = "https://github.com/madara-alliance/madara"
+version = "0.7.0"
```
jbcaron: why this crate does not take the attributes of the workspace?

jbcaron (changes requested):
In the future it would be nice to rename MadaraBackend, MadaraStorageError... more simply

crates/client/sync/Cargo.toml
antiyro: changed it to match workspace attribute

scripts/deoxys
antiyro: done

jbcaron (approved):

PR #245, opened 2024-09-05
fix: adding the --starting-from arg back by ignoring ordered state

antiyro:
fixing an error happening when syncinc closed blocks when syncing pendings that makes the node crash in full mode

adding the --starting-from arg back by ignoring ordered state


crates/node/src/cli/sync.rs
```diff
@@ -48,6 +48,10 @@ pub struct SyncParams {
     /// Periodically create a backup, for debugging purposes. Use it with `--backup-dir <PATH>`.
     #[clap(long, value_name = "NUMBER OF BLOCKS")]
     pub backup_every_n_blocks: Option<u64>,
+
+    /// Ignore the order of the blocks to allow starting at some height.
+    #[clap(long)]
+    pub ignore_block_order: bool,
```
cchudant: i'd prefer if this was implied when using --starting-block, because i can't see any other case where you would want --starting-block without --ignore-block-order

cchudant: also, i think we should put a huge warning with starting-block saying that it'll definitely break your db. Maybe rename --starting-block to --unsafe-starting-block, because that may be a bit surprising if the user provides --starting-block expecting some kind of fork-testing behavior and getting a totally broken db
and maybe even hide the option in clap hidden because I don't want users to actually use it lol

cchudant (approved):
lgtm :partying_face::partying_face::partying_face:

crates/client/sync/src/lib.rs
```diff
@@ -24,14 +24,18 @@ pub async fn sync(
     telemetry: TelemetryHandle,
     pending_block_poll_interval: Duration,
 ) -> anyhow::Result<()> {
-    let starting_block = if let Some(starting_block) = starting_block {
-        starting_block
+    let (starting_block, ignore_block_order) = if let Some(starting_block) = starting_block {
```
Trantorian1, line 27: A bit hard to read :face_with_spiral_eyes:

crates/client/block_import/src/verify_apply.rs
```diff
@@ -160,7 +162,7 @@ fn check_parent_hash_and_num(
     };
 
     let block_number = if let Some(block_n) = unverified_block_number {
-        if block_n != expected_block_number {
+        if block_n != expected_block_number && !validation.ignore_block_order {
```
Trantorian1, line 165: Doesn't this mean we will still be having the same error unless ignoring block order?

antiyro: yes if for x reason the block order is disrupted then it will throw an error. If unswafe-starting-block is set then we assume you dont want your state to be ordered

Trantorian1 (approved):

PR #247, opened 2024-09-06
feat: RPC Versioning

akhercha:
# Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?

Introduction of versions for our RPC API.

Versions are defined in the `mc-rpc::versions` module, organized as follow:
```bash
versions
├── mod.rs
└── v0_7_1
    ├── mod.rs
    ├── api.rs               # contains the API definition of the current version
    └── methods              # contains all the methods implementation
         ├── mod.rs
         ├── read/
         ├── trace/
         └── write/
```

All the implementations are then merged in our `RpcModule` in the `versioned_rpc_api` function.

Methods will be version-prefixed to avoid collisions, for example:
* `starknet_V0_7_1_blockNumber`,
* `starknet_V0_8_0_blockNumber`
* etc...

The "routing" is done in the `VersionMiddlewareLayer` middleware, where depending on the version provided in the route by the request's user, we forward to the right method.

If no version is provided, the latest RPC version available is used, defined in `RpcVersion::RPC_VERSION_LATEST`.

No

## Possible improvement

I don't like the current process of adding a new version ; it is:

1. Create a new constant in the `RpcVersion` primitive, like: `pub const RPC_VERSION_0_7_1: RpcVersion = RpcVersion([0, 7, 1]);`
2. Add it in the `SUPPORTED_RPC_VERSIONS`  array,
3. Add the version API + implementation in the `mc-rpc::versions` module,
4. Add the module when merging the API methods:
```rust
pub fn versioned_rpc_api(starknet: &Starknet, read: bool, write: bool, trace: bool) -> anyhow::Result<RpcModule<()>> {
    let mut rpc_api = RpcModule::new(());

    merge_rpc_versions!(
        rpc_api, starknet, read, write, trace,
        v0_7_1, // We can add new versions by adding the version module below
        // , v0_8_0 (for example)
    );
// ...
```

Lot of dependency between objects, the middleware also uses the `SUPPORTED_RPC_VERSIONS` array for routing.

It is fine though because we don't have new versions everyday, that's an uncommon event.
But happy to hear your suggestions!


antiyro (approved):
Approving this since it has been reviewed correctly in #242

Trantorian1 (approved):

PR #248, opened 2024-09-06
fix: L1 network mismatch

### LINKED ISSUE #246: Detecting L1 / L2 chain id mismatch

### Motivation

@Trantorian1 got this issue when running the node:
```rs
Error: Getting initial ethereum state

Caused by:
    0: buffer overrun while deserializing
    1: buffer overrun while deserializing

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/error.rs:565:25
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/result.rs:1964:27
   2: mc_eth::client::EthereumClient::get_last_verified_block_number::{{closure}}
             at ./crates/client/eth/src/client.rs:104:28
```
It is cause by the fact that he has put a mainnet l1 rpc endpoint whereas he was running a sepolia node.

The buffer overrun is because the node tries to read the l1 core contract

### Request

Detect the network mismatch

### Solution

Either:
- Get the chain ID from eth RPC, check that it matches
- Better, I think, if the L1 core contract does not exist, throw an error that reads something like: `The L1 Core Contract could not be found. Check that the L2 chain matches the L1 RPC endpoint.`
(we need to do some hand-holding here, point to the solution in the error message somehow)

Yes!

akhercha:
# Pull Request type

- Bugfix

Resolves: #246 

## What is the new behavior?

- Now correctly checks the existence of the L1 Core Contract:

image

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: L1 rpc <=> network mismatch
 - feat: rpc versioning
 - feat: bumping Starknet version from v0.13.2 to Starknet v0.13.2.1
 - fix: replaced old namings with adapted namings
--- a/crates/client/eth/src/client.rs
+++ b/crates/client/eth/src/client.rs
@@ -150,6 +150,23 @@ pub mod eth_client_getter_test {
     const L2_BLOCK_HASH: &str = "563216050958639290223177746678863910249919294431961492885921903486585884664";
     const L2_STATE_ROOT: &str = "1456190284387746219409791261254265303744585499659352223397867295223408682130";
 
+    #[rstest]
+    #[tokio::test]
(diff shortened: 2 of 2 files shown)


cchudant (approved):

cchudant:
i haven't tested, and I don't know if this warrants a test case, but it looks correct!

thank you!

antiyro (approved):
lgtm thanks for your reactivity!

PR #255, opened 2024-09-10
feat: configurable chain config

### LINKED ISSUE #240: Configurable chain config (duh)

### Motivation

Last week with @Mohiiit we talked about where to put new configuration params - sometimes the line between what should be put in `ChainConfig` vs what should be put in cli arguments is a bit blurry
Let's fix that!

### Request

I believe the difference should be related to how the user uses madara, not about technical details or whatnot.

CLI args should be how node operators (people running madara nodes) configure madara, and chain config should be how appchain developers (kakarot pragma & cie) configure their chain.
This means that block times, and other details like pending block tick time should be in chain config - even though it may be a debatable. The batch size for block production should also be in chain config instead of cli arguments as it is very related to pending block tick time -- we should expect people to tune both at the same time. This may seem a bit strange because this is a performance setting though.

In addition,
- Chain config should be configurable from a .yml (or .toml?) file that we pass as argument `--chain-config ./chain.yml`.
- The docs should provide an annotated (with comments) example file
- Sometimes we would like to override some of the chain config via cli args -- for example, to set a special chain_id for a devnet or e2e test. It would look like `--chain-config-override chain_id=MADARA_TEST`.
  - technical details: this is easy to do, we would just need to load the chain config as a `serde_yaml::Value`, override the value we want by writing fields into that untyped `Value`, and then use `serde_yaml::from_value` to get a ChainConfig struct. This means that this would reuse the same code for deserializing Felts and stuff as the config file, the deserialization is done by serde.
  - stuff like versioned_constants may require more thought to deserialize from cli args - let's just not support that because I don't really see any usecase for that

Plus,
- We may want "presets" in the future, to configure in a single cli arg a devnet/full node syncing from mainnet/other chains
  - i don't know how this would really look in practice, maybe we can get away with just switching between different preset chain configs? the question is whether presets should affect cli arg config.

Thoughts? :)

I am not sure if yml is better than toml, I think starknet people are more familiar with toml? what do you think?
I don't want json as it can't have comments.

### Solution

_No response_

Yes!

azurwastaken:
# Pull Request type

- Feature

Resolves: #240 

## What is the new behavior?

- We can now provide a yaml with chain config description or use presets to populate chain config struct

No


config.yaml.example
```diff
@@ -0,0 +1,33 @@
+chain_name: "Starknet Mainnet"
```
antiyro: Could you drop some documentation describing each of these fields?

azurwastaken: Done

crates/client/eth/src/l1_messaging.rs
```diff
@@ -315,6 +315,21 @@ mod l1_messaging_tests {
         }
     );
 
+    #[fixture]
```
Trantorian1: Couldn't this be refactored to be more reusable given your are using the same code in multiple places?

crates/node/src/cli/chain_config_overrides.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[arg(long, requires = "chain_config_override", value_name = "OVERRIDED MAX NONCE VALIDATION FOR SKIP")]
+    pub max_nonce_for_validation_skip: Option<u64>,
+    #[arg(long, requires = "chain_config_override", value_name = "OVERRIDED ETH CORE CONTRACT")]
+    pub eth_core_contract_address: Option<String>,
+}
+
+impl ChainConfigOverrideParams {
+    pub fn override_cfg(&self, mut chain_config: ChainConfig) -> ChainConfig {
+        let params = self.clone();
+        (
```
Trantorian1: This is pretty hard to decypher, consider just assigning the variables individually

crates/node/src/cli/mod.rs
```diff
@@ -76,6 +97,21 @@ impl RunCmd {
         self.name.as_ref().expect("Name was just set")
     }
 
+    pub fn get_config(&self) -> anyhow::Result<Arc<ChainConfig>> {
+        let mut chain_config: ChainConfig = match &self.preset {
+            Some(preset_name) => ChainConfig::from_preset(preset_name.as_str())?,
+            None => {
+                ChainConfig::from_yaml(&self.chain_config_path.clone().expect("Failed to retrieve chain config path"))?
```
Trantorian1, line 104: Maybe a more explicit error for when a chain config has not been specified at all and no preset name was given?

Trantorian1 (commented):
All seems good. Check comments for a few changes to change, maily refactoring.

crates/node/src/cli/mod.rs
azurwastaken: if neither a yaml or preset is given, it throw an error due to #[clap(group(ArgGroup::new("chain_config").args(["chain_config_path", "preset"]).required(true)))]

crates/client/eth/src/l1_messaging.rs
azurwastaken: created a test_common in the utils crate crates/primitives/utils/src/tests_common.rs

crates/node/src/cli/chain_config_overrides.rs
azurwastaken: changed

EvolveArt (approved):

Trantorian1 (approved):
All good :+1:

PR #257, opened 2024-09-11
feat: add coveralls to ci

EvolveArt:
# Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
X Other (please describe): CI

Resolves: #NA

## What is the new behavior?

- Adds a new clippy check in CI
- Push coverage report to coveralls
- Spawns anvil from code in all tests, no need to spawn it on the side

No

## Other information

It was investigated to check number of unwraps but it doesn't make any sense to do in rust, it's fine to just make sure by ourselves there are no unsafe unwraps. We can always come back and add it where most unsafe ones have been removed.
As for the criterion benchmarking it should come in a separate PR as we first need to precisely know what should be benchmarked.


.github/workflows/coverage.yml
```diff
@@ -61,3 +62,9 @@ jobs:
         with:
           GITHUB_TOKEN: ${{ github.token }}
           COVERAGE_FILE: coverage.cobertura.xml
+
+      - name: Coveralls
+        uses: coverallsapp/github-action@v2
+        with:
+          files: coverage.cobertura.xml
+          debug: true
```
antiyro, line 51: Could we add a coverage fail treshold here? something like: 60%?

EvolveArt: this is done automatically through coveralls UI

antiyro:
could you clean CI for merge

antiyro:
Actually I would take this PR to replace the comment from github actions with one from coveralls wdyt?

EvolveArt:
> Actually I would take this PR to replace the comment from github actions with one from coveralls wdyt?

done

antiyro (approved):
lgtm

PR #258, opened 2024-09-11
fix: behavior when pending block is not present in db

cchudant:
# Pull Request type

- Bugfix

RPC endpoints return BlockNotFound when doing any call on Pending, when the block is in db

## What is the new behavior?

Match the pathfinder and juno behavior: behave as if there is a pending block in db in that case. The new behavior makes sense to me for call, estimate, simulate etc - but I was surprised to see that they also return a fake pending block for get_block rpcs too - all the endpoints, really. We need to match their behavior even if it's questionable because that's what the clients expect.

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: pending block must always be returned in rpc even if none is in db
 - docs: fixed Docker Compose instructions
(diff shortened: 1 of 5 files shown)


Trantorian1 (approved):

antiyro (approved):
lgtm thanks this will unlock kakarot tests

PR #261, opened 2024-09-13
fix(devnet): generate a fixed set of public and private keys for devnet

antiyro:


crates/node/src/service/block_production.rs
```diff
@@ -79,7 +79,10 @@ impl Service for BlockProductionService {
                     .context("Building genesis block from devnet config")?;
 
                 block_import
-                    .add_block(genesis_block, BlockValidationContext::new(backend.chain_config().chain_id.clone()))
+                    .add_block(
+                        genesis_block,
+                        BlockValidationContext::new(backend.chain_config().chain_id.clone()).trust_class_hashes(true),
```
cchudant, line 84: if you add that here, can you also add it in devnet/src/lib.rs in the tests?

cchudant (approved):
do we want:
- a mode where the keys are always random (?)
- a cli arg to change the seed?

i can't see any use cases for those though

crates/node/src/service/block_production.rs
antiyro: yes I'll also remove any from_random usage on the tests to match what we have in the rest of the codebase

cchudant: i dont think there are any :)
we use the DevnetKeys struct directly to access the secrets

Trantorian1 (approved):

PR #262, opened 2024-09-13
fix: flaky test

Mohiiit:
There were flaky tests in gas price worker because of the anvil instance, now each test has it's own anvil instance, it's not super beautiful but it works :) and it's stable (I hope)

# Pull Request type

- Test

In current case we launch an anvil instance and this is shared among tests. Because of the rust lifetimes, sometimes the anvil instance get out of scope before each test runs.

This was adopted because it was a code repetition.

Resolves: #NA

## What is the new behavior?

- each test has it's own anvil instance

No


antiyro (approved):
lgtm

ocdbytes (approved):
lgtm 🚀

PR #266, opened 2024-09-16
Fix pending

cchudant:
# Pull Request type

- Bugfix

Pending block returns an error when fetched

## What is the new behavior?

Pending block is fetched correctly, and some tests are added.

No


Trantorian1 (approved):

antiyro (approved):
lgtm

PR #268, opened 2024-09-17
Fix(class_hash): Fixed class hash computation for legacy contracts

antiyro:
This fix was tested against mainnet and it appears that we currently fail on 89 class hashes computation over the first 100k blocks and succeding over 14k class hashes computations. All of this only on legacy contracts, sierra contracts works fine. I will merge this one to unlock legacy computes and open a detailed issue to solve those exceptions.


antiyro:
I would like to make it work for kakarot before merging it

antiyro:
merged in #271

PR #269, opened 2024-09-17
refactor: calculate class hash in devnet

apoorvsadana:
In the current setup, the class hashes for devnet are hard coded. This isn't really a big issue but everytime we update the Cairo compiler we need to manually change these class hashes in the devnet. This PR calculates the class hash instead of taking the hardcoded values.

### DIFF
--- a/crates/client/devnet/src/classes.rs
+++ b/crates/client/devnet/src/classes.rs
@@ -94,7 +94,9 @@ impl InitiallyDeclaredClasses {
 
     /// Load the classes into `DeclaredClass`es.
     pub fn into_loaded_classes(self) -> Vec<DeclaredClass> {
-        self.0.into_values().map(|class| match class {
+        self.0
+            .into_values()
+            .map(|class| match class {
                 InitiallyDeclaredClass::Sierra(c) => DeclaredClass::Sierra(SierraDeclaredClass {
                     class_hash: c.class_hash,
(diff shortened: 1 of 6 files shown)


Trantorian1 (approved):
All good :+1:

crates/client/devnet/src/classes.rs
```diff
@@ -24,11 +26,23 @@ pub enum InitiallyDeclaredClass {
 }
 
 impl InitiallyDeclaredClass {
```
jbcaron, line 28: just out of curiosity, why aren’t the classes definition stored compressed directly?

apoorvsadana: I didn't understand the question sorry, the compressed version is stored now right?

antiyro:
lgtm have you tested it with different scarb versions?

apoorvsadana:
> lgtm have you tested it with different scarb versions?

Yes!

antiyro (approved):

crates/client/devnet/src/classes.rs
jbcaron: it's just that we produce uncompressed class definitions with scarb

apoorvsadana: Oh ok, I am not really sure but if Scarb has that option then this could be an optimisation, yes.

PR #271, opened 2024-09-18
fix(block-production): fix bouncer, error reporting, debug messages

cchudant:
This is on antiyro's current legacy class hash branch but i can rebase it on main if we want it merged before

# Pull Request type

- Bugfix

Bouncer fraction is wrong

## What is the new behavior?

- Correct bouncer fraction
- Better error reporting around mempool and block production
- Better debug messages and info messages

No


.github/workflows/coverage.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - uses: software-mansion/setup-scarb@v1
         with:
-          scarb-version: "2.7.0"
+          scarb-version: "2.8.2"
```
Trantorian1, line 28: Doesn't this pose an issue with `devnet` as scarb 2.8.2 is not yet compatible with starknet? You can check with @apoorvsadana, I remember he had an issue related to this.

crates/client/block_import/src/pre_validate.rs
```diff
@@ -172,7 +174,17 @@ fn class_conversion(
...
             log::trace!("Converting legacy class with hash {:#x}", legacy.class_hash);
-            // TODO: verify that the class hash is correct
+            if !validation.trust_class_hashes {
+                let class_hash = legacy
+                    .contract_class
+                    .compute_class_hash()
+                    .map_err(|e| BlockImportError::ComputeClassHash { class_hash: legacy.class_hash, error: e })?;
+                if class_hash != legacy.class_hash {
+                    // TODO: For now we skip the exceptions for the legacy class hash mismatch
+                    log::debug!("Class hash mismatch: got {:#x}, expected {:#x}", class_hash, legacy.class_hash,);
```
Trantorian1, line 184: Ehhhh, is this a good idea?

crates/client/db/src/db_metrics.rs
```diff
@@ -1,15 +1,34 @@
...
 impl DbMetrics {
     pub fn register(registry: &MetricsRegistry) -> Result<Self, PrometheusError> {
         Ok(Self {
+            db_size: registry.register(Gauge::new("db_size", "Node storage usage in GB")?)?,
             column_sizes: registry
                 .register(IntGaugeVec::new(Opts::new("column_sizes", "Sizes of RocksDB columns"), &["column"])?)?,
         })
     }
+
+    /// Returns the total storage size
```
Trantorian1, line 19: Do we currently have the metrics behind a compilation flag? I'm just wondering as it could make sense if the resulting performance hit is noticeable. Might not be needed though, I saw we check the db only once every 200 blocks. This is more of a general observation and request for feedback.

crates/client/mempool/src/block_production.rs
```diff
@@ -287,35 +326,53 @@ impl BlockProductionTask {
...
-        let config_bouncer = self.executor.bouncer.bouncer_config.block_max_capacity;
-        let frac = n_pending_ticks_per_block / current_pending_tick; // div by zero: current_pending_tick has been checked for 0 above
+        let config_bouncer = self.backend.chain_config().bouncer_config.block_max_capacity;
+
+        // reduced_gas = gas * current_pending_tick/n_pending_ticks_per_block
+        // - we're dealing with integers here so prefer having the division last
+        // - use u128 here because the multiplication would overflow
+        // - div by zero: current_pending_tick has been checked for 0 above
+        let reduced_cap =
+            |v: usize| (v as u128 * current_pending_tick as u128 / n_pending_ticks_per_block as u128) as usize;
```
Trantorian1, line 340: `n_pending_ticks_per_block` should be checked instead of `current_pending_tick`!

crates/primitives/class/src/convert.rs
```diff
@@ -0,0 +1,86 @@
...
+                .map(|item| parse_legacy_entrypoint(&item, is_pre_0_11_0))
+                .collect(),
+        },
+        program,
+    })
+}
+
+fn parse_legacy_entrypoint(entrypoint: &LegacyContractEntryPoint, pre_0_11_0: bool) -> RawLegacyEntryPoint {
+    RawLegacyEntryPoint {
+        // This doesn't really matter as it doesn't affect class hashes. We simply try to guess as
```
Trantorian1, line 77: How so?

Trantorian1 (changes requested):
> :warning: You have an **incorectly guarded division by 0**, please fix this

See comments below for more details

Trantorian1 (approved):

.github/workflows/coverage.yml
cchudant: I just applied the same version everywhere, someone changed it but forgot to change the other places

crates/client/block_import/src/pre_validate.rs
cchudant: @antiyro

crates/client/db/src/db_metrics.rs
cchudant: why would you want it behind a compilation flag?

yep it's once every 200 blocks and I think the way it's handled is bad - it should instead be like the maybe db flush thing -- at the end of every block, check that 5s(?) has elapsed since the last metrics update and if so, update
it would make much more sense to handle it that way so that it continues to work fine when the chain is fully synced and new blocks don't come that often

but also, it's kind of inexpensive i believe - it's stats that rocksdb already tracks, and updating the prometheus metrics lowers to nothing but a few atomic stores under the hood so i'm also inclined to not bother and put it every block.

crates/client/mempool/src/block_production.rs
cchudant: :bangbang:

crates/primitives/class/src/convert.rs
cchudant: @antiyro

crates/client/db/src/db_metrics.rs
cchudant: i changed it to the 5s thinggy

crates/client/block_import/src/pre_validate.rs
antiyro: So it appears that for the first 100k blocks (the one containing most of the cairo 0 computations) we only pass this 80 times. Which is a lot but we have not identified yet the impact of those exceptions. So we could either crash the sync or continue and pass the Sequencer class which is in my opinion the best option while we fix it. It unlocks 98% of our class hash computations on sync.

crates/primitives/class/src/convert.rs
antiyro: This comes from [Starkli](https://github.com/xJonathanLEI/starkli/blob/fa4f0e3508f77df344482fce8f57e21fb2295008/src/utils.rs#L214) implementation.

cchudant:
closes #278

antiyro (approved):

cchudant:
i hate the ci's guts

crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -189,6 +215,21 @@ impl ChainConfig {
...
+    /// Verify that the chain config is valid for block production.
+    pub fn precheck_block_production(&self) -> anyhow::Result<()> {
+        // block_time != 0 implies that n_pending_ticks_per_block != 0.
+        if self.sequencer_address == ContractAddress::default() {
+            bail!("Sequencer address cannot be 0x0 for block production.")
+        }
+        if self.block_time.as_millis() == 0 {
+            bail!("Block time cannot be zero for block production.")
+        }
+        if self.pending_block_update_time.as_millis() == 0 {
```
EvolveArt, line 227: we should also check that pending_block_update_time is < block_time right ?

cchudant: good idea yea

PR #272, opened 2024-09-19
fix(cli): clean command line argument + columns metrics

antiyro:
Removed required presets and chain config if a network was specified

### DIFF
--- a/crates/node/src/cli/mod.rs
+++ b/crates/node/src/cli/mod.rs
@@ -24,7 +24,7 @@ use url::Url;
 
 /// Madara: High performance Starknet sequencer/full-node.
 #[derive(Clone, Debug, clap::Parser)]
-#[clap(group(ArgGroup::new("chain_config").args(["chain_config_path", "preset"]).required(true)))]
+#[clap(group(ArgGroup::new("chain_config")))]
 pub struct RunCmd {
     /// The human-readable name for this node.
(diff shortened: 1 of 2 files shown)


crates/node/src/cli/mod.rs
```diff
@@ -98,19 +98,31 @@ impl RunCmd {
...
+        } else if let Some(path) = &self.chain_config_path {
+            ChainConfig::from_yaml(path)?
+        } else {
+            // Neither preset nor chain_config_path is provided
+            match self.network {
+                NetworkType::Main => ChainConfig::from_preset("mainnet")?,
+                NetworkType::Test => ChainConfig::from_preset("sepolia")?,
+                NetworkType::Devnet => ChainConfig::from_preset("test")?,
+                NetworkType::Integration => ChainConfig::from_preset("integration")?,
+                _ => {
```
jbcaron, line 112: why the wildcard? all cases are already covered

crates/node/src/cli/sync.rs
```diff
@@ -67,7 +67,7 @@ impl SyncParams {
             feeder_gateway,
             chain_id,
             sound,
-            verify: !self.disable_root,
+            verify: self.disable_root,
```
Trantorian1, line 70: This seems a bit strange to me

PR #274, opened 2024-09-20
fix: spec version

Mohiiit:
spec version rpc call's response updated and l1 url updated for anvil forking

# Pull Request type

- Bugfix

currently the spec version call returns 0_7_0

Resolves: #NA

## What is the new behavior?

- after the change it would return 0.7.0

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Next release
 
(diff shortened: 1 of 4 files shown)


antiyro (approved):
thanks!

apoorvsadana (approved):

PR #275, opened 2024-09-20
refactor: Optimize RocksDB opening process

doanthekhiem:


cchudant (commented):
Hi :wave: 
Thank you for your willingness to contribute to madara :partying_face: 
From a surface level look of the changes, I don't see how this changes anything about how rocksdb is opened
Care to explain what this change does?

Also, I think this doesn't compile, Colum::iter is not a function as far as I remember

doanthekhiem:
Yes, it's not really necessary. I just wanted to separate the functions for easier maintenance. So, I'll close this PR. Thank you.

PR #276, opened 2024-09-21
Add log message when no pending block to improve debugging

doanthekhiem:
I suggest adding a log message when there is no pending block to improve debugging.


Trantorian1:
Please refrain from presenting inconsequential changes to the codebase in a pull request.

PR #279, opened 2024-09-23
fix(cli): Fix and clean client modes and fetch presets

antiyro:


crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -27,18 +28,57 @@ pub enum ChainPreset {
...
+        if Path::new(local_path).exists() {
+            ChainConfig::from_yaml(Path::new(local_path))
+                .context(format!("Failed to read local config file: {}", local_path))
+        } else {
+            log::info!(
+                "📝 Local Chain config file for: {:?} not found. Fetching from remote URL.",
+                self.get_preset_name()
+            );
+            let remote_url = match self {
+                ChainPreset::Mainnet => "https://raw.githubusercontent.com/madara-alliance/madara/main/crates/primitives/chain_config/presets/mainnet.yaml",
```
cchudant: id prefer if we used include bytes here please
or, i think it's just better to revert all of it - remake the functions as they were before
```rs
    pub fn starknet_sepolia() -> Self {
        Self {
            chain_name: "Starknet Sepolia".into(),
            chain_id: ChainId::Sepolia,
            eth_core_contract_address: eth_core_contract_address::SEPOLIA_TESTNET.parse().expect("parsing a constant"),
            ..Self::starknet_mainnet()
        }
    }
```
like that
it's much better i think, we really dont need to parse a file here

crates/node/src/cli/mod.rs
```diff
@@ -98,22 +126,68 @@ impl RunCmd {
...
-        let mut chain_config: ChainConfig = match &self.preset {
-            Some(preset_name) => ChainConfig::from_preset(preset_name.as_str())?,
+        let chain_config = match &self.preset {
+            Some(preset_name) => ChainConfig::from_preset(preset_name.as_str()).map_err(|err| {
+                log::error!("Failed to load config from preset '{}': {}", preset_name, err);
+                anyhow::anyhow!(err)
+            })?,
             None => {
-                ChainConfig::from_yaml(&self.chain_config_path.clone().expect("Failed to retrieve chain config path"))?
+                let path = self.chain_config_path.clone().ok_or_else(|| {
```
cchudant: can we just support --devnet without specifying any kind of preset please

crates/node/src/cli/mod.rs
```diff
@@ -98,22 +126,68 @@ impl RunCmd {
...
             }
         };
 
-        // Override stuff if flag is setted
-        if self.chain_config_override {
-            chain_config = self.chain_params.override_cfg(chain_config);
-        }
+        // Override stuff if flag is set
+        let chain_config =
+            if self.chain_config_override { self.chain_params.override_cfg(chain_config) } else { chain_config };
```
cchudant: id prefer if you remade the chain config override system from scratch and follow how i described it in #240 

which means, no new cli struct or field by field override by hand

crates/node/src/main.rs
```diff
@@ -132,14 +141,22 @@ async fn main() -> anyhow::Result<()> {
...
-                        run_cmd.network.feeder_gateway(),
-                        chain_config.chain_id.to_felt(),
-                    ))),
-                )
+                ServiceGroup::default().with(sync_service),
+                // TODO(rate-limit): we may get rate limited with this unconfigured provider?
+                Arc::new(ForwardToProvider::new(SequencerGatewayProvider::new(
+                    run_cmd
+                        .network
+                        .expect(
```
cchudant: use context not expect here please

cchudant (changes requested):
list of stuff from #255 i'd like reverted/changed:
- revert the set_workdir addition in the tests
- update the example config file
- revert the built in configs to functions like
```rs

    pub fn starknet_sepolia() -> Self {
        Self {
            chain_name: "Starknet Sepolia".into(),
            chain_id: ChainId::Sepolia,
            eth_core_contract_address: eth_core_contract_address::SEPOLIA_TESTNET.parse().expect("parsing a constant"),
            ..Self::starknet_mainnet()
        }
    }
```
and they do not need to return results
The preset can just use those functions instead of reading any file
- remove any fs read except in tests, and when the chain config file is supplied as argument
- please add this comment at the top of chain_config.rs
```rs
// Note: We are NOT using fs read for constants, as they NEED to be included in the resulting
// binary. Otherwise, using the madara binary without cloning the repo WILL crash, and that's very very bad.
// The binary needs to be self contained! We need to be able to ship madara as a single binary, without
// the user needing to clone the repo.
// Only use `fs` for constants when writing tests.
```
- change the chain config override system (see my comment below)
- Duration parsing needs to accept `2s`, `200ms` -- stings like that, and should not accept naked numbers. Change all the cli that currently accept durations to that format too.
- I think we should not use test config for devnet, but a config named devnet. The test preset can be removed as it's only used in tests using `ChainConfig::test_config`
- there are a few `except` and error handling stuff that could be better handled.

cchudant:
i think preset can be moved to node cli args and not inside the primitive crate altogether -- to a clap enum. It would allow clap to show a proper error message when the preset is unknown, with a list of the possible values. Much better UX

cchudant:
also, i dont know what the default impl on ChainConfig means, try to remove it and see if it was useful
(i just dont know why it's there it feels weird)

crates/node/src/cli/mod.rs
antiyro: Why would you want that, we would like to have custom devnets also (the kkt use case were they need to custom steps is already a use case for that)

cchudant: well, i think --devnet without --chain-config implies --preset devnet

antiyro:
This PR will be handled by @akhercha from now on

crates/primitives/chain_config/src/chain_config.rs
akhercha: Reverted the read file stuff!

crates/node/src/cli/mod.rs
akhercha: As said in DMs, I've tried to implement a different override mechanism but it's far from perfect at the moment because `ChainConfig` doesn't implement `Serialize` (since our flow is _1._ generate `ChainConfig` _2._ override if possible , we need to convert the pre-built `ChainConfig` to `Value`).

Instead of going down the rabbit hole of implementing wrapper structs around Blockifiers one & implementing Serialize myself, I've created a wrapper struct `OverridableChainConfig` that represents the fields we're able to override.

I'm not really fan of this tbh but it works at least. Any thoughts? Implementing Serialize would appear the cleanest here probably but didn't try yet - can do if you judge it better

crates/node/src/main.rs
akhercha: doneeee

akhercha:
Everything's done, just not sure about the ChainConfig overrides as I said above + this comment:
> - update the example config file

Wdym? I updated the comments inside but happy to do more. @cchudant

crates/node/src/cli/mod.rs
akhercha: `--devnet` alone is now accepted - preset will be `devnet` by default & it's still possible to specify another config

configs/presets/devnet.yaml
```diff
@@ -9,8 +9,8 @@ versioned_constants:
   "0.13.2": "crates/primitives/chain_config/resources/versioned_constants_13_2.json"
 eth_core_contract_address: "0xE2Bb56ee936fd6433DC0F6e7e3b8365C906AA057"
 latest_protocol_version: "0.13.2"
-block_time: 360
-pending_block_update_time: 2
+block_time: "6min"
```
antiyro: why 6mins block time over chain configs, we have around 30sec avg on mainnet

configs/presets/devnet.yaml
```diff
@@ -30,5 +30,5 @@ bouncer_config:
     message_segment_length: 18446744073709551615
     n_events: 18446744073709551615
     state_diff_size: 131072
-sequencer_address: "0x123"
+sequencer_address: "0x211b748338b39fe8fa353819d457681aa50ac598a3db84cacdd6ece0a17e1f3"
```
antiyro: what is this address exactly?

antiyro (commented):
Tested locally and lgtm

configs/presets/devnet.yaml
akhercha: Woops, it was meant to be used for the test config. Removed!

configs/presets/devnet.yaml
akhercha: Not sure, it was like this before - but updated to 30s, u're right!

configs/presets/devnet.yaml
```diff
@@ -30,5 +30,5 @@ bouncer_config:
     message_segment_length: 18446744073709551615
     n_events: 18446744073709551615
     state_diff_size: 131072
-sequencer_address: "0x123"
+sequencer_address: "0x0"
```
antiyro: Sequencer address cannot be zero for block production. We should use another one for the devnet config

akhercha: done!

akhercha (approved):
LGTM

PR #281, opened 2024-09-24
fix docker file

apoorvsadana:
The Docker file wasn't working correctly for the devnet. It should work now with this. We would need to chain config and versioned constants into the Docker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: docker file fixes for devnet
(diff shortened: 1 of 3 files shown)


antiyro:
omg I was working on this issue for the last hour T_T

Trantorian1:
## Behold, for the flaky gods have bestowed upon us a passing CI :pray: 

...Now whatever you do don't commit you fools.

image

antiyro (approved):
lgtm i will merge it since it doesnt introduce any breaking change (the current dockerfile is broken)

PR #282, opened 2024-09-24
feat(cli): updating madara launcher to support full, sequencer, devnet modes

antiyro:


Dockerfile
```diff
@@ -17,18 +17,21 @@ COPY crates crates
...
-ENV SCARB_REPO="https://github.com/software-mansion/scarb/releases/download"
-ENV PLATFORM="x86_64-unknown-linux-gnu"
-ENV SCARB_TARGET="/usr/src/scarb.tar.gz"
-
-RUN curl -fLS -o $SCARB_TARGET                                          \
-    $SCARB_REPO/$SCARB_VERSION/scarb-$SCARB_VERSION-$PLATFORM.tar.gz && \
-    tar -xz -C /usr/src/ --strip-components=1 -f $SCARB_TARGET &&       \
-    mv /usr/src/bin/scarb /bin
+# Installing Scarb
+# Dynamically detect the architecture
```
Trantorian1, line 21: Why are we checking for architecture inside of a Dockerfile? This will always resolve to whatever image we are building this under and should be set to a fixed architecture instead.

Trantorian1 (changes requested):

Dockerfile
antiyro: for some reason this error was thrown on mac builds:

image

PR #283, opened 2024-09-25
feat(test): block conversion on sync tests

azurwastaken:
# Pull Request type

- Testing

Resolves: #?

## What is the new behavior?

- Add test for block conversion

No


crates/client/sync/src/l2.rs
```diff
@@ -247,3 +247,209 @@ pub async fn sync(
...
+    #[tokio::test]
+    async fn test_l2_block_conversion_task(test_setup: Arc<MadaraBackend>) {
+        let backend = test_setup;
+        let (updates_sender, updates_receiver) = mpsc::channel(100);
+        let (output_sender, mut output_receiver) = mpsc::channel(100);
+        let block_import =
+            Arc::new(BlockImporter::new(backend.clone(), &MetricsRegistry::dummy(), None, true).unwrap());
+        let validation = BlockValidationContext::new(backend.chain_config().chain_id.clone());
+
+        let mock_block = UnverifiedFullBlock {
```
Mohiiit: maybe we can have a helper function inside the `crates/client/block_import/src/tests/block_import_utils` which would return `UnverifiedFullBlock`. This helper might be helpful for other tests as well

crates/client/sync/src/l2.rs
```diff
@@ -247,3 +247,209 @@ pub async fn sync(
...
+    #[tokio::test]
+    async fn test_l2_verify_and_apply_task(test_setup: Arc<MadaraBackend>) {
+        let backend = test_setup;
+        let (block_conv_sender, block_conv_receiver) = mpsc::channel(100);
+        let block_importer =
+            Arc::new(BlockImporter::new(backend.clone(), &MetricsRegistry::dummy(), None, true).unwrap());
+        let validation = BlockValidationContext::new(backend.chain_config().chain_id.clone());
+        let telemetry = TelemetryService::new(true, vec![]).unwrap().new_handle();
+
+        let mock_block = UnverifiedFullBlock {
```
Mohiiit: same, can have a helper function

Mohiiit (commented):
Very clean, just a few changes

Mohiiit (approved):

crates/client/block_import/Cargo.toml
```diff
@@ -8,6 +8,9 @@ repository.workspace = true
 version.workspace = true
 license.workspace = true
 
+[features]
+default = []
+testing = []
```
antiyro, line 13: what is this for?

EvolveArt (approved):

crates/client/block_import/Cargo.toml
EvolveArt: for some utils only under testing feature

PR #284, opened 2024-09-25
feat(test): adding JS tests to the CI

Mohiiit:
Adding starknet js tests 
# Pull Request type

- Testing

Resolves: #NA

## What is the new behavior?

- Ideally js tests should run in CI to check nothing is broken

No


.github/workflows/starknet-js-test.yml
```diff
@@ -0,0 +1,32 @@
...
+          path: target/release/madara
+          key: ${{ runner.os }}-madara-${{ github.sha }}
+          fail-on-cache-miss: true
+      - name: Restore Cairo build artifacts
+        uses: actions/cache@v3
+        with:
+          path: cairo/target
+          key: ${{ runner.os }}-cairo-${{ github.sha }}
+          fail-on-cache-miss: true
+      - name: Setup and run dev chain
```
apoorvsadana:
```suggestion
      - name: Setup dev chain and run tests
```

js_tests/basic.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    },
+    {
+      nonce,
+    },
+  );
+
+  await provider.waitForTransaction(declareResponse.transaction_hash);
+
+  // Add assertion to check if the class hash is valid
+  expect(declareResponse.class_hash).toBeTruthy();
```
apoorvsadana: we should get class hash and see if it's on chain

js_tests/basic.test.ts
```diff
@@ -0,0 +1,174 @@
...
+        readContractSierra("madara_contracts_HelloStarknet"),
+      ),
+    },
+    {
+      nonce,
+    },
+  );
+
+  // Add assertion to check if the contract address is valid
+  expect(deployResult.contract_address[0]).toBeTruthy();
```
apoorvsadana: we can maybe call get class hast at and check if it's registered on chain

js_tests/basic.test.ts
```diff
@@ -0,0 +1,174 @@
...
+        readContractSierra("madara_contracts_AccountUpgradeable"),
+      ),
+    },
+    {
+      nonce,
+    },
+  );
+
+  // Add assertion to check if the account address is valid
+  expect(deployResult.contract_address[0]).toBeTruthy();
```
apoorvsadana: same as above

apoorvsadana (commented):
small review :)

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: Swatinem/rust-cache@v2
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.78.0
+      - name: Setup build deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang llvm libudev-dev protobuf-compiler
+      - uses: rui314/setup-mold@v1
+      - uses: software-mansion/setup-scarb@v1
```
antiyro, line 24: Any reason why you are calling 2 times software-mansion/setup-scarb@v1 for version 2.8.2?

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          scarb-version: "2.8.2"
+      - uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+      - uses: software-mansion/setup-scarb@v1
```
antiyro: also here

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+          cd cairo
+          scarb build
+      - name: Build
+        run: |
+          cargo build --release --bin madara
+      - name: Cache Madara binary
+        uses: actions/cache@v3
+        with:
+          path: target/release/madara
+          key: ${{ runner.os }}-madara-${{ github.sha }}
```
antiyro: Here it appears that you are generating you key over the commit hash this doesnt invalidate the cache on each commit?

antiyro: I have checked maybe you can use:

`key: ${{ runner.os }}-madara-${{ hashFiles('Cargo.lock') }}`

to generate a key based on the lock changes or take in account even more files

.github/workflows/pull-request.yml
```diff
@@ -37,3 +36,13 @@ jobs:
...
+
+  build:
+    name: Build Madara
+    uses: ./.github/workflows/build.yml
+    needs: changelog
+
+  js_test:
+    name: Run JS Tests
+    uses: ./.github/workflows/starknet-js-test.yml
+    needs: build
```
antiyro, line 49: why does js_test depends on build? could we parallelize that?

Mohiiit: we need a build of madara to have it's instance running before running tests on it

so the flow is that we run madara and we run tests on it, so for that we need the latest build

js_tests/basic.test.ts
Mohiiit: technically we don't need it because getClassAt is the first call that is made when trying to deploy a contract.

here is the calls that are made when trying to deploy a contract:
```
[2024-09-27 11:47:01 HTTP] 🌐 starknet_V0_7_1_getClassAt 200 112069 bytes - 3.492ms
[2024-09-27 11:47:02 HTTP] 🌐 starknet_V0_7_1_estimateFee 200 162 bytes - 8.587ms
[2024-09-27 11:47:02 HTTP] 🌐 starknet_V0_7_1_addInvokeTransaction 200 123 bytes - 6.325ms
```

.github/workflows/pull-request.yml
antiyro: okk nice

antiyro: could we move the js_tests unde tests/js_tests?

Mohiiit: you mean inside the crates?

.github/workflows/starknet-js-test.yml
Mohiiit: resolved

js_tests/basic.test.ts
Mohiiit: resolved

js_tests/basic.test.ts
Mohiiit: resolved

js_tests/basic.test.ts
Mohiiit: resolved

.github/workflows/build.yml
Mohiiit: my bad, removed

.github/workflows/build.yml
Mohiiit: resolved

.github/workflows/build.yml
Mohiiit: resolved

.github/workflows/pull-request.yml
antiyro: no in the root folder directly, so we can add more tests in the futur

Mohiiit: resolved

antiyro (approved):
lgtm

apoorvsadana (approved):

PR #286, opened 2024-09-26
fix(rpc): estimate fee should throw an error if any txn fails

apoorvsadana:
Currently, `estimate_fee` returns the estimate even if the txn is failing. This is incorrect and `estimate_fee` should fail with `TxnExecutionError` instead.

Pending: add test cases


antiyro (approved):

crates/client/rpc/src/versions/v0_7_1/methods/read/estimate_fee.rs
```diff
@@ -46,8 +46,16 @@ pub async fn estimate_fee(
 
     let execution_results = exec_context.re_execute_transactions([], transactions, true, validate)?;
 
-    let fee_estimates =
-        execution_results.iter().map(|result| exec_context.execution_result_to_fee_estimate(result)).collect();
+    let fee_estimates = execution_results.iter().enumerate().try_fold(Vec::new(), |mut acc, (index, result)| {
```
Trantorian1: Consider using `Vec::with_capacity` to avoid unnecessary re-allocation since the max size is known.

Trantorian1 (changes requested):

Trantorian1 (approved):

PR #292, opened 2024-10-01
feat(rpc): storage proofs for snos and v0.8.0

### LINKED ISSUE #415: feat(starknet_getStorageProof)

### Motivation

### Request

Implement [`starknet_getStorageProof`](https://github.com/starkware-libs/starknet-specs/blob/c94df2c5866e11c866abd3d234b0d5df681073c3/api/starknet_api_openrpc.json#L910) from rpc v0.8.0-rc0.

> [!NOTE]
> From the specs: _Get merkle paths in one of the state tries: global state, classes, individual contract. A single request can query for any mix of the three types of storage proofs (classes, contracts, and storage)._.

Yes!

cchudant:
# Pull Request type

- Feature

Resolves: #415 

Yes (deleted some unused rocksdb columns - old database won't open.)


HermanObst:
@antiyro @cchudant hey there! I started reviewing this PR, but I would really appreciate a description that summarize the changes I need to expect in the PR.

crates/client/block_import/src/pre_validate.rs
```diff
@@ -365,7 +365,7 @@ fn compute_merkle_root<H: StarkHash + Send + Sync>(values: &[Felt]) -> Felt {
     let config = bonsai_trie::BonsaiStorageConfig::default();
     let bonsai_db = bonsai_trie::databases::HashMapDb::<bonsai_trie::id::BasicId>::default();
     let mut bonsai_storage =
-        bonsai_trie::BonsaiStorage::<_, _, H>::new(bonsai_db, config).expect("Failed to create bonsai storage");
+        bonsai_trie::BonsaiStorage::<_, _, H>::new(bonsai_db, config, /* max tree height */ 64);
```
HermanObst, line 372: Im interested about this tree (specifically about the height). What is for?

crates/client/db/src/bonsai_db.rs
```diff
@@ -22,23 +28,29 @@ impl DatabaseKeyMapping {
...
         write_opt.disable_wal(true);
-        Self { db, column_mapping, write_opt }
+        Self { db, column_mapping, write_opt, snapshots }
     }
 }
 
-impl BonsaiDatabase for BonsaiDb<'_> {
+impl fmt::Debug for BonsaiDb {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "<database>")
```
shamsasari: Something like `"BonsaiDb { ... }"` would be better. `database` is too generic for debug output.

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +160,121 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+pub struct BonsaiTransaction {
+    snapshot: Arc<SnapshotWithDBArc<DB>>,
+    changed: BTreeMap<(usize, ByteVec), Option<ByteVec>>,
+    db: Arc<DB>,
+    column_mapping: DatabaseKeyMapping,
+}
+
+impl fmt::Debug for BonsaiTransaction {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "<tx>")
```
shamsasari: ditto

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +160,121 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+            DatabaseKey::TrieLog(_) => 2,
+        },
+        k.as_slice().into(),
+    )
+}
+
+pub struct BonsaiTransaction {
+    snapshot: Arc<SnapshotWithDBArc<DB>>,
+    changed: BTreeMap<(usize, ByteVec), Option<ByteVec>>,
+    db: Arc<DB>,
```
shamsasari: Doesn't `SnapshotWithDBArc<DB>` already contain a reference to the same `DB`?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +160,121 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+            DatabaseKey::Trie(_) => 0,
+            DatabaseKey::Flat(_) => 1,
+            DatabaseKey::TrieLog(_) => 2,
+        },
+        k.as_slice().into(),
+    )
+}
+
+pub struct BonsaiTransaction {
+    snapshot: Arc<SnapshotWithDBArc<DB>>,
```
shamsasari: Why is the `Arc` needed here if `SnapshotWithDBArc` has an `Arc` already?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +160,121 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    fn snapshot(&mut self, id: BasicId) {
+        tracing::debug!("Generating RocksDB snapshot");
+        self.snapshots.create_new(id);
+    }
+
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    fn transaction(&self, id: BasicId) -> Option<(BasicId, Self::Transaction<'_>)> {
+        tracing::trace!("Generating RocksDB transaction");
+        if let Some((id, snapshot)) = self.snapshots.get_closest(id) {
```
shamsasari: `map` would avoid the need to manually return `None`.

crates/client/rpc/src/lib.rs
```diff
@@ -18,23 +18,43 @@ use std::sync::Arc;
...
 use mp_block::{MadaraMaybePendingBlock, MadaraMaybePendingBlockInfo};
-use mp_chain_config::{ChainConfig, RpcVersion};
+use mp_chain_config::ChainConfig;
 use mp_convert::ToFelt;
 
 pub use errors::{StarknetRpcApiError, StarknetRpcResult};
 use providers::AddTransactionProvider;
 use utils::ResultExt;
 
+// TODO: maake it actually configurable.
```
shamsasari: typo: make

crates/client/rpc/src/lib.rs
```diff
@@ -18,23 +18,43 @@ use std::sync::Arc;
...
+#[derive(Clone)]
```
shamsasari: Debug?

crates/client/rpc/src/versions/v0_8_0/get_storage_proof.rs
```diff
@@ -0,0 +1,185 @@
...
+            .map(|(node_hash, n)| NodeHashToNodeMappingItem {
+                node_hash,
+                node: match n {
+                    mc_db::ProofNode::Binary { left, right } => MerkleNode::Binary { left, right },
+                    mc_db::ProofNode::Edge { child, path } => {
+                        MerkleNode::Edge { child, path: path_to_felt(&path), length: path.len() }
+                    }
+                },
+            })
+            .collect(),
```
shamsasari: Can you extract this out into a variable please. It's doing too much to be inlined.

crates/client/db/src/bonsai_db.rs
cchudant: So, I don't use this fmt::Debug. This is used on the bonsai-trie side for debug printing the whole database, but in a real world where we have more than a few entries in db, i don't want to dump the whole database to the terminal by accident

crates/client/db/src/bonsai_db.rs
cchudant: same

crates/client/db/src/bonsai_db.rs
cchudant: the trait looks like
trait BonsaiDb: fmt::Debug
which forces us to implement Debug, the only way to avoid forcin people to impl Debug here would be a separate copy-pasted definition of the whole trait with #[cfg(not(test)))] and without the fmt::Debug bound, which sounds silly

crates/client/rpc/src/versions/v0_8_0/get_storage_proof.rs
cchudant: sure

crates/client/db/src/bonsai_db.rs
shamsasari: > i don't want to dump the whole database to the terminal by accident

Yes, of course. My point was simply have "BonsaiDb" in the debug output and not a generic "database" so that it's clear to anyone looking at the debug output what the type is.

cchudant:
@HermanObst @shamsasari I don't 100% know where to start really, but the basic idea is this:
Bonsai-trie has a trie-log and uses snapshots to get the historical state of the global tries. The way it's handled today in bonsai-trie bothers me quite a lot, and this is the stop-gap solution that uses what's already there in bonsai-trie, with some drawbacks:
- we can't get a storage proof for a block that is earlier than the block the node restarted at (due to using rocksdb snapshots)
- the trie logs are not compact at all and handled weirdly in bonsai-trie
- i had to remove a few features temporarily in bonsai-trie to make all of this work

The snapshot arc thing is a thing that stores a rocksdb Snapshot in an Arc, so that it can be stored in the MadaraBackend. Otherwise, it has an annoying lifetime. See https://github.com/rust-rocksdb/rust-rocksdb/issues/937 and https://github.com/rust-rocksdb/rust-rocksdb/issues/936

That's the hardest part, getting a MerkleTree out of the trie log. Otherwise, generating the merkle proof is as simple as calling getStorageProof which is implemented in my new tree traversal Iterator struct from https://github.com/madara-alliance/bonsai-trie/pull/36

I started reworking the whole trie-log idea in bonsai-trie from scratch in a way that would fix all of the drawbacks & not use rocksdb snapshots. That won't allow us to get rid of our rocksdb fork though, since I will use ArcSnapshot for other things.

Feel free to ask more questions though, this pr is a weird one.

Cargo.toml
```diff
@@ -85,12 +85,11 @@ version = "0.7.0"
...
 [workspace.dependencies]
-rocksdb = { version = "0.22", features = [
-  # "multi-threaded-cf",
-] }
+rocksdb = { version = "0.22", git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
+librocksdb-sys = { version = "0.17.0", git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
 
 # Bonsai trie dependencies
-bonsai-trie = { default-features = false, git = "https://github.com/cchudant/bonsai-trie.git", branch = "fix_inserts_remove_leaks", features = [
+bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "trie-log", features = [
```
shamsasari: The base branch is `oss`. Is the plan to merge `trie-log` into `oss` or is `oss` used for something else?

Cargo.toml
```diff
@@ -85,12 +85,11 @@ version = "0.7.0"
 license = "Apache-2.0"
 
 [workspace.dependencies]
-rocksdb = { version = "0.22", features = [
-  # "multi-threaded-cf",
-] }
+rocksdb = { version = "0.22", git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
+librocksdb-sys = { version = "0.17.0", git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
```
shamsasari: We're using different versions of the same branch?

crates/client/db/src/rocksdb_snapshot.rs
```diff
@@ -0,0 +1,169 @@
...
+        let readopts = ReadOptions::default();
+        self.get_cf_opt(cf, key.as_ref(), readopts)
+    }
+
+    /// Returns the bytes associated with a key value and given read options.
+    pub fn get_opt<K: AsRef<[u8]>>(&self, key: K, mut readopts: ReadOptions) -> Result<Option<Vec<u8>>, Error> {
+        // Safety: the snapshot originates from the `db`, and it is not dropped during the lifetime of the `readopts` variable.
+        unsafe {
+            readopts.set_raw_snapshot(self.inner);
+        }
```
shamsasari: Extract this out into a private function. It's repeated several times, and would keep the unsafe call into one place.

crates/client/db/src/rocksdb_snapshot.rs
```diff
@@ -0,0 +1,169 @@
+use librocksdb_sys as ffi;
+use rocksdb::{AsColumnFamilyRef, DBAccess, DBPinnableSlice, Error, ReadOptions};
+use std::sync::Arc;
+
+/// A copy of [`rocksdb::SnapshotWithThreadMode`] with an Arc<DB> instead of an &'_ DB reference
+pub struct SnapshotWithDBArc<D: DBAccess> {
```
shamsasari, line 11: Can you provide more information for why this needs to be done. Your comment in the PR is good, but it also needs to be in the code. Links to the GH issues you referenced would be helpful as well.

crates/client/db/src/rocksdb_snapshot.rs
```diff
@@ -0,0 +1,169 @@
...
+#[allow(dead_code)]
+impl<D: DBAccess> SnapshotWithDBArc<D> {
+    /// Creates a new `SnapshotWithDBArc` of the database `db`.
+    pub fn new(db: Arc<D>) -> Self {
+        let snapshot = unsafe { db.create_snapshot() };
+        Self { db, inner: snapshot }
+    }
+
+    /// Returns the bytes associated with a key value with default read options.
+    pub fn get<K: AsRef<[u8]>>(&self, key: K) -> Result<Option<Vec<u8>>, Error> {
```
shamsasari, line 38: How much clutter is it for the user of this API if we removed these overloads which don't have the `ReadOptions` specified, and the user calls the "opt" versions with `ReadOptions::default()`? It would tidy up the API here a bit, but then not sure if it's worth it.

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,51 @@
...
+            if let Some(max_saved_snapshots) = self.max_saved_snapshots {
+                if snapshots.len() > max_saved_snapshots {
+                    snapshots.pop_first();
+                }
+            }
+        }
+    }
+
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    pub fn get_closest(&self, id: BasicId) -> Option<(BasicId, Arc<SnapshotWithDBArc<DB>>)> {
```
shamsasari: Can we introduce a type alias for `Arc<SnapshotWithDBArc<DB>>`. It's used a few times and type signature is unwieldy.

shamsasari (commented):
Thanks for the comment, it was very useful. Something of it needs to be in the code as well.

crates/client/db/src/bonsai_db.rs
cchudant: oh okay

crates/client/db/src/bonsai_db.rs
cchudant: true

crates/client/db/src/bonsai_db.rs
cchudant: that's because the snapshot needs to live in the Snapshots struct which holds and manage them, but we don't want to hold a lock the whole time a snapshot is used in a bonsai transaction
so, instead, the Snapshots struct will hold the lock only for the time of cloning the Arc to the snapshot

in addition, the snapshot holding an Arc of the database is very important for safety reasons - this ensures that the snapshot cannot outlive the db. Otherwise the SnapshotWithDBArc should be marked unsafe.

these two arcs have very different goals.

shamsasari: This comment in the code would be very useful.

crates/client/db/src/rocksdb_snapshot.rs
cchudant: actually i don't agree with that
that would technically mean that this private function should be unsafe to call, since it can't ensure the readopts lifetime requirement is upheld by the caller
so we would still need an unsafe block at all of these places and a safety comment to go along with them

cchudant: actually there's a smarter way to avoid repeating that safety comment a bunch of times, it looks like this

```rs
    // This function allows us to not repeat the unsafe block a bunch of times. It takes ownership of the
    // readoptions to ensure its lifetime is stictly smaller than self.
    fn readopts_with_raw_snapshot<R>(&self, mut readopts: ReadOptions, f: impl FnOnce(&ReadOptions) -> R) -> R {
        // Safety: the snapshot originates from the `db`, and it is not dropped during the lifetime of the `readopts` variable.
        unsafe {
            readopts.set_raw_snapshot(self.inner);
        }

        f(&readopts)
    }
```
that way the function is safe

crates/client/db/src/snapshots.rs
cchudant: done

crates/client/db/src/rocksdb_snapshot.rs
cchudant: this is a copy of https://docs.rs/rocksdb/latest/rocksdb/struct.SnapshotWithThreadMode.html and the upstream type has it, so i copied the interface entirely
this file is supposed to go away once the issue is dealt with upstream, so i dont really want to change the api

crates/client/db/src/rocksdb_snapshot.rs
cchudant: added the links to gh issues :)

Cargo.toml
cchudant: im not sure why they do that but they do that upstream, their bindings and rocksdb crate have different versions
might be because librocksdb-sys follows the official rocksdb versions and the rocksdb crate doesnt

Cargo.toml
cchudant: this has been fixed :)

crates/client/db/src/bonsai_db.rs
cchudant: done :)

crates/client/db/src/bonsai_db.rs
```diff
@@ -94,13 +105,13 @@ impl BonsaiDatabase for BonsaiDb<'_> {
         // NB: we don't need old value as the trie log is not used :)
         // this actually speeds up things quite a lot
```
shamsasari: This comment isn't relevant anymore?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +159,117 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+    }
+
+    fn insert(
+        &mut self,
+        key: &DatabaseKey,
+        value: &[u8],
+        _batch: Option<&mut Self::Batch>,
+    ) -> Result<Option<ByteVec>, Self::DatabaseError> {
+        self.changed.insert(to_changed_key(key), Some(value.into()));
+        Ok(None)
```
shamsasari, line 231: `self.changed.insert` gives you the old value. Why is it not returned to the caller?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +159,117 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+    }
+
+    fn remove(
+        &mut self,
+        key: &DatabaseKey,
+        _batch: Option<&mut Self::Batch>,
+    ) -> Result<Option<ByteVec>, Self::DatabaseError> {
+        self.changed.insert(to_changed_key(key), None);
```
shamsasari, line 239: ditto

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +159,117 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+        Ok(None)
+    }
+
+    fn remove_by_prefix(&mut self, _prefix: &DatabaseKey) -> Result<(), Self::DatabaseError> {
+        unreachable!("unused yet")
+    }
+
+    fn write_batch(&mut self, _batch: Self::Batch) -> Result<(), Self::DatabaseError> {
+        // TODO: this is not really used yet, this whole abstraction does not really make sense anyway, this needs to be modified
```
shamsasari: Use `unreachable!` then?

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,55 @@
...
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    pub fn create_new(&self, id: BasicId) {
+        if self.max_saved_snapshots == Some(0) {
+            return;
+        }
+
+        let mut snapshots = self.snapshots.write().expect("Poisoned lock");
+
+        if let btree_map::Entry::Vacant(entry) = snapshots.entry(id) {
+            tracing::debug!("Making snap at {id:?}");
```
shamsasari: snap -> snapshot

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl Snapshots {
+    pub fn new(db: Arc<DB>, max_saved_snapshots: Option<usize>) -> Self {
+        Self { db, snapshots: Default::default(), max_saved_snapshots }
+    }
+
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    pub fn create_new(&self, id: BasicId) {
```
shamsasari: This is a confusing name. I think `make_snapshot` would be better.

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+pub type SnapshotRef = Arc<SnapshotWithDBArc<DB>>;
+
+/// This struct holds the snapshots. To avoid holding the lock the entire time the snapshot is used, it's behind
+/// an Arc. Getting a snapshot only holds the lock for the time of cloning the Arc.
+pub struct Snapshots {
+    db: Arc<DB>,
+    snapshots: RwLock<BTreeMap<BasicId, SnapshotRef>>,
+    max_saved_snapshots: Option<usize>,
```
shamsasari: Why "saved"? They're not persisted, right?

crates/client/rpc/src/versions/user/v0_8_0/methods/read/mod.rs
```diff
@@ -1,2 +1,30 @@
...
+
 pub mod get_compiled_casm;
-pub mod lib;
+pub mod get_storage_proof;
+
+#[async_trait]
+impl StarknetReadRpcApiV0_8_0Server for Starknet {
+    fn spec_version(&self) -> RpcResult<String> {
+        Ok(RpcVersion::RPC_VERSION_0_8_0.to_string())
+    }
```
shamsasari, line 15: I think you need to rebase, as this is already in main

crates/node/src/cli/db.rs
```diff
@@ -13,4 +13,21 @@ pub struct DbParams {
     /// Restore the database at startup from the latest backup version. Use it with `--backup-dir <PATH>`
     #[clap(env = "MADARA_RESTORE_FROM_LATEST_BACKUP", long)]
     pub restore_from_latest_backup: bool,
+
+    /// Maximal number of trie logs saved.
```
shamsasari: Maximum

shamsasari: Will users know what trie logs are?

crates/node/src/cli/db.rs
```diff
@@ -13,4 +13,21 @@ pub struct DbParams {
...
+    /// Blocks older than this limit will not be stored for retrieving historical merkle trie state.
+    /// By default, the value 0 means that no historical merkle trie state access is allowed.
+    #[clap(env = "MADARA_DB_MAX_SAVED_TRIE_LOGS", long, default_value_t = 0)]
+    pub db_max_saved_trie_logs: usize,
+
+    /// How many of the latest snapshots are saved, older ones are discarded.
```
shamsasari: snapshots of what? This is not explained.

crates/node/src/cli/db.rs
```diff
@@ -13,4 +13,21 @@ pub struct DbParams {
...
+    /// Higher values cause more database space usage, while lower values prevent the efficient reverting and histoical access for
+    /// the global state trie at older older blocks.
```
shamsasari: older repeated

crates/node/src/cli/db.rs
```diff
@@ -13,4 +13,21 @@ pub struct DbParams {
...
+    #[clap(env = "MADARA_DB_MAX_SNAPSHOTS", long, default_value_t = 0)]
+    pub db_max_saved_snapshots: usize,
+
+    /// A database snapshot is created every `db_snapshot_interval` commits.
+    /// See `db_max_saved_snapshots` to understand what snapshots are used for.
```
shamsasari: See --db-max-saved-snapshots

shamsasari (changes requested):

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +159,117 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
-    Self: 'db,
-{
-    type Transaction = Self;
+fn to_changed_key(k: &DatabaseKey) -> (usize, ByteVec) {
+    (
+        match k {
+            DatabaseKey::Trie(_) => 0,
+            DatabaseKey::Flat(_) => 1,
+            DatabaseKey::TrieLog(_) => 2,
+        },
```
Trantorian1, line 165: Why return a `uszie` if we are only ever using `0`, `1` or `2` as a value?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +159,117 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+            )
+        })
+    }
+
+    fn merge<'a>(&mut self, _transaction: Self::Transaction<'a>) -> Result<(), Self::DatabaseError>
+    where
+        Self: 'a,
+    {
+        unreachable!("unused for now")
+    }
```
Trantorian1, line 284: Perhaps we need to split up this trait to avoid all these `unreachable` functions

crates/client/db/src/snapshots.rs
Trantorian1: If I'm not mistaken we keep a certain amount of snaphots in RAM to be able to fetch historical values much faster than with a simple iterative lookup from the tip of the chain. These are lost on node restart though.

crates/client/db/src/lib.rs
```diff
@@ -249,6 +245,19 @@ impl DatabaseExt for DB {
...
+pub struct TrieLogConfig {
+    pub max_saved_trie_logs: usize,
+    pub max_saved_snapshots: usize,
+    pub snapshot_interval: u64,
+}
+
+impl Default for TrieLogConfig {
+    fn default() -> Self {
+        Self { max_saved_trie_logs: 0, max_saved_snapshots: 0, snapshot_interval: 5 }
+    }
```
Trantorian1: Does it make sense to have `max_saved_snapshots: 0` but `snapshot_interval: 5`?

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+        }
+    }
+
+    #[tracing::instrument(skip(self), fields(module = "BonsaiDB"))]
+    pub fn get_closest(&self, id: BasicId) -> Option<(BasicId, SnapshotRef)> {
+        tracing::debug!("get closest {id:?} {self:?}");
+        let snapshots = self.snapshots.read().expect("Poisoned lock");
+        snapshots.range(..&id).next().map(|(id, snapshot)| (*id, Arc::clone(snapshot)))
```
Trantorian1: Don't we need to call `.rev()` on this?

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+    let slice = &mut BitSlice::from_slice_mut(&mut arr)[5..];
+    slice[..path.len()].copy_from_bitslice(path);
+    Felt::from_bytes_be(&arr)
+}
+
+/// Returns (root hash, nodes)
+fn make_trie_proof<H: StarkHash + Send + Sync>(
+    block_n: u64,
+    trie: &mut GlobalTrie<H>,
+    _trie_name: StorageProofTrie,
```
Trantorian1: Remove this field?

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+            kind: StorageProofLimit::MaxKeys,
+            limit: starknet.storage_proof_config.max_keys,
+            got: proof_keys,
+        }
+        .into());
+    }
+
+    let n_tries = (!class_hashes.is_empty() as usize)
+        + (!contract_addresses.is_empty() as usize)
+        + contracts_storage_keys.iter().map(|keys| (!keys.storage_keys.is_empty() as usize)).sum::<usize>();
```
Trantorian1: :warning: This should be `saturating_add` as we are dealing with user-controlled input!

Trantorian1 (changes requested):

crates/client/db/src/lib.rs
cchudant: no but it allows this
```rs
TrieLogConfig { max_saved_trie_logs: 50, max_saved_snapshots: 10, ..Default::default() }
```


i don't like these 3 arguments but that's how the bonsai trie works right now and i'd love to spend the time to redo all of this

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: good catch

crates/client/db/src/bonsai_db.rs
cchudant: yes, added a comment as to why

crates/client/db/src/snapshots.rs
cchudant: yup, i have changed the trie to work backwards now. This is fixed

cchudant:
I have modified upstream bonsai-trie and this branch to apply the trie-logs backwards. This is important because it allows us to serve proofs for blocks that are older than the current oldest snapshot

cchudant:
i also fixed the review comments

Trantorian1 (approved):
Seems good, great work :+1:

crates/client/db/src/bonsai_db.rs
```diff
@@ -91,16 +102,13 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
-
-        // let old_value = self.db.get_cf(&handle, key.as_slice())?;
+        let old_value = self.db.get_cf(&handle, key.as_slice())?;
         if let Some(batch) = batch {
             batch.put_cf(&handle, key.as_slice(), value);
         } else {
             self.db.put_cf_opt(&handle, key.as_slice(), value, &self.write_opt)?;
         }
-        Ok(None)
+        Ok(old_value.map(Into::into))
```
HermanObst, line 111: Is this used for something later?

crates/client/db/src/bonsai_db.rs
```diff
@@ -148,144 +156,119 @@ impl BonsaiDatabase for BonsaiDb<'_> {
...
+        },
+        k.as_slice().into(),
+    )
+}
+
+pub struct BonsaiTransaction {
+    snapshot: SnapshotRef,
+    changed: BTreeMap<(u8, ByteVec), Option<ByteVec>>,
+    column_mapping: DatabaseKeyMapping,
+}
```
HermanObst: Could you add documentation for this struct? It would be helpful to clarify:

Field Descriptions: For example, what does the changed key represent? (I assume u8 represents the column and ByteVec is the database key?) And why is the value optional?

Usage Examples: Providing examples of how this struct is intended to be used would help others understand its purpose and functionality.

In general, adding more documentation would greatly enhance the code's readability and make it easier for others to understand and work with efficiently :)

crates/client/db/src/lib.rs
```diff
@@ -249,6 +246,19 @@ impl DatabaseExt for DB {
     }
 }
 
+#[derive(Debug)]
```
HermanObst, line 248: Link to references/documentation?

crates/client/db/src/snapshots.rs
```diff
@@ -0,0 +1,86 @@
...
+struct SnapshotsInner {
+    historical: BTreeMap<u64, SnapshotRef>,
+    /// Current snapshot for the latest block.
+    head: SnapshotRef,
+    head_block_n: Option<u64>,
+}
+
+/// This struct holds the snapshots. To avoid holding the lock the entire time the snapshot is used, it's behind
+/// an Arc. Getting a snapshot only holds the lock for the time of cloning the Arc.
+pub struct Snapshots {
```
HermanObst, line 19: Can we add some docs on how the Snapshots relates to being able to retrieve storage proofs ? 
And also slightly document the fields of the struct?

crates/client/rpc/src/errors.rs
```diff
@@ -17,6 +17,21 @@ pub enum StarknetTransactionExecutionError {
     ContractError,
 }
 
+#[derive(Clone, Copy, Serialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum StorageProofLimit {
```
HermanObst, line 22: Can you add documentation on the purpose of 
`MaxUsedTries` and `MaxKeys`?
What is the limit of a Storage proof?

crates/client/rpc/src/errors.rs
```diff
@@ -17,6 +17,21 @@ pub enum StarknetTransactionExecutionError {
...
+    MaxUsedTries,
+    MaxKeys,
+}
+
+#[derive(Clone, Copy, Serialize, Debug, PartialEq, Eq)]
+#[serde(tag = "trie", content = "contract_address", rename_all = "snake_case")]
+pub enum StorageProofTrie {
+    Classes,
+    Contracts,
+    ContractStorage(Felt),
```
HermanObst, line 33: `ContractStorage(Felt)`
I guess that felt is the contract address of that Storage Slot?

crates/client/rpc/src/lib.rs
```diff
@@ -12,29 +12,46 @@ pub mod utils;
...
-pub use errors::{StarknetRpcApiError, StarknetRpcResult};
 use providers::AddTransactionProvider;
+use starknet_types_core::felt::Felt;
+use std::sync::Arc;
 use utils::ResultExt;
 
+pub use errors::{StarknetRpcApiError, StarknetRpcResult};
+
+#[derive(Clone, Debug)]
+pub struct StorageProofConfig {
```
HermanObst, line 30: Again, some documentation of what is the goal of each field would be awesome :)

crates/client/rpc/src/versions/user/v0_8_0/api.rs
```diff
@@ -1,10 +1,84 @@
...
+    #[serde_as(as = "UfeHex")]
+    pub contract_address: Felt,
+    #[serde_as(as = "Vec<UfeHex>")]
+    pub storage_keys: Vec<Felt>,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum MerkleNode {
```
HermanObst, line 23: Maybe `PatriciaNode` its better here, but Im just being annoying
Feel free to reject

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+pub fn get_storage_proof(
+    starknet: &Starknet,
+    block_id: BlockId,
+    class_hashes: Option<Vec<Felt>>,
+    contract_addresses: Option<Vec<Felt>>,
+    contracts_storage_keys: Option<Vec<ContractStorageKeysItem>>,
+) -> RpcResult<GetStorageProofResult> {
+    // Pending block does not have a state root, so always fallbacck to latest.
```
HermanObst: fall back

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,202 @@
...
+        .get_block_n(&block_id)
+        .or_internal_server_error("Resolving block number")?
+        .ok_or(StarknetRpcApiError::NoBlocks)?;
+
+    let Some(latest) = starknet.backend.get_latest_block_n().or_internal_server_error("Getting latest block in db")?
+    else {
+        return Err(StarknetRpcApiError::BlockNotFound.into());
+    };
+
+    if latest.saturating_sub(block_n) > starknet.storage_proof_config.max_distance {
```
HermanObst, line 100: This means that in the Default for `storage_proof_config` (`max_distance = 0`) the only possible block on which we can get storage proofs is the current one, right?

Which is the maximum `max_distance` possible? (would be cool if we can explain it relation with snapshots and trielogs - if any)

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,202 @@
...
+    let contracts_storage_keys = contracts_storage_keys.unwrap_or_default();
+
+    // Check limits.
+
+    let proof_keys = saturating_sum(
+        iter::once(class_hashes.len())
+            .chain(iter::once(contract_addresses.len()))
+            .chain(contracts_storage_keys.iter().map(|v| v.storage_keys.len())),
+    );
+    if proof_keys > starknet.storage_proof_config.max_keys {
```
HermanObst, line 121: Is there some limit to this?

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,202 @@
...
+            got: proof_keys,
+        }
+        .into());
+    }
+
+    let n_tries = saturating_sum(
+        iter::once(!class_hashes.is_empty() as usize)
+            .chain(iter::once(!contract_addresses.is_empty() as usize))
+            .chain(contracts_storage_keys.iter().map(|keys| (!keys.storage_keys.is_empty() as usize))),
+    );
```
HermanObst, line 134: In SNOS, the default usage needs to retrieve proofs for all the tries. 
Is there a limitation here?

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,202 @@
...
+    if n_tries > starknet.storage_proof_config.max_tries {
+        return Err(StarknetRpcApiError::ProofLimitExceeded {
+            kind: StorageProofLimit::MaxUsedTries,
+            limit: starknet.storage_proof_config.max_tries,
```
HermanObst, line 138: Is there a limit to this? 
Usually several contract storage tries are queried

crates/node/src/cli/db.rs
```diff
@@ -13,4 +13,25 @@ pub struct DbParams {
     /// Restore the database at startup from the latest backup version. Use it with `--backup-dir <PATH>`
     #[clap(env = "MADARA_RESTORE_FROM_LATEST_BACKUP", long)]
     pub restore_from_latest_backup: bool,
+
+    /// This is the number of blocks for which you can get storage proofs using the storage proof endpoints.
+    /// Blocks older than this limit will not be stored for retrieving historical merkle trie state. By default,
+    /// the value 0 means that no historical merkle trie state access is allowed.
+    #[clap(env = "MADARA_DB_MAX_SAVED_TRIE_LOGS", long, default_value_t = 0)]
+    pub db_max_saved_trie_logs: usize,
```
HermanObst, line 21: What is the maximum number?

crates/node/src/cli/rpc.rs
```diff
@@ -135,6 +136,26 @@ pub struct RpcParams {
...
+    #[arg(env = "MADARA_RPC_STORAGE_PROOF_MAX_DISTANCE", long, default_value_t = 0)]
+    pub rpc_storage_proof_max_distance: u64,
+
+    /// Limit how many keys can be queried in a storage proof rpc request. Default: 1024.
+    #[arg(env = "MADARA_RPC_STORAGE_PROOF_MAX_KEYS", long, default_value_t = 1024)]
+    pub rpc_storage_proof_max_keys: usize,
+
+    /// Limit how many tries can be used within a single storage proof rpc request. Default: 5.
+    /// The global class trie and global contract tries count each as one, and every contract whose
+    /// storage is queried count as one each.
```
HermanObst, line 156: Is there a max value to this?

HermanObst (changes requested):
This is looking great. Good job team! 
Im just requiring some documentation for others to have a faster understanding of the code. The rest looks ok!

I would also would like some test on the class trie, but we can left that to other PR

crates/client/block_import/src/pre_validate.rs
cchudant: https://docs.starknet.io/architecture-and-concepts/network-architecture/block-structure/#transactions_events_receipts_commitments

in-memory merkle tree within every block
max tree height is 64 since the indices are u64s.
will add a comment about that

crates/client/db/src/bonsai_db.rs
cchudant: what do you mean?

crates/node/src/cli/rpc.rs
cchudant: not really

crates/node/src/cli/db.rs
cchudant: there's no maximum number

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: yes, for snos you would just put something very big
these limits are just limits that you put for eg. public rpcs so that they don't get DoSed by having to handle huge requests

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: not really

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: not really

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: max distance can be usize::MAX if you really wanted to. But we run into the risk of this kind of scenario:

current block is 1000
we have 25 snapshots, one every 4 blocks, meaning the oldest snapshot we have is on block 900
user asks for a storage proof on a block very far into the past, way farther than the oldest snapshot we have, like block_n=500

in this scenario, the endpoint will actually still work - it will in fact answer with a storage proof for block_n=500. But to do so it will have to get the snapshot at block_n=900, and apply the trie log in reverse up until block_n=500.
This max_distance is useful to reject requests like these to avoid getting DoSed in public rpc configurations for example.
For SNOS you could totally put usize::MAX, as with max_tries and max_keys

will add a comment about that

crates/client/rpc/src/versions/user/v0_8_0/api.rs
cchudant: it's the name in the specs (:

crates/client/db/src/bonsai_db.rs
cchudant: well, most of this file needs to be redone anyway, this abstraction is not that good.
will add some comments, but this is a hack for now

crates/client/db/src/snapshots.rs
cchudant: mmh, i'm having a hard time figuring out what needs more documentation here

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: where do you think i should put this explanation? There's an explanation in RpcParams::rpc_storage_proof_max_distance that looks like it's enough, what do you think?

crates/client/db/src/bonsai_db.rs
cchudant: no it's used now, the old value will be added to the trie log

cchudant:
@HermanObst I am not really sure what you have in mind for the documentation you are asking.
I have made some modifications, but:
- Do you think the documentation in RpcParams and DbParams (which is user-facing) is enough, or should I copy paste all of that all over the codebase everytime one of those fields pop up again? I would assume people read the user-facing doc before jumping into the code, or if they have questions about these args they would just look at how they're defined and find the user-facing RpcParams/DbParams struct pretty easily. I also think that these should have a dedicated page in the user-facing documentation, when we have that.
- the snapshots struct in particular looks very self explanatory and I struggle to find any comment that I could add that would explain it better than looking at its code directly. I may be biased as it's my code and I understand it pretty well, but what do you not understand here?

antiyro (approved):

PR #296, opened 2024-10-01
feat: Revert state to previous block

### LINKED ISSUE #202: feat(reorg): Add support for possible l1 and l2 reorgs

### Motivation

Currently, on Starknet, reorgs can occur due to a bug in the sequencer that might invoke a manual reorg until a stable state is reached, or due to an L1 reorg affecting a segment of the state containing an L2 verification. In the event of a reorg, we need to revert the entire state to match the L2 state. Detecting a reorg involves verifying if the parent block hash equals the latest block hash sync. If it doesn't, we need to sequentially revert all elements stored in the database (n-1) and revert the commitments on the Bonsai lib.

### Request

We need to implement support for handling possible L1 and L2 reorgs on Starknet. This includes detecting reorgs by checking the parent block hash against the latest block hash sync and performing necessary reverts on the database and Bonsai lib commitments.

A deep understanding of storage and its implications is necessary to resolve this issue. Some changes at the level of the Bonsai lib may come into play.

### Solution

1. **Detection**: Implement a mechanism to check if the parent block hash matches the latest block hash sync. If they don't match, a reorg is detected.
2. **Revert State**: Sequentially revert all elements stored in the database (n-1) to match the L2 state.
3. **Revert Commitments**: Revert the commitments on the Bonsai lib accordingly.

Feel free to discuss this issue under this discussion.

Yes!

notlesh:
This PR adds basic support for reverting state during a reorg.

It adds a `fn reorg()` which supports reorganizing a chain by reverting to a historical block (more on the name in my comment below). This triggers the bonsai tries to be reverted as well as the rocksdb to drop all blocks and associated state for the part of the chain that becomes orphaned.

There are fairly extensive tests which cover this new fn in actual reorg scenarios (blocks are pruned and then a new fork added).

Resolves: #202

No, it introduces a fn that is not yet used outside of test cases.


madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -325,6 +328,52 @@ fn block_hash(
     Ok((block_hash, header))
 }
 
+/// Reorgs the blockchain from its current tip back to `new_tip` from which a new fork can then be
+/// played on top.
+///
+/// Returns the result of the reorg, which describes the part of the chain that was orphaned.
+#[allow(dead_code)] // not used yet outside of tests
+fn reorg(
```
notlesh: Thoughts on calling this `reorg()`?

It's really just reverting. Other than some test cases, there is no actual reorg going on.

It might help to have some clear distinction between the two, so I'll define it here (let me know if you disagree):

* `Revert`: Set the current tip of the blockchain to some direct ancestor, dropping all blocks in between
* `Reorg`: Set the current tip of the blockchain to a block on another fork of the blockchain. Achieved by first reverting to the most recent common ancestor, then applying new blocks until the desired one is reached.

With this distinction in mind, I think this should be called revert and that this current PR isn't so much about reorgs (again, other than the tests).

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -325,6 +328,52 @@ fn block_hash(
...
+
+    backend
+        .revert_to(block_number)
+        .map_err(|error| BlockImportError::Internal(Cow::Owned(format!("error reverting block db: {}", error))))?;
+
+    // TODO: ensure there is no race condition here...? (e.g. we reverted and someone else appends a block before this next call)
+    let latest_block_info = backend
+        .get_block_info(&BlockId::Tag(BlockTag::Latest))
+        .map_err(make_db_error("getting latest block info"))?
+        .ok_or(BlockImportError::Internal(Cow::Owned("no latest block after reorg".to_string())))?;
```
notlesh: Thoughts on this -- esp. whether it's an issue and how to prevent it?

madara/crates/client/db/src/block_db.rs
```diff
@@ -355,6 +355,56 @@ impl MadaraBackend {
...
+            let block_hash_encoded = bincode::serialize(&block_info.block_hash)?;
+
+            tx.delete_cf(&block_n_to_block, &block_n_encoded);
+            tx.delete_cf(&block_hash_to_block_n, &block_hash_encoded);
+            tx.delete_cf(&block_n_to_block_inner, &block_n_encoded);
+            tx.delete_cf(&block_n_to_state_diff, &block_n_encoded);
+        }
+
+        let latest_block_n_encoded = bincode::serialize(&revert_to)?;
+        tx.put_cf(&meta, ROW_SYNC_TIP, latest_block_n_encoded);
```
notlesh, line 410: We set the `ROW_SYNC_TIP` here, but do we need to do anything else related to the tip of the chain? Anything with pending blocks?

notlesh (commented):
I think the current `fn reorg()` belongs somewhere else. Originally it was in `verify_apply.rs` because it was intended to *detect* reorgs during block import, but that is no longer the case. Any thoughts on where would be more appropriate?

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -325,6 +326,52 @@ fn block_hash(
...
 
+/// Reorgs the blockchain from its current tip back to `new_tip` from which a new fork can then be
+/// played on top.
+///
+/// Returns the result of the reorg, which describes the part of the chain that was orphaned.
+#[allow(dead_code)] // not used yet outside of tests
+fn reorg(
+    backend: &MadaraBackend,
+    new_tip: &PreValidatedBlock, // TODO: we don't need a PreValidatedBlock, do we?
+) -> Result<ReorgResult, BlockImportError> {
```
whichqua: I think `Result<ReorgResult, ..>` is a little confusing. Maybe rename `ReorgResult` to `ReorgData`?

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -325,6 +326,52 @@ fn block_hash(
...
 }
 
+/// Reorgs the blockchain from its current tip back to `new_tip` from which a new fork can then be
+/// played on top.
+///
+/// Returns the result of the reorg, which describes the part of the chain that was orphaned.
+#[allow(dead_code)] // not used yet outside of tests
+fn reorg(
+    backend: &MadaraBackend,
+    new_tip: &PreValidatedBlock, // TODO: we don't need a PreValidatedBlock, do we?
```
whichqua: Seems like you are using new tip to get the block_number and block_hash maybe you should let the caller handle that?

madara/crates/client/block_import/src/verify_apply.rs
raynaudoe: maybe using a lock for the entire reorg fn and any insert attempt should check that lock?
I think the scenario where a pending block is being inserted while a reorg happens is possible 🤔

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Agreed, I was trying to be consistent with other structs defined in `types.rs` but `ReorgData` is less ambiguous and consistent with the RPC spec.

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Yes, I think it needs some sort of locking mechanism, but I want to make sure I reuse something that exists for this purpose rather than create my own :)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Changed in [4ec928a](https://github.com/madara-alliance/madara/pull/296/commits/4ec928ac6e1550756b3d9a66d10153f1865ffcd6)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Cleaned this up in [0b44e12](https://github.com/madara-alliance/madara/pull/296/commits/0b44e12a06c2fdcd68a6cb306da3ab78044d7276):

* reorg only wants a hash now
* we look up block number from hash, which implicitly ensures the block exists in our db
* cleaned up the test code a bit

madara/crates/client/block_import/src/types.rs
```diff
@@ -218,3 +218,17 @@ pub struct BlockImportResult {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct PendingBlockImportResult {}
+
+/// Output of a [`crate::reorg`] operation. Fields are designed after `ReorgData` from the v0.8 RPC
+/// spec (https://github.com/starkware-libs/starknet-specs/blob/6e5a577802a5e117da747296bed0545ee41f451d/api/starknet_ws_api.json#L348)
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ReorgData {
+    /// Hash of the first known block of the orphaned chain
+    pub starting_block_hash: Felt,
```
cchudant, line 227: why do we need block hashes here?
i think if you're just modifying the backend, you should probably just have a revert_to(block_n: u64) function in the backend, it would be the responsability of sync to catch a reorg.

Well, unless you're making a broadcast channel in the backend to listen to reorgs - but i thought you were still working on the database reverting part.

madara/crates/client/block_import/src/verify_apply.rs
cchudant: should go into the backend imo, and be called revert_to

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -88,6 +88,10 @@ impl MadaraBackend {
         Ok(())
     }
 
+    pub fn revert_to(&self, revert_to: u64) -> Result<(), MadaraStorageError> {
+        self.block_db_revert(revert_to)
```
cchudant: reverting block_db is not sufficient, but I think you're aware of it as we talked about it before
(sorry if this pr was supposed to be a draft)

cchudant (changes requested):

madara/crates/client/block_import/src/types.rs
notlesh: The link above explains why -- it's to prepare for the reorg subscription RPC.

But it also gives more information than just block number, I'd rather provide as much context as possible to the caller.

It's also used in tests IIRC.

madara/crates/client/db/src/storage_updates.rs
notlesh: Can you elaborate? We also need to revert the bonsai tries, but that's done elsewhere. Is that what you mean?

notlesh: [d8b045f](https://github.com/madara-alliance/madara/pull/296/commits/d8b045f221e069a29b06c2b1ec6c7efa2a5fa364) adds support for removing declared class info and rolls back contract state. It's not optimized but should suffice for small reorgs, or at least for early testing :)

notlesh:
> should go into the backend imo, and be called revert_to

I tried moving this to `mc_db` but I don't think it really fits there because it's doing a lot with the bonsai trie, and the backend doesn't touch bonsai tries much (even though they are accessed through `MadaraBackend`).

The current reorg logic (which, yes, should be renamed to `revert`) still seems pretty closely tied with all all of the logic in `verify_apply.rs`, it's basically the opposite of `verify_apply_inner()`:

* it cares about the current block hash/num (similar to the `check_parent_hash_and_num()` call)
* it reverts trie updates (opposite of the `update_tries()` call)
* it reverts db state (opposite of `backend.store_block()` call)

In addition, testing it the way I've done so far relies on building a blockchain by calls to `verify_apply_inner()`.

Let me know if I'm still thinking about this the wrong way.

(CC @cchudant)

madara/crates/client/block_production/src/lib.rs
```diff
@@ -623,6 +623,53 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
     }
 }
 
+#[cfg(any(test, feature = "testing"))]
```
cchudant: why here?

madara/crates/client/db/src/block_db.rs
```diff
@@ -375,6 +375,7 @@ impl MadaraBackend {
         tx.delete_cf(&meta, ROW_PENDING_STATE_UPDATE);
 
         let latest_block_n = self.get_latest_block_n()?.unwrap(); // TODO: unwrap
+        let mut state_diffs = Vec::with_capacity((latest_block_n - revert_to) as usize);
         for block_n in (revert_to + 1..latest_block_n + 1).rev() {
```
cchudant: (revert_to + 1..=latest_block_n).rev()

madara/crates/client/db/src/block_db.rs
```diff
@@ -389,6 +390,11 @@ impl MadaraBackend {
 
             let block_hash_encoded = bincode::serialize(&block_info.block_hash)?;
 
+            // get state diff for this block before removing it
+            let state_diff_serialized = self.db.get_cf(&block_n_to_state_diff, block_n_encoded.clone())?.unwrap(); // TODO: unwrap, should we expect an entry for every single block here, or will some be None?
```
cchudant: yes, but we should put make a MadaraStorageError IMO instead of panicking

cchudant: to make it consistent with how we deal with errors everywhere else in mc-db

madara/crates/client/db/src/class_db.rs
```diff
@@ -213,6 +214,49 @@ impl MadaraBackend {
         Ok(())
     }
 
+    pub(crate) fn classes_db_revert(
```
cchudant: class_db_revert to make it consistent with the other methods IMO?

madara/crates/client/db/src/class_db.rs
```diff
@@ -213,6 +214,49 @@ impl MadaraBackend {
...
+        &self,
+        _revert_to: u64,
+        state_diffs: &Vec<StateDiff>,
+    ) -> Result<(), MadaraStorageError> {
+        let classes_info_col = self.db.get_column(Column::ClassInfo);
+        let classes_compiled_col = self.db.get_column(Column::ClassCompiled);
+
+        let mut deprecated_declared_class_hashes = HashSet::new();
```
cchudant: why 3 for loops and intermediary hashset?

just loop through the `state_diffs` param and insert to the write batch as you go no?

madara/crates/client/db/src/contract_db.rs
```diff
@@ -202,6 +203,82 @@ impl MadaraBackend {
...
+        revert_to: u64,
+        state_diffs: &Vec<StateDiff>,
+    ) -> Result<(), MadaraStorageError> {
+        let contract_to_class_hashes_col = self.db.get_column(Column::ContractToClassHashes);
+        let contract_to_nonces_col = self.db.get_column(Column::ContractToNonces);
+        let contract_storage_col = self.db.get_column(Column::ContractStorage);
+
+        let mut contract_to_class_hashes_keys = HashSet::new();
+        let mut contract_to_nonce_keys = HashSet::new();
+        let mut contract_storage_keys = HashSet::new();
```
cchudant: ditto, i dont think these are necessary this can be simplified, just go through the input vec and push stuff into the change batch

madara/crates/client/db/src/contract_db.rs
```diff
@@ -202,6 +203,82 @@ impl MadaraBackend {
...
+
+            // contract storage is a compound key (contract_address:storage_address)
+            diff.storage_diffs.iter().for_each(|diff_item| {
+                diff_item.storage_entries.iter().for_each(|entry| {
+                    contract_storage_keys.insert((diff_item.address, entry.key));
+                });
+            });
+        }
+        let latest_block_n = self.get_latest_block_n()?.unwrap(); // TODO: unwrap - Option here probably relates to genesis block
+        for block_n in (revert_to + 1..latest_block_n + 1).rev() {
```
cchudant: ..=

madara/crates/client/block_production/src/lib.rs
cchudant: ah you moved it
idk if it should stay here if you're using from another crate, this feels messy

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -330,37 +330,42 @@ fn block_hash(
...
-
-    // TODO: ensure there is no race condition here...? (e.g. we reverted and someone else appends a block before this next call)
-    let block_info = backend
+pub fn revert_to(backend: &MadaraBackend, new_tip_block_hash: &Felt) -> Result<ReorgData, BlockImportError> {
+    let target_block_info = backend
         .get_block_info(&BlockId::Hash(*new_tip_block_hash))
         .map_err(make_db_error("getting block info for new tip"))?
-        .ok_or(BlockImportError::Internal(Cow::Owned(format!("no block found for requested tip {}", new_tip_block_hash).to_string())))?;
+        .ok_or(BlockImportError::Internal(Cow::Owned(
+            format!("no block found for requested tip {}", new_tip_block_hash).to_string(),
```
cchudant: format! returns a String, the to_string here is unnecessary

also, please use ok_or_else if you're using format! in the else branch, the cost of fomat! and allocation shouldnt be paid for when taking the non-erroring path

cchudant:
```rs
.ok_or_else(|| BlockImportError::Internal(format!("no block found for requested tip {new_tip_block_hash:#x}").into()))
```
into is handy for Cow

cchudant: also, notice the :#x
when formatting felts, you need the :#x otherwise it will be formatted as a decimal number instead of hexadecimal and it's very ugly
(starknet-types-core's fault)

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -330,37 +330,42 @@ fn block_hash(
...
+        .map_err(make_db_error("getting current tip block info"))?
+        .ok_or(BlockImportError::Internal(Cow::Owned("on blockchain tip in storage".to_string())))?;
+
+    let target_block_number = target_block_info.block_n().expect("Target block must have a block number");
+    let target_block_id = BasicId::new(target_block_number);
 
-    let block_number = block_info.block_n().expect("Block must have a block number");
-    let block_id = BasicId::new(block_number);
+    let previous_tip_block_number =
+        previous_tip_block_info.block_n().expect("Previous tip block must have a block number");
```
cchudant: to make it consistent with the rest of the codebase, it would be great to remove the expect and return errors
(not saying this was a great choice to begin with, maybe all of those internal errors should be using .expect() but the ship has already been sailed and i would prefer it to be consistent)

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -371,7 +376,7 @@ fn reorg(
```
cchudant, line 382: same here, ok_or should be a ok_or_else and idk why you're using to_string on a litteral to put that into a Cow::Owned, you can just make a Cow::Borrowed


or use into, which does the Cow for you
.ok_or_else(|| BlockImportError::Internal("no latest block after reorg".into()))?;

cchudant: (same for the rest of the pr)

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -330,37 +330,42 @@ fn block_hash(
...
-        .revert_to(block_id)
+        .revert_to(target_block_id, previous_tip_block_id)
         .map_err(|error| BlockImportError::Internal(Cow::Owned(format!("error reverting class trie: {}", error))))?;
 
     backend
-        .revert_to(block_number)
+        .revert_to(target_block_number)
         .map_err(|error| BlockImportError::Internal(Cow::Owned(format!("error reverting block db: {}", error))))?;
 
     // TODO: ensure there is no race condition here...? (e.g. we reverted and someone else appends a block before this next call)
```
cchudant: this is the caller's responsability. it'll be much more clear when sync-refactor pr is merged, as i wrote an extensive doc on this subject

but to be brief, i would say this is either: block_production's responsability - and it's very easy for it to ensure it is not doing anything silly
or, mc-sync when we're running sync instead of block production.

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -371,7 +376,7 @@ fn reorg(
 
     Ok(ReorgData {
         starting_block_hash: *new_tip_block_hash,
-        starting_block_number: block_number,
+        starting_block_number: target_block_number,
         ending_block_hash: latest_block_info.block_hash().expect("how would a block not have a hash?"), // TODO: better error message, but srsly, how?
```
cchudant: latest_block_info returns a "maybe pending block", because you can also ask for the pending block. The pending block has no block_hash as it's not closed yet. This is because there's a concept of a "pending block" in starknet.

You can cast the MadaraMaybePendingBlockInfo using .as_nonpending().

the MadaraBackend right now is a bit silly regarding that, you can't query a "closed block" (meaning, non-pending) without having to deal with the fact that you might have asked for a pending one - ideally we would have a get_closed_block_info taking a block_n: u64, or a kind of block tag that excludes pending blocks.

i really want to refactor the MadaraBackend at some point because it's full of dumb stuff like that (which is why you have to deal with soo many error cases right now)

cchudant:
```rs
let block_info = backend
  .get_block_info(&BlockId::Tag(BlockTag::Latest))
  .map_err(make_db_error("getting latest block info"))?
  .ok_or(BlockImportError::Internal("No latest block after reorg".into()))?;
let block_info = block_info.as_nonpending()
  .ok_or(BlockImportError::Internal("Asked for a closed block; got a pending one".into()))?;
// and then you can just have `block_info.block_hash` which is a `Felt`.
```

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -89,7 +89,10 @@ impl MadaraBackend {
     }
 
     pub fn revert_to(&self, revert_to: u64) -> Result<(), MadaraStorageError> {
```
cchudant, line 91: you also need to clear the pending block here

cchudant (changes requested):
one quick test that I would really love to see is a very dumb test that tries to import some blocks, then revert all of them and at that point the db should be completely empty

this would make sure we're not missing anything that should be reverted

cchudant:
(my review is not final i probably missed some stuff, i'll continue tomorrow I think)

madara/crates/client/db/src/contract_db.rs
```diff
@@ -202,6 +203,82 @@ impl MadaraBackend {
         Ok(())
     }
 
+    // TODO: does this need rayon thread pool? what about #[tracing::instrument()] ?
```
cchudant: i don't know about tracing::instrument - currently it's sprinkled without much thought in the codebase and we don't really have a clear rule. We should probably ask @heemankv what the rule should be to be honest.

cchudant: does this need rayon thread pool? => right now I don't think you're using anything that uses rayon, so you don't really need to put that comment.

context:
the reason the use of rayon is documented on top of the functions is that when calling a rayon function from the tokio pool, it'll block and starve the tokio worker, which is suboptimal.
We use rayon functions (par_iter, etc) for storage update functions because it allows us to offload compute-heavy stuff into a pool that's made for it ; and rayon has a very very low parallelization overhead (but can only be used for compute stuff) - all of this without starving the tokio workers which ar for IO stuff.

It's the caller responsability to make sure it's not calling a function that uses rayon stuff from the tokio pool.
Tokio pool handles IO stuff and is where async/.await can be used.

Tokio has a section explaining what I mean: https://docs.rs/tokio/latest/tokio/#cpu-bound-tasks-and-blocking-code

madara/crates/client/db/src/contract_db.rs
```diff
@@ -202,6 +203,82 @@ impl MadaraBackend {
...
+            diff.storage_diffs.iter().for_each(|diff_item| {
+                diff_item.storage_entries.iter().for_each(|entry| {
+                    contract_storage_keys.insert((diff_item.address, entry.key));
+                });
+            });
+        }
+        let latest_block_n = self.get_latest_block_n()?.unwrap(); // TODO: unwrap - Option here probably relates to genesis block
+        for block_n in (revert_to + 1..latest_block_n + 1).rev() {
+            // for each entry in the keys we collected above, create a db key with that entry and the block number and delete it
+            // TODO: we may be able to leverage RocksDb's tree-based iterator here to avoid looping over each block
```
cchudant: true!

cchudant (changes requested):

cchudant:
> > should go into the backend imo, and be called revert_to
> 
> I tried moving this to `mc_db` but I don't think it really fits there because it's doing a lot with the bonsai trie, and the backend doesn't touch bonsai tries much (even though they are accessed through `MadaraBackend`).
> 
> The current reorg logic (which, yes, should be renamed to `revert`) still seems pretty closely tied with all all of the logic in `verify_apply.rs`, it's basically the opposite of `verify_apply_inner()`:
> 
>     * it cares about the current block hash/num (similar to the `check_parent_hash_and_num()` call)
> 
>     * it reverts trie updates (opposite of the `update_tries()` call)
> 
>     * it reverts db state (opposite of `backend.store_block()` call)
> 
> 
> In addition, testing it the way I've done so far relies on building a blockchain by calls to `verify_apply_inner()`.
> 
> Let me know if I'm still thinking about this the wrong way.
> 
> (CC @cchudant)

i think it's fine for now yes

madara/crates/client/block_production/src/lib.rs
notlesh: Yeah, seems messy. I'm up for suggestions

notlesh:
> one quick test that I would really love to see is a very dumb test that tries to import some blocks, then revert all of them and at that point the db should be completely empty
> 
> this would make sure we're not missing anything that should be reverted

The tests do that, but an explicit check to ensure there is no state whatsoever is a good idea. This would work for both the db and bonsai tries...

madara/crates/client/db/src/contract_db.rs
notlesh: [1b9a023](https://github.com/madara-alliance/madara/pull/296/commits/1b9a023d212f021cbdde8fb59627ed0c31f66dff)

madara/crates/client/db/src/block_db.rs
notlesh: [1b9a023](https://github.com/madara-alliance/madara/pull/296/commits/1b9a023d212f021cbdde8fb59627ed0c31f66dff)

madara/crates/client/db/src/class_db.rs
notlesh: [702aee5](https://github.com/madara-alliance/madara/pull/296/commits/702aee5f2cccdd94c5e0f037dfebd43cb3776a9d)

madara/crates/client/db/src/class_db.rs
notlesh: Good catch, this case can be simplified (done in [91ec01c](https://github.com/madara-alliance/madara/pull/296/commits/91ec01c0a5b3f69c69d085e5def75b536f262331)).

The `HashMap`s were used to dedupe things, but that's not relevant for declared classes since that can only happen once.

Note that the similar case in `contract_db.rs` is still a relevant optimization, see below.

madara/crates/client/db/src/contract_db.rs
notlesh: This is an optimization to dedupe things. However, without the RocksDb tree iterator optimization it's actually creating more work..! Good catch, I'll go ahead and work on adding it.

notlesh: I started to implement this and realized it isn't necessarily an optimization -- it would probably perform worse in most cases. It might perform better in cases where a value is changed a lot (e.g. storage items in a popular dex).

Given that, I'd prefer the simplest approach, and so I ripped out the dedupe code as you suggested.

This leaves some caveats, namely that some important things aren't validated:
* `StateDiff`s passed to `contract_db_revert` need a corresponding block number, and there's no way to enforce this. That said, I forced the caller to pair them together `Vec<(u64, StateDiff)>`, and we can be reasonably sure `block_db_revert()` assembles this properly.
* The diffs should be contiguous and end on the current blockchain tip
* `RocksDb` doesn't appear to give you any way to know if something was deleted or how many things were deleted, which would otherwise be a great/simple/low-overhead sanity check to ensure the revert code is removing the exact items it expects

`*_db_revert()` fns are all `pub(crate)` and are documented with these caveats at least.

See [cd27f22](https://github.com/madara-alliance/madara/pull/296/commits/cd27f2203d8d15688eb8ac2e08393890b70e3968)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Ok, thanks for clarifying. Is there anything about this I should mention in the fn documentation, or is this obvious enough?

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Good points, thanks. I believe `clippy` can handle some of this, it would be good to add to CI...

[e33d5b9](https://github.com/madara-alliance/madara/pull/296/commits/e33d5b90867e5ee6980a11706afe4098db2b0355)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: [e33d5b9](https://github.com/madara-alliance/madara/pull/296/commits/e33d5b90867e5ee6980a11706afe4098db2b0355)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: [e33d5b9](https://github.com/madara-alliance/madara/pull/296/commits/e33d5b90867e5ee6980a11706afe4098db2b0355) and [e17319b](https://github.com/madara-alliance/madara/pull/296/commits/e17319baa14f6a99baaea0e4e8ae82515cde57bc)

madara/crates/client/block_import/src/verify_apply.rs
notlesh: Ah, this gives some good context, thanks.

One problem: the non-pending block doesn't have a block number :laughing: 

```rust
pub struct MadaraBlockInfo {
    pub header: Header,
    pub block_hash: Felt,
    pub tx_hashes: Vec<Felt>,
}
```

notlesh: Ah, thanks, this actually gives me a lot of good context. Cleaned this up more in [807f040](https://github.com/madara-alliance/madara/pull/296/commits/807f040131a17455a3845823878a3d6a598c919c)

madara/crates/client/db/src/block_db.rs
notlesh: [807f040](https://github.com/madara-alliance/madara/pull/296/commits/807f040131a17455a3845823878a3d6a598c919c)

madara/crates/client/block_import/src/verify_apply.rs
cchudant: it was obvious to me, but maybe adding a quick "caller should ensure noone is saving or reverting any block while this function is running" can be good
(although, store_block & cie also have the same problem and it's not documented)

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -334,20 +334,23 @@ pub fn revert_to(backend: &MadaraBackend, new_tip_block_hash: &Felt) -> Result<R
...
+            BlockImportError::Internal(format!("no block found for requested tip {new_tip_block_hash:#x}").into())
+        })?;
+    let target_block_info =
+        target_block_info.as_nonpending().ok_or(BlockImportError::Internal("target block cannot be pending".into()))?;
 
     let previous_tip_block_info = backend
         .get_block_info(&BlockId::Tag(BlockTag::Latest))
         .map_err(make_db_error("getting current tip block info"))?
-        .ok_or(BlockImportError::Internal(Cow::Owned("on blockchain tip in storage".to_string())))?;
+        .ok_or_else(|| BlockImportError::Internal(Cow::Borrowed("on blockchain tip in storage")))?;
```
cchudant: BlockImportError::Internal("There are no blocks in the database".into())

cchudant: oh by the way, on starknet the genesis block is block 0 and is executable
we import it and is mostly treated like any other block

when doing `.get_block_info(&BlockId::Tag(BlockTag::Latest))` it can indeed return None, if the genesis block has not been imported yet

(in that case, means there is no latest block in db, and get_latest_block_n is None)

madara/crates/client/block_import/src/verify_apply.rs
```diff
@@ -334,20 +334,23 @@ pub fn revert_to(backend: &MadaraBackend, new_tip_block_hash: &Felt) -> Result<R
...
 
-    let target_block_number = target_block_info.block_n().expect("Target block must have a block number");
+    let target_block_number = target_block_info.header.block_number;
     let target_block_id = BasicId::new(target_block_number);
 
-    let previous_tip_block_number =
-        previous_tip_block_info.block_n().expect("Previous tip block must have a block number");
+    let previous_tip_block_number = previous_tip_block_info
+        .block_n()
+        .ok_or_else(|| BlockImportError::Internal(Cow::Borrowed("Previous tip block must have a block number")))?;
```
cchudant: nit: Cow Borrowed can be replaced with .into() on the litteral

madara/crates/client/db/src/block_db.rs
```diff
@@ -391,9 +396,12 @@ impl MadaraBackend {
             let block_hash_encoded = bincode::serialize(&block_info.block_hash)?;
 
             // get state diff for this block before removing it
-            let state_diff_serialized = self.db.get_cf(&block_n_to_state_diff, block_n_encoded.clone())?.unwrap(); // TODO: unwrap, should we expect an entry for every single block here, or will some be None?
+            let state_diff_serialized =
+                self.db.get_cf(&block_n_to_state_diff, block_n_encoded.clone())?.ok_or_else(|| {
+                    MadaraStorageError::InconsistentStorage(Cow::Owned(format!("Block {block_n} has no StateDiff")))
```
cchudant, line 398: nit: Cow::Owned can be replaced with .into() on the result of format!

madara/crates/client/db/src/storage_updates.rs
cchudant: self.clear_pending_block() iirc

madara/crates/client/block_production/Cargo.toml
```diff
@@ -83,3 +83,6 @@ tracing = { workspace = true }
 tracing-core = { workspace = true, default-features = false }
 tracing-opentelemetry = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+
+# testing
+rstest = { workspace = true, optional = true }
```
cchudant: this needs to be in dev-dependencies

madara/crates/client/block_production/src/lib.rs
cchudant: it would make more sense in block import but i dont really have a strong opinion, it's probably fine

madara/crates/client/db/src/block_db.rs
cchudant: IIRC no

madara/crates/client/db/src/block_db.rs
```diff
@@ -355,6 +357,68 @@ impl MadaraBackend {
...
+        let block_hash_to_block_n = self.db.get_column(Column::BlockHashToBlockN);
+        let block_n_to_block = self.db.get_column(Column::BlockNToBlockInfo);
+        let block_n_to_block_inner = self.db.get_column(Column::BlockNToBlockInner);
+        let block_n_to_state_diff = self.db.get_column(Column::BlockNToStateDiff);
+        let meta = self.db.get_column(Column::BlockStorageMeta);
+
+        // clear pending
+        tx.delete_cf(&meta, ROW_PENDING_INFO);
+        tx.delete_cf(&meta, ROW_PENDING_INNER);
+        tx.delete_cf(&meta, ROW_PENDING_STATE_UPDATE);
```
cchudant: we also need to clear the pending block in classdb and blockdb, but I think it's just easier to remove the clearing of pending block from here and just call the .clear_pending_block() fn in the main db revert_to() function

cchudant (changes requested):
clearing the pending block is the only remaining issue I think, the rest is mostly style nitpicks

madara/crates/client/block_import/src/verify_apply.rs
notlesh: [f080b5c](https://github.com/madara-alliance/madara/pull/296/commits/f080b5c08b130542ecdc42cbfadaaef8dd32f2ae)

madara/crates/client/db/src/storage_updates.rs
notlesh: [7518bc7](https://github.com/madara-alliance/madara/pull/296/commits/7518bc7d7dce70a51929cd4d7c1ac80732ef8026)

notlesh: A bit off topic, but why do we store pending blocks in the db? A pending block is just a block that is being built, right? Why not keep it in RAM until it's become a proper block?

madara/crates/client/db/src/block_db.rs
notlesh: Removed this (since `clear_pending_block()` is being called now) and added some clarification about it in [6b49010](https://github.com/madara-alliance/madara/pull/296/commits/6b4901044efe1d400b6a48a5bbfc2b31812479d7)

madara/crates/client/block_production/Cargo.toml
notlesh: I ended up just copying the fixtures that I needed in [9b0e528](https://github.com/madara-alliance/madara/pull/296/commits/9b0e528788bfe1776091dec56852b36b862b0c59). I couldn't find a clean way to share them across crates with just `#[cfg(test)]` so I figured this was the lesser of two evils.

notlesh:
> one quick test that I would really love to see is a very dumb test that tries to import some blocks, then revert all of them and at that point the db should be completely empty

This was a great idea. I applied it to the `bonsai-trie`s in [694c76d](https://github.com/madara-alliance/madara/pull/296/commits/694c76da0b715ebf22ab9d7560801e889d4fbe13) and discovered [this bug](https://github.com/madara-alliance/bonsai-trie/issues/42).

~~I tried applying it to the `RocksDb`s, but it doesn't appear that there's easy ways to do things like query the number of items stored. Of course, this can be done, but it may be tedious / inefficient (e.g. iterate through each key by prefix `''`). Let me know if you have suggestions on this.~~

I added a test-only way to query the number of items in a column in [60ba88e](https://github.com/madara-alliance/madara/pull/296/commits/60ba88e896a7e6db6a61fd715d0ae635993b4b91) and tested the relevant ones within my existing tests.

This also uncovered [what I think is another bug](https://github.com/madara-alliance/madara/issues/573) and added a workaround in [c6e4054](https://github.com/madara-alliance/madara/pull/296/commits/c6e40543371461fb27cfa00e29238d7793720331).

madara/crates/client/db/src/storage_updates.rs
cchudant: because we want to be able to keep the pending block around across restarts

I have a PR that will be up soon that replaces the db reads to the pending block to read into RAM, meaning db will only be involved when the pending block changes and needs to be saved again

cchudant (approved):
cool!

PR #298, opened 2024-10-02
fix(docs): updated readme and fixed launcher

antiyro:


Dockerfile
```diff
@@ -1,60 +1,50 @@
...
-    else \
-        echo "Unsupported architecture: $ARCH"; exit 1; \
-    fi && \
-    curl -fLS -o /usr/src/scarb.tar.gz \
-        $SCARB_REPO/$SCARB_VERSION/scarb-$SCARB_VERSION-$PLATFORM.tar.gz && \
-    tar -xz -C /usr/local --strip-components=1 -f /usr/src/scarb.tar.gz
+
+# Installing scarb, new since devnet integration
+# Installation steps are taken from the scarb build script
+# https://github.com/software-mansion/scarb/blob/main/install.sh
```
apoorvsadana:
```suggestion
```

EvolveArt (approved):

Dockerfile
```diff
@@ -1,60 +1,48 @@
...
-
 # Set the working directory
 WORKDIR /usr/local/bin
-
 # Copy the compiled binary from the builder stage
 COPY --from=builder /usr/src/madara/target/release/madara .
 
-# Chain presets to be mounted at startup
-VOLUME crates/primitives/chain_config/presets
-VOLUME crates/primitives/chain_config/resources
```
apoorvsadana, line 57: we don't need to mount this anymore?

apoorvsadana (approved):

Dockerfile
antiyro: nop this is unused and was added as a fix for unaccessible chain configs but now we fetch them (iirc)

PR #304, opened 2024-10-03
fix(sync): pending block retrying mechanism & flaky tests

cchudant:
# Pull Request type

- Bugfix

BlockNotFound, sometimes on sepolia

## What is the new behavior?

Pending block works
Tests have been changed also, so that madara instances are not leaked. What I did to fix it is a bit (actually very) messy though - but it works :) and it's in the tests anyway so it should be fine..?

None


antiyro (approved):
lgtm

apoorvsadana (approved):

PR #306, opened 2024-10-03
feat: gas fee price flag added in cli

Mohiiit:
gas fee price flag added in the cli, developers would be able to fix gas either 0, a number or they can pass on a l1 endpoint to fetch the current gas price as well.

# Pull Request type

- Feature

Currently the only option is to disable it or to let it sync from l1

Resolves: #NA

## What is the new behavior?

- developers would be able to fix the gas price for their use case

No


crates/node/src/service/l1.rs
```diff
@@ -52,13 +52,13 @@ impl L1SyncService {
             None
         };
 
-        let gas_price_sync_enabled = authority && !config.gas_price_sync_disabled;
+        let gas_price_sync_enabled = authority && config.gas_price.is_none();
```
Trantorian1: Should this not be `is_some()`?

!gas_price_sync_disabled = gas_price_enabled

Trantorian1 (changes requested):
Inverted a negation in `l1.rs`

crates/node/src/service/l1.rs
Mohiiit: so as per the new changes, we can either set some values to the gas price, like it can be zero or it can be some other fix fees, so in case it's fixed we don't want to call the gas sync worker, so hence if we have the authority enabled and we have nothing set to gas price, we will call the gas price worker

does it make sense, sorry for not providing the enough information beforehand

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -79,6 +83,8 @@ async fn update_l1_block_metrics(eth_client: &EthereumClient, l1_gas_provider: G
     let current_gas_price = l1_gas_provider.get_gas_prices();
     let eth_gas_price = current_gas_price.eth_l1_gas_price;
 
+    log::debug!("Gas prices fetched is: {:?}", eth_gas_price);
```
apoorvsadana:
```suggestion
    log::debug!("Gas price fetched is: {:?}", eth_gas_price);
```

crates/node/src/main.rs
```diff
@@ -93,10 +93,14 @@ async fn main() -> anyhow::Result<()> {
     );
 
     let l1_gas_setter = GasPriceProvider::new();
+
+    if let Some(fix_gas) = run_cmd.l1_sync_params.gas_price {
+        l1_gas_setter.update_eth_l1_gas_price(fix_gas as u128);
+        l1_gas_setter.update_eth_l1_data_gas_price(fix_gas as u128);
```
apoorvsadana: should we have two different flags for these in the CLI?

apoorvsadana (commented):
Can we add some test case for this? Maybe an e2e one which starts Madara with `--gas-price=X` and checks it's X with a normal fee estimate? This might be harder now that we have `SharedMadara` in e2e though

crates/node/src/service/l1.rs
Trantorian1: Great makes sense! Maybe add that in a comment for future readability :)

Mohiiit: resolved

crates/client/eth/src/l1_gas_price.rs
Mohiiit: resolved

crates/node/src/main.rs
Mohiiit: resolved

apoorvsadana (approved):

Trantorian1 (approved):

PR #307, opened 2024-10-04
fix(fgw): sync from other nodes and block signature

### LINKED ISSUE #72: feat(sync): Utilize Signature Verification Methods to Secure L2-only Blocks
# Feature Request

## Summary

We need to enhance the security of blocks that exist solely on Layer 2 by implementing signature verification methods.

## Action Required

- Implement Signature Verification: Utilize the get_signature and get_public_key endpoints to verify block data on L2.
- Use ECDSA on the STARK Curve for all cryptographic signatures to ensure data integrity and authenticity.

## Goals

- Ensure Data Integrity: Verify that all L2 block data (block_hash & state_diff_commitment) is accurate and unchanged.
- Increase Security: Protect against unauthorized data changes and potential security threats.

## Impact

- Nodes will require updates to handle new verification methods.
- Operators must learn how to use these new features effectively.

## Feedback Requested
- Please provide feedback on this plan, especially any concerns about implementation or potential impacts on system performance.

## Note

- This mechanism is temporary and will be replaced by the Starknet Consensus protocol once it is released
- The key pair will be updated every year (unless the key is compromised, in which case we will replace it immediately)

Trantorian1:
# Pull Request type

- Bugfix

The current Madara feeder gateway presents several issues and does not currently work when synchronizing from another node. This is due to invalid JSON response formats and missing fgw methods which other nodes like Pathfinder or Juno use. This pr aims to fix that.

## What is the new behavior?

- Performs manual serialization of legacy contract ABI into json as part of [`serialize_to_json`](https://github.com/madara-alliance/madara/blob/761f2c1e231bcfedbb9c03a1ceaa74553334ba75/crates/primitives/class/src/compile.rs#L28). This has to be done manually since `bincode`, which is used to store contract ABI in db, does not support the `#[serde(tag = "type")]` [container attribute](https://serde.rs/container-attrs.html).

- Missing fgw methods `get_signature` `get_compiled_class_by_class_hash`, `get_contract_addresses` and `get_public_key` have been added. This does not currently include all possible fgw methods. However, in the absence of a clear spec, this will at least work for the current versions of Juno and Pathfinder.

- Calls to the madara fgw now have their url normalized so partially invalid request such as `/foo///abc` will be evaluated correctly.

- Changes to how certain types are serialized to JSON by the fgw has been updated to more closely match the behavior of the official starknet.io fgw. In the future it would be nice to have this based on specs instead.

- Proper block signing has has been implemented in the fgw: Madara now has a private key it uses to sign the hash of blocks it provides through the fgw. This is still not very secure and we are looking into ways to zeroize the private key and load it from env on use to limit its availabiltiy. Note that this fixes #72 

Yes. Due to changes in some storage structure, existing dbs will need to be re-synchronized from genesis.


antiyro:
Could we take this issue to support Juno and Pathfinder? @Trantorian1

Trantorian1:
> Could we take this issue to support Juno and Pathfinder? @Trantorian1

On it :+1:

crates/client/gateway/src/error.rs
```diff
@@ -71,6 +98,8 @@ impl std::fmt::Display for StarknetError {
 pub enum StarknetErrorCode {
     #[serde(rename = "StarknetErrorCode.BLOCK_NOT_FOUND")]
     BlockNotFound,
+    #[serde(rename = "StarknetErrorCode.NO_BLOCK_HEADER")]
+    NoBlockHeader,
```
jbcaron, line 102: Is this really a sequencer error (even if we don't have specs) and not a custom error

crates/client/gateway/src/server/router.rs
```diff
@@ -3,53 +3,97 @@ use std::{convert::Infallible, sync::Arc};
...
 
 // Main router to redirect to the appropriate sub-router
+#[allow(clippy::too_many_arguments)]
 pub(crate) async fn main_router(
     req: Request<Body>,
     backend: Arc<MadaraBackend>,
     add_transaction_provider: Arc<dyn AddTransactionProvider>,
     feeder_gateway_enable: bool,
     gateway_enable: bool,
+    eth_core_contract_address: H160,
```
jbcaron: I think this data should be in the config chain

crates/client/gateway/src/server/router.rs
```diff
@@ -3,53 +3,97 @@ use std::{convert::Infallible, sync::Arc};
...
+        (&Method::GET, "feeder_gateway/get_contract_addresses") => {
+            Ok(handle_get_contract_addresses(eth_core_contract_address, eth_gps_statement_verifier)
+                .await
+                .unwrap_or_else(Into::into))
+        }
+        (&Method::GET, "feeder_gateway/get_public_key") => {
+            Ok(handle_get_publick_key(public_key).await.unwrap_or_else(Into::into))
+        }
+        _ => {
+            log::warn!("Feeder gateway received invalid request: {path}");
```
jbcaron: it is not a good idea to display a warning every time a client makes a bad request

crates/client/gateway/src/server/router.rs
```diff
@@ -3,53 +3,97 @@ use std::{convert::Infallible, sync::Arc};
...
     add_transaction_provider: Arc<dyn AddTransactionProvider>,
 ) -> Result<Response<Body>, Infallible> {
     match (req.method(), req.uri().path()) {
-        (&Method::POST, "/feeder/add_transaction") => {
+        (&Method::POST, "feeder/add_transaction") => {
             Ok(handle_add_transaction(req, add_transaction_provider).await.unwrap_or_else(Into::into))
         }
-        _ => Ok(not_found_response()),
+        _ => {
+            log::warn!("Gateway received invalid request: {path}");
```
jbcaron: it is not a good idea to display a warning every time a client makes a bad request

crates/client/gateway/src/server/router.rs
jbcaron: it would be interesting to have the same logs as the rpc

crates/client/gateway/src/server/service.rs
```diff
@@ -11,18 +11,24 @@ use hyper::{
...
 
+#[allow(clippy::too_many_arguments)]
 pub async fn start_server(
     db_backend: Arc<MadaraBackend>,
     add_transaction_provider: Arc<dyn AddTransactionProvider>,
     feeder_gateway_enable: bool,
     gateway_enable: bool,
     gateway_external: bool,
     gateway_port: u16,
+    eth_core_contract_address: H160,
```
jbcaron: the chain config is already in MadaraBackend
Remove this

crates/primitives/gateway/src/receipt.rs
```diff
@@ -146,12 +146,24 @@ pub struct ExecutionResources {
     pub builtin_instance_counter: BuiltinCounters,
     pub n_steps: u64,
     pub n_memory_holes: u64,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub data_availability: Option<L1Gas>,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
jbcaron, line 153: This is not a problem with bincode?

crates/client/gateway/src/server/handler.rs
```diff
@@ -21,37 +21,89 @@ use crate::error::StarknetError;
...
+        .or_internal_server_error(format!("Retrieving block info for block {block_id}"))?
+        .ok_or(StarknetError::block_not_found())?;
+
+    match block_info {
+        MadaraMaybePendingBlockInfo::Pending(_) => Err(GatewayError::InternalServerError(format!(
+            "Retrieved pending block info from db for non-pending block {block_id}"
+        ))),
+        MadaraMaybePendingBlockInfo::NotPending(block_info) => {
+            let signature =
+                ProviderBlockSignature { block_hash: block_info.block_hash, signature: block_info.signature };
```
jbcaron: I wonder if we shouldn't calculate the signature just in time because the sequencer key can be renewed

jbcaron:
this PR finally solves #72 🥳

crates/client/gateway/src/error.rs
Trantorian1: Yes, this happens when calling `getBlock?blockNumber=pending&headerOnly=true`. See starknet.io for reference.

crates/client/gateway/src/server/router.rs
Trantorian1: Will move it to debug with a custom target. Was handy during debugging but I see how this could be an issue.

crates/primitives/gateway/src/receipt.rs
Trantorian1: Since the gateway primitives are never stored on disk but only used for data serialization to and from fgw responses, this never ends up interacting with bincode and is therefore safe. I am a bit worried though at how easy it is for us to call bincode on unsupported data formats at the moment, and if it might not be wiser to switch to newer versions of bincode which do not have this ambiguity with serde.

crates/client/gateway/src/server/handler.rs
Trantorian1: I think this is fine for now, given that the signature has only been very loosely implemented and is only a static value in the chain config. We will want to add better edge case handling once block signing has been fully implemented into Madara though.

Trantorian1:
> this PR finally solves #72 🥳

This pr implements a sequencer key very loosely, in the sense that it is only a static value passed to the chain config. Other than that, there is no verification logic, and so I do not think we can count this as fixing #72 :/

crates/client/gateway/src/server/handler.rs
jbcaron: the chainConfig is not completely static, you can restart the sequencer with a different version, a different sequencer address, etc. but with the same db

jbcaron: https://community.starknet.io/t/introducing-p2p-authentication-and-mismatch-resolution-in-v0-12-2/97993

> The key pair will be updated every year (unless the key is compromised, in which case we will replace it immediately)

crates/client/gateway/src/server/service.rs
Trantorian1: Done

crates/client/gateway/src/server/handler.rs
Trantorian1: This has been implemented :+1:

crates/client/gateway/src/client/methods.rs
```diff
@@ -190,6 +205,12 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    #[ignore]
```
jbcaron, line 209: is this still useful?

crates/primitives/utils/Cargo.toml
```diff
@@ -13,13 +13,15 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 
+starknet-core.workspace = true
+
 # Other
 anyhow.workspace = true
 async-trait.workspace = true
 futures.workspace = true
 rayon.workspace = true
-rstest = { workspace = true }
+rstest.workspace = true
```
jbcaron: [dev-dependencies]

crates/client/gateway/src/client/methods.rs
Trantorian1: Just serves as a handy stub which we can uncomment to set up fgw tests. I'm in agreement its not very clean, and I think a combinations of macros and a gateway mock.rs to automatically download and compress fgw responses would be cleaner (would also be nice as I have seen other areas of the codebase rely either on json or manually typing large structs fields). Imo this works until a better system is in place.

crates/primitives/utils/Cargo.toml
Trantorian1: Done :+1:

jbcaron (approved):

Mohiiit (approved):

PR #312, opened 2024-10-07
fix: override cli

### LINKED ISSUE #297: feat(config): Add support for nested chain config override
Right now I am only able to use `--chain-config-override` for the parent fields of the chain config like:

```
 --chain-config-override sequencer_address=0x123
```

I would like to be able to modify nested fields in order for exemple to modify the steps number of the chain config like:

```
--chain-config-override bouncer_config.block_max_capacity.n_steps=100000
```

It apears that [this function](https://github.com/madara-alliance/madara/blob/main/crates/node/src/cli/chain_config_overrides.rs#L21) does not support it

jbcaron:
# Pull Request type

- Bugfix

Resolves: #297 

## What is the new behavior?

- add default versioned_constants
- fix: override_chain_config

No


configs/presets/integration.yaml
```diff
@@ -30,5 +26,5 @@ bouncer_config:
     message_segment_length: 18446744073709551615
     n_events: 18446744073709551615
     state_diff_size: 131072
-sequencer_address: "0x0"
+sequencer_address: "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8"
```
antiyro, line 28: where those this address comes from?

configs/presets/mainnet.yaml
```diff
@@ -30,5 +26,5 @@ bouncer_config:
     message_segment_length: 18446744073709551615
     n_events: 18446744073709551615
     state_diff_size: 131072
-sequencer_address: "0x0"
+sequencer_address: "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8"
```
antiyro, line 28: Same here

configs/presets/sepolia.yaml
```diff
@@ -30,5 +26,5 @@ bouncer_config:
     message_segment_length: 18446744073709551615
     n_events: 18446744073709551615
     state_diff_size: 131072
-sequencer_address: "0x0"
+sequencer_address: "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8"
```
antiyro, line 28: and here

crates/node/src/cli/chain_config_overrides.rs
```diff
@@ -1,73 +1,51 @@
...
+    deserialize_bouncer_config, deserialize_starknet_version, serialize_bouncer_config, serialize_starknet_version,
+    ChainConfig, StarknetVersion,
+};
+use mp_utils::parsers::parse_key_value_yaml;
+use mp_utils::serde::{deserialize_duration, serialize_duration};
 
 /// Override chain config parameters.
-/// Format: "--chain-config-override chain_id=NEW_MADARA --chain-config-override chain_name=NEW_NAME"
-#[derive(clap::Parser, Clone, Debug)]
+/// Format: "--chain-config-override chain_id=SN_MADARA,chain_name=MADARA,block_time=1.5"
```
akhercha: `block_time=1.5` is only a typo right? It should be "1s", "500ms" etc...

jbcaron: yes that was before using our own `deserialize_duration` method

antiyro (approved):

akhercha (approved):
lgtm, thanks!

PR #314, opened 2024-10-07
fix: block production panic

cchudant:
# Pull Request type

- Bugfix

## What is the new behavior?

Fix a panic during block production: when a transaction is rejected,it is not removed properly from the current execution batch. The error can also happen when the execution has multiple batches.
That specific case has been fixed and cannot happen anymore, but just in case, the panic has been removed.

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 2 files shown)


antiyro (approved):
lgtm

jbcaron (approved):

PR #316, opened 2024-10-08
bug(tests): Fix local test scripts

shamsasari:
The test scripts in `scripts` didn't work due to missing environment variables needed by the tests.

The tests also relied on the working directory being set such that test artifacts in source could be read. This has been fixed with `include_bytes!` and `include_str!` isntead, and the the working directory for the test is no longer modified.

### DIFF
--- a/crates/tests/src/lib.rs
+++ b/crates/tests/src/lib.rs
@@ -172,26 +172,14 @@ impl MadaraCmdBuilder {
     }
 
     pub fn run(self) -> MadaraCmd {
-        let output = std::process::Command::new("cargo")
-            .arg("locate-project")
-            .arg("--workspace")
-            .arg("--message-format=plain")
-            .output()
-            .expect("Failed to execute command");
-
-        let cargo_toml_path = String::from_utf8(output.stdout).expect("Invalid UTF-8");
-        let project_root = PathBuf::from(cargo_toml_path.trim()).parent().unwrap().to_path_buf();
-
-        env::set_current_dir(&project_root).expect("Failed to set working directory");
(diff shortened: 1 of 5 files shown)


crates/tests/src/lib.rs
```diff
@@ -183,7 +183,7 @@ impl MadaraCmdBuilder {
         let project_root = PathBuf::from(cargo_toml_path.trim()).parent().unwrap().to_path_buf();
 
         env::set_current_dir(&project_root).expect("Failed to set working directory");
-        let target_bin = option_env!("COVERAGE_BIN").unwrap_or("./target/debug/madara");
+        let target_bin = option_env!("COVERAGE_BIN").expect("COVERAGE_BIN env");
```
cchudant: just use env! instead of option_env! in this case no?
also, why do you need that? if we're setting the current dir to project root it should be fine right?

scripts/e2e-tests.sh
```diff
@@ -11,6 +11,8 @@ anvil --fork-url https://eth.merkle.io --fork-block-number 20395662 &
 subshell() {
     set -e
     cargo build --bin madara --profile dev
+    export COVERAGE_BIN=$(realpath target/debug/madara)
+    export ETH_FORK_URL=https://eth.merkle.io
```
cchudant, line 16: put it in e2e-coverage.sh in that case too imo

cchudant (approved):
thanks! ive always just ignored these tests because i couldnt be bothered to fix this :laughing:

antiyro (approved):

PR #318, opened 2024-10-09
fix(sync): pipeline stalling on machines with few cpu cores

cchudant:
# Pull Request type

- Bugfix

can be tested using numactl to force the usage of only cpu 0
```
numactl -C 0 ./target/debug/madara --no-l1-sync --full -n sepolia
```
The pipeline stalls

## What is the new behavior?

fixed the deadlock

No


cchudant:
at some point when i stop being lazy i will instrument all of these mutexes with tracing (also all of the important tasks too..)

Cargo.toml
```diff
@@ -156,7 +156,11 @@ hyper = { version = "0.14", features = ["server"] }
 ip_network = "0.4"
 lazy_static = { version = "1.4", default-features = false }
 once_cell = "1.19"
-log = { version = "0.4", features = ["std", "kv_std"] }
+log = { version = "0.4", features = [
+  "std",
+  "kv_std",
+  "release_max_level_debug",
```
Trantorian1, line 162: Should't this be set to `release_max_level_info` for a prod build?

crates/client/block_import/src/rayon.rs
```diff
@@ -19,26 +22,44 @@ impl Default for RayonPool {
...
     where
         F: FnOnce() -> R + Send + 'static,
         R: Send + 'static,
     {
-        let _permit = self.semaphore.acquire().await;
+        let max_tasks = self.max_tasks;
+        let permit_id = self.permit_id.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        log::debug!("acquire permit {permit_id}");
+        let permit = self.semaphore.acquire().await.expect("Poisoned semaphore");
+        let n_acquired_permits = self.n_acquired_permits.fetch_add(1, std::sync::atomic::Ordering::SeqCst) + 1;
```
Trantorian1, line 38: Are `permit` and `n_aquired_permits` still needed? They seem to be only used for debug output.

Trantorian1 (commented):
Tested locally, all seems good. Consider removing some debug info though

Cargo.toml
cchudant: bsaically, trace! expands to an `if level >= log::STATIC_MAX_LEVEL && level >= get_level(target)`, and STATIC_MAX_LEVEL is a const which expands to debug when this feature flag is set + when `debug_assertions` is off

so, in production & release builds, trace! calls will be optimized out as deadcode, their arguments won't even be evaludated.

crates/client/block_import/src/rayon.rs
cchudant: yep it's for debug output but i was thinking of letting them here just in case, if we ever need to debug this part of the code again?

cchudant: this should have no performance impact, this is called only once per block

Trantorian1: Yea fair :)

Cargo.toml
cchudant: ah, your question is about info

i think we should keep debug logs in prod builds just in case, and i don't see the perf benefits of removing them

Trantorian1 (approved):

Cargo.toml
```diff
@@ -53,7 +53,7 @@ default-members = [
 
 [profile.dev]
 incremental = true
-panic = "abort"
+# panic = "abort"
```
apoorvsadana:
```suggestion
```

crates/client/block_import/src/rayon.rs
```diff
@@ -19,26 +22,44 @@ impl Default for RayonPool {
 impl RayonPool {
     pub fn new() -> Self {
         let n_cores = thread::available_parallelism().expect("Getting the number of cores").get();
-
         let max_tasks = n_cores * 2;
```
apoorvsadana, line 25: Just thinking loud here (and probably this isn't for this issue) - Juno has an interesting feature where I can set the max concurrent vms that will be used for `starknet_calls`. Playing around with that allowed use to configure Juno fro Kakarot in a different way than Starknet because Juno has a lot of heavy calls. Can the configuring `2` number here also help chains that have blocks of different sizes?

apoorvsadana (approved):

crates/client/block_import/src/rayon.rs
Trantorian1: If I am correct that might be a different problem entirely as the `RayonPool` is specifically not made to be used in a RPC context.

apoorvsadana: I see but my point was more around configuring number of cores you use depending on the application needs. From what I understand, the `RayonPool` is used for block imports. So the number of block imports you want to work with concurrently might change depending upon how heavy the block is maybe? Anyways, this is not urgent and maybe I should find a breaking case and then we can add this :)

cchudant: @apoorvsadana 

> So the number of block imports you want to work with concurrently might change depending upon how heavy the block is maybe

That's a good question, but it's not relevant here - this is a failsafe that allows backpressure in the pipeline. It basically only helps to stop a DoS vector when we'll have p2p and people will be able to push blocks to us.
This should not affect the speed of the pipeline.

The question about tuning the numbers wrt. network speed, memory, size of blocks is higher up in the pipeline: we handle 10 fetches (including prevalidate which is class compilation, conversion & hashing) in parallel, and it's still in sync/l2.rs. It'll be moved to block import, someday:tm: and yea we forgot about setting a cli (or chainconfig?) param for it.

PR #263, opened 2024-09-14
fix : events from call info / UDC cairo 0 migration

ocdbytes:
- Events emitting fix.
- UDC migration to cairo 0. (to support starknet.js `deploy` function)

Pending : 

- need to add unit test cases.

PR #324, opened 2024-10-11
fix(mempool): validator errors were ignored in `mempool/rsc/lib.rs`

Trantorian1:
# Pull Request type

- Bugfix

All validator errors were being ignore in [`crates/client/mempool/src/lib.rs`](https://github.com/madara-alliance/madara/blob/main/crates/client/mempool/src/lib.rs) after #263.

Resolves: #NA

## What is the new behavior?

- Mempool validator errors have been re-enabled


antiyro (approved):

cchudant:
perhaps it would be great if you add a test so that it cannot happen again in the future - wdyt @Trantorian1?

cchudant (approved):
very cool! thank you

Cargo.lock
jbcaron, line 1: is this a push error?

PR #327, opened 2024-10-13
chore: Padded devnet addresses display to 64 characters

akhercha:
# Pull Request type

- Other (please describe): QoL

Resolves: #NA

## What is the new behavior?

Some dev tools check that the provided account addresses are 64 characters long (instead of padding it with zeros). This PR is a smol QoL change allowing us to copy addresses without worrying about any missing leading zeros.

No

### DIFF
--- a/crates/client/devnet/src/predeployed_contracts.rs
+++ b/crates/client/devnet/src/predeployed_contracts.rs
@@ -27,7 +27,7 @@ impl fmt::Display for DevnetKeys {
         writeln!(f, "==== DEVNET PREDEPLOYED CONTRACTS ====")?;
         writeln!(f)?;
         for (i, contract) in self.0.iter().enumerate() {
-            writeln!(f, "(#{}) Address: {:#x}", i + 1, contract.address,)?;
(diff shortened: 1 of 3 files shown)


antiyro (approved):

crates/client/devnet/src/predeployed_contracts.rs
```diff
@@ -27,7 +27,7 @@ impl fmt::Display for DevnetKeys {
         writeln!(f, "==== DEVNET PREDEPLOYED CONTRACTS ====")?;
         writeln!(f)?;
         for (i, contract) in self.0.iter().enumerate() {
-            writeln!(f, "(#{}) Address: {:#x}", i + 1, contract.address,)?;
+            writeln!(f, "(#{}) Address: 0x{:0>64}", i + 1, format!("{:x}", contract.address),)?;
```
jbcaron: why not use the `to_fixed_hex_string` method?

akhercha: didn't even know it exists - updated, thanks!

jbcaron (approved):

PR #328, opened 2024-10-14
fix: block production nonce error

### LINKED ISSUE #323: bug: nonce inconsistency
**Current Behavior**

Currently when using an account everything will work fine until some random point where the nonce returned by the `getNonce` RPC endpoint is not the one that the execution engine expects thus leading to an error and the account being unusable.
There is no way to recover the account after this.

image

**Expected Behavior**
It should not happen.

**Steps to reproduce**

Hard to reproduce because it happens randomly but it's linked to the errors listed on #319

cchudant:
# Pull Request type

Bugfix
- Solves #323
- Added a clear_pending_block when block production on_block_time errors, just in case (unreleated to 323)
- Re-added mempool prevalidation, which was broken by #263 - this pr is a duplicate with #324, sorry didnt see it

closes #323 

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix(db): storing a block needs to clear the current pending block
 - feat(script): added more capabilities to the launcher script
 - fix(fgw): sync from other nodes and block signature
 - fix: added more launcher capabilities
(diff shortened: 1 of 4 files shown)


cchudant:
fixes #323

antiyro (approved):

akhercha (approved):
LGTM - thanks!

PR #332, opened 2024-10-14
fix(devnet): devnet predeployed contracts stable address across systems

cchudant:
# Pull Request type

- Bugfix

Predeployed contract address is not stable across machines. That's because it was using StdRng - replace it with SmallRng which is pure-rust.

YES - all the deployed contract addresses are different now!


Mohiiit:
JS test have hardcoded accounts and private key from the devnet earlier, have to update that as well

antiyro:
> JS test have hardcoded accounts and private key from the devnet earlier, have to update that as well

on it

antiyro (approved):

Mohiiit (approved):

PR #333, opened 2024-10-14
fix: pathfinder block hash mismatch block 0 (sequencer mode)

apoorvsadana:
The `compute_hash` method for state commitments was not sorting the storage entries. Maybe this wasn't caught because the feeder gateway always returns state diffs in a sorted order. I have fixed this and added a test case.

The current code iterates over storage_diffs twice (once to sort storage entries and then to sort storage diffs as well). I feel this is cleaner compared to another alternative of writing custom logic for sorting which takes a mutable and also this doesn't have that much of an overhead. Happy to hear thoughts though!


crates/primitives/state_update/src/lib.rs
```diff
@@ -100,6 +100,7 @@ impl StateDiff {
 
         let storage_diffs_sorted = {
             let mut storage_diffs = self.storage_diffs.clone();
+            storage_diffs.iter_mut().for_each(|storage_diff| storage_diff.sort());
```
Trantorian1: Consider renaming this function to something more obvious, like `sort_storage`

Trantorian1 (commented):
Seems good. Check comments for a minor nitpick.

jbcaron (approved):

crates/primitives/state_update/src/lib.rs
apoorvsadana: fixed!

antiyro (approved):

Trantorian1 (approved):

PR #345, opened 2024-10-15
fix(db): max rocksdb LOG files count and size and add more memory metrics

cchudant:
# Pull Request type

- Bugfix

rocksdb's LOG file is unbounded and takes wayyyy too much space in production

## What is the new behavior?

limit the number of rocksdb LOG files, and their size.
Set the rocksdb level to WARN, which will print a summary of all column stats (~100Ko) at startup only, whereas the INFO level will print it very regularily.
Plus, add a few more memtable metrics to prometheus endpoint.

No


antiyro (approved):
lgtm thanks lets see how it works in full mode

apoorvsadana (approved):

PR #346, opened 2024-10-15
fix(devnet): devnet predeployed contracts stable address across systems (re)

cchudant:
# Pull Request type

- Bugfix

#332 is wrong because of the docs of [SmallRng](https://docs.rs/rand/latest/rand/rngs/struct.SmallRng.html) says:
> The algorithm is deterministic but should not be considered reproducible due to dependence on platform and possible replacement in future library versions.

YES: all the predeployed contracts addresses are changed, once again :)


akhercha (approved):
LGTM, thankssss

antiyro (approved):

PR #347, opened 2024-10-15
feat: added special address with block number and block hash

Mohiiit:
Small fix for SNOS to run, it requires special address in the state diff
# Pull Request type

- Feature

Currently we were not adding the block hash and block number of previous block with special address which was breaking the SNOS

Resolves: #NA

## What is the new behavior?

- The above mention issue is resolved

No

### DIFF
--- a/crates/client/mempool/src/block_production.rs
+++ b/crates/client/mempool/src/block_production.rs
@@ -417,7 +417,10 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
                     let address = Felt::ONE;
                     new_state_diff.storage_diffs.push(ContractStorageDiffItem {
                         address,
-                        storage_entries: vec![StorageEntry { key: Felt::from(block_n), value: prev_block_hash }],
+                        storage_entries: vec![StorageEntry {
+                            key: Felt::from(prev_block_number),
+                            value: prev_block_hash,
+                        }],
                     });
                 }
                 Ok(None) => {
--- a/crates/client/mempool/src/block_production.rs
+++ b/crates/client/mempool/src/block_production.rs
@@ -405,8 +405,31 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
 
         // Complete the block with full bouncer capacity.
         let start_time = Instant::now();
-        let (new_state_diff, _n_executed) =
+        let (mut new_state_diff, _n_executed) =
             self.continue_block(self.backend.chain_config().bouncer_config.block_max_capacity)?;
+
+        if block_n >= 10 {
+            let prev_block_number = block_n - 10;
+            match self.backend.get_block_hash(&BlockId::Number(prev_block_number)) {
+                Ok(Some(prev_block_hash)) => {
+                    let address = Felt::ONE;
+                    new_state_diff.storage_diffs.push(ContractStorageDiffItem {
+                        address,
+                        storage_entries: vec![StorageEntry {
+                            key: Felt::from(prev_block_number),
+                            value: prev_block_hash,
+                        }],
+                    });
+                }
+                Ok(None) => {
+                    // this shouldn't happen, ideally should panic
+                    log::error!("No block hash found for block number {}", prev_block_number);
+                }
+                Err(e) => {
+                    log::error!("Error fetching block hash for block {}: {:?}", prev_block_number, e);
+                }
+            }
+        }
 
         // Convert the pending block to a closed block and save to db.
 
--- a/crates/client/mempool/src/block_production.rs
+++ b/crates/client/mempool/src/block_production.rs
@@ -408,6 +408,8 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         let (mut new_state_diff, _n_executed) =
             self.continue_block(self.backend.chain_config().bouncer_config.block_max_capacity)?;
 
+        // SNOS requirement: For blocks >= 10, the hash of the block 10 blocks prior
+        // at address 0x1 with the block number as the key
         if block_n >= 10 {
             let prev_block_number = block_n - 10;
             match self.backend.get_block_hash(&BlockId::Number(prev_block_number)) {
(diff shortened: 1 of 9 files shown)


crates/client/mempool/src/block_production.rs
```diff
@@ -405,9 +405,32 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
+                        address,
+                        storage_entries: vec![StorageEntry {
+                            key: Felt::from(prev_block_number),
+                            value: prev_block_hash,
+                        }],
+                    });
+                }
+                Ok(None) => {
+                    // this shouldn't happen, ideally should panic
+                    log::error!("No block hash found for block number {}", prev_block_number);
```
apoorvsadana: let's panic ig?

crates/client/mempool/src/block_production.rs
```diff
@@ -405,9 +405,32 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
         // Complete the block with full bouncer capacity.
         let start_time = Instant::now();
-        let (new_state_diff, _n_executed) =
+        let (mut new_state_diff, _n_executed) =
             self.continue_block(self.backend.chain_config().bouncer_config.block_max_capacity)?;
 
+        if block_n >= 10 {
+            let prev_block_number = block_n - 10;
+            match self.backend.get_block_hash(&BlockId::Number(prev_block_number)) {
+                Ok(Some(prev_block_hash)) => {
```
apoorvsadana: let's add a comment

crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -263,6 +263,8 @@ impl ChainConfig {
             chain_name: "Madara".into(),
             chain_id: ChainId::Other("MADARA_DEVNET".into()),
             sequencer_address: Felt::from_hex_unchecked("0x123").try_into().unwrap(),
+            block_time: Duration::from_secs(4),
+            pending_block_update_time: Duration::from_secs(1),
```
apoorvsadana:
```suggestion

```

apoorvsadana (approved):

crates/client/mempool/src/block_production.rs
Mohiiit: added

crates/client/mempool/src/block_production.rs
Mohiiit: resolved

crates/primitives/chain_config/src/chain_config.rs
Mohiiit: resolved

apoorvsadana (approved):

jbcaron:
This logic for block hashes is already implemented for blockifier with the StateAdapter, it should be changed accordingly while keeping a compatibility of re-execution of the mainnet.
https://github.com/madara-alliance/madara/blob/fe0993210671a21c3f19dbc654d8e5aab0f37c90/crates/client/exec/src/blockifier_state_adapter.rs#L33-L55

apoorvsadana:
@jbcaron, I am mostly missing something but do we need this custom logic in the reader? The feeder gateway does return state diffs where it says `0x1` has the following storage keys and values. In full node mode, if we are applying these state diffs in the DB, won't normal fetching from DB work same as this custom logic we're using here?

jbcaron:
we should just delete this code in the same PR but be sure that we can re-execute the mainnet blocks (I don't know from when this feature was introduced)
I checked on sepolia, there is no issue

apoorvsadana:
> we should just delete this code in the same PR but be sure that we can re-execute the mainnet blocks 

makes sense, we can delete and see if the test cases pass + that we can resync a few hundred blocks on mainnet + get_storage_at on 0x1 address works as expected

EDIT: @Mohiiit has also already informed the visoft team for e2e cases for this specific case

Trantorian1 (dismissed):

crates/client/mempool/src/block_production.rs
```diff
@@ -405,9 +405,34 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
         let start_time = Instant::now();
-        let (new_state_diff, _n_executed) =
+        let (mut new_state_diff, _n_executed) =
             self.continue_block(self.backend.chain_config().bouncer_config.block_max_capacity)?;
 
+        // SNOS requirement: For blocks >= 10, the hash of the block 10 blocks prior
+        // at address 0x1 with the block number as the key
+        if block_n >= 10 {
+            let prev_block_number = block_n - 10;
+            match self.backend.get_block_hash(&BlockId::Number(prev_block_number)) {
```
cchudant: this match can be simplified to
```rs
let prev_block_hash = self.backend.get_block_hash(&BlockId::Number(prev_block_number))
  .map_err(|err| Error::Unexpected(
      format!("Error fetching block hash for block {prev_block_number}: {err:#}").into()
  ))?
  .ok_or_else(|| Error::Unexpected(
      format!("No block hash found for block number {prev_block_number}").into()
  ))?;
let address = Felt::ONE;
new_state_diff.storage_diffs.push(ContractStorageDiffItem {
    address,
    storage_entries: vec![StorageEntry { key: Felt::from(block_n), value: prev_block_hash }],
});
  ```

Comment (Mohiiit) [reply crates/client/mempool/src/block_production.rs]: resolved

File: crates/client/rpc/src/versions/v0_7_1/methods/read/get_storage_at.rs [hunk 5]
```diff
@@ -45,17 +45,25 @@ pub fn get_storage_at(
...
+        block_number,
+        Some(DbBlockId::Number(num)) if num >= 10 && contract_address == Felt::ONE
+    );
+
+    if !skip_contract_check {
+        starknet
+            .backend
+            .get_contract_class_hash_at(&block_id, &contract_address)
+            .or_internal_server_error("Failed to check if contract is deployed")?
+            .ok_or(StarknetRpcApiError::ContractNotFound)?;
```
jbcaron, line 60: I think we should only return 0x0 in this case
we can therefore remove this check
this is already the implementation in the state adapter but maybe this logic should be in the db
@cchudant  what do you think?

Mohiiit: I checked with the juno rpc call and it has this behaviour hence implemented this

cchudant: apparently you have to return ContractNotFound
https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L216

but we need to bring this up into full node collab channel - getNonce also returns ContractNotFound but I think it's possible to increment the nonce of an undeclared contract

jbcaron (approved):

PR #348, opened 2024-10-16
Removes Prometheus and adds OTEL for metrics/logs/traces

heemankv:
# This PR removes prometheus and adds OTEL for trace/logs/metrics collection.

### Metrics:
- Prometheus is used as a metrics collector.
- Prometheus registry is passed to appropriate services and used.
- Data is collected and sent to Prometheus backend which connects to dashboards (Grafana) for visualisation.
### Logs:  
- `utils/setup_logger` is used to format the logs.
- uses `log::info/trace/debug/warn/error` to show logs on CLI.
### Traces: 
- currently not supported

## What is the new behavior?
- Implemented `Analytics` that initiates OTEL `metrics/logs/traces` providers and sets them globally for use.
- `Analytics` is initiated at the top of `main` function of  `madara node`.
- Each service implements it's own metric services that uses global provider set above.
- These services then creates appropriate `meter` and `instruments` and register data.
- Data is batched and channeled to `otel-collector` service running externally.

- Moreover, `otel-collector` can be configured to export data to numerous backend and dashboards.  
- A dashboard created on Signoz is also attached.

### NOTE: 
- `Analytics` doesn't implement `Service` like any other client, because : 
     1) It's setup is a sync function.
     2) It needs to be started before anything, adding it to current service setup  fails to promise that.
- Any logs that were defined like `logs::info` are changed to use [tracing lib](https://github.com/tokio-rs/tracing) like : `tracing::info`
- Appropriate functions are instrumented for traces using `#[tracing::instrument()]` feature of `tracing`.
- Prometheus Gauge/Counter instruments are replaced with their Otel correspondent.
- Addition of `#[tracing::instrument()]` is incremental, as we use the services more, we'll add more tracing.

No, but the analytics services currently setup with prometheus will stop working.

## Other information

Example of tracing : 
Screenshot 2024-10-17 at 4 04 58 PM
Example of new format for logs :
Screenshot 2024-10-17 at 4 06 52 PM


heemankv (commented):
Self review completed, fixes can be found [here](https://github.com/madara-alliance/madara/pull/348).

crates/client/analytics/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+  "metrics",
+  "logs",
+] }
+opentelemetry-semantic-conventions = { workspace = true }
+opentelemetry-stdout = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio", "logs"] }
+tokio = { workspace = true }
+tracing = { workspace = true }
+tracing-core = { workspace = true, default-features = false }
+tracing-opentelemetry = "0.26.0"
```
Mohiiit: let's use workspace here as well

crates/client/analytics/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "mc-analytics"
+description = "Madara client for open telemetry analytics service"
```
Mohiiit:
```suggestion
description = "Madara client open telemetry analytics service"
```

crates/client/block_import/Cargo.toml
```diff
@@ -44,6 +45,23 @@ starknet-core.workspace = true
...
+  "tonic",
+  "metrics",
+  "logs",
+] }
+opentelemetry-semantic-conventions = { workspace = true }
+opentelemetry-stdout = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio", "logs"] }
+tracing = { workspace = true }
+tracing-core = { workspace = true, default-features = false }
+tracing-opentelemetry = "0.26.0"
```
Mohiiit: use workspace

crates/client/db/Cargo.toml
```diff
@@ -48,6 +48,23 @@ tokio = { workspace = true, features = [
...
+  "tonic",
+  "metrics",
+  "logs",
+] }
+opentelemetry-semantic-conventions = { workspace = true }
+opentelemetry-stdout = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio", "logs"] }
+tracing = { workspace = true }
+tracing-core = { workspace = true, default-features = false }
+tracing-opentelemetry = "0.26.0"
```
Mohiiit: use workspace

crates/client/devnet/Cargo.toml
```diff
@@ -53,3 +52,19 @@ anyhow.workspace = true
...
+  "tonic",
+  "metrics",
+  "logs",
+] }
+opentelemetry-semantic-conventions = { workspace = true }
+opentelemetry-stdout = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio", "logs"] }
+tracing = { workspace = true }
+tracing-core = { workspace = true, default-features = false }
+tracing-opentelemetry = "0.26.0"
```
Mohiiit: same as earlier

crates/client/devnet/src/lib.rs
```diff
@@ -313,7 +316,7 @@ mod tests {
             )
             .unwrap();
 
-        log::debug!("{:?}", backend.get_block_info(&BlockId::Tag(BlockTag::Latest)));
+        tracing::debug!("{:?}", backend.get_block_info(&BlockId::Tag(BlockTag::Latest)));
```
Mohiiit: let's add a proper comment here?

crates/client/rpc/src/versions/v0_7_1/methods/read/get_block_with_receipts.rs
```diff
@@ -11,7 +11,7 @@ pub fn get_block_with_receipts(
     starknet: &Starknet,
     block_id: BlockId,
 ) -> StarknetRpcResult<MaybePendingBlockWithReceipts> {
-    log::debug!("block_id {block_id:?}");
+    tracing::debug!("block_id {block_id:?}");
```
Mohiiit: let's update the comment here

crates/node/src/cli/analytics.rs
```diff
@@ -0,0 +1,28 @@
+use clap::Args;
+use url::Url;
+
+/// Parameters used to config analytics.
+#[derive(Debug, Clone, Args)]
+pub struct AnalyticsParams {
+    /// Name of the service.
+    #[arg(env = "MADARA_ANALYTICS_SERVICE_NAME", long, alias = "analytics", default_value = "madara")]
```
Mohiiit:
```suggestion
    #[arg(env = "MADARA_ANALYTICS_SERVICE_NAME", long, alias = "analytics", default_value = "madara-analstics")]
```

crates/node/src/main.rs
```diff
@@ -194,21 +192,24 @@ async fn main() -> anyhow::Result<()> {
...
             // This log is immediately flooded with devnet accounts and so this can be missed.
             // Should we add a delay here to make this clearly visisble?
-            log::warn!("You're running a devnet with the network config of {:?}. This means that devnet transactions can be replayed on the actual network.", run_cmd.network);
+            tracing::warn!("You're running a devnet with the network config of {:?}. This means that devnet transactions can be replayed on the actual network.", run_cmd.network);
         }
     }
 
     app.start_and_drive_to_end().await?;
+
+    tracing::info!("Shutting down analytics");
```
Mohiiit, line 202: is it getting killed while shutting down just like other services tho?

Mohiiit (commented):
Although all the logs are being converted to tracing, if the scope of the PR also includes adding detailed tracing, we might have to add more logs.

cchudant:
Very cool, I've wanted to move to tracing entirely for some time
I have one small thing to note: i want the logs to be kept as clean as possible if you start madara without just to test it or develop with it, which is why i've done all this https://github.com/madara-alliance/madara/blob/main/crates/node/src/util.rs#L40 which is a bit cursed (for example, I check the log target to show "HTTP" in purple instead of the blue "INFO", I hide the target in some cases (rpc internal server error), add emojis to the warn/errors 
and also the HTTP logs show the time elapsed in color etc etc. All of this is very terminal-specific and isn't made to work in production with OTEL stack - it's really for development and devnet uses.

I'd like to keep that behavior by doing a custom tracing terminal format, which would be disabled / hooked up differently for production where the full OTEL stack makes more sense, wdyt @heemankv

crates/client/analytics/Cargo.toml
heemankv: Great catch @Mohiiit !
I have updated it [here](https://github.com/madara-alliance/madara/pull/348/commits/f6b21b54ed153b788abdf30176ebfde4833783b7).

crates/client/block_import/Cargo.toml
heemankv: Updated [here](https://github.com/madara-alliance/madara/pull/348/commits/f6b21b54ed153b788abdf30176ebfde4833783b7).

crates/client/db/Cargo.toml
heemankv: Updated [here](https://github.com/madara-alliance/madara/pull/348/commits/f6b21b54ed153b788abdf30176ebfde4833783b7).

crates/client/devnet/Cargo.toml
heemankv: Updated [here](https://github.com/madara-alliance/madara/pull/348/commits/f6b21b54ed153b788abdf30176ebfde4833783b7).

crates/client/devnet/src/lib.rs
heemankv: I can come up with :
```tracing::debug!("block imported {:?}", backend.get_block_info(&BlockId::Tag(BlockTag::Latest)));```
what do you suggest ?

File: crates/client/rpc/src/versions/v0_7_1/methods/read/get_block_with_receipts.rs (hunk 7 above)
Comment (heemankv) [reply crates/client/rpc/src/versions/v0_7_1/methods/read/get_block_with_receipts.rs]: Updated to 
```tracing::debug!("get_block_with_receipts called with {:?}", block_id);```
[here](https://github.com/madara-alliance/madara/pull/348/commits/7da86a8fc2fc83f7da37ba8b78344a66f7863657).

Comment (heemankv) [reply crates/client/rpc/src/versions/v0_7_1/methods/read/get_block_with_receipts.rs]: Updated [here](https://github.com/madara-alliance/madara/pull/348/commits/7da86a8fc2fc83f7da37ba8b78344a66f7863657).

File: crates/node/src/main.rs (hunk 9 above)
Comment (heemankv) [reply crates/node/src/main.rs:202]: As mentioned in the PR description :
>Analytics doesn't implement Service like any other client, because :
>It's setup is a sync function.
>It needs to be started before anything, adding it to current service setup fails to promise that.

Analytics is not a `Service` as per's this codebase's definition is concerned.
And it needs to end at the very end to ensure that all ending logs are updated as well. 
Hence the separation in setup and shutdown

**Comment by heemankv** [comment]
> Although all the logs are being converted to tracing, if the scope of the PR also includes adding detailed tracing, we might have to add more logs.

As mentioned in the PR description :
> Addition of #[tracing::instrument()] is incremental, as we use the services more, we'll add more tracing.

Let's keep adding more logs and tracing incremental, and mark this PR as a start for that effort, 
we can do more rounds of PRs to add tracing and logs, after which any new PR needs to handle that by itself.

WDYT @Mohiiit

**Comment by Mohiiit** [comment]
> > Although all the logs are being converted to tracing, if the scope of the PR also includes adding detailed tracing, we might have to add more logs.
> 
> As mentioned in the PR description :
> 
> > Addition of #[tracing::instrument()] is incremental, as we use the services more, we'll add more tracing.
> 
> Let's keep adding more logs and tracing incremental, and mark this PR as a start for that effort, we can do more rounds of PRs to add tracing and logs, after which any new PR needs to handle that by itself.
> 
> WDYT @Mohiiit

cool, my bad

File: crates/client/analytics/src/lib.rs [hunk 10]
```diff
@@ -0,0 +1,226 @@
...
+    meter_provider: Option<SdkMeterProvider>,
+    service_name: String,
+    log_level: Level,
+    collection_endpoint: Option<Url>,
+}
+
+impl Analytics {
+    pub fn new(service_name: String, log_level: String, collection_endpoint: Option<Url>) -> anyhow::Result<Self> {
+        let log_level = Level::from_str(&log_level).unwrap_or(Level::INFO);
+        Ok(Self { meter_provider: None, service_name, log_level, collection_endpoint })
```
Trantorian1:
```suggestion
    pub fn new(service_name: String, log_level: tracing::Level, collection_endpoint: Option<Url>) -> anyhow::Result<Self> {
        Ok(Self { meter_provider: None, service_name, log_level, collection_endpoint })
```

This seems less error-prone than relying on a string here. You can then use `tracing::Level::ERROR` and related.

crates/client/analytics/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            .with_thread_names(true)
+            .with_thread_ids(true)
+            .with_file(false)
+            .with_timer(tracing_subscriber::fmt::time::SystemTime)
+            .with_ansi(true);
+
+        let tracing_subscriber = tracing_subscriber::registry()
+            .with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))
+            .with(tracing_subscriber::fmt::layer().event_format(format))
+            .with(EnvFilter::from_default_env());
```
Trantorian1: Doesn't
```rust
.with(EnvFilter::from_default_env());
```

Risk overriding the log level set as
```rust
.with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))
```

crates/client/analytics/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        tracing::info!("OTEL initialized");
+        Ok(())
+    }
+
+    fn init_tracer_provider(&self) -> anyhow::Result<Tracer> {
+        //  Guard clause if otel is disabled
+        let Some(otel_endpoint) = self.collection_endpoint.clone() else {
+            return Err(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel tracer provider"));
+        };
```
Trantorian1:
```suggestion
        let otel_endpoint = self
            .collection_endpoint
            .clone()
            .ok_or(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel tracer provider"))?;
```

crates/client/analytics/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        global::set_tracer_provider(provider.clone());
+        Ok(provider.tracer(format!("{}{}", self.service_name, "_subscriber")))
+    }
+
+    fn init_metric_provider(&self) -> anyhow::Result<SdkMeterProvider> {
+        //  Guard clause if otel is disabled
+        let Some(otel_endpoint) = self.collection_endpoint.clone() else {
+            return Err(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel metric provider"));
+        };
```
Trantorian1:
```suggestion
        let otel_endpoint = self
            .collection_endpoint
            .clone()
            .ok_or(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel metric provider"))?;
```

crates/client/analytics/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            .build();
+        global::set_meter_provider(provider.clone());
+        Ok(provider)
+    }
+
+    fn init_logs(&self) -> anyhow::Result<LoggerProvider> {
+        //  Guard clause if otel is disabled
+        let Some(otel_endpoint) = self.collection_endpoint.clone() else {
+            return Err(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel log provider"));
+        };
```
Trantorian1:
```suggestion
        let otel_endpoint = self
            .collection_endpoint
            .clone()
            .ok_or(anyhow::anyhow!("OTEL endpoint is not set, not initializing otel log provider"))?;
```

crates/client/block_import/src/metrics.rs
```diff
@@ -70,16 +138,17 @@ impl BlockMetrics {
...
 
-        self.l2_block_number.set(block_header.block_number as f64);
-        self.transaction_count.set(f64::from_u64(block_header.transaction_count).unwrap_or(0f64));
-        self.event_count.set(f64::from_u64(block_header.event_count).unwrap_or(0f64));
+        self.l2_sync_time.record(total_sync_time, &[]);
+        self.l2_latest_sync_time.record(latest_sync_time, &[]);
+        self.l2_avg_sync_time.record(total_sync_time / (block_header.block_number - self.starting_block) as f64, &[]);
+
+        self.l2_block_number.record(block_header.block_number as f64, &[]);
+        self.transaction_count.record(f64::from_u64(block_header.transaction_count).unwrap_or(0f64), &[]);
```
Trantorian1: Why not set these to `Gauge<u64>` instead?

crates/client/block_import/src/metrics.rs
```diff
@@ -70,16 +138,17 @@ impl BlockMetrics {
...
+        self.event_count.record(f64::from_u64(block_header.event_count).unwrap_or(0f64), &[]);
 
-        self.l1_gas_price_wei.set(f64::from_u128(block_header.l1_gas_price.eth_l1_gas_price).unwrap_or(0f64));
-        self.l1_gas_price_strk.set(f64::from_u128(block_header.l1_gas_price.strk_l1_gas_price).unwrap_or(0f64));
+        self.l1_gas_price_wei.record(f64::from_u128(block_header.l1_gas_price.eth_l1_gas_price).unwrap_or(0f64), &[]);
```
Trantorian1, line 150: Same here, `Gauge<u128>`?

crates/client/db/src/bonsai_db.rs
```diff
@@ -156,13 +163,13 @@ impl BonsaiDatabase for BonsaiDb<'_> {
 //     }
 
 //     fn get(&self, key: &DatabaseKey) -> Result<Option<Vec<u8>>, Self::DatabaseError> {
```
Trantorian1, line 165: Consider removing theses comments.

crates/client/db/src/class_db.rs
```diff
@@ -25,7 +26,7 @@ impl MadaraBackend {
         nonpending_col: Column,
     ) -> Result<Option<V>, MadaraStorageError> {
         // todo: smallint here to avoid alloc
-        log::debug!("get encoded {key:#x}");
+        tracing::debug!("get encoded {key:#x}");
```
Trantorian1: Something more descriptive? This doesn't feel very clear.

crates/client/db/src/class_db.rs
```diff
@@ -35,7 +36,7 @@ impl MadaraBackend {
                 return Ok(Some(bincode::deserialize(&res)?)); // found in pending
             }
         }
-        log::debug!("get encoded: not in pending");
+        tracing::debug!("get encoded: not in pending");
```
Trantorian1: Same here

crates/client/db/src/class_db.rs
```diff
@@ -73,25 +75,27 @@ impl MadaraBackend {
         if !valid {
             return Ok(None);
         }
-        log::debug!("valid");
+        tracing::debug!("valid");
```
Trantorian1: This log should be made more explicit.

crates/client/db/src/class_db.rs
```diff
@@ -73,25 +75,27 @@ impl MadaraBackend {
...
 
         Ok(Some(info.class_info))
     }
 
+    #[tracing::instrument(skip(self, class_hash), fields(module = "ClassDB"))]
```
Trantorian1: We probably want to keep the class hash as part of the generated `span`

crates/client/db/src/class_db.rs
```diff
@@ -107,6 +111,7 @@ impl MadaraBackend {
     }
 
     /// NB: This functions needs to run on the rayon thread pool
+    #[tracing::instrument(skip(self, block_id, converted_classes, col_info, col_compiled), fields(module = "ClassDB"))]
```
Trantorian1: We probably want to keep the block id as part of the generated `span`

crates/client/db/src/class_db.rs
```diff
@@ -169,6 +174,7 @@ impl MadaraBackend {
     }
 
     /// NB: This functions needs to run on the rayon thread pool
+    #[tracing::instrument(skip(self, block_number, converted_classes), fields(module = "ClassDB"))]
```
Trantorian1: We probably want to keep the block number as part of the generated `span`

crates/client/db/src/contract_db.rs
```diff
@@ -82,6 +83,7 @@ impl MadaraBackend {
         }
     }
 
+    #[tracing::instrument(skip(self, id, contract_addr), fields(module = "ContractDB"))]
```
Trantorian1: We probably want to keep the contract address as part of the generated `span`

crates/client/db/src/contract_db.rs
```diff
@@ -91,6 +93,7 @@ impl MadaraBackend {
         Ok(self.get_contract_class_hash_at(id, contract_addr)?.is_some())
     }
 
+    #[tracing::instrument(skip(self, id, contract_addr), fields(module = "ContractDB"))]
```
Trantorian1, line 96: We probably want to keep the contract address as part of the generated `span`

crates/client/db/src/contract_db.rs
```diff
@@ -105,6 +108,7 @@ impl MadaraBackend {
         )
     }
 
+    #[tracing::instrument(skip(self, id, contract_addr), fields(module = "ContractDB"))]
```
Trantorian1: We probably want to keep the contract address as part of the generated `span`

crates/client/db/src/contract_db.rs
```diff
@@ -115,6 +119,7 @@ impl MadaraBackend {
         })
     }
 
+    #[tracing::instrument(skip(self, id, contract_addr, key), fields(module = "ContractDB"))]
```
Trantorian1: We probably want to keep the contract address  and key as part of the generated `span`

crates/client/db/src/db_metrics.rs
```diff
@@ -45,18 +81,18 @@ impl DbMetrics {
...
         builder.add_db(db);
         let mem_usage = builder.build().context("Getting memory usage")?;
-        self.mem_table_total.set(mem_usage.approximate_mem_table_total() as _);
-        self.mem_table_unflushed.set(mem_usage.approximate_mem_table_unflushed() as _);
-        self.mem_table_readers_total.set(mem_usage.approximate_mem_table_readers_total() as _);
-        self.cache_total.set(mem_usage.approximate_cache_total() as _);
+        self.mem_table_total.record(mem_usage.approximate_mem_table_total() as _, &[]);
+        self.mem_table_unflushed.record(mem_usage.approximate_mem_table_unflushed() as _, &[]);
+        self.mem_table_readers_total.record(mem_usage.approximate_mem_table_readers_total() as _, &[]);
+        self.cache_total.record(mem_usage.approximate_cache_total() as _, &[]);
```
Trantorian1:
```suggestion
        self.mem_table_total.record(mem_usage.approximate_mem_table_total(), &[]);
        self.mem_table_unflushed.record(mem_usage.approximate_mem_table_unflushed(), &[]);
        self.mem_table_readers_total.record(mem_usage.approximate_mem_table_readers_total(), &[]);
        self.cache_total.record(mem_usage.approximate_cache_total(), &[]);
```

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -83,11 +83,12 @@ async fn update_l1_block_metrics(eth_client: &EthereumClient, l1_gas_provider: G
...
 
-    log::debug!("Gas price fetched is: {:?}", eth_gas_price);
+    tracing::debug!("Gas price fetched is: {:?}", eth_gas_price);
 
     // Update the metrics
-    eth_client.l1_block_metrics.l1_block_number.set(latest_block_number as f64);
-    eth_client.l1_block_metrics.l1_gas_price_wei.set(eth_gas_price as f64);
+
+    eth_client.l1_block_metrics.l1_block_number.record(latest_block_number, &[]);
+    eth_client.l1_block_metrics.l1_gas_price_wei.record(eth_gas_price as u64, &[]);
```
Trantorian1, line 91: Why not make this a `Gauge<u128>`?

crates/node/src/cli/analytics.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Parameters used to config analytics.
+#[derive(Debug, Clone, Args)]
+pub struct AnalyticsParams {
+    /// Name of the service.
+    #[arg(env = "MADARA_ANALYTICS_SERVICE_NAME", long, alias = "analytics", default_value = "madara-analstics")]
+    pub analytics_service_name: String,
+
+    /// Log level of the service.
+    #[arg(env = "RUST_LOG", long, default_value = "info")]
+    pub analytics_log_level: String,
```
Trantorian1: Consider using `tracing::Level` with a custom clap parser.

https://docs.rs/clap/4.5.20/clap/_derive/index.html#arg-attributes
https://docs.rs/clap/4.5.20/clap/struct.Arg.html#method.value_parser

crates/node/src/cli/analytics.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub analytics_collection_endpoint: Option<Url>,
+}
+
+#[derive(Debug, thiserror::Error)]
+enum AnalyticsError {
+    #[error("invalid collection endpoint: {0}")]
+    CollectionEndpointParsingError(String),
+}
+
+fn parse_collection_endpoint(s: &str) -> Result<Url, AnalyticsError> {
```
Trantorian1: Consider using the existing [`parse_url`](https://github.com/madara-alliance/madara/blob/ab98e514ef2c470a5304ba3563de46835e1ec02a/crates/primitives/utils/src/parsers.rs#L24)

crates/node/src/service/rpc/metrics.rs
```diff
@@ -2,85 +2,111 @@ use std::time::Instant;
...
 
     pub(crate) fn ws_disconnect(&self, now: Instant) {
         let micros = now.elapsed().as_secs();
 
         if let Some(counter) = self.ws_sessions_closed.as_ref() {
-            counter.inc()
+            counter.add(1, &[]);
         }
-        self.ws_sessions_time.with_label_values(&["ws"]).observe(micros as _);
+        self.ws_sessions_time.record(micros as f64, &[]);
```
Trantorian1, line 99: Why not use a `Histogram<u64>` instead?

crates/node/src/service/rpc/metrics.rs
```diff
@@ -91,29 +117,33 @@ impl RpcMetrics {
...
-                req.method_name(),
-                // the label "is_error", so `success` should be regarded as false
-                // and vice-versa to be registered correctly.
-                if rp.is_success() { "false" } else { "true" },
-                if is_rate_limited { "true" } else { "false" },
-            ])
-            .inc();
+
+        self.calls_time.record(
+            micros as f64,
```
Trantorian1, line 132: Consider using `Histogram<u128>` instead.

crates/node/src/service/rpc/middleware.rs
```diff
@@ -140,12 +140,12 @@ where
...
             let response_time = now.elapsed();
 
-            log::info!(
+            tracing::info!(
                 target: "rpc_calls",
                 method = method,
                 status = status,
                 res_len = res_len,
-                response_time = response_time.as_micros();
+                response_time = response_time.as_micros() as u64,
```
Trantorian1: Why downcast to a smaller type?

Trantorian1 (changes requested):
Good changes, looking forwards to being able to use `tracing` in the codebase :)

Some things still need to be updated. I was being a bit picky with the function instrumentation and the way certain key arguments are `skip`, but then I saw how many they were and didn't have time to check them all :skull:. Imo it might still be better to update these over time but still, maybe un-skip certain arguments.

Some metrics are being cast to more imprecise types, is there a _really_ good reason for this? To me it seems like more of an artefact of the previous metrics. We really want to avoid a scenario were we get some very creative imaginary bugs because of integer overflows in the metrics -yikes!

Also, you should probably update the parsing logic for the log level and collection endpoints.

crates/client/devnet/src/lib.rs
heemankv: Resolving since this is already committed [here](https://github.com/madara-alliance/madara/pull/348/commits/7da86a8fc2fc83f7da37ba8b78344a66f7863657).

crates/client/analytics/src/lib.rs
heemankv: Valid, moreover I have reworked the `AnalyticsParams` to take in `analytics_log_level` as Level itself.
see [here](https://github.com/madara-alliance/madara/pull/348/commits/476d41186fb42842d63f5b98a62e1e4377f553e4).

crates/client/analytics/src/lib.rs
heemankv: Oh this is a very good catch,
```.with(EnvFilter::from_default_env());``` takes directly from `RUST_LOG` env,
whereas
```.with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))``` takes from self.log_level which is being defined from RUST_LOG in out codebase,

IMO it's better to keep `.with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))`

I don't see any added loss in choosing either of these implementations,
I'll go with ```.with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))``` 
there's no specific reason.
WDYT ?

Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/476d41186fb42842d63f5b98a62e1e4377f553e4).

crates/client/analytics/src/lib.rs
heemankv: Valid, this is much cleaner. 
Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/476d41186fb42842d63f5b98a62e1e4377f553e4).

crates/client/block_import/src/metrics.rs
heemankv: I tried that, but otel is the bottleneck here, it only supports `u64` for unsigned integers. See [this](https://docs.rs/opentelemetry/latest/opentelemetry/metrics/struct.Meter.html).

Can you think of anything that we can do to custom create these instruments?

crates/client/db/src/bonsai_db.rs
heemankv: I am not the author of this code, 
@cchudant what do you think ? 
Can we remove these comments ?

crates/client/block_import/src/metrics.rs
heemankv: IMO : 
1. `l2_avg_sync_time` should be a `f64` since `total_sync_time` is a `f64`.
2. `l2_block_number` should be used as an `u64` and is changed accordingly.
3. `transaction_count` should be used as an `u64` and is changed accordingly.

Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/analytics/src/lib.rs
heemankv: Valid, this is much cleaner.
Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/analytics/src/lib.rs
heemankv: Valid, this is much cleaner.
Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/class_db.rs
heemankv: Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2), wdyt ?

crates/client/db/src/class_db.rs
heemankv: Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/class_db.rs
heemankv: Committed [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/class_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/class_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/class_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/contract_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/contract_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/contract_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/db/src/db_metrics.rs
heemankv: Valid, applied.

crates/client/db/src/contract_db.rs
heemankv: Valid, added [here](https://github.com/madara-alliance/madara/pull/348/commits/731f7933090a42efa9052d1ab6f3ecfee2ab83c2).

crates/client/eth/src/l1_gas_price.rs
heemankv: As mentioned above, otel meter doesn't have support for u128.

crates/node/src/cli/analytics.rs
heemankv: Implemented [here](https://github.com/madara-alliance/madara/pull/348/commits/476d41186fb42842d63f5b98a62e1e4377f553e4).

crates/node/src/cli/analytics.rs
heemankv: Valid, simplified [here](https://github.com/madara-alliance/madara/pull/348/commits/fafe791ee0b8cb64bd64b36bf4496fcdb9c8cd14).

crates/node/src/service/rpc/middleware.rs
heemankv: Must be a mistake, corrected [here](https://github.com/madara-alliance/madara/pull/348/commits/fafe791ee0b8cb64bd64b36bf4496fcdb9c8cd14).

heemankv:
> Good changes, looking forwards to being able to use `tracing` in the codebase :)
> 
> Some things still need to be updated. I was being a bit picky with the function instrumentation and the way certain key arguments are `skip`, but then I saw how many they were and didn't have time to check them all 💀. Imo it might still be better to update these over time but still, maybe un-skip certain arguments.
> 
> Some metrics are being cast to more imprecise types, is there a _really_ good reason for this? To me it seems like more of an artefact of the previous metrics. We really want to avoid a scenario were we get some very creative imaginary bugs because of integer overflows in the metrics -yikes!
> 
> Also, you should probably update the parsing logic for the log level and collection endpoints.

Thanks for the detailed review @Trantorian1, I have replied to majorly all of your comments, we can start a re-review.
1. Valid, instrumentation can be made better, this PR focuses on introducing OTEL instrumentation to the code base, we can have multiple (module based maybe) PRs to properly add instrumentation and accordingly we can re-review `skip`.
2. Yes the metrics are indeed an artefact of the pre-existing Prometheus metrics, keeping your concerns of overflow in mind, I have implemented all the suggestions provided.
3. Yes the parsing logic is now much simplified, thanks to your reviews.

heemankv:
> Very cool, I've wanted to move to tracing entirely for some time I have one small thing to note: i want the logs to be kept as clean as possible if you start madara without just to test it or develop with it, which is why i've done all this https://github.com/madara-alliance/madara/blob/main/crates/node/src/util.rs#L40 which is a bit cursed (for example, I check the log target to show "HTTP" in purple instead of the blue "INFO", I hide the target in some cases (rpc internal server error), add emojis to the warn/errors and also the HTTP logs show the time elapsed in color etc etc. All of this is very terminal-specific and isn't made to work in production with OTEL stack - it's really for development and devnet uses.
> 
> I'd like to keep that behavior by doing a custom tracing terminal format, which would be disabled / hooked up differently for production where the full OTEL stack makes more sense, wdyt @heemankv

Interesting, 
The codebase currently checks if the `otel_endpoint` is available or not,
If it is available then it initialises OTEL, else it just initialises the tracing_subscriber.
We can tweak this logic to initialise our custom format if the OTEL is not being initialised. 

I currently use [`fmt().format`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/index.html) to configure and format the logs to the new version this PR introduces. I believe we can try [`format_event`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.FormatEvent.html) to format the logs as per our custom logic.

Is this close to what you are trying to solve for ?
WDYT @cchudant

crates/client/analytics/src/lib.rs
Trantorian1: Seems fine! My issue with using `EnvFilter::from_default_env` after `tracing_subscriber::filter::LevelFilter::from_level(self.log_level)` is that `EnvFilter::from_default_env` can also override the target I think. Though you are right that in our case it doesn't change much if out log level is derived from `RUST_LOG` anyway.

crates/client/block_import/src/metrics.rs
Trantorian1: If OTEL doesn't support u128 that's fine :+1:

Trantorian1 (commented):
Good, will merge as soon as we have some docs for Signoz :+1:

crates/client/analytics/src/lib.rs
heemankv: Valid, hence the change.

crates/node/src/service/rpc/metrics.rs
heemankv: Hey, seems like a miscommunication, we are indeed using Historgram for `ws_sessions_time`
```RUST
 /// Histogram over RPC websocket sessions.
 ws_sessions_time: Histogram<f64>,
````

crates/node/src/service/rpc/metrics.rs
heemankv: See [here](https://github.com/madara-alliance/madara/blob/b3b0f3aa84e884c8d06ba41aa4de91459259f49a/crates/node/src/service/rpc/metrics.rs#L28).

crates/node/src/service/rpc/metrics.rs
heemankv: See [here](https://github.com/madara-alliance/madara/blob/b3b0f3aa84e884c8d06ba41aa4de91459259f49a/crates/node/src/service/rpc/metrics.rs#L28).

Trantorian1 (approved):
Great!

cchudant:
@heemankv yes, as long as the devnet / development experience stays nice and clean it's okay

heemankv:
> @heemankv yes, as long as the devnet / development experience stays nice and clean it's okay

@cchudant I reworked the tracing logger to work similar to the logging code already written, I hope this satisfies all the requirements.

image

Mohiiit (approved):

antiyro (approved):

PR #349, opened 2024-10-18
Feat/strk price flag

Mohiiit:
STRK gas price cli param added which will let user set a fix value of strk gas

# Pull Request type

- Feature

Currently we only have it for the eth gas

Resolves: #NA

## What is the new behavior?

- now devs would be able to set a fix strk gas price as well

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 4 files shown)


antiyro (approved):

ocdbytes (approved):
lgtm

PR #350, opened 2024-10-19
refactor(rpc): re-worked rpc tower server and added proper websocket support

Trantorian1:
# Pull Request type

Current RPC tower server was inherited from Substrate and is a bit of a mess, especially in how method versioning and websockets are handled. This pr fixes this.

## What is the new behavior?

- Versioning is now a seperate rpc middleware layer. This means it will be applied to both http and ws rpc requests. 

> [!NOTE]
> Previously, the `VersionMiddleware` would only apply to http rpc requests and so any ws request to a starknet method would have to be of the format `starknet_{version}_{method}`

- Rpc middleware has been seperated into `RpcMiddlewareLayerMetrics` and `RpcMiddlewareLayerRateLimit` for ease of development. 

> [!WARNING]
> The current way by which I am communicating rate limiting between layers is questionable and any feedback is much appreciated.

- The `versioned_starknet_rpc` macro has been rewritten to be more hygenic and no longer overwrites macro attributes other than `#[method]`.

- Rpc call `rpc_methods` has had its output reworked to more closely resemble the actual rpc call api.

- Added some documentation on the rpc service explaining how this all ties together.

No

## Other information

Some debug logging has been left as part of an effort in instrumenting the rpc server with more runtime debug information. This mostly serves as a placeholder atm, as I would like to add more logging once #348 has been merged.

Please let me know what you think about this. I personally believe more observability would be quite nice (could enable output to a log file through a cmd line flag for example). Note that this does not even need to have any runtime overhead as we could disable debug logging under the production build flag (I'm not sure that would be a good idea though).


crates/node/src/service/rpc/server.rs
```diff
@@ -204,31 +177,53 @@ pub async fn start_server(
...
     }
 }
 
-pub(crate) fn build_rpc_api<M: Send + Sync + 'static>(mut rpc_api: RpcModule<M>) -> RpcModule<M> {
-    let mut available_methods = rpc_api.method_names().collect::<Vec<_>>();
+pub(crate) fn build_rpc_api<M: Send + Sync + 'static>(mut rpc_api: jsonrpsee::RpcModule<M>) -> jsonrpsee::RpcModule<M> {
+    let mut available_methods = rpc_api
+        .method_names()
+        .map(|name| {
+            let mut split = name.split("_");
```
jbcaron, line 190: rename v1, v2, v3
 major, minor, patch

Trantorian1: Done.

akhercha (approved):

crates/node/src/service/rpc/server.rs
```diff
@@ -77,102 +62,77 @@ pub async fn start_server(
...
+        .inactive_limit(Duration::from_secs(60))
+        .max_failures(3);
 
     let http_middleware = tower::ServiceBuilder::new()
-		.option_layer(host_filter)
-		// Proxy `GET /health` requests to internal `system_health` method.
-		// .layer(ProxyGetRequestLayer::new("/health", "system_health")?)
-        .layer(VersionMiddlewareLayer)
-		.layer(try_into_cors(cors.as_ref())?);
+        .option_layer(host_filtering(cors.is_some(), local_addr))
```
jbcaron, line 78: There is a potential issue with this logic: if cors is not defined but rpc-external is enabled, the host filtering will reject all external connections. This means that without CORS configured, valid external RPC requests may be unnecessarily blocked.

crates/node/src/service/rpc/server.rs
jbcaron: this has not been modified

crates/node/src/service/rpc/server.rs
Trantorian1: This has been changed so that if `rpc_external` is set, `CORS` will default to be permissive, and otherwise will allow only trafic from localhost. The idea is for it to match how permissive rpc calls are.

cchudant (approved):
cool!

jbcaron (approved):

PR #351, opened 2024-10-21
fix(network): added FGW url to chain config

antiyro:
Now both of this commands should work the same way:

`cargo run --release -- --full --preset mainnet`
`cargo run --release -- --full --network mainnet`

This ofc adds a bit of overhead but the idea is to keep it high level for the users and differenciate Starknet from app chains using Network.

We could in the future imagine that a sequencer will sync mainnet with` --network mainnet `

### DIFF
--- a/crates/node/src/cli/mod.rs
+++ b/crates/node/src/cli/mod.rs
@@ -24,7 +24,6 @@ use clap::ArgGroup;
 use mp_chain_config::ChainConfig;
 use std::path::PathBuf;
 use std::sync::Arc;
-use url::Url;
 
 /// Madara: High performance Starknet sequencer/full-node.
 #[derive(Clone, Debug, clap::Parser)]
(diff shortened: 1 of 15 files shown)


antiyro:
Note that when using devnet and test preset you must have a fgw url endpoint on port 8080, if not then you should use a custom config

Mohiiit (approved):

crates/node/src/cli/mod.rs
```diff
@@ -202,23 +201,6 @@ pub enum NetworkType {
...
-
-    pub fn gateway(&self) -> Url {
-        Url::parse(&format!("{}/gateway/", self.uri())).expect("Invalid uri")
-    }
-
-    pub fn feeder_gateway(&self) -> Url {
-        Url::parse(&format!("{}/feeder_gateway/", self.uri())).expect("Invalid uri")
-    }
-
     pub fn chain_id(&self) -> ChainId {
```
jbcaron, line 204: why keep this?

Trantorian1: This is still in use [here](https://github.com/madara-alliance/madara/blob/d7eb7bf00c412a52bb33dcb1d9c4a0d90f05866f/crates/node/src/main.rs#L197)

Trantorian1 (approved):

Mohiiit (approved):

PR #353, opened 2024-10-21
Feat/deprecated classes

Mohiiit:
during `finalize_execution_state` cairo 0 class hash were getting dropped, so refactored the logic in order to make sure that the state diff has the old contracts present at `deprecated_declared_classes`


jbcaron:
I believe the issue is more related to the structure of CommitmentStateDiff itself, wdyt?

Mohiiit:
I agree, we thought this would be the faster way to sort it out, but ideally it shouldn't get dropped

jbcaron:
I suggest you modify the function
csd_to_state_diff to directly transform StateMaps into StateDiff

jbcaron (approved):
👍
you can rename csd to state_map

crates/client/mempool/src/block_production.rs
```diff
@@ -59,14 +59,22 @@ pub enum Error {
 fn csd_to_state_diff(
     backend: &MadaraBackend,
     on_top_of: &Option<DbBlockId>,
-    csd: &CommitmentStateDiff,
+    state_map: StateMaps,
 ) -> Result<StateDiff, Error> {
+    // `StateMaps` to `CommitmentStateDiff` conversion drops declared_contracts. This causes
+    // Cairo 0 contracts to be removed from the state diff. So we keep track of them separately.
+    let mut deprecated_declared_contracts = state_map.declared_contracts.clone();
```
Trantorian1: This should not be cloned. Update the conversion `From<StateMap> for CommitmentStateDiff` to include `deprecated_classes` as part of `CommitmentStateDiff`. This can be done zero-copy by mutating `state_map.declared_contracts` in place as the conversion takes ownership of the `StateMap`.

crates/client/mempool/src/block_production.rs
```diff
@@ -59,14 +59,22 @@ pub enum Error {
...
         address_to_class_hash,
         address_to_nonce,
         storage_updates,
         class_hash_to_compiled_class_hash,
-    } = csd;
+    } = CommitmentStateDiff::from(state_map);
+
+    // Cairo 1 declared contracts are present in both the declared_contracts and the compiled_class_hashes maps.
+    // So we remove them from `deprecated_declared_contracts` as we only use them for Cairo 0 contracts
+    deprecated_declared_contracts.retain(|key, _| !class_hash_to_compiled_class_hash.contains_key(key));
```
Trantorian1: Move this logic to `CommitmentStateDiff` and update it to be zero-copy.

Trantorian1 (changes requested):
This code needs to be part of the conversion `From<StateMap> for CommitmentStateDif`.

crates/client/mempool/src/block_production.rs
Trantorian1: My bad, since `CommitmentStateDiff` is part of blockifier this should be implemented via an extension trait instead.

jbcaron (approved):

Trantorian1 (approved):

PR #356, opened 2024-10-22
feat(evm): Support EVM deployment with Madara

antiyro:
This pull request adds the necessary environment to easily deploy an EVM sequencer (devnet) or a specific full node with Kakarot v0.9.2


Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN cargo build --release
 FROM debian:bookworm
 # Install runtime dependencies
 RUN apt-get -y update && \
-    apt-get install -y openssl ca-certificates &&\
+    apt-get install -y openssl ca-certificates tini &&\
```
Trantorian1, line 29: This is not necesasry as we do not use tine currently. Either remove it or consider updating the Dockerfile as per [this guide](https://github.com/krallin/tini?tab=readme-ov-file#using-tini)

evm/docker-compose.yml
```diff
@@ -0,0 +1,145 @@
+version: "3.8"
```
Trantorian1: Any way to test locally that this works?

Trantorian1 (changes requested):

evm/docker-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+      context: ../
+      dockerfile: Dockerfile
+    restart: unless-stopped
+    ports:
+      - 9944:9944
+    networks:
+      - internal_prod
+    volumes:
+      - madara_files_prod:/var/lib/madara
+    entrypoint: ["tini", "--"]
```
antiyro, line 13: here @Trantorian1

Dockerfile
antiyro: the evm compose uses it

jbcaron: it would be good to set it as the entrypoint

evm/docker-compose.yml
jbcaron: this definition is no longer necessary

antiyro:
I will implement configuration file to improve user experience on EVM configuration

antiyro:
Force merging this since it doesnt affect the core Madara

PR #358, opened 2024-10-24
fix(namespace): versioning now works for methods without `starknet` namespace

Trantorian1:
# Pull Request type

- Bugfix

Following #350, change were made to the rpc method versioning middleware that did not take into account methods with a unique namespace such as `madara_addDeclareV0Transaction`.

## What is the new behavior?

Methods with arbitrary namespaces can be versioned.

No

### DIFF
--- a/crates/node/src/service/rpc/middleware.rs
+++ b/crates/node/src/service/rpc/middleware.rs
@@ -211,7 +211,7 @@ where
                 return inner.call(req).await;
             }
 
-            let Ok(version) = RpcVersion::from_request_path(&path) else {
+            let Ok(version) = RpcVersion::from_request_path(&path).map(|v| v.name()) else {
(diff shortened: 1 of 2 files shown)


Mohiiit (approved):

crates/node/src/service/rpc/middleware.rs
```diff
@@ -233,7 +233,7 @@ where
                 );
             };
 
-            let method_new = format!("starknet_{}_{}", version.name(), method_without_namespace);
+            let method_new = format!("{namespace}_{version}_{method}");
```
Mohiiit, line 236:
```suggestion
            let method_new = format!("{}_{}_{}", namespace, version, method);
```

antiyro (approved):

jbcaron (approved):

Issue #340, opened 2024-10-15
feat(starknet_unsubscribe)


Trantorian1:
### Is there an existing issue?

- [X] I have searched the existing issues

### Motivation

See #335 for more information about implementing rpc v0.8.0-rc0 in Madara.

### Request

Implement [`starknet_unsubscribe`](https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_ws_api.json#L254) from rpc v0.8.0-rc0.

> [!NOTE]
> From the specs: _Close a previously opened ws stream, with the corresponding subscription id_.

### Are you willing to help with this request?

Yes!

PR #359, opened 2024-10-24
feat(v0.8.0-rc0): starknet_subscribeNewHeads

### LINKED ISSUE #336: feat(starknet_subscribeNewHeads)

### Motivation

### Request

Implement [`starknet_subscribeNewHeads`](https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_ws_api.json#L10) from rpc v0.8.0-rc0.

> [!NOTE]
> From the specs: _Creates a WebSocket stream which will fire events for new block headers_.

Yes!

Trantorian1:
# Pull Request type

- Feature

Resolves: #336 #340

## What is the new behavior?

- Added support for `starknet_subscribeNewHeads` rpc websocket method
- Updated `MadaraBackend` to allow streaming new information to websocket connections directly

No.


crates/client/db/src/block_db.rs
```diff
@@ -243,12 +260,17 @@ impl MadaraBackend {
...
-        tx.put_cf(&block_hash_to_block_n, block_hash_encoded, &block_n_encoded);
         tx.put_cf(&block_n_to_block, &block_n_encoded, bincode::serialize(&block.info)?);
+        tx.put_cf(&block_hash_to_block_n, block_hash_encoded, &block_n_encoded);
         tx.put_cf(&block_n_to_block_inner, &block_n_encoded, bincode::serialize(&block.inner)?);
         tx.put_cf(&block_n_to_state_diff, &block_n_encoded, bincode::serialize(state_diff)?);
         tx.put_cf(&meta, ROW_SYNC_TIP, block_n_encoded);
 
+        // susbcribers
+        if self.sender_block_info.receiver_count() > 0 {
+            let _ = self.sender_block_info.send(block.info.clone());
```
jbcaron: maybe log errors

crates/client/db/src/lib.rs
```diff
@@ -395,6 +396,7 @@ impl MadaraBackend {
             last_flush_time: Default::default(),
             chain_config,
             db_metrics: DbMetrics::register(&MetricsRegistry::dummy()).unwrap(),
+            sender_block_info: tokio::sync::broadcast::channel(1024).0,
```
jbcaron: I think we can reduce the buffer to only a few blocks

crates/client/rpc/src/errors.rs
```diff
@@ -208,3 +210,56 @@ impl From<StarknetApiError> for StarknetRpcApiError {
...
+    fn code(&self) -> i32 {
+        match self {
+            Self::TooManyBlocksBack => 68,
+            Self::Internal { context: _ } => jsonrpsee::types::error::INTERNAL_ERROR_CODE,
+        }
+    }
+    fn message(&self) -> &str {
+        match self {
+            Self::TooManyBlocksBack => "Cannot go back more than 1024 blocks",
+            Self::Internal { context } => context,
```
jbcaron: be careful, we do not want to give information to the client when there is an internal problem, only log the problem in madara

crates/client/db/src/block_db.rs
Trantorian1: Fixed.

crates/client/db/src/lib.rs
Trantorian1: Fixed.

crates/client/rpc/src/errors.rs
Trantorian1: Fixed.

crates/client/db/src/block_db.rs
cchudant, line 1: Why these?
```rs
    pub fn get_block_info_from_block_latest(&self) -> Result<Option<MadaraBlockInfo>> {
        let Ok(Some(block_n)) = &self.get_latest_block_n() else {
            return Ok(None);
        };

        let col = self.db.get_column(Column::BlockNToBlockInfo);
        let res = self.db.get_cf(&col, bincode::serialize(block_n)?)?;
        let Some(res) = res else { return Ok(None) };
        let block = bincode::deserialize(&res)?;
        Ok(Some(block))
    }

    pub fn get_block_info_from_block_hash(&self, block_hash: Felt) -> Result<Option<MadaraBlockInfo>> {
        let Ok(Some(block_n)) = self.block_hash_to_block_n(&block_hash) else { return Ok(None) };
        self.get_block_info_from_block_n(block_n)
    }
```
same for `pub fn block_hash_to_block_n` et `pub fn get_block_info_from_block_n` where you added the `pub` keyword; there's a reason I have left them private
The problem is that you can make so many functions like these, for every permutation of what you want and what you have to address - it's just endless!

Instead, I recommend you do

`get_block_n(BlockId::Hash(your hash))` et `get_block_info(BlockId::Number(your number)` for the two where you added the `pub` keyword, and for the 2 new ones `get_block_info(BlockId::Tag(BlockTag::Latest))` and `get_block_info(BlockId::Hash(your hash))`

that's why there's this whole DbBlockIdResolvable trait and DbBlockId thinggy - it's so that you can get whatever you want from any kind of resolvable id in the fewest lookups possible.

Also, if you want to reuse a block id to address multiple things in the db, you can factor out the resolve block_id => db_block_id using `resolve_block_id` - it returns a DbBlockId, and using it afterwards, saves a lookup by doing the BlockId::Hash -> DbBlockId::Number lookup only once for example.

crates/client/db/src/block_db.rs
```diff
@@ -243,12 +260,19 @@ impl MadaraBackend {
...
 
-        tx.put_cf(&block_hash_to_block_n, block_hash_encoded, &block_n_encoded);
         tx.put_cf(&block_n_to_block, &block_n_encoded, bincode::serialize(&block.info)?);
+        tx.put_cf(&block_hash_to_block_n, block_hash_encoded, &block_n_encoded);
         tx.put_cf(&block_n_to_block_inner, &block_n_encoded, bincode::serialize(&block.inner)?);
         tx.put_cf(&block_n_to_state_diff, &block_n_encoded, bincode::serialize(state_diff)?);
         tx.put_cf(&meta, ROW_SYNC_TIP, block_n_encoded);
 
+        // susbcribers
+        if self.sender_block_info.receiver_count() > 0 {
```
cchudant, line 267: we should separate db from backend at some point in the future, not necessarily here though i think that's fine for now?

crates/client/rpc/src/versions/v0_8_0/api.rs
```diff
@@ -0,0 +1,12 @@
...
+use starknet_types_core::felt::Felt;
+
+use crate::errors::WsResult;
+
+pub(crate) type NewHead = starknet_types_rpc::BlockHeader<Felt>;
+
+#[versioned_rpc("V0_8_0", "starknet")]
+pub trait StarknetWsRpcApi {
+    #[subscription(name = "subscribeNewHeads", unsubscribe = "unsubscribe", item = NewHead, param_kind = map)]
+    async fn subscribe_new_heads(&self, block_id: starknet_core::types::BlockId) -> WsResult;
```
cchudant: wait, if you use /v0.8.0, you won't get all the older rpc methods then?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+    async fn subscribe_new_heads(
+        &self,
+        pending: jsonrpsee::PendingSubscriptionSink,
+        block_id: starknet_core::types::BlockId,
+    ) -> WsResult {
+        let Ok(sink) = pending.accept().await else {
+            return WsResult::Err(StarknetWsApiError::internal(Cow::from("Failed to establish websocket connection")));
+        };
+
+        let mut block_n = match block_id {
```
cchudant, line 18: what? no please use backend.get_block_n(block_id) it directly accepts starknet core types BlockId and returns the block n

why do you even need the block info, you only need the block n here?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+            }
+        };
+
+        let mut rx = self.backend.subscribe_block_info();
+        for n in block_n.. {
+            if sink.is_closed() {
+                return WsResult::Ok;
+            }
+
+            let block_info = match self.backend.get_block_info_from_block_n(n) {
```
cchudant: i'm okay with that but I thought you made a rocksdb iterator-based thinggy for that?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+
+            let res = send_block_header(&sink, block_info, block_n).await;
+            if matches!(res, WsResult::Err(_)) {
+                return res;
+            }
+
+            block_n += 1;
+        }
+
+        loop {
```
cchudant, line 86: i think i know why you're doing two loops here, and i don't like it
also, you need to select! here too, otherwise if a client sends a block n far into the future, and closes the connection your task will hang and that's no good

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+            let Ok(block_info) = rx.recv().await else {
+                return WsResult::Err(StarknetWsApiError::internal(Cow::from("Failed to retrieve block info")));
+            };
+
+            if block_info.header.block_number == block_n {
+                let res = send_block_header(&sink, block_info, block_n).await;
+                if matches!(res, WsResult::Err(_)) {
```
cchudant: isnt that the ? operator?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+                }
+
+                break;
+            }
+        }
+
+        loop {
+            tokio::select! {
+                block_info = rx.recv() => {
+                    let Ok(block_info) = block_info else {
```
cchudant: map_err

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+use crate::{
+    errors::{StarknetWsApiError, WsResult},
+    versions::v0_8_0::StarknetWsRpcApiV0_8_0Server,
+};
+
+use super::BLOCK_PAST_LIMIT;
+
+#[jsonrpsee::core::async_trait]
+impl StarknetWsRpcApiV0_8_0Server for crate::Starknet {
+    async fn subscribe_new_heads(
```
cchudant, line 10: everywhere here in this func you're dropping the Err values instead of displaying them as an internal server error

id advise you to either
```rs
.map_err(|e| {
  display_internal_server_error(format!("some context blah blah: {err:#}");
  StarknetWsApiError::InternalServerError
})
```
for Results, and something similar for unwrapping Options where None means an error.

A better way to do this is to do like we already do in the RPCs: `or_internal_server_error` / `or_else_internal_server_error` for unwrapping results, and `ok_or_internal_server_error` / `ok_or_else_internal_server_error` for unwrapping options.
You can search the codebase to see how they work, but that's the best way to handle internal servers errors in RPCs IMO.

Note: we don't want to show the reason of an internal server error to the client! as a general rule, it's not a good idea - only to the console

cchudant:
```rs
let Ok(sink) = pending.accept().await else {
  return WsResult::Err(StarknetWsApiError::internal(Cow::from("Failed to establish websocket connection")));
```
would look like
```rs
let sink = pending.accept().await
  .or_internal_server_error("Failed to establish websocket connection")?;
```

and 
```rs
let Ok(Some(block_info)) = self.backend.get_block_info_from_block_latest() else {
    return WsResult::Err(StarknetWsApiError::internal(Cow::from(format!(
        "Failed to retrieve block info for block {block_n}",
    ))));
};
```
would look like
```rs
let block_info = self.backend.get_block_info(xxx)
  .or_else_internal_server_error(|| format!("Failed to retrieve block info for block {block_n}"))?
  .ok_or_internal_server_error("There are no blocks in the database")?;
```
which, now that i realise, is not an internal server error at all - you should return a clean error if there is no latest block, which can only mean there are no blocks in db yet. For the standard rpc api there's a NoBlocks error for that.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+                        return WsResult::Err(StarknetWsApiError::internal(Cow::from("Failed to retrieve block info")));
+                    };
+
+                    let res = send_block_header(&sink, block_info, block_n).await;
+                    if matches!(res, WsResult::Err(_)) {
+                        return res;
+                    }
+                },
+                _ = sink.closed() => {
+                    return WsResult::Ok
```
cchudant: wait WsResult is not Result? why?

cchudant (changes requested):

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: That turned out not to be possible due to the block info being out-of-order (relative to the block number) in that rocksdb collumn

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: We can't use the `?` operator inside this function as `WsResult` is not a proper error type. This is due to `jsonrpsee` forcing specific trait bounds for custom error messages. The `Try` and `Fromresidual` traits in nightly allow for custom implementations of `?` but we are in stable.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: `jsonrpsee` forces you to provide your own implemtation of `IntoSubscriptionCloseResponse` to return custom errors and error codes, which is required in starknet specs :/

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,398 @@
...
+                    ))))
+                }
+            };
+
+            let res = send_block_header(&sink, block_info, block_n).await;
+            if matches!(res, WsResult::Err(_)) {
+                return res;
+            }
+
+            block_n += 1;
```
cchudant: comment why this addition cannot overflow even though it's a user controlled input?

crates/client/rpc/src/versions/v0_8_0/api.rs
Trantorian1: Yup, forgot to fix this in the macro sorry

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: This is already handled [here](https://github.com/madara-alliance/madara/blob/0b0cf309e31b7cde6f533b9d5c693bd25f02e106/crates/client/rpc/src/errors.rs#L237) but I think I'll more this code to the init of `InternalError` instead of the `message` function.

crates/client/db/src/block_db.rs
Trantorian1: Understood, updating the code accordingly

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: No this is a very good point, this would actually be possible with the way I've currently set up the function.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+        let sink = pending.accept().await.or_internal_server_error("Failed to establish websocket connection")?;
+
+        let mut block_n = match block_id {
+            starknet_core::types::BlockId::Number(block_n) => {
+                let err = || format!("Failed to retrieve block info for block {block_n}");
+                let block_latest = self
+                    .backend
+                    .get_block_n(&mp_block::BlockId::Tag(mp_block::BlockTag::Latest))
+                    .or_else_internal_server_error(err)?
+                    .ok_or_else_internal_server_error(err)?;
```
cchudant: None here happens when there are no blocks in DB, which should not internal server error, it should return a soft error kinda like StarknetApiError::NoBlocks in the standard rpc.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
cchudant: this isnt resolved right?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+
+                block_n
+            }
+            starknet_core::types::BlockId::Hash(block_hash) => {
+                let err = || format!("Failed to retrieve block info at hash {block_hash:#x}");
+                let block_latest = self
+                    .backend
+                    .get_block_n(&mp_block::BlockId::Tag(mp_block::BlockTag::Latest))
+                    .or_else_internal_server_error(err)?
+                    .ok_or_else_internal_server_error(err)?;
```
cchudant: that's a block hash not found error, should not internal server error

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
cchudant: all of this match is just
```rs
let mut block_n = starknet.get_block_n(block_id)
  .ok_or_internal_server_error("Resolving block id")?
  .ok_or(StarknetWsApiError::BlockNotFound)?;
```

cchudant (changes requested):

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: The reason I went for this is that it requires one less db read in the case where `BlockId::Tag(BlockTag::Latest)` is passed. Furthermore, we will have to check for `BlockId::Tag(BlockTag::Pending)` anyway, and there should basically be no overhead between that and a match, so we might as well go this route.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: Fixed

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: Fixed

crates/client/rpc/src/versions/v0_8_0/api.rs
Trantorian1: Fixed

crates/client/db/src/block_db.rs
Trantorian1: For now I think its fine, we want the block info to be sent to websocket receivers as soon as they are stored anyway so it feels like this makes sense in the current setup

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+            }
+        };
+
+        let mut rx = self.backend.subscribe_block_info();
+        for n in block_n.. {
+            if sink.is_closed() {
+                return Ok(());
+            }
+
+            let block_info = match self.backend.get_block_info(&mp_block::BlockId::Number(n)) {
```
cchudant, line 69:
```rs
let Some(block_info) = self.backend.get_block_info(&mp_block::BlockId::Number(n))
  .or_else_internal_server_error(|| format!("Failed to retrieve block info for block {n}"))?
else {
    break
};
let block_info = block_info.as_nonpendin_owned()
  .ok_or_else_internal_server_error(|| format!("Got a pending block for block {n} when querying a closed one"))?;

File: crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs [hunk 17]
```diff
@@ -0,0 +1,370 @@
...
+                }
+                Ok(None) => break,
+                Err(e) => {
+                    let err = format!("Failed to retrieve block info for block {n}: {e}");
+                    return Err(StarknetWsApiError::internal_server_error(err).into());
+                }
+            };
+
+            send_block_header(&sink, block_info, block_n).await?;
+            block_n = block_n.saturating_add(1);
```
cchudant, line 82: i think you meant `n` here and not block_n?

cchudant (approved):
cool!

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: Idk, personally I find a match to be more readable in this case  ¯\_(ツ)_/¯

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: No, this should indeed be `block_n` so that we are left with the height of the latest block when exiting the loop

crates/client/rpc/src/errors.rs
jbcaron, line 1: 👍

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+    versions::v0_8_0::StarknetWsRpcApiV0_8_0Server,
+};
+
+use super::BLOCK_PAST_LIMIT;
+
+#[jsonrpsee::core::async_trait]
+impl StarknetWsRpcApiV0_8_0Server for crate::Starknet {
+    async fn subscribe_new_heads(
+        &self,
+        pending: jsonrpsee::PendingSubscriptionSink,
```
jbcaron: maybe rename it more explicitly

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+                    .backend
+                    .get_block_n(&mp_block::BlockId::Tag(mp_block::BlockTag::Latest))
+                    .or_else_internal_server_error(err)?
+                    .ok_or(StarknetWsApiError::BlockNotFound)?;
+
+                let block_n = self
+                    .backend
+                    .get_block_n(&block_id)
+                    .or_else_internal_server_error(err)?
+                    .ok_or_else_internal_server_error(err)?;
```
jbcaron: wouldn't that be a block_not_found error?

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+                    return Err(StarknetWsApiError::TooManyBlocksBack.into());
+                }
+
+                block_n
+            }
+            starknet_core::types::BlockId::Tag(starknet_core::types::BlockTag::Latest) => self
+                .backend
+                .get_latest_block_n()
+                .or_internal_server_error("Failed to retrieve block info for latest block")?
+                .ok_or_internal_server_error("Failed to retrieve block info for latest block")?,
```
jbcaron: same

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+
+        // We need to check the block number at each iteration as the first
+        // time this is exectued we might already have received some blocks
+        // from the backend which we manually fecthed from db
+        loop {
+            tokio::select! {
+                block_info = rx.recv() => {
+                    let block_info = block_info.or_internal_server_error("Failed to retrieve block info")?;
+                    if block_info.header.block_number == block_n {
```
jbcaron, line 90: can be added a check also to close the connection if a block is missing

crates/node/src/service/rpc/server.rs
```diff
@@ -191,14 +191,22 @@ pub(crate) fn build_rpc_api<M: Send + Sync + 'static>(mut rpc_api: jsonrpsee::Rp
...
+                format!("rpc/{namespace}_{method}")
+            } else {
+                // versioned method
+                let namespace = split[0];
+                let major = split[1];
+                let minor = split[2];
+                let patch = split[3];
+                let method = split[4];
+                format!("rpc/{major}_{minor}_{patch}/{namespace}_{method}")
+            }
```
jbcaron, line 209: better 😁

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: Done!

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: Fixed.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: This whould be a `StarknetWsApiError::NoBlocks`. Fixed.

crates/client/rpc/src/versions/v0_8_0/methods/ws/lib.rs
Trantorian1: This is already the case as we will return an internal server error if we cannot retrieve the latest block from the backend block info broadcast channel, which will automatically close the connection.

> [!NOTE]
> The actual connection lifecycle is handled by `jsonrpsee` and not us: the connection is simply dropped whenever we exit the method.

crates/node/src/service/rpc/server.rs
Trantorian1: :saluting_face:

jbcaron (approved):

PR #360, opened 2024-10-24
fix(compile): wrong struct field being used in state map conversion

Trantorian1:
# Pull Request type

- Bugfix

Small mistake in #353 

No


Mohiiit (approved):
Thanks for this

jbcaron (approved):

PR #365, opened 2024-10-25
fix(compilation): do not compile devnet contract artifacts when doing `cargo build`

cchudant:
# Pull Request type

- Bugfix/feat?

## What is the new behavior?

TLDR:
- devnet contract addresses are now 100% stable
- starkli should now work with the devnet predeployed accounts
- the Dockerfile does not require `scarb` anymore, now only running the tests require it
- `cargo build` does not compile e2e-tests anymore

In more details:
- Remove `cairo0` folder, and ERC20/AccountUpgradable compilation from the cairo 1 scarb project. Instead, put these contracts in a new cairo-artifacts folder. Compilation of these artifact is done ahead of time, and a README is provided to check the integrity. The Account contracts are now compiled with a version that matches starkli's allowed class hash list.
- The remaining cairo contracts that are compiled are only used for testing. Remove the need to compile these artifacts when doing `cargo build` using a new crate that is in dev-dependencies of the crates that need the test contracts. This effectively means that we don't need scarb 
- Remove the e2e tests and this new cairo-test-contracts package from the default-members of the workspace, meaning the command `cargo build` will not build them. This is also the case for `cargo test`, which is a problem; so you now need to do `cargo test --workspace` now to run the e2e tests.
- By removing the compilation of the contracts, the devnet contract addresses are now stable.

YES, the predeployed contract addresses are all modified once again. For the last time! hopefully.
Their new address is also now documented in devnet/README.md.


Trantorian1 (approved):
Great to have gotten rid of `scarb` in the build process, this will make docker builds so much easier :fire:

antiyro (approved):

PR #368, opened 2024-10-28
feat(gas) : fetch strk price from oracle

### LINKED ISSUE #352: feat(): fetch strk price from pragma oracle

### Motivation

Currently we set a fixed strk price by CLI which is inaccurate.

### Request

Retrieve strk price from pragma oracle to have a dynamic price

### Solution

_No response_

Yes!

azurwastaken:
Pull Request type

Feature
What is the current behavior?

Resolves: https://github.com/madara-alliance/madara/issues/352


cchudant (approved):
Just wondering, how would you run madara in production without oracle? Maybe, for chains which don't want an L2 token, only eth? if we set its price to zero, people would be able to send transactions without fees
That would be for later PR though

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,132 @@
...
+            .send()
+            .await
+            .expect("failed to retrieve price from pragma oracle");
+
+        let oracle_api_response = response.json::<PragmaApiResponse>().await.expect("failed to parse api response");
+        let eth_strk_price = u128::from_str_radix(oracle_api_response.price.trim_start_matches("0x"), 16)?;
+
+        Ok((eth_strk_price, oracle_api_response.decimals))
+    }
+}
```
jbcaron: Does the node crash in case of unavailable service?

azurwastaken:
> Just wondering, how would you run madara in production without oracle? Maybe, for chains which don't want an L2 token, only eth? if we set its price to zero, people would be able to send transactions without fees That would be for later PR though

Yes, oracle are just here to do live conversion between ETH and L2 token so, in my opinion, the solution would be to disable the ability to pay gas with STRK/L2token.

crates/primitives/oracle/src/pragma.rs
azurwastaken: oh yes probably, my bad, i changed expect to context to propagate the error to crates/client/eth/src/l1_gas_price.rs:17

EvolveArt:
> Just wondering, how would you run madara in production without oracle? Maybe, for chains which don't want an L2 token, only eth? if we set its price to zero, people would be able to send transactions without fees That would be for later PR though

That would fck up all the rpc endpoints that have the strk gas price tho so not a good idea.

antiyro:
> > Just wondering, how would you run madara in production without oracle? Maybe, for chains which don't want an L2 token, only eth? if we set its price to zero, people would be able to send transactions without fees That would be for later PR though
> 
> That would fck up all the rpc endpoints that have the strk gas price tho so not a good idea.

we could default it anyways

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -67,6 +68,21 @@ async fn update_gas_price(eth_client: &EthereumClient, l1_gas_provider: GasPrice
     l1_gas_provider.update_eth_l1_gas_price(*eth_gas_price);
     l1_gas_provider.update_eth_l1_data_gas_price(avg_blob_base_fee);
 
+    // fetch eth/strk price and update
+    if let Some(oracle_provider) = &l1_gas_provider.oracle_provider {
+        let (eth_strk_price, decimals) =
+            oracle_provider.fetch_eth_strk_price().await.context("failed to retrieve ETH/STRK price")?;
+        let strk_gas_price = (BigDecimal::new((*eth_gas_price).into(), decimals.into())
+            / BigDecimal::new(eth_strk_price.into(), decimals.into()))
```
antiyro, line 77: if `eth_strk_price` price is 0 here there is a possible 0 division?

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub async fn fetch_eth_strk_price(&self) -> anyhow::Result<(u128, u32)> {
+        let response = reqwest::Client::new()
+            .get(self.get_fetch_url(String::from("eth"), String::from("strk")))
+            .header("x-api-key", self.api_key.clone())
+            .send()
+            .await
+            .context("failed to retrieve price from pragma oracle")?;
+
+        let oracle_api_response = response.json::<PragmaApiResponse>().await.context("failed to parse api response")?;
+        let eth_strk_price = u128::from_str_radix(oracle_api_response.price.trim_start_matches("0x"), 16)?;
```
antiyro: if the price is not fetched in the expected format what will happen?

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -67,6 +68,21 @@ async fn update_gas_price(eth_client: &EthereumClient, l1_gas_provider: GasPrice
...
+        .as_bigint_and_exponent();
+        let strk_data_gas_price = (BigDecimal::new(avg_blob_base_fee.into(), decimals.into())
+            / BigDecimal::new(eth_strk_price.into(), decimals.into()))
+        .as_bigint_and_exponent();
+
+        l1_gas_provider.update_strk_l1_gas_price(strk_gas_price.0.to_str_radix(10).parse::<u128>()?);
```
antiyro: I would expect this not to overflow but would be great to ensure clean code

antiyro (changes requested):

crates/node/src/main.rs
```diff
@@ -108,6 +109,18 @@ async fn main() -> anyhow::Result<()> {
...
+    if let Some(ref oracle_url) = run_cmd.l1_sync_params.oracle_url {
+        if let Some(ref oracle_api_key) = run_cmd.l1_sync_params.oracle_api_key {
+            let oracle = Oracle::new("Pragma", oracle_url.to_string(), oracle_api_key.clone())?;
+            l1_gas_setter.set_oracle_provider(oracle);
+        }
+    }
+
+    if l1_gas_setter.is_oracle_needed() && l1_gas_setter.oracle_provider.is_none() {
+        log::error!("STRK gas is not fixed and oracle is not provided");
+        panic!();
```
antiyro: I would personally handle the case where we dont want the strk price by defaulting it

crates/client/eth/src/l1_gas_price.rs
azurwastaken: yeah you are right, added a check

crates/primitives/oracle/src/pragma.rs
azurwastaken: it fail to parse and propagate an error without panicking

crates/node/src/main.rs
cchudant: also, could you use anyhow::bail! instead of panic and put the error message inside the bail?

crates/primitives/oracle/src/pragma.rs
cchudant: could you add .context() to the error then?

crates/client/eth/src/l1_gas_price.rs
cchudant: same thing, could you add .context() to the error here too?

cchudant:
@EvolveArt 
> That would fck up all the rpc endpoints that have the strk gas price tho so not a good idea.

i don't think i understand, rpcs never use gas prices providers for anything, it's only block prod?
for rpcs we always use the gas prices in the block headers, it's only block prod that needs gas prices

EvolveArt:
> @EvolveArt
> 
> > That would fck up all the rpc endpoints that have the strk gas price tho so not a good idea.
> 
> i don't think i understand, rpcs never use gas prices providers for anything, it's only block prod? for rpcs we always use the gas prices in the block headers, it's only block prod that needs gas prices

Yes and the block headers value is updated with the strk price

crates/client/eth/src/l1_gas_price.rs
Trantorian1: This still seems to be an issue!

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -67,6 +68,21 @@ async fn update_gas_price(eth_client: &EthereumClient, l1_gas_provider: GasPrice
...
 
+    // fetch eth/strk price and update
+    if let Some(oracle_provider) = &l1_gas_provider.oracle_provider {
+        let (eth_strk_price, decimals) =
+            oracle_provider.fetch_eth_strk_price().await.context("failed to retrieve ETH/STRK price")?;
+        let strk_gas_price = (BigDecimal::new((*eth_gas_price).into(), decimals.into())
+            / BigDecimal::new(eth_strk_price.into(), decimals.into()))
+        .as_bigint_and_exponent();
+        let strk_data_gas_price = (BigDecimal::new(avg_blob_base_fee.into(), decimals.into())
+            / BigDecimal::new(eth_strk_price.into(), decimals.into()))
```
Trantorian1, line 80: Same issue here

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Supported Aggregation Methods
+pub enum AggregationMethod {
+    #[serde(rename = "median")]
+    Median,
+    #[serde(rename = "mean")]
+    Mean,
+    #[serde(rename = "twap")]
+    #[default]
+    Twap,
+}
```
Trantorian1: Maybe some short comments on what each one of these do? `Median` and `Mean` seem pretty self-explanatory but `Twap` a little less so.

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,135 @@
...
+pub enum Interval {
+    #[serde(rename = "1min")]
+    OneMinute,
+    #[serde(rename = "15min")]
+    FifteenMinutes,
+    #[serde(rename = "1h")]
+    OneHour,
+    #[serde(rename = "2h")]
+    #[default]
+    TwoHours,
```
Trantorian1, line 111: Would it be sensible to add an extra `Custom` aggregation interval?

Trantorian1 (changes requested):

crates/client/eth/src/l1_gas_price.rs
azurwastaken: added the check here : crates/primitives/oracle/src/pragma.rs:59

when the api return a 0 it say that we failed to fetch the price for this specific reason

crates/primitives/oracle/src/pragma.rs
azurwastaken: those aggregation intervals are those provided by our api and its not possible to query custom one

crates/primitives/oracle/src/pragma.rs
azurwastaken: I can if its mandatory but I don't know if its relevant as twap is also a pretty common aggregation method in the financial world.

jbcaron: `#[serde(rename_all = "snake_case")]` You can simplify the renaming by using this attribute at the enum level, instead of applying `#[serde(rename)]` to each variant individually. This ensures consistency and reduces boilerplate.

crates/client/eth/src/l1_gas_price.rs
Trantorian1: Fair, but I still don't very like the fact that this function can be called with data that results in a division by 0. Imo a check should still be added here as it is still possible to introduce hard to detect crashes if this function were to be misused in the future.

crates/primitives/oracle/src/pragma.rs
jbcaron: imo, `Twap` is self-explanatory. Just add a doc comment explaining the acronym

crates/client/eth/src/l1_gas_price.rs
azurwastaken: I'm not sure to get it. It can't have a division by 0 because the utils versions that retrieve the price return an error   if its zero.
if we want a to add other oracle in the future, the utils function that use it must check its own return data.

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+use anyhow::bail;
+use serde::{Deserialize, Serialize};
+
+mod pragma;
+
+use pragma::*;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "oracle_name", content = "config")]
+pub enum Oracle {
```
antiyro: could you document this enum

antiyro (changes requested):

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl Oracle {
+    pub fn new(oracle_name: &str, url: String, key: String) -> anyhow::Result<Self> {
+        match oracle_name {
+            "Pragma" => Ok(Oracle::Pragma(PragmaOracle::new(url, key))),
+            _ => bail!("Unknown Oracle name"),
+        }
+    }
+
+    pub fn set_base_url(&mut self, url: String) {
```
jbcaron, line 22:
```suggestion
    pub fn set_base_url(&mut self, url: String) -> &Self {
```
allows the chaining of constructors by returning a reference to the current instance

crates/client/mempool/src/l1.rs
```diff
@@ -22,9 +24,19 @@ impl GasPriceProvider {
...
     }
 
+    pub fn is_oracle_needed(&self) -> bool {
+        self.strk_gas_price_sync_enabled.load(Ordering::Relaxed)
+            || self.strk_data_gas_price_sync_enabled.load(Ordering::Relaxed)
+    }
+
+    pub fn set_oracle_provider(&mut self, oracle_provider: Oracle) {
+        self.oracle_provider = Some(Arc::new(oracle_provider));
+    }
```
jbcaron:
```suggestion
    pub fn set_oracle_provider(&mut self, oracle_provider: Oracle) -> &mut Self {
        self.oracle_provider = Some(Arc::new(oracle_provider));
        self
    }
```
allows method chaining by returning a reference to the current instance

crates/client/mempool/src/l1.rs
```diff
@@ -11,6 +12,7 @@ pub struct GasPriceProvider {
     data_gas_price_sync_enabled: Arc<AtomicBool>,
     strk_gas_price_sync_enabled: Arc<AtomicBool>,
     strk_data_gas_price_sync_enabled: Arc<AtomicBool>,
+    pub oracle_provider: Option<Arc<Oracle>>,
 }
 
 impl GasPriceProvider {
```
jbcaron, line 19: consider updating all setter methods to return &mut Self to enable method chaining and improve API usability

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use anyhow::{bail, Context};
+use serde::{Deserialize, Serialize};
+
+pub const DEFAULT_API_URL: &str = "https://api.dev.pragma.build/node/v1/data/";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PragmaOracle {
+    #[serde(default = "default_oracle_api_url")]
+    pub api_url: String,
```
jbcaron: why not use a Url type?

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,134 @@
+use std::fmt;
+
+use anyhow::{bail, Context};
+use serde::{Deserialize, Serialize};
+
+pub const DEFAULT_API_URL: &str = "https://api.dev.pragma.build/node/v1/data/";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PragmaOracle {
+    #[serde(default = "default_oracle_api_url")]
```
jbcaron, line 14: the default value here seems unnecessary or unclear—consider whether it makes sense to provide a default for oracle_api_url or if it should be explicitly required instead.

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,134 @@
...
+            interval: Interval::OneMinute,
+            price_bounds: Default::default(),
+        }
+    }
+
+    pub fn get_fetch_url(&self, base: String, quote: String) -> String {
+        format!("{}{}/{}?interval={}&aggregation={}", self.api_url, base, quote, self.interval, self.aggregation_method)
+    }
+
+    pub async fn fetch_eth_strk_price(&self) -> anyhow::Result<(u128, u32)> {
```
jbcaron: please add documentation to explain the returned tuple

crates/primitives/oracle/src/pragma.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Default, Debug, Serialize, Deserialize, Clone)]
+/// Supported Aggregation Methods
+#[serde(rename_all = "snake_case")]
+pub enum AggregationMethod {
+    Median,
+    Mean,
+    #[default]
+    // Time weighted average price
+    Twap,
+}
```
jbcaron:
```suggestion
#[derive(Default, Debug, Serialize, Deserialize, Clone)]
/// Supported Aggregation Methods
#[serde(rename_all = "snake_case")]
pub enum AggregationMethod {
pub enum AggregationMethod {
    /// Computes the median value from the data.
    Median,
    /// Computes the mean (average) value from the data.
    Mean,
    /// Time Weighted Average Price. This is the default option.
    #[default]
    Twap,
}
```
'//' comments are not considered documentation and do not appear in the IDE

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+        match self {
+            Oracle::Pragma(pragma_oracle) => pragma_oracle.api_url = url,
+        }
+    }
```
jbcaron:
```suggestion
    pub fn set_base_url(&mut self, url: String) -> &mut Self {
        match self {
            Oracle::Pragma(pragma_oracle) => pragma_oracle.api_url = url,
        }
        self
    }
```

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+        match self {
+            Oracle::Pragma(pragma_oracle) => pragma_oracle.fetch_eth_strk_price().await,
+        }
+    }
+
+    pub fn set_api_key(&mut self, key: String) {
+        match self {
+            Oracle::Pragma(pragma_oracle) => pragma_oracle.api_key = key,
+        }
+    }
```
jbcaron:
```suggestion
    pub fn set_api_key(&mut self, key: String) -> &mut Self {
        match self {
            Oracle::Pragma(pragma_oracle) => pragma_oracle.api_key = key,
        }
        self
    }
```

crates/primitives/oracle/src/lib.rs
jbcaron: Instead of using the Oracle enum for defining different oracle configurations, it might be more flexible and extensible to define a trait that represents the Oracle interface

jbcaron (commented):
you can add documentation to structs to improve clarity

crates/primitives/oracle/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+
+# Other
+anyhow = { workspace = true }
+num-bigint = { workspace = true }
+reqwest = { workspace = true }
+serde = { workspace = true }
```
jbcaron:
```suggestion
serde = { workspace = true, features = ["derive"] }
```

crates/client/mempool/src/l1.rs
jbcaron: we could also consider a Builder pattern which would provide the same benefits plus immutability and validation

crates/primitives/oracle/src/pragma.rs
azurwastaken: changed

crates/client/mempool/src/l1.rs
azurwastaken: done

crates/primitives/oracle/src/lib.rs
azurwastaken: done

crates/primitives/oracle/src/pragma.rs
azurwastaken: done

crates/primitives/oracle/Cargo.toml
azurwastaken: done

crates/client/mempool/src/l1.rs
azurwastaken: done

crates/primitives/oracle/src/lib.rs
azurwastaken: done

crates/client/mempool/src/l1.rs
```diff
@@ -12,6 +13,7 @@ pub struct GasPriceProvider {
     data_gas_price_sync_enabled: Arc<AtomicBool>,
     strk_gas_price_sync_enabled: Arc<AtomicBool>,
     strk_data_gas_price_sync_enabled: Arc<AtomicBool>,
+    pub oracle_provider: Option<Arc<Box<dyn Oracle>>>,
```
jbcaron:
```suggestion
    pub oracle_provider: Option<Arc<dyn Oracle>>,
```
Because Arc already provides heap allocation and dynamic dispatch for trait objects, making Box redundant here. This double indirection adds unnecessary complexity.
Need a specific reason to keep both Arc and Box?

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+use async_trait::async_trait;
+
+pub mod pragma;
+
+#[async_trait]
+pub trait Oracle: Send + Sync {
```
jbcaron, line 6: 👍

crates/primitives/oracle/src/lib.rs
```diff
@@ -0,0 +1,10 @@
...
+    async fn fetch_eth_strk_price(&self) -> anyhow::Result<(u128, u32)>;
+    fn get_fetch_url(&self, base: String, quote: String) -> String;
+    fn get_api_key(&self) -> &String;
```
jbcaron: I don't understand why these methods are part of the trait

crates/client/mempool/src/l1.rs
```diff
@@ -23,9 +25,20 @@ impl GasPriceProvider {
...
+            oracle_provider: None,
         }
     }
 
+    pub fn is_oracle_needed(&self) -> bool {
+        self.strk_gas_price_sync_enabled.load(Ordering::Relaxed)
+            || self.strk_data_gas_price_sync_enabled.load(Ordering::Relaxed)
+    }
+
+    pub fn set_oracle_provider(&mut self, oracle_provider: Box<dyn Oracle>) -> &mut Self {
```
jbcaron:
```suggestion
    pub fn set_oracle_provider(&mut self, oracle_provider: impl Oracle + 'static) -> &mut Self {
```

crates/node/src/main.rs
```diff
@@ -106,6 +107,20 @@ async fn main() -> anyhow::Result<()> {
         l1_gas_setter.update_strk_l1_data_gas_price(strk_fix_blob_gas as u128);
         l1_gas_setter.set_strk_data_gas_price_sync_enabled(false);
     }
+    if let Some(ref oracle_url) = run_cmd.l1_sync_params.oracle_url {
+        if let Some(ref oracle_api_key) = run_cmd.l1_sync_params.oracle_api_key {
+            let oracle = PragmaOracleBuilder::new()
+                .with_api_url(oracle_url.clone())
+                .with_api_key(oracle_api_key.clone())
+                .build();
+            l1_gas_setter.set_oracle_provider(Box::new(oracle));
```
jbcaron:
```suggestion
            l1_gas_setter.set_oracle_provider(oracle);
```

jbcaron (changes requested):
Please review my comments and fix the CI issues. Let me know if you need clarification

antiyro (approved):

jbcaron (approved):

crates/primitives/oracle/Cargo.toml
```diff
@@ -18,7 +18,8 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 # Other
 anyhow = { workspace = true }
+async-trait = "0.1.83"
```
Trantorian1: Import this from workspace instead.

crates/primitives/oracle/src/pragma.rs
```diff
@@ -124,12 +135,45 @@ impl Default for PriceBounds {
...
+    }
+
+    pub fn with_api_key(mut self, api_key: String) -> Self {
+        self.api_key = Some(api_key);
+        self
+    }
+
+    pub fn build(self) -> PragmaOracle {
+        PragmaOracle {
+            api_url: self.api_url.unwrap_or_else(default_oracle_api_url),
```
Trantorian1: Shouldn't it be a failing case when no url is specified?

Trantorian1 (changes requested):

crates/client/eth/Cargo.toml
```diff
@@ -35,6 +35,7 @@ starknet_api = { workspace = true }
 # Other
 alloy = { workspace = true }
 anyhow = "1.0.75"
+bigdecimal = "0.4.5"
```
Trantorian1: This should be imported from workspace.

Trantorian1 (changes requested):

crates/client/eth/Cargo.toml
azurwastaken: bigdecimal is only used in crate eth

crates/primitives/oracle/src/pragma.rs
azurwastaken: in fact, the builder cannot be reached if url or key is not provided via cli.

removed those useless unwrap and stuff

Trantorian1 (approved):

PR #370, opened 2024-10-30
feat(confg): added chain config template and fgw example

antiyro:


antiyro:
force merged this since it doesnt affect the core

PR #371, opened 2024-10-30
fix: l2-to-l1 messaging

cchudant:
# Pull Request type

- Bugfix

Receipt is produced wrong

## What is the new behavior?

We need to do the same thing for both events and l2-to-l1 messages: iterate recursively into the callinfos.
We already had a function to do that for events, but it looked kind of ugly - so I replaced it with the recursive iterator that blockifier provides. I moved events to use that too.


antiyro (approved):
this should effectively solve l2>l1 messaging @shamsasari

shamsasari:
Should I wait for the merge into `feat/evm`?

crates/primitives/receipt/src/from_blockifier.rs
```diff
@@ -82,8 +79,13 @@ pub fn from_blockifier_execution_info(res: &TransactionExecutionInfo, tx: &Trans
...
 
-    let messages_sent = res
-        .non_optional_call_infos()
+    fn recursive_call_info_iter(res: &TransactionExecutionInfo) -> impl Iterator<Item = &CallInfo> {
+        res
+            .non_optional_call_infos() // all root callinfos
+            .flat_map(|call_info| call_info.iter()) // flatmap over the roots' recursive inner call infos
+    }
+
+    let messages_sent = recursive_call_info_iter(res)
```
jbcaron, line 88: why not do these 2 operations at the same time?

cchudant: sure okay

cchudant: it would be a bit messy with flat_map to unzip the iterator into two collections like this, do you have a suggestion that would look good?
i can't come up with a cleaner way than two iterations here and i don't think the perf implications are very important

shamsasari:
> this should effectively solve l2>l1 messaging @shamsasari

As I mentioned offline, this PR unfortunately doesn't make the Kakarot `TestL2ToL1Messages` test pass. This could be an issue with the test itself though, since it waits for a custom `MessageHashesAddedFromL2` event to be emitted. I had a look at Katana, which is what Kakarot is using for their tests, and it seems to have code which emits this custom event.

I'm not sure what to make of this. Is this behaviour somehow expected of a Starknet node?

crates/primitives/receipt/src/from_blockifier.rs
```diff
@@ -82,8 +79,13 @@ pub fn from_blockifier_execution_info(res: &TransactionExecutionInfo, tx: &Trans
         _ => None,
     };
 
-    let messages_sent = res
-        .non_optional_call_infos()
+    fn recursive_call_info_iter(res: &TransactionExecutionInfo) -> impl Iterator<Item = &CallInfo> {
+        res
+            .non_optional_call_infos() // all root callinfos
+            .flat_map(|call_info| call_info.iter()) // flatmap over the roots' recursive inner call infos
```
Trantorian1: Are you sure this is recursive? Looking at the source code in Blockifier, it seems that `CallInfo::iter` indeed yields an iterator over its `inner_calls: Vec<CallInfo>`, but this does not recurse to return the `inner_calls` of _these_ `CallInfo`s.

This means the iterator will work but we will be missing inner call infos which are more than 1 level of recursion deep. This might be why this pr does not pass the kkrt tests.

crates/primitives/receipt/src/from_blockifier.rs
```diff
@@ -171,102 +184,3 @@ impl From<GasVector> for L1Gas {
...
-        OrderedEvent {
-            order,
-            event: EventContent { keys: vec![EventKey(Felt::ZERO); order], data: EventData(vec![Felt::ZERO; order]) },
-        }
-    }
-
-    fn event(order: usize) -> Event {
-        Event { from_address: Default::default(), keys: vec![Felt::ZERO; order], data: vec![Felt::ZERO; order] }
-    }
-}
```
Trantorian1: Why remove this test? Now we do not check if `recursive_call_info_iter` is actually operating recursively.

Trantorian1 (changes requested):
There seems to be some logic issues with the way in which the `CallInfo` iterator is being used, and I am not convinced it is operating recursively as is expected.

crates/primitives/receipt/src/from_blockifier.rs
cchudant: See https://docs.rs/blockifier/latest/src/blockifier/execution/call_info.rs.html#178-181
The `iter` in CallInfo returns a CallInfoIter, whose inner vec queue starts with `self` in it.
The CallInfoIter https://docs.rs/blockifier/latest/src/blockifier/execution/call_info.rs.html#229 here implements Iterator by popping the last call info from its queue, adding the child call infos to the queue, and returning the popped call info. This effectively implements a Breadth first search algorithm into the call info tree, as an Iterator returning &CallInfo. This goes into all of the children inner calls.

crates/primitives/receipt/src/from_blockifier.rs
cchudant: okay i'll leave it

cchudant:
> > this should effectively solve l2>l1 messaging @shamsasari
> 
> As I mentioned offline, this PR unfortunately doesn't make the Kakarot `TestL2ToL1Messages` test pass. This could be an issue with the test itself though, since it waits for a custom `MessageHashesAddedFromL2` event to be emitted. I had a look at Katana, which is what Kakarot is using for their tests, and it seems to have code which emits this custom event.
> 
> I'm not sure what to make of this. Is this behaviour somehow expected of a Starknet node?

The TestL2ToL1Messages event, do you mean it's a l1 core contract event or an event in L2?

shamsasari:
`TestL2ToL1Messages` is the name of the kkt test class. In it it [waits](https://github.com/shamsasari/kakarot/blob/4c4235c9eab0ca4dc54f0518bc6e8b785e00f0e7/tests/end_to_end/L1L2Messaging/test_messaging.py#L92) for `MessageHashesAddedFromL2` events to be emitted from L2. Obviously Madara isn't emitting these, but Katana [does](https://github.com/dojoengine/dojo/blob/72578062d93ac12ff5f57c24c158b0a4b13bc6fd/crates/katana/contracts/messaging/solidity/src/StarknetMessagingLocal.sol#L56)!! Nor is this an event from the Starknet Core contract from what I can tell. I may be wrong and missing something here, but to me this looks like a poorly designed kkt test.

cchudant:
> `TestL2ToL1Messages` is the name of the kkt test class. In it it [waits](https://github.com/shamsasari/kakarot/blob/4c4235c9eab0ca4dc54f0518bc6e8b785e00f0e7/tests/end_to_end/L1L2Messaging/test_messaging.py#L92) for `MessageHashesAddedFromL2` events to be emitted from L2. Obviously Madara isn't emitting these, but Katana [does](https://github.com/dojoengine/dojo/blob/72578062d93ac12ff5f57c24c158b0a4b13bc6fd/crates/katana/contracts/messaging/solidity/src/StarknetMessagingLocal.sol#L56)!! Nor is this an event from the Starknet Core contract from what I can tell. I may be wrong and missing something here, but to me this looks like a poorly designed kkt test.

oh right that's an event produced by the core contract in L1.
Madara doesnt deploy L1 test core contract yet, it's expected to be deployed by the madara bootstrapper. I think we should do something similar to katana and actually deploy it in devnet mode.

i believe that we should change kakarot to have it listen to https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580rc66e004922f3/src/starkware/starknet/solidity/StarknetMessaging.sol#L142 that event instead, and our solidity devnet test core contract (that would be deployed on l1 at startup when running madara in devnet mode) would produce this event instead.
I'm not sure I understand their motivation for their `MessageHashesAddedFromL2` event, they talk about it above the function you linked but i don't see the difference with the official ConsumedMessageToL1 that i just linked?

shamsasari:
Yeah that makes more sense.

shamsasari:
Sorry, I forgot to ask, where is this L1 test core contract defined? If you say to listen to `ConsumedMessageToL1` events instead then why need anything from this test contract?

cchudant:
i'll explain it again
the l1 core contract is a solidity contract, which is not in madara
I am talking about this core contract https://etherscan.io/address/0xc662c410c0ecf747543f5ba90660f6abebd9c8c4 here is the one for the starknet mainnet chain, this is an ethereum contract on eth mainnet.
the core contract address is specified in the chain_config, and differs for each starknet chain
 
as of now, the only interaction we have with the l1 core contract is that we watch it in mc-eth for state updates (in full node mode) and l1 to l2 messages and message cancellations for sequencing mode.

we do not publish anything to the core contract, as this is the responsability of the madara-orchestrator - but in a real chain we should publish the SNOS proof (to another contract called gps statement verifier actually, but it's linked to the state update that is later published on the core contract), the state update as blob, and this is where the l2-to-l1 messages are consumed

for devnet mode, we probably want to deploy a fake core contract to L1, and have this publish to core contract thing in the sequencer instead of the bootstrapper
I kind of don't want the madara node to have this logic for production, as it would mean we need to have private keys to an ethereum wallet inside the madara node - and I really dont want that, i think it's way safer if it's the orchestrator who has it
also, if we were to manage the core contract in the madara node, this would mean that everything that orchestrator is doing (running SHARP, making the pie, publishing DA etc) would be in the madara node and that goes against the model we were thinking of with karnot

this means that right now, there is nothing that consumed the produced blocks in l2 by the madara node and updates the core contract
i'd like to have that logic in for devnet only in madara node, and even deploy that core contract
this would work very similarly to what katana is doing, which is deploying a core contract that does not require any stark proof, it's just a dumb contract for devnet that actually doesnt secure anything

the `MessageHashesAddedFromL2` is an event that is produced on L1, the file you linked is a solidity file, not cairo. it's an L1 core contract event.

crates/primitives/receipt/src/from_blockifier.rs
Trantorian1: Okay yep I got it, thanks for the explenation :+1:

Trantorian1 (approved):

PR #372, opened 2024-10-31
 feat(mempool): mempool transaction limits, saving to db and continue block prod from pending block

### LINKED ISSUE #373: bug: pending txs should be stored in graceful shutdown

### Description of bug

If some transactions have been processed due to ticks but haven't been included in a block because the block hasn't sealed, then if you close the sequencer and restart, the txs will be lost.

### Steps to reproduce

To reproduce

1. Start a madara sequencer with 2s tick time and 1hr block time
2. Send a basic txn and wait for receipt => works
3. Restart madara (before block is created so anytime before 1hr works)
4. Get receipt again => fails

This can break tooling like explorers or other apps that index pending blocks. Maybe not a good design but we should still try to track processed txs across restarts.

cchudant:
# Pull Request type

- Feature

Add limits to the mempool:
- transaction age
- max number of transactions in the mempool
- additional limit of declare transactions in the mempool

fixes #373


cchudant:
Added:
- Mempool transaction are now saved to and recovered from DB
- Block production can now continue the current pending block (fixes #373)
- Added e2e tests that restart madara devnet to test these two features.

configs/chain_config.example.yaml
```diff
@@ -69,3 +69,10 @@ bouncer_config:
 # /!\ Only used for block production.
 # Address of the sequencer (0x0 for a full node).
 sequencer_address: "0x0"
+
+# Transaction limit in the mempool.
+mempool_tx_limit: 10000
+# Transaction limit in the mempool, additionnal limit for declare transactions.
```
shamsasari: typo: additional

crates/client/analytics/src/lib.rs
```diff
@@ -42,7 +42,7 @@ impl Analytics {
 
         let tracing_subscriber = tracing_subscriber::registry()
             .with(tracing_subscriber::filter::LevelFilter::from_level(self.log_level))
-            .with(tracing_subscriber::fmt::layer().event_format(custom_formatter));
+            .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr).event_format(custom_formatter));
```
shamsasari: What does this change do? Logging to stderr?

crates/client/block_import/src/types.rs
```diff
@@ -142,13 +142,14 @@ pub struct UnverifiedCommitments {
...
 /// An unverified pending full block as input for the block import pipeline.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Default, Serialize, Deserialize)]
 pub struct UnverifiedPendingFullBlock {
     pub header: UnverifiedHeader,
     pub state_diff: StateDiff,
     pub transactions: Vec<Transaction>,
     pub receipts: Vec<TransactionReceipt>,
     pub declared_classes: Vec<DeclaredClass>,
+    pub visited_segments: Option<Vec<(Felt, Vec<usize>)>>,
```
shamsasari: Can we have a struct so that the types are better described

crates/client/block_production/src/finalize_execution_state.rs
```diff
@@ -0,0 +1,328 @@
...
+};
+use starknet_api::core::{ClassHash, CompiledClassHash, ContractAddress, Nonce};
+
+use crate::{Error, VisitedSegments};
+
+#[derive(Debug, thiserror::Error)]
+#[error("Error converting state diff to state map")]
+pub struct StateDiffToStateMapError;
+
+pub fn state_diff_to_state_map(diff: StateDiff) -> Result<StateMaps, StateDiffToStateMapError> {
```
shamsasari, line 22: Can each field value be assigned its own variable please. The inlining here is killing me!

crates/client/block_production/src/lib.rs
```diff
@@ -209,35 +105,89 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
     }
 
-    #[tracing::instrument(
-        skip(backend, importer, mempool, l1_data_provider, metrics),
-        fields(module = "BlockProductionTask")
-    )]
+    // #[tracing::instrument(
+    //     skip(backend, importer, mempool, l1_data_provider, metrics),
+    //     fields(module = "BlockProductionTask")
+    // )]
```
shamsasari: delete?

shamsasari:
This is quite a large PR. Can it not be broken down, especially if there's two features?

crates/client/analytics/src/lib.rs
cchudant: yep
it's a fix for the pr that brought tracing in, the logs are now in stdout instead of stderr
its a subtle change but its kinda a breaking change when you think about it, i'm just reverting it
and that actually bothers me because the e2e tests inherit stderr but pipe stdout, and i couldnt see anything of the node logs when adding e2e tests

shamsasari: But shouldn't logs print out to stdout?

crates/client/block_import/src/types.rs
jbcaron: maybe just add documentation

crates/client/analytics/src/lib.rs
cchudant: it makes more sense to have logs be printed in stderr right?

shamsasari: ok sure. Plus as you say this is reverting behaviour back.

jbcaron: in my opinion, stdout should be reserved for program output, while stderr is best suited for diagnostics, including logs

cchudant:
> This is quite a large PR. Can it not be broken down, especially if there's two features?

i am thinking about it
i think it's just easier at this point to have it as one pull request, all of the changes here are in the mempool. 

I believe if i were to split this pr in 2 or 3, one of these pr would necessarily still have >50 files changed.. that would be the one fixing saving the txs to the db, it's kind of a pain to split that because it touches all of the mempool crate, adds a db columns, and refactors all of the to_blockifier stuff

cchudant:
the mempool limits and continue block prod from pending block should only change a few files really, it's the mempool saving that's making it complex

cchudant:
@shamsasari i did the fixes you mentioned, except the splitting the pr one

antiyro (approved):

crates/client/block_production/src/finalize_execution_state.rs
```diff
@@ -0,0 +1,322 @@
...
+                segments: contract_class.get_visited_segments(class_visited_pcs)?,
+            })
+        })
+        .collect::<Result<_, Error>>()?;
+
+    Ok(VisitedSegments(visited_segments))
+}
+
+pub(crate) fn finalize_execution_state<S: StateReader>(
+    _executed_txs: &[MempoolTransaction],
```
Trantorian1: Why is this field still here if it is not being used?

crates/client/block_production/src/lib.rs
```diff
@@ -335,20 +287,19 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
+            &mut self.executor,
+            &self.backend,
+            &on_top_of,
+        )?;
 
-        let (state_diff, _visited_segments, _weights) =
-            finalize_execution_state(&executed_txs, &mut self.executor, &self.backend, &on_top_of)?;
+        // Add back the unexecuted transactions to the mempool.
+        stats.n_re_added_to_mempool = txs_to_process.len();
+        self.mempool.re_add_txs(txs_to_process, executed_txs);
```
Trantorian1, line 298: We only seem to be exiting the above loop when `txs_to_process.is_empty()`, does this do anything?

crates/client/db/src/mempool_db.rs
```diff
@@ -0,0 +1,77 @@
...
+            .db
+            .iterator_cf(&col, IteratorMode::Start)
+            .map(|kv| {
+                let (k, v) = kv?;
+                let hash: Felt = bincode::deserialize(&k)?;
+                let tx: TransactionWithConvertedClass = bincode::deserialize(&v)?;
+
+                Result::<_>::Ok((hash, tx.tx, tx.converted_class))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
Trantorian1: Maybe this could be returned directly as an iterator to avoid allocating all in one go? I am assuming we will be applying some extra transformation on top of it so this might helpful.

crates/client/devnet/src/lib.rs
```diff
@@ -628,4 +637,133 @@ mod tests {
...
+            )))
+        )
+    }
+
+    #[rstest]
+    fn test_mempool_age_limit() {
+        let max_age = Duration::from_millis(1000);
+        let mut chain =
+            chain_with_mempool_limits(MempoolLimits { max_age, max_declare_transactions: 2, max_transactions: 5 });
+        println!("{}", chain.contracts);
```
Trantorian1: Consider removing this `println` (nitpick)

crates/client/exec/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod call;
 pub mod execution;
 mod fee;
 mod trace;
+// pub mod transaction;
```
Trantorian1: Remove this (nitpick)

crates/client/mempool/src/inner/deployed_contracts.rs
```diff
@@ -0,0 +1,31 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct DeployedContracts(HashMap<ContractAddress, u64>);
+impl DeployedContracts {
+    pub fn decrement(&mut self, address: ContractAddress) {
+        match self.0.entry(address) {
+            hash_map::Entry::Occupied(mut entry) => {
+                *entry.get_mut() -= 1;
+                if entry.get() == &0 {
+                    entry.remove();
+                }
```
Trantorian1: Is it possible for the entry to start at 0?

crates/client/mempool/src/inner/limits.rs
```diff
@@ -0,0 +1,148 @@
...
+pub(crate) struct TransactionCheckedLimits {
+    check_tx_limit: bool,
+    check_declare_limit: bool,
+    check_age: bool,
+    tx_arrived_at: SystemTime,
+}
+
+impl TransactionCheckedLimits {
+    // Returns which limits apply for this transaction.
+    // This struct is also used to update the limits after insretion, without having to keep a clone of the transaction around.
```
Trantorian1: Typo (nitpick)

crates/client/mempool/src/inner/limits.rs
```diff
@@ -0,0 +1,148 @@
...
+                check_declare_limit: false,
+                check_age: true,
+                tx_arrived_at: tx.arrived_at,
+            },
+            TransactionType::L1Handler => TransactionCheckedLimits {
+                check_tx_limit: true,
+                check_declare_limit: false,
+                check_age: false,
+                tx_arrived_at: tx.arrived_at,
+            },
```
Trantorian1: Why are we not checking the age limit on L1Handler transactions? Is it possible to pollute the mempool with a large number of L1Handler transactions which never get removed because of this?

crates/client/mempool/src/inner/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    contract_addr: Felt,
+    timestamp: ArrivedAtTimestamp,
+}
+
+impl PartialEq for AccountOrderedByTimestamp {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other).is_eq()
+    }
+}
+impl Eq for AccountOrderedByTimestamp {}
```
Trantorian1: Why not just derive those?

crates/client/mempool/src/inner/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+        // Important: Fallback on contract addr here.
+        // There can be timestamp collisions.
+        self.timestamp.cmp(&other.timestamp).then_with(|| self.contract_addr.cmp(&other.contract_addr))
+    }
+}
+impl PartialOrd for AccountOrderedByTimestamp {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
Trantorian1, line 44: Same here.

crates/client/mempool/src/inner/nonce_chain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[derive(Debug, Clone)]
+pub struct OrderMempoolTransactionByNonce(pub MempoolTransaction);
+
+impl PartialEq for OrderMempoolTransactionByNonce {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other).is_eq()
+    }
+}
+impl Eq for OrderMempoolTransactionByNonce {}
```
Trantorian1, line 15: Why are therse not derived?

crates/client/mempool/src/inner/nonce_chain.rs
```diff
@@ -0,0 +1,147 @@
...
+impl Ord for OrderMempoolTransactionByNonce {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
+        self.0.nonce().cmp(&other.0.nonce())
+    }
+}
+impl PartialOrd for OrderMempoolTransactionByNonce {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
Trantorian1, line 25: Same here

crates/client/mempool/src/inner/nonce_chain.rs
```diff
@@ -0,0 +1,147 @@
...
+            if let Some((previous, _)) = self.transactions.remove_entry(&mempool_tx) {
+                let previous = previous.0.clone();
+                let inserted = self.transactions.insert(mempool_tx, ());
+                debug_assert!(inserted.is_none());
+                ReplacedState::Replaced { previous }
+            } else {
+                let inserted = self.transactions.insert(mempool_tx, ());
+                debug_assert!(inserted.is_none());
+                ReplacedState::NotReplaced
+            }
```
Trantorian1, line 102: Couldn't you just use `BtreeMap::entry` in combination with `Entry::Vacant` and `Entry::Occupied` to avoid the double lookup?

crates/client/mempool/src/inner/proptest.rs
```diff
@@ -0,0 +1,269 @@
...
+                    };
+
+                    assert_eq!(expected, res);
+
+                    if expected.is_ok() {
+                        if insert.0.tx.tx_type() == TransactionType::DeployAccount {
+                            new_contracts.insert(insert.0.contract_address());
+                        }
+                        inserted.insert(insert.0.tx_hash());
+                        inserted_contract_nonce_pairs.insert((insert.0.nonce(), insert.0.contract_address()));
```
Trantorian1, line 234: Should we maybe not also add an assertion to check that the inserted transaction matches?

crates/client/mempool/src/lib.rs
```diff
@@ -137,19 +174,28 @@ impl Mempool {
...
         if !is_only_query(&tx) {
-            tracing::debug!("Adding to mempool tx_hash={:#x}", tx_hash(&tx).to_felt());
-            // Finally, add it to the nonce chain for the account nonce
+            tracing::debug!("Adding to inner mempool tx_hash={:#x}", tx_hash);
+            // Add to db
+            let saved_tx = blockifier_to_saved_tx(&tx, arrived_at);
+            self.backend.save_mempool_transaction(&saved_tx, tx_hash, &converted_class)?;
+
+            // Add it to the inner mempool
             let force = false;
```
Trantorian1, line 182: Is this variable really necessary?

crates/primitives/block/src/lib.rs
```diff
@@ -274,6 +274,15 @@ impl From<MadaraBlock> for MadaraMaybePendingBlock {
...
 
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct VisitedSegments(pub Vec<VisitedSegmentEntry>);
+
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct VisitedSegmentEntry {
+    pub class_hash: Felt,
+    pub segments: Vec<usize>,
+}
+
```
Trantorian1, line 291: Maybe add a few comments explaining how this works? I'm still not sure I got what thos does exactly :face_with_spiral_eyes:

crates/primitives/transactions/src/broadcasted_to_blockifier.rs
```diff
@@ -19,182 +137,58 @@ pub enum BroadcastedToBlockifierError {
...
     #[error("Failed to convert transaction to starkneti-api: {0}")]
     ConvertToTxApiError(#[from] TransactionApiError),
     #[error("Failed to convert transaction to blockifier: {0}")]
     ConvertTxBlockifierError(#[from] TransactionExecutionError),
     #[error("Failed to convert contract class: {0}")]
     ConvertContractClassError(#[from] ContractClassError),
-    #[error("Declare legacy contract classes are not supported")]
-    LegacyContractClassesNotSupported,
+    // #[error("Declare legacy contract classes are not supported")]
+    // LegacyContractClassesNotSupported,
```
Trantorian1: remove these comments

crates/tests/src/lib.rs
```diff
@@ -252,3 +260,195 @@ async fn madara_can_sync_a_few_blocks() {
...
+        "--chain-config-path",
+        "test_devnet.yaml",
+        "--chain-config-override",
+        "block_time=5min,pending_block_update_time=5min",
+    ]);
+    let mut node = cmd_builder.clone().run();
+    node.wait_for_ready().await;
+
+    // should find receipt
+    let _receipt = node.json_rpc().get_transaction_receipt(res.transaction_hash).await.unwrap();
```
Trantorian1, line 480: Maybe replace this `unwrap` with an `assert`?

crates/client/db/src/mempool_db.rs
```diff
@@ -0,0 +1,77 @@
...
+impl MadaraBackend {
+    #[tracing::instrument(skip(self), fields(module = "MempoolDB"))]
+    pub fn get_mempool_transactions(&self) -> Result<Vec<(Felt, SavedTransaction, Option<ConvertedClass>)>> {
+        let col = self.db.get_column(Column::MempoolTransactions);
+        let txs = self
+            .db
+            .iterator_cf(&col, IteratorMode::Start)
+            .map(|kv| {
+                let (k, v) = kv?;
+                let hash: Felt = bincode::deserialize(&k)?;
```
shamsasari: bincode prepends a 8-byte length value, so it outputs 40 bytes instead of 32. I would suggest we use `to_bytes_be` directly.

Trantorian1 (commented):
A few minute points might need changing. This is really cool work! (Also please for the love of god and everything dear _do not force push_, I don't want to have to sift through all these files again next time I review :sweat_smile: )

crates/client/exec/src/transaction.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[error("{0}")]
+    Internal(Cow<'static, str>),
+    TransactionExecutionError(#[from] )
+}
+
+/// Convert an starknet-api Transaction to a blockifier Transaction
+///
+/// **note:** this function does not support deploy transaction
+/// because it is not supported by blockifier
+pub fn to_blockifier_transactions(
```
shamsasari, line 29: `to_blockifier_transaction`, i.e. not plural

crates/client/block_production/src/lib.rs
cchudant: ouch woah good catch

i think the
```rs
if block_now_full {
  break;
}
```
breaks out the wrong loop! it needs to be a break with label
uuh i'll add a test before fixing that because that's pretty bad that i did not find this earlier

crates/client/devnet/src/lib.rs
cchudant: this was intended actually, this is the println! that writes the devnet contracts keys to the console with the proper formatting and everything

i thought it made more sense to put it in stdout instead of inside a normal log, so that it can be parsed by other programs easily or whatnot
i'm fine with changing it but i was thinking that this made more sense that way

(this is not a change by the way, it was like that since the day i made the devnet)

crates/client/mempool/src/lib.rs
cchudant: no, but you can't see from the github interface what that `false` is supposed to mean if it's written like `mempool.insert(tx, false)`
so instead, for these kind of parameters that are very hard to understand without seeing the called function, i just make a variable like this
it's a good habit to have imo and it's something i would probably even require when reviewing a pull request (though this has never come up yet in any review i've done in madara i believe)
if there was a clippy warning forcing it that way i would be even happier actually

crates/tests/src/lib.rs
cchudant: i think it would look more janky, what do you have in mind?
assert result is_ok()? that wouldnt show the error to the console if the assertion fails

notlesh: How about an `expect`?

cchudant: it's in a test, i usually never use expect in tests and just use unwraps
I think it's fine to unwrap in tests

crates/client/db/src/mempool_db.rs
cchudant: that's very fair

crates/client/block_production/src/lib.rs
cchudant: no, it's breaking from the correct loop afaik, i misread
this `loop` has two exits: either txs_to_process is empty (mempool has no more transactions for us)
or block_now_full is true, meaning blockifier has returned fewer transactions than what we asked for, meaning bouncer cap has been reached.
this re_add is useful for the later case only, to put the remaining items back into the mempool

crates/client/db/src/mempool_db.rs
cchudant: we should do that everywhere tbh, i'll postpone that for later

cchudant: #403

crates/client/db/src/mempool_db.rs
cchudant: that's true, i was thinking this is only startup code so i didnt bother much with performance at all

cchudant: i changed it to iterator :)

crates/client/mempool/src/inner/deployed_contracts.rs
cchudant: it's not, the count will be one on insertion

crates/client/mempool/src/inner/limits.rs
cchudant: it's because we don't want to miss l1 handler transactions - these are transactions added on the L1 core contract.

I fixed the check_tx_limit which did count for L1Handler, i missed that before. Now it still counts but won't reject any transactions for this reason

crates/client/mempool/src/inner/mod.rs
cchudant: we want Ord to behave a specific way to make the .first()/pop_first on the btreemap give the transactions in the FCFS order (by timestamp)
i want it to be explicit here

crates/client/mempool/src/inner/nonce_chain.rs
cchudant: we only want the nonce to be compared, not the rest.
if two txs are equal, we want the two mempool transactions to appear as equal to the btreemap, so that we find that the entry is not empty for this nonce - this is how we check for nonce collision.
this goes for partialord/ord/partial_eq/eq, the OrderMempoolTransactionByNonce is a wrapper around MempoolTransaction that just orders them in the specific order so that the NonceChain::transactions btreemap works like we want it to

crates/client/mempool/src/inner/nonce_chain.rs
cchudant: we can't update the keys using the btreemap Entry api unfortunately, we can only update the value

i think we should remove the `OrderMempoolTransactionByNonce` struct, and make the field a BTreeMap<Nonce, MempoolTransaction> instead
that would make so much more sense, I did it that way because I was using a BTreeSet before
added a todo for that

crates/client/mempool/src/inner/proptest.rs
cchudant: i think we can't, because forced transactions can update the nonce chain and these will not match anymore

crates/primitives/block/src/lib.rs
cchudant: added:

/// Visited segments are the class segments that are visited during the execution of the block.
/// This info is an input of SNOS and used for proving.

i don't know that much about it myself, but as far as i understand it's just the portions of the classes that are visited during execution

Trantorian1 (approved):
All good! Great work :fire:

PR #377, opened 2024-11-05
fix(error): Added a comment for non archive node L1 keys

antiyro:
This gives more context to the user in case they are using an L1 RPC key that points to a non archive client.

### DIFF
--- a/crates/client/eth/src/l1_messaging.rs
+++ b/crates/client/eth/src/l1_messaging.rs
@@ -62,7 +62,9 @@ pub async fn sync(
         .to_block(BlockNumberOrTag::Finalized)
         .watch()
         .await
-        .context("Failed to watch event filter")?
+        .context(
+            "Failed to watch event filter - Ensure you are using an L1 RPC endpoint that points to an archive node",
+        )?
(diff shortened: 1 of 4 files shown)


crates/client/eth/src/l1_messaging.rs
```diff
@@ -62,7 +62,9 @@ pub async fn sync(
         .to_block(BlockNumberOrTag::Finalized)
         .watch()
         .await
-        .context("Failed to watch event filter")?
+        .context(
+            "Failed to watch event filter - Ensure you are using an L1 RPC endpoint that points to an archive node",
+        )?
```
Trantorian1, line 67: Interesting. I'm not familiar with Eth l1 non-archive nodes, what are the differences with typical archive nodes?

Trantorian1 (approved):
All good!

crates/client/eth/src/l1_messaging.rs
antiyro: L1 Full nodes only contain pruned state.

cchudant (approved):

PR #378, opened 2024-11-06
perf(get events)

Trantorian1:
# Pull Request type

The current implementation of `starknet_getEvents` in Madara suffers from very poor performance compared to other nodes. This pr's goal is to improve this (will benchmark this again before un-drafting the pull request).

image
_`starknet_getEvents` performance across nodes, bench measured with sync from a local fgw_

## What is the new behavior?

- Refactored block event retrieval to avoid unnecessary allocations, clones and key checks
- Better handling of `StarknetRpcApiError::BlockNotFound` edge cases.
- Added tests to `starknet_getEvents` to ensure proper functionality

> [!NOTE]
> Further new behavior will be documented as this PR is developed

Yes. Calls to `starknet_getEvent` would previously return `Ok()` if `from_block` or `to_block` were higher than the latest block. This has been changed to return `StarknetRpcApiError::BlockNotFound`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ starknet-e2e-test/contracts/build
 # rpc output
 output*
 
+# test output
+test_output*
+
 tmp/
 
 *.info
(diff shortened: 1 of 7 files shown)


Trantorian1:
This pr is no longer in the works.

Issue #231, opened 2024-08-19
feat(docs): Document rocksdb configuration


cchudant:
### Is there an existing issue?

- [X] I have searched the existing issues

### Request

from a comment raised in #220's review: We should document our choices around disabling WAL, atomic manual flushes, the threading config etc.

Assigning this to myself

### Are you willing to help with this request?

Yes!

heemankv:
Status update from codebase review on 2026-05-05.

Classification: Still relevant.

Current state:
I did not find a dedicated user-facing RocksDB configuration document; most of the real detail is still buried in code and internal notes.

Keeping this issue open based on the current tree.

PR #379, opened 2024-11-06
fix(db): fix number of files in db, startup hang, ram issues and flushing issues

cchudant:
# Pull Request type

- Bugfix

Pragma devnet has >1M sst files. This is because compaction couldn't keep up with the atomic flushes we do after every block. This resulted in db startup taking an absurd amount of time (sometimes tens of minutes, scanning most of those >1M files). In addition, the flushes were also taking a lot of time (~10s!).

## What is the new behavior?

I could replicate the issue by copying the pragma db locally, but I could also replicate the issue by syncing a node with a modified version of block-import that force-flushes every block (this mimicks the behavior of the sequencer which does the same) and watching the number of files in the db. This was much easier.

The tiered compaction is not adapted for our needs, so I switched the db to universal compaction. I managed to sync 100k blocks with only a hundred sst files instead of around 30 thousand files in the db with this compaction mode. Hopefully, this should fixes all of these issues + the RAM taken by rocksdb should now be bounded.

Theorically, no - but i think you should upgrade your sequencer db by syncing a full-node on it, and replacing the sequencer's db with the full-node's to benefit of all of the new db options. Using an old db will work, but it won't try to update the old sst files very much I think.


crates/client/db/src/rocksdb_options.rs
```diff
@@ -0,0 +1,73 @@
...
+    options.create_missing_column_families(true);
+    let cores = std::thread::available_parallelism().map(|e| e.get() as i32).unwrap_or(1);
+    options.increase_parallelism(cores);
+    options.set_max_background_jobs(cores);
+
+    options.set_atomic_flush(true);
+    options.set_max_subcompactions(cores as _);
+
+    options.set_max_log_file_size(10 * MiB);
+    options.set_max_open_files(2048);
```
antiyro, line 24: could you add a comment describing each of those numbers?

crates/client/db/src/rocksdb_options.rs
```diff
@@ -0,0 +1,73 @@
...
+                    contract_db::CONTRACT_NONCES_PREFIX_EXTRACTOR,
+                ));
+            }
+            _ => {}
+        }
+
+        options.set_compression_type(DBCompressionType::Zstd);
+        match self {
+            Column::BlockNToBlockInfo | Column::BlockNToBlockInner => {
+                options.optimize_universal_style_compaction(1 * GiB);
```
antiyro, line 65: same here why 1GB?

antiyro (approved):
lgtm if you could just give some comments on the current used parameters

crates/client/db/src/rocksdb_options.rs
```diff
@@ -0,0 +1,73 @@
...
+    options.increase_parallelism(cores);
+    options.set_max_background_jobs(cores);
+
+    options.set_atomic_flush(true);
+    options.set_max_subcompactions(cores as _);
+
+    options.set_max_log_file_size(10 * MiB);
+    options.set_max_open_files(2048);
+    options.set_keep_log_file_num(3);
+    options.set_log_level(rocksdb::LogLevel::Warn);
```
shamsasari, line 26: Why warn and not info?

crates/client/db/src/rocksdb_options.rs
shamsasari: Yes, or at least a doc for the function explaining how it addresses the issue this PR is solving.

shamsasari (changes requested):

crates/client/db/src/rocksdb_options.rs
cchudant: this is not a change, it's marked as a change because git bad
 
info is the default, and it will print column stats every so often, making huuuge log files with absolutely useless information

see #345 where it was introduced

crates/client/db/src/rocksdb_options.rs
cchudant: the max open files was added in #367 
i'll add info to the options, i may also solve #231 while i'm at it if it doesnt take me too long

crates/client/db/src/rocksdb_options.rs
cchudant: i can't access to the column sizes metrics as of now, so I eyeballed it
we can tune all of this later, the basic idea is that this arg is the RAM budget of the column - i like to think that a normal machine should run madara in 4Go RAM max, and base these tuning params around that

I have not run any test to see if these numbers are optimal they just looked okay to me.

Trantorian1 (approved):

shamsasari (approved):

shamsasari:
Is this good to merge? I will need to deal with the merge conflicts in https://github.com/madara-alliance/madara/pull/388.

PR #380, opened 2024-11-13
build: update blockifier to 0.8.0 and other dependencies

jbcaron:
## Pull Request type

- Build-related changes

Dependencies are outdated.

Resolves: #NA

## What is the new behavior?

- update blockifier to stable release 0.8.0.
- update dependencies:
  - starknet-rs
  - starknet-types
  - cairo
- removed obsolete conversions for starknet-providers types
- leveraged new native conversions for felt to primitive types, improving code readability and reducing custom conversion logic

No


Cargo.toml
```diff
@@ -231,8 +231,5 @@ tracing-subscriber = { version = "0.3.18", features = [
 tracing-opentelemetry = "0.26.0"
 
 [patch.crates-io]
-starknet-core = { git = "https://github.com/kasarlabs/starknet-rs.git", branch = "fork" }
-starknet-providers = { git = "https://github.com/kasarlabs/starknet-rs.git", branch = "fork" }
-
 # wait for PR https://github.com/starknet-io/types-rs/pull/76 to be merged
-starknet-types-core = { git = "https://github.com/kasarlabs/types-rs.git", branch = "feat-deserialize-v0.1.5" }
+starknet-types-core = { git = "https://github.com/kasarlabs/types-rs.git", branch = "feat-deserialize-v0.1.7" }
```
Trantorian1, line 229: Are we still using the functionality added into this fork?

crates/primitives/receipt/src/lib.rs
```diff
@@ -227,6 +226,12 @@ pub struct FeePayment {
     pub unit: PriceUnit,
 }
 
+impl From<Felt> for FeePayment {
+    fn from(fee: Felt) -> Self {
+        Self { amount: fee, unit: PriceUnit::Wei }
```
Trantorian1, line 232: Why the default to `Wei`? Maybe isolate this into a more explicit function like `felt_to_fee_wei`? Nitpick

Trantorian1 (approved):
Seems good!

Cargo.toml
jbcaron: yes it is an optimization for the serialization of Felt with bincode

crates/primitives/receipt/src/lib.rs
jbcaron: I think this implementation dates back to before V3. I’ll check if it needs to be updated

jbcaron: the PriceUnit value needs to be dynamic and depend on the transaction version. For versions ZERO, ONE, and TWO, it should serialize as `Wei`. For all other versions, it should serialize as `Fri`. This ensures compatibility with differing transaction specifications.

I propose addressing this issue in a separate pull request to keep the changes isolated and easier to review.

Trantorian1: Sounds good, keeping this conversation open in the meantime as a point of reference

cchudant: you can make a comment in the code / lift this concern to an issue

cchudant (commented):
what feature is this pr adding? (why is it called feat and not fix)
is it just a pr cleaning up stuff and upgrading some versions or does it do something else?

crates/primitives/receipt/src/lib.rs
jbcaron: added, see #385

crates/primitives/class/src/compile.rs
```diff
@@ -134,9 +133,15 @@ fn parse_sierra_version(program: &[Felt]) -> Result<SierraVersion, ClassCompilat
         [first, ..] if first == &VERSION_0_1_0_AS_SHORTSTRING => Ok(SierraVersion(0, 1, 0)),
         [a, b, c, ..] => {
             let (a, b, c) = (
-                felt_to_u64(a).map_err(|_| ClassCompilationError::ParsingSierraVersion("malformed version".into()))?,
-                felt_to_u64(b).map_err(|_| ClassCompilationError::ParsingSierraVersion("malformed version".into()))?,
-                felt_to_u64(c).map_err(|_| ClassCompilationError::ParsingSierraVersion("malformed version".into()))?,
+                a.to_owned()
```
cchudant: this to_owned makes the &Felt -> Felt conversion right? maybe change to .clone() because that was making me confused

cchudant (approved):
cool

crates/primitives/class/src/compile.rs
jbcaron: done
replaced `to_owned` with dereferencing

PR #383, opened 2024-11-14
feat(fgw): add transaction methods for GatewayProvider

jbcaron:
## Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?
- split the provider crate into two separate crates:
  - server
  - client
- renamed FeederClient to GatewayProvider
- added methods to GatewayProvider:
  - add_invoke_transaction
  - add_declare_transaction
  - add_deploy_account_transaction
- refactored ForwardToProvider to utilize the new methods instead of relying on starknet-providers

No


crates/client/gateway/client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-description = "Madara client rpc service"
-name = "mc-gateway"
+description = "Madara gateway provider"
```
Trantorian1: This is a nitpick but maybe rename the enclosing folder to `provider` instead?

Trantorian1 (approved):
Seems good to me!

crates/client/gateway/client/Cargo.toml
jbcaron: fixed

crates/client/rpc/src/errors.rs
```diff
@@ -167,36 +167,26 @@ impl From<StarknetRpcApiError> for jsonrpsee::types::ErrorObjectOwned {
...
-            StarknetError::CompiledClassHashMismatch => StarknetRpcApiError::CompiledClassHashMismatch,
-            StarknetError::UnsupportedTxVersion => StarknetRpcApiError::UnsupportedTxnVersion,
-            StarknetError::UnsupportedContractClassVersion => StarknetRpcApiError::UnsupportedContractClassVersion,
-            StarknetError::UnexpectedError(data) => StarknetRpcApiError::ErrUnexpectedError { data },
-            StarknetError::NoTraceAvailable(_) => StarknetRpcApiError::InternalServerError,
-            StarknetError::TransactionExecutionError(error) => StarknetRpcApiError::TxnExecutionError {
-                tx_index: error.transaction_index as usize,
-                error: error.execution_error.to_string(),
-            },
+        match err.code {
```
cchudant, line 170: i am confused, whtat's the difference? why don't the code and enum variants match? :sweat_smile:

crates/primitives/gateway/src/user_transaction.rs
```diff
@@ -15,21 +15,31 @@ use starknet_types_core::felt::Felt;
...
+    Declare(UserDeclareTransaction),
     InvokeFunction(UserInvokeFunctionTransaction),
     DeployAccount(UserDeployAccountTransaction),
 }
 
 impl From<UserTransaction> for BroadcastedTransaction {
     fn from(transaction: UserTransaction) -> Self {
         match transaction {
-            UserTransaction::DeclareV1(v1) => BroadcastedTransaction::Declare(v1.into()),
+            UserTransaction::Declare(v1) => BroadcastedTransaction::Declare(v1.into()),
```
cchudant, line 26: rename the variable v1 => tx

cchudant (approved):
how is this a 46-files changed pr whaaat

crates/client/rpc/src/providers/forward_to_provider.rs
```diff
@@ -1,26 +1,27 @@
...
-    pub fn new(provider: P) -> Self {
+impl ForwardToProvider {
+    pub fn new(provider: GatewayProvider) -> Self {
         Self { provider }
     }
 }
 
 #[async_trait]
-impl<P: Provider + Send + Sync> AddTransactionProvider for ForwardToProvider<P> {
+impl AddTransactionProvider for ForwardToProvider {
```
cchudant, line 24: note for later: i'd love to move AddTransactionProvider to its own crate

jbcaron: I also asked myself the same question
I split the gateway crate into client and server (which is a good thing) due to cyclic dependencies caused by this 😅

cchudant: ah yes
i think of this addtxprovider trait as an abstraction from the frontend services and backend services in madara, and it would allow mc-eth (or other future l1 watchers) / mc-rpc / mc-gateway / mc-p2p not to have to depend on mc-rpc
i think if we do this we should redefine our own error enum for add tx provider, and implement the Into starknet types rpc error in mc-rpc for it (and in gateway, p2p too then)

PR #384, opened 2024-11-15
refactor: replace starknet-rs BlockId with types-rs BlockId

jbcaron:
## Pull Request type

Resolves: #NA

## What is the new behavior?

- remove mp_block::BlockId
-

No


crates/client/block_import/src/verify_apply.rs
```diff
@@ -5,18 +5,21 @@ use crate::{
...
-use starknet_core::types::Felt;
+use starknet_types_core::felt::Felt;
 use starknet_types_core::hash::{Poseidon, StarkHash};
+use starknet_types_rpc::BlockTag;
 use std::{borrow::Cow, sync::Arc};
 
 mod classes;
 mod contracts;
 
+type BlockId = starknet_types_rpc::BlockId<Felt>;
```
cchudant: why is it generic?

cchudant: i think this type alias should live in mp-block, it would be less messy wdyt?

cchudant (approved):
whats the motivation for removing our BlockId/BlockTag vs the other stuff where we keep our structs?

jbcaron:
> whats the motivation for removing our BlockId/BlockTag vs the other stuff where we keep our structs?

the motivation behind removing our custom BlockId/BlockTag type is that it didn’t provide any tangible benefits
by harmonizing with the common type, we avoid unnecessary conversions between provider, RPC, and internal representations. This simplification reduces overhead and ensures consistency across the codebase

now, we only maintain two BlockId types:
- the unified one for most operations
- the specialized one for the DB

crates/client/block_import/src/verify_apply.rs
jbcaron: I was actually torn between these two approaches. I don’t like the fact that the current solution exposes this type publicly and introduces additional boilerplate. Moving the type alias to mp-block would indeed help encapsulate it better and reduce clutter in the main codebase.

as  for why the type is generic, it’s designed that way to allow flexibility in representing the block hash with any type, depending on the context.

Trantorian1 (approved):
Seems good!

PR #386, opened 2024-11-18
feat(endpoint): added extra `admin` rpc endpoint for sensitive rpc calls

Trantorian1:
# Pull Request type

- Feature

Currently, all rpc methods are implemented under a single endpoint. This poses issues as we want to add privilidged methods which allow node operators and developers to manipulate their node at a distance, and currently these methods would be exposed to users as well. 

## What is the new behavior?

A few changes have been made, in order of importance:

- Added an extra `admin` endpoint running on port `9943`. See [here](https://github.com/madara-alliance/madara/blob/5e4f779dae68902e84a943ed686c0d883e478116/crates/node/src/service/rpc/mod.rs#L42) for when this endpoint is enabled (perhaps this need to be clearer?). Currently, only the `MadarRpcWriteApi` methods have been moved there but we plan to add more methods soon.
- Removed the rate limiting layer from madara's rpc service. More on that below.
- Updated `RPC.md`

> [!NOTE]
> The idea is that node operators and developer will (and should!) be running their own proxy for handling features such as rate limiting, api tokens and the like. Separating admin methods into their own endpoint makes it easier to limit their access in this way.

Yes. If you were relying on the `madara_addDeclareV0Transaction` rpc method, please note that this has been moved to the `admin` endpoint on port `9943`.

## Other information

I think it would be a good idea in the near future to bundle a sample proxy configuration with Madara which node operators and devs can use as a starting point to their own production environment. I am unfamiliar with this topic so if you are already using your own proxy with Madara for the reasons discussed above, please share!


crates/node/src/cli/rpc.rs
```diff
@@ -185,6 +181,16 @@ impl RpcParams {
         SocketAddr::new(listen_addr.into(), self.rpc_port)
     }
 
+    pub fn addr_admin(&self) -> SocketAddr {
+        let listen_addr = if self.rpc_external {
```
jbcaron: a specific bind configuration is required for the admin interface
we may want to expose the normal rpc methods externally, while restricting the admin rpc methods to local access, even if they are on a different port

however, ensure that both normal and admin RPC methods remain accessible externally when running inside Docker

crates/node/src/service/rpc/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+            (RpcMethods::Unsafe, _) => (true, true),
+            (RpcMethods::Auto, false) => (true, true),
+            (RpcMethods::Auto, true) => {
+                tracing::warn!(
+                    "Option `--rpc-external` will hide node operator endpoints. To enable them, please pass \
+                     `--rpc-methods unsafe`."
+                );
+                (true, false)
+            }
+        };
```
jbcaron: ensure this logic is explicitly reflected in the CLI

jbcaron (commented):
great

crates/node/src/cli/rpc.rs
Trantorian1: Yep, I'm assuming this is the responsibility of the node operator. For example, it might be worthwhile opening the admin methods to 0.0.0.0, but with a proxy requiring an admin api token and ip whitelist.

crates/node/src/service/rpc/mod.rs
Trantorian1: I have updated the CLI docs to more accuratly reflect how this works :+1:

crates/node/src/cli/rpc.rs
Trantorian1: I think I see more clearly what you were talking about. I have updated the `rpc-methods` cli arg (which is now rpc-endpoints) to handle this:

- Auto: `admin` and `user` are exposed on `localhost` by default. If `rpc-external` is _true_, `user` is exposed on `0.0.0.0` but `admin` stays exposed on `localhost`.
- Safe: `admin` is disabled and is _not_ exposed, even if `rpc-external` is set to _false_.
- Unsafe: `admin` is exposed on `0.0.0.0` if `rpc-external` is set to true.

crates/node/src/service/rpc/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+                server_handle_admin: None,
+            });
+        }
+
+        let (rpcs, node_operator) = match config.rpc_endpoints {
+            RpcEndpoints::Safe => (true, false),
+            RpcEndpoints::Auto | RpcEndpoints::Unsafe => (true, true),
+        };
+
+        let (read, write, trace, admin, ws) = (rpcs, rpcs, rpcs, node_operator, rpcs);
```
cchudant: can you remove this atrocity
(it's my fault originally so i'm allowed to say that ahah)

cchudant (commented):
would it make more sense if it worked like this:
- `--rpc-disabled` disables the normal rpcs (unchanged)
- `--rpc-external` makes the normal rpcs external
- `--node-operator-rpc` enables the node operator rpcs (disabled by default, but i'm thinking in devnet it should be enabled by default tbh - but maybe for a later pr since we don't have any endpoint that specifically useful for devnet yet, like manual revert and stuff)
- `--unsafe-node-operator-rpc-external` allows the node operator rpcs to listen to 0.0.0.0. this is very necessary in docker setups for example since your reverse proxy will have another local ip

that would mean we remove all of this --rpc-endpoints unsafe nonsense? this --rpc-endpoints thinggy is something i made by taking inspiration from substrate when we switched, that's how they do it - but contrary to them we chose to expose the node operator endpoints on another port, which i think makes this design a bit weird

crates/node/src/cli/rpc.rs
```diff
@@ -185,6 +189,16 @@ impl RpcParams {
         SocketAddr::new(listen_addr.into(), self.rpc_port)
     }
 
+    pub fn addr_admin(&self) -> SocketAddr {
+        let listen_addr = if self.rpc_external && self.rpc_endpoints == RpcEndpoints::Unsafe {
```
jbcaron: It seems unusual to require the user to enable RpcEndpoints::Unsafe to expose admin RPCs under Docker

jbcaron (approved):

crates/node/src/cli/rpc.rs
Trantorian1: To me this seems fine when considering a lot of existing Madara options can be quite confusing as is. Something worth investigating in the future would be configuration presets for node operators and docker for example, so calling somehting like `madara --preset docker` vs `madara --preset operator` will automatically set this up. I see several issues in the way in how clap works that might make this quite complicated though.

Trantorian1:
> would it make more sense if it worked like this:
> 
> * `--rpc-disabled` disables the normal rpcs (unchanged)
> * `--rpc-external` makes the normal rpcs external
> * `--node-operator-rpc` enables the node operator rpcs (disabled by default, but i'm thinking in devnet it should be enabled by default tbh - but maybe for a later pr since we don't have any endpoint that specifically useful for devnet yet, like manual revert and stuff)
> * `--unsafe-node-operator-rpc-external` allows the node operator rpcs to listen to 0.0.0.0. this is very necessary in docker setups for example since your reverse proxy will have another local ip
> 
> that would mean we remove all of this --rpc-endpoints unsafe nonsense? this --rpc-endpoints thinggy is something i made by taking inspiration from substrate when we switched, that's how they do it - but contrary to them we chose to expose the node operator endpoints on another port, which i think makes this design a bit weird

Imo we should avoid adding to many cli flags to Madara, it can already be confusing enough as is :sweat_smile: I chose to use `rpc-endpoints` to keep this encapsulated behind a single flag for ease of use.

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
```diff
@@ -0,0 +1,16 @@
+use m_proc_macros::versioned_rpc;
+use mp_transactions::BroadcastedDeclareTransactionV0;
+
+use jsonrpsee::core::RpcResult;
+use starknet_core::types::DeclareTransactionResult;
+
+/// This is an admin method, so semver is different!
```
shamsasari, line 7: I think this should match the version of the madara node itself.

cchudant (approved):

crates/client/rpc/src/lib.rs
```diff
@@ -91,32 +91,22 @@ impl Starknet {
...
-        rpc_api.merge(versions::v0_7_1::MadaraWriteRpcApiV0_7_1Server::into_rpc(starknet.clone()))?;
-    }
-    if ws {
-        rpc_api.merge(versions::v0_8_0::StarknetWsRpcApiV0_8_0Server::into_rpc(starknet.clone()))?;
-    }
+    rpc_api.merge(versions::user::v0_7_1::StarknetReadRpcApiV0_7_1Server::into_rpc(starknet.clone()))?;
+    rpc_api.merge(versions::user::v0_8_0::StarknetReadRpcApiV0_8_0Server::into_rpc(starknet.clone()))?;
+    rpc_api.merge(versions::user::v0_7_1::StarknetWriteRpcApiV0_7_1Server::into_rpc(starknet.clone()))?;
+    rpc_api.merge(versions::user::v0_7_1::StarknetTraceRpcApiV0_7_1Server::into_rpc(starknet.clone()))?;
+    rpc_api.merge(versions::user::v0_8_0::StarknetWsRpcApiV0_8_0Server::into_rpc(starknet.clone()))?;
```
shamsasari, line 101: This might be nitpicking, but why `user` and not `starknet`?

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
-    /// Listen to all network interfaces. This usually means that the RPC server will be accessible externally.
-    /// Please note that some endpoints should not be exposed to the outside world - by default, enabling remote access
-    /// will disable these endpoints. To re-enable them, use `--rpc-methods unsafe`
-    #[arg(env = "MADARA_RPC_EXTERNAL", long)]
+    /// Disables the user rpc endpoint. This includes all methods which are part
+    /// of the official starknet specs.
+    #[arg(env = "MADARA_RPC_DISABLE", long, default_value_t = false)]
+    pub rpc_disable: bool,
+
+    /// Exposes the user rpc endpoint on port `0.0.0.0`. This generally means
```
shamsasari: rpc -> RPC. There are several instances of this in this file.

shamsasari: The backticks don't do anything in the `--help` output and are printed, making it look weird.

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
-    ///
-    /// This is disabled by default.
-    ///
-    /// For example `--rpc-rate-limit 10` will maximum allow
-    /// 10 calls per minute per connection.
-    #[arg(env = "MADARA_RPC_RATE_LIMIT", long)]
-    pub rpc_rate_limit: Option<NonZeroU32>,
+    /// Enables the admin rpc endpoint. This includes additional rpc methods
+    /// which are not part of the official specs and can be used by node admins
+    /// to control their node at a distance. By default, this is exposed to
```
shamsasari: "By default, this is only exposed on localhost."

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
+    /// `localhost`.
+    #[arg(env = "MADARA_RPC_EXTERNAL", long, default_value_t = false)]
+    pub rpc_admin: bool,
+
+    /// Exposes the admin rpc endpoint on port `0.0.0.0`. This is especially
```
shamsasari: port -> address

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
+    /// useful when running Madara from inside a docker container. Be very
+    /// careful however when exposing this endpoint to the outside world.
```
shamsasari, line 76: Add elobration that this endpoint is secured and requires the user to add a proxy infront of it, etc. Essentially the note you have in the PR description should be summarised here for the user.

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
+    #[arg(env = "MADARA_RPC_EXTERNAL", long, default_value_t = false)]
```
shamsasari: MADARA_RPC_ADMIN_EXTERNAL

crates/node/src/cli/rpc.rs
```diff
@@ -67,36 +54,28 @@ impl FromStr for Cors {
...
-    ///
-    /// For example `--rpc-rate-limit 10` will maximum allow
-    /// 10 calls per minute per connection.
-    #[arg(env = "MADARA_RPC_RATE_LIMIT", long)]
-    pub rpc_rate_limit: Option<NonZeroU32>,
+    /// Enables the admin rpc endpoint. This includes additional rpc methods
+    /// which are not part of the official specs and can be used by node admins
+    /// to control their node at a distance. By default, this is exposed to
+    /// `localhost`.
+    #[arg(env = "MADARA_RPC_EXTERNAL", long, default_value_t = false)]
```
shamsasari: MADARA_RPC_ADMIN

crates/node/src/cli/rpc.rs
```diff
@@ -114,6 +93,10 @@ pub struct RpcParams {
     #[arg(env = "MADARA_RPC_PORT", long, value_name = "PORT", default_value_t = RPC_DEFAULT_PORT)]
     pub rpc_port: u16,
 
+    /// The RPC port to listen at for admin rpc calls.
+    #[arg(env = "MADARA_RPC_PORT_ADMIN", long, value_name = "ENGINE PORT", default_value_t = RPC_DEFAULT_PORT_ADMIN)]
```
shamsasari: ENGINE? PORT should be sufficient

shamsasari (changes requested):

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
Trantorian1: Why? These are not tied by a common spec and we might want to make changes to admin methods which constitute a version change: we cannot do this if we tie admin and other rpc methods under the same semver.

crates/client/rpc/src/lib.rs
Trantorian1: Personal preference I guess. User came more naturally in opposition of 'admin'

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
shamsasari: No, that's not what I meant. The madara binary has a version, and since this is a madara-specific API, it would make sense to have them be the same. Major or breaking changes would reflect naturally with the binary.

Actually, it doesn't look like the binary is versioned. I don't see anything in the `--help` output and there's no `--version` flag. Anyway, _assuming_ the madara node is versioned, it's admin RPC version should be the same. Again, I'm not talking about other RPC methods.

crates/node/src/cli/rpc.rs
Trantorian1: This is already clear in the sentence 'Be very careful however when exposing this endpoint to the outside world' no? The user can do what they want with that afterwards. Plus, as mentioned a proxy might not be necessary if the node is simply running inside a docker container. Imo it's up to the user to decide how they secure their infra, but the fact this is dangerous is clearly stated.

shamsasari (approved):

PR #388, opened 2024-11-19
feat(rpc): v0.8.0 `getMessagesStatus` method

### LINKED ISSUE #414: feat(starknet_getMessageStatus)

### Motivation

### Request

Implement [`starknet_getMessageStatus`](https://github.com/starkware-libs/starknet-specs/blob/c94df2c5866e11c866abd3d234b0d5df681073c3/api/starknet_api_openrpc.json#L252) from rpc v0.8.0-rc0.

> [!NOTE]
> From the specs: _Given an l1 tx hash, returns the associated l1_handler tx hashes and statuses for all L1 -> L2 messages sent by the l1 transaction, ordered by the l1 tx sending orde_.

> [!WARNING]
> Please remember to update the [README](https://github.com/madara-alliance/madara?tab=readme-ov-file#supported-json-rpc-methods) section on supported RPC methods once this has been implemented!

Yes!

shamsasari:
## Pull Request type

- Feature

Resolves: #414 

## What is the new behavior?

- Added v0.8.0 `starket_getMessagesStatus` RPC method
- L1 transaction hash to L1 handler transaction hash mappings is stored in db in a new column.
- The L1 -> L2 processing logic was refactored, with the bulk of the event loop extracted out into a separate method
- This also includes better logging
- The env variable for the `--analytics_collection_endpoint` config was also fixed

New database column added.

## Other information

Since there can be multiple handler transactions per L1 transaction, and because the processing of L1 -> L2 messaging is done on a per log event basis, a composite DB key is used so that new entries for the same L1 tx can be added without having to re-serialize and update the same key. Retrieving the handler tx hashes is done by a prefix scan on the L1 transaction hash which is an efficient operation with RocksDB.

Since rejections are currently not supported, the RPC response doesn't populate the `failure_reason` field.


jbcaron:
⚠️ Note: This PR introduces a breaking change due to the addition of a new column in the database

crates/client/db/src/l1_db.rs
```diff
@@ -128,4 +129,32 @@ impl MadaraBackend {
...
+        Ok(l1_handler_tx_hashes)
+    }
+
+    /// Store mapping from L1 transaction to L1 handler transaction (on the L2). A unique order
+    /// value is required to ensure the handler transactions are retreived in the correct order.
+    pub fn add_l1_handler_tx_hash_mapping(
+        &self,
+        l1_tx_hash: TxHash,
+        l1_handler_tx_hash: Felt,
+        order: u64,
```
jbcaron: Can the index be a type smaller than u64?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -66,114 +68,83 @@ pub async fn sync(
...
+
+    // Check if cancellation was initiated
+    let event_hash = get_l1_to_l2_msg_hash(&event);
+    let cancellation_timestamp = client.get_l1_to_l2_message_cancellations(event_hash).await?;
+    if cancellation_timestamp != Felt::ZERO {
+        tracing::info!("⟠ L1 message was cancelled at timestamp {:?}", cancellation_timestamp.to_biguint());
+        return Ok(());
+    }
+
+    let l1_handler_transaction = parse_handle_l1_message_transaction(&event)?;
```
jbcaron: I wonder if it would be better to use our own primitive, mp_transaction::L1HandlerTransaction, even if it requires integrating starknet-api for blocking purposes.

crates/client/rpc/src/versions/v0_8_0/methods/read/get_messages_status.rs
```diff
@@ -0,0 +1,43 @@
...
+    let l1_handler_tx_hashes = starknet
+        .backend
+        .get_l1_handler_tx_hashes(transaction_hash)
+        .or_internal_server_error("Retrieving L1 handler transactions from database")?;
+    if l1_handler_tx_hashes.is_empty() {
+        return Err(StarknetRpcApiError::TxnHashNotFound);
+    }
+    let mut message_statuses = vec![];
+    for l1_handler_tx_hash in l1_handler_tx_hashes {
+        let finality_status = match get_transaction_status(starknet, l1_handler_tx_hash) {
```
jbcaron: I think we will need to review the get_transaction_status function in sequencer mode, as we need to handle transactions in both the mempool and the pending state

jbcaron (commented):
Does this implementation account for the entire message history on L1, even if the production sequencer has been inactive for an extended period?

antiyro:
> ⚠️ Note: This PR introduces a breaking change due to the addition of a new column in the database

I think that it maybe a good idea to wait for the rolling updates PR from @Trantorian1 to merge this one.

crates/client/db/src/l1_db.rs
shamsasari: The reason it's `u64` is because it's using the log index. Not sure why that needs to be that big, even u32 is more than enough to cover logs in a block I would have thought. I can reduce if you think it's safe.

shamsasari:
> Does this implementation account for the entire message history on L1, even if the production sequencer has been inactive for an extended period?

No it doesn't. @antiyro let me know if it should in this PR.

shamsasari:
> Does this implementation account for the entire message history on L1, even if the production sequencer has been inactive for an extended period?

It picks up from the last log event that was processed (which is stored in the db). So yes, I think it does.

crates/client/db/src/l1_db.rs
```diff
@@ -128,4 +129,32 @@ impl MadaraBackend {
...
         Ok(())
     }
+
+    pub fn get_l1_handler_tx_hashes(&self, l1_tx_hash: TxHash) -> Result<Vec<Felt>, DbError> {
+        let l1_l2_mappings_column = self.db.get_column(Column::L1MessagingHandlerTxHashes);
+        let mut l1_handler_tx_hashes = vec![];
+        for kv_bytes in self.db.prefix_iterator_cf(&l1_l2_mappings_column, l1_tx_hash.as_slice()) {
+            let l1_handler_tx_hash = Felt::from_bytes_be_slice(&kv_bytes?.1);
+            l1_handler_tx_hashes.push(l1_handler_tx_hash);
+        }
```
Trantorian1: Consider replacing this with a `collect` on `self.db.prefix_iterator_cf` to mitigate resizing incurred by `Vec::push`.

crates/client/db/src/l1_db.rs
```diff
@@ -128,4 +129,32 @@ impl MadaraBackend {
...
+        l1_tx_hash: TxHash,
+        l1_handler_tx_hash: Felt,
+        order: u64,
+    ) -> Result<(), DbError> {
+        let l1_l2_mappings_column = self.db.get_column(Column::L1MessagingHandlerTxHashes);
+        let mut key = [0u8; 40];
+        key[..32].copy_from_slice(l1_tx_hash.as_slice());
+        key[32..].copy_from_slice(&order.to_be_bytes()); // BE is important for the lexographic sorting
+        let mut writeopts = WriteOptions::default();
+        writeopts.disable_wal(true);
```
Trantorian1: Shouldn't WAL be enabled so we don't risk loosing this on a node restart?

crates/client/eth/src/l1_messaging.rs
```diff
@@ -67,114 +69,83 @@ pub async fn sync(
...
         Some(Fee(event.fee.try_into()?)),
         None,
         false,
     )?;
-    let res = mempool.accept_l1_handler_tx(blockifier_transaction)?;
+    mempool.accept_l1_handler_tx(blockifier_transaction)?;
+
+    let l1_tx_hash = log.transaction_hash.ok_or_else(|| anyhow!("Missing transaction hash"))?;
+    let block_number = log.block_number.ok_or_else(|| anyhow!("Event missing block number"))?;
+    let log_index = log.log_index.ok_or_else(|| anyhow!("Event missing log index"))?;
```
Trantorian1:
```suggestion
    let l1_tx_hash = log.transaction_hash.ok().context("Missing transaction hash")?;
    let block_number = log.block_number.ok().context("Event missing block number")?;
    let log_index = log.log_index.ok().context("Event missing log index")?;
```

crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_messages_status.rs
```diff
@@ -0,0 +1,43 @@
...
+            Err(e) => return Err(e),
+        };
+        message_statuses.push(MessageStatus {
+            transaction_hash: l1_handler_tx_hash,
+            finality_status,
+            // TODO Update this once get_transaction_status supports rejections
+            failure_reason: None,
+        })
+    }
+    Ok(message_statuses)
```
Trantorian1: Consider using `iter().try_fold(Vec::with_capacity(l1_handler_tx_hashes.len(), ...))`

Trantorian1 (changes requested):
A few changes related to iteration, otherwise all seems good :+1:

crates/client/db/src/l1_db.rs
shamsasari: Storing the mapping is guarded behind `messaging_update_last_synced_l1_block_with_event`, which also has WAL disabled. The sync will restart from the last stored `LastSyncedEventBlock`. So I think that should also be enabled? @cchudant

shamsasari:
PTAL

Trantorian1 (approved):
Seems good!

crates/madara/client/db/src/lib.rs
```diff
@@ -230,6 +232,7 @@ impl Column {
             ContractStorage => "contract_storage",
             L1Messaging => "l1_messaging",
             L1MessagingNonce => "l1_messaging_nonce",
+            L1MessagingHandlerTxHashes => "l1_messaging_handler_tx_hashes",
```
Trantorian1, line 237: This pr needs to update the db version as it introduces a new schema!

crates/madara/client/eth/src/l1_messaging.rs
```diff
@@ -67,105 +70,74 @@ pub async fn sync(
...
+        &Transaction::L1Handler(l1_handler_transaction.clone()),
+        chain_id,
+        &l1_handler_transaction.version,
+    )?;
 
-    // TODO: remove unwraps
-    // Ques: shall it panic if no block number of event_index?
-    let block_sent = LastSyncedEventBlock::new(l1_block_number.unwrap(), event_index.unwrap());
-    backend.messaging_update_last_synced_l1_block_with_event(block_sent)?;
+    mempool.tx_accept_l1_handler(l1_handler_transaction.clone().into(), fees)?;
```
Trantorian1, line 125: We only need `l1_handler_transaction` for the `tracing::info` below, there should be a way to avoid cloning here.

crates/madara/client/mempool/src/lib.rs
```diff
@@ -364,11 +360,14 @@ impl MempoolProvider for Mempool {
...
-        let nonce_target =
-            self.backend.get_l1_messaging_nonce_latest()?.map(|nonce| nonce.try_increment()).unwrap_or(Ok(nonce))?;
+        let nonce_target = self.backend.get_l1_messaging_nonce_latest()?.unwrap_or(nonce);
         let nonce_info = if nonce != nonce_target {
             NonceInfo::pending(nonce, nonce_next)
         } else {
+            let _ = self
+                .backend
+                .set_l1_messaging_nonce(nonce)
+                .map_err(|_err| MempoolError::StorageError(MadaraStorageError::InvalidNonce));
```
Trantorian1: I don't think this should be here. We should leave this kind of logic to the l1 sync.

crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_messages_status.rs
```diff
@@ -0,0 +1,36 @@
...
+    l1_handler_tx_hashes.iter().try_fold(
+        Vec::with_capacity(l1_handler_tx_hashes.len()),
+        |mut acc, l1_handler_tx_hash| {
+            let finality_status = match get_transaction_status(starknet, *l1_handler_tx_hash) {
+                Ok(tx_status) => tx_status.finality_status,
+                Err(StarknetRpcApiError::TxnHashNotFound) => {
+                    tracing::error!("L1 handler tx {l1_handler_tx_hash:?} for L1 tx {transaction_hash:?} not found");
+                    return Err(StarknetRpcApiError::InternalServerError);
+                }
+                Err(e) => return Err(e),
```
Trantorian1: We probably want to log this error and just return an internal server error to the user.

Trantorian1 (changes requested):

Trantorian1 (approved):
Sees good!

jbcaron:
@HermanObst please format TOML files using Taplo v0.9.3.

HermanObst:
> @HermanObst please format TOML files using Taplo v0.9.3.

@jbcaron fixed! Thanks for the review

crates/madara/client/db/src/rocksdb_options.rs
```diff
@@ -56,6 +58,9 @@ impl Column {
                     contract_db::CONTRACT_NONCES_PREFIX_EXTRACTOR,
                 ));
             }
+            Column::L1MessagingHandlerTxHashes => {
+                options.set_prefix_extractor(SliceTransform::create_fixed_prefix(TxHash::LEN));
```
Mohiiit, line 62: given we are adding support for l2<>l3 as well in the madara, do you think it make sense to make it generic?

Mohiiit, line 62: given we are adding support for l2<>l3 as well in the madara, do you think it make sense to make it generic?

crates/madara/client/eth/src/l1_messaging.rs
```diff
@@ -67,105 +70,76 @@ pub async fn sync(
...
-    let transaction = parse_handle_l1_message_transaction(event)?;
-    let tx_nonce = transaction.nonce;
-    let fees: u128 = event.fee.try_into()?;
+    client: &EthereumClient,
+    chain_id: &ChainId,
+    mempool: &Arc<Mempool>,
+    event: LogMessageToL2,
+    log: Log,
+) -> anyhow::Result<()> {
+    tracing::debug!("⟠ Processing L1 -> L2 message event {event:#?}, contract address: {}, block number: {:?}, transaction index: {:?}, transaction hash: {:?}, log index: {:?}",
```
Mohiiit, line 94: nitpick: not sure if this is relevant but this was `info` earlier

Mohiiit (approved):
perfect!!

Trantorian1 (changes requested):
Don't merge there are some bugs related to the handling of l1 handler nonces! See #486

raynaudoe:
HI @Mohiiit, is this PR still relevant?

Mohiiit:
@raynaudoe sorry for the late reply, yes this PR will be relevant given we are updating madara to the latest spec, we would need this method

PR #390, opened 2024-11-20
feat(debug): service cancelling and profiling build

Trantorian1:
# Pull Request type

- Feature

Each capability in Madara starts as a `Service`: currently there is no way for services to communicate shutdown intent between each other. That is to say for example the rpc service can't ask all the other services to gracefully shutdown, and can only manage its own life cycle.

## What is the new behavior?

This pr adds a shared `tokio_util::sync::CancellationToken` to each service in a `ServiceGroup`. This is directly passed to `graceful_shutdown` so that if any service requests a cancellation, other services will also enter shutdown mode.

Besides that, this pr adds a profiling build target for use in benchmarks and performance analysis. This mostly adds debug symbols to a release-equivalent build and uses `symbol-mangling` to demangle them for clearer perf reports.

> [!WARNING]
> This is _not_ an error handling mechanism. The idea is not to allow a service to gracefully shutdown related services in case of an error, which is why `tokio_util::sync::CancellationToken::child_token` is never used. This is only targeted at features which would allow the global halting of Madara from a local service (such as an admin rpc endpoints).

No.


crates/node/src/cli/sync.rs
```diff
@@ -62,6 +62,13 @@ pub struct SyncParams {
     #[clap(env = "MADARA_N_BLOCKS_TO_SYNC", long, value_name = "NUMBER OF BLOCKS")]
     pub n_blocks_to_sync: Option<u64>,
 
+    /// Gracefully shutdown Madara once it has finished to synchronize all
```
shamsasari: Grammer: finished synchronizing all blocks.

crates/node/src/cli/sync.rs
```diff
@@ -62,6 +62,13 @@ pub struct SyncParams {
...
+    /// blocks. This can either be once the node has caught up with the head of
+    /// the chain or when it has synced as many blocks as specified by
+    /// `n-blocks-to-sync`.
+    #[clap(env = "MADARA_STOP_ON_SYNC", long, value_name = "STOP ON SYNC", default_value_t = false)]
```
shamsasari: The value_name isn't used for bool flags - you can see that in the `--help` output.

crates/node/src/cli/sync.rs
```diff
@@ -62,6 +62,13 @@ pub struct SyncParams {
     #[clap(env = "MADARA_N_BLOCKS_TO_SYNC", long, value_name = "NUMBER OF BLOCKS")]
     pub n_blocks_to_sync: Option<u64>,
 
+    /// Gracefully shutdown Madara once it has finished to synchronize all
+    /// blocks. This can either be once the node has caught up with the head of
+    /// the chain or when it has synced as many blocks as specified by
+    /// `n-blocks-to-sync`.
```
shamsasari: The backticks don't do anything and appear in the `--help` output, making it look a bit weird to the user. Change this to `--n-blocks-to-sync`

crates/node/src/service/block_production.rs
```diff
@@ -104,7 +108,7 @@ impl Service for BlockProductionService {
             };
 
             // display devnet welcome message :)
-            // we display it to stdout instead of stderr
+            // we display itto stdout instead of stderr
```
shamsasari: revert

shamsasari (changes requested):

shamsasari (approved):
LGTM

antiyro:
lgtm but I would use `cancellation_token.is_cancelled()` instead of `CTRL_C` to make it cleaner and easier to maintain. If you receive `ctrl_c` and `sigterm` you could use cancelation token instead (completely removing `CTRL_C` if possible). Check https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.is_cancelled

Trantorian1:
> lgtm but I would use `cancellation_token.is_cancelled()` instead of `CTRL_C` to make it cleaner and easier to maintain. If you receive `ctrl_c` and `sigterm` you could use cancelation token instead (completely removing `CTRL_C` if possible). Check https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.is_cancelled

Fixed! We don't even need to use `CancellationToken::is_cancelled` as `wait_or_graceful_shutdown` already races `CancellationToken::cancelled` and the future to be executed, which has the same effect.

antiyro (approved):
lgtm now thanks!

PR #392, opened 2024-11-20
dev docs: add documentation to the rocksdb configuration

### LINKED ISSUE #231: feat(docs): Document rocksdb configuration

cchudant:
# Pull Request type

- (developer) documentation content changes

resolves #231
fixes the outstanding review comments in #379 

cc @shamsasari 

(note:
the thing i mention about snapshots for consistent db reads in the module docs is something i wanted to fix for a very long time, it's not a new concern. i was waiting for #292 to be merged to fix it.
it can't cause any problem in block production mode so that's why it hasn't yet been fixed)


antiyro (approved):
lgtm thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- dev docs: add documentation to the rocksdb configuration
```
shamsasari, line 5: Not related to this PR, but you know something is not quite right when you have to update the changelog for developer comments!

crates/client/db/src/rocksdb_options.rs
```diff
@@ -1,3 +1,50 @@
+//! # Rocksdb options
+//!
+//! We configure the rocksdb database in a very specific way. We want to guarantee a few things:
+//!
+//! ## Fault tolerance
+//!
+//! The madara node can be stopped at any time, and it should be able to recover when restarting.
+//! In particular, in case of a power/hardware failure, the database should not be corrupted beyond repair. We cannot
+//! run code that would allow us to gracefully shutdown in these cases.
```
shamsasari, line 9: > We cannot run code that would allow us to gracefully shutdown in these cases.

Is that correct? I feel like there's a "not" missing, or something.

crates/client/db/src/rocksdb_options.rs
```diff
@@ -1,3 +1,50 @@
...
+//!
+//! ## Consistent view of the database
+//!
+//! The usual way it is achieved is by batching all of the changes into a single batch or a single rocksdb transaction so
+//! that they are applied to the database atomically.
+//!
+//! ## Multithreaded commit to the database
+//!
+//! This is main reason we differ from the default rocksdb configuration: rocksdb Transaction objects are not thread safe,
+//! and committing a transaction also isn't multithreaded at all. This means that if we want to write a whole block into the
```
shamsasari, line 23: > also isn't multithreaded at all

You mean, "isn't also thread-safe"?

crates/client/db/src/rocksdb_options.rs
```diff
@@ -1,3 +1,50 @@
...
+//! - disabling the WAL, as we don't need it with atomic flushing
+//! - disabling auto flushing, so that a block cannot be written into the db half way
+//! - do not use any transactions, do all the writes directly to the database.
+//! - use explicit [Snapshot]s for reads: all reads use the latest snapshot; and between every block we do a new snapshot and drop the old one.
+//!   This ensures we don't read the database when a block has been half-way written
+//!
+//! That last point isn't yet implemented, as we need some changes to the rust rocksdb bindings, see [rust-rocksdb#937]. No issue has yet been
+//! found with this as our db is almost only append-only - but this should nonetheless be fixed. (FIXME)
+//!
+//! This configuration makes a lot of sense because we making a blockchain node. All of our db writes are very big, all or nothing and infrequent.
```
shamsasari, line 39: This configuration makes a lot of sense in the context of a blockchain node.

crates/client/db/src/rocksdb_options.rs
```diff
@@ -13,22 +60,30 @@ pub fn rocksdb_global_options() -> Result<Options> {
...
-    options.increase_parallelism(cores);
-    options.set_max_background_jobs(cores);
 
+    // See module documentation.
     options.set_atomic_flush(true);
-    options.set_max_subcompactions(cores as _);
 
+    // By default rocksdb will spam a lot of info about every column very regularily, making huge files that take
+    // multiple gigabytes. Using log level warn, this info will not be printed to the log file. In addition,
+    // we limit the size and number of files.
```
shamsasari, line 69: I think you intended this below for the `set_log_level`?

crates/client/db/src/rocksdb_options.rs
```diff
@@ -13,22 +60,30 @@ pub fn rocksdb_global_options() -> Result<Options> {
...
+    options.set_max_open_files(2048);
+
+    // Concurrency options
+    let cores = std::thread::available_parallelism().map(|e| e.get() as i32).unwrap_or(1);
+    options.increase_parallelism(cores);
+    options.set_max_background_jobs(cores);
+    options.set_max_subcompactions(cores as _);
     let mut env = Env::new().context("Creating rocksdb env")?;
-    // env.set_high_priority_background_threads(cores); // flushes
+    // env.set_high_priority_background_threads(cores); // flushes - our flushes are manual so this option is not useful.
```
shamsasari, line 85: So can you just remove the commented out code?

shamsasari (changes requested):

CHANGELOG.md
cchudant: that's right!
ive said it offline plenty of times: if the changelog is a copy of the git log, then the changelog has absolutely no purpose

the changelog should be user-facing and non-verbose with migration notes and plenty of details for all the new features. It's probably best written by looking at the git log just before a release

The [bevy changelog](https://bevyengine.org/news/bevy-0-14/) is a very good example of a very well made changelog - imo, it's actually the best example ive ever seen of a changelog, it's always a delightful read and probably gargantuan task to write it.

we don't do releases that often yet though, and we are not the same kind of software as bevy, we don't expect the same kind of audience - i would be in favor of just removing that changelog file and the github action for now, and probably do a simple feature-by-feature user facing changelog for real releases.

crates/client/db/src/rocksdb_options.rs
cchudant: unsure if my grammar is correct but it looks correct to me: if there's a power failure, or the node is killed with SIGKILL for example, there is no graceful shutdown - the program just abruptly stops

in WAL mode, rocksdb will look at the WAL on startup and fix the potential corruption (ie. the node stopped during a write)
in case of a graceful shutdown (SIGTERM/SIGINT/...) the rocksdb `DB` object gets dropped, and in that destructor the db is flushed and there is no need to recover from WAL the next time on startup

crates/client/db/src/rocksdb_options.rs
cchudant: i actually mean the commit is done in a single-threaded fashion

Well, in a write-prepared transaction (the default), all of the changes are buffered in memory and then calling commit() on the transaction will apply them using the current thread only. this is what I mean: the commit is not multithreaded (which makes sense - it would be weird for rocksdb to start using all of the threads when doing commit() on a transaction)
In a write-unprepared transaction, all of the changes are applied directly to the database instead of being buffered. The commit will just update the latest sequence number that you read from on the database if i understand correctly?
This is very close to what we need actually! but we're still hitting a problem, which is that the transaction object is not thread-safe (here I'm talking about the struct being `Sync`) - also it uses &mut self for all of its operations. That might be a problem of the rust `rocksdb` crate though, but looking at the official documentation https://github.com/facebook/rocksdb/wiki/WriteUnprepared-Transactions it doesn't look like they say it's thread safe on the C++ side as well. There's this "Key Locking" section on that page that mention threads and locking but I'm not sure I 100% understand if it's relevant or not. I haven't looked at that part of the rocksdb C++ code.

Hope that clears up this section, rocksdb is sometimes a bit confusing. I'll try to be a bit more clear about that, thank you!

crates/client/db/src/rocksdb_options.rs
cchudant: i mean that for the 3 following settings, but i'll swap the lines so that the log level warn is just below this comment and the other two (which is what this "In addition" refers to) follow just after that

crates/client/db/src/rocksdb_options.rs
cchudant: thank you very much

CHANGELOG.md
shamsasari: I 100% second this!

crates/client/db/src/rocksdb_options.rs
```diff
@@ -1,3 +1,50 @@
...
+//! The way we work around that is by:
+//! - enabling atomic flushing
+//! - disabling the WAL, as we don't need it with atomic flushing
+//! - disabling auto flushing, so that a block cannot be written into the db half way
+//! - do not use any transactions, do all the writes directly to the database.
+//! - use explicit [Snapshot]s for reads: all reads use the latest snapshot; and between every block we do a new snapshot and drop the old one.
+//!   This ensures we don't read the database when a block has been half-way written
+//!
+//! That last point isn't yet implemented, as we need some changes to the rust rocksdb bindings, see [rust-rocksdb#937]. No issue has yet been
+//! found with this as our db is almost only append-only - but this should nonetheless be fixed. (FIXME)
```
Trantorian1, line 37: "is almost exclusively append-only"

crates/client/db/src/rocksdb_options.rs
```diff
@@ -13,22 +60,30 @@ pub fn rocksdb_global_options() -> Result<Options> {
...
 
+    // By default rocksdb will spam a lot of info about every column very regularily, making huge files that take
+    // multiple gigabytes. Using log level warn, this info will not be printed to the log file. In addition,
+    // we limit the size and number of files.
     options.set_max_log_file_size(10 * MiB);
-    options.set_max_open_files(2048);
     options.set_keep_log_file_num(3);
     options.set_log_level(rocksdb::LogLevel::Warn);
 
+    // Max number of open files for rocksdb. This number has been chosen a bit arbitrarily, but it should low enough
```
Trantorian1, line 74: "it should be low enough"

Trantorian1 (changes requested):
A few spelling mistakes. Also some of @shamsasari's RFCs have not been addressed yet.

HermanObst:
@cchudant could you address RFC in this PR?

raynaudoe:
Hi @cchudant , are there any pending RFCs in this PR ?

PR #393, opened 2024-11-21
feat(warp): added `warp update` to madara and other changes

Trantorian1:
# Pull Request type

- Feature

There currently is no way to migrate the Madara db on a breaking change. This pr fixes this, adding some more quality of life features in the process.

## What is the new behavior?

- Added warp update, which allows Madara to migrate its database at speeds much faster than a re-sync from genesis. This works by running a local fgw node and a second node which syncs from it with higher than normal parallelization. This has been documented in the `README`, where you can find more detailed information on how to test this feature.

- Added cli parameter presets. This is quite primitive atm, and only functions by overriding other cli args _after_ they have been parsed. This sadly means the user cannot currently override options set in the presets (also this is all very manual). Imo it is still a good start though in making some common args configuration such as rpc, gateway (or just lengthy ones like warp sync) easier to use.

- Db flush no longer occurs on a fixed-interval timer of 5s. Instead, `maybe_flush` has been renamed to `flush` and is called every `n` blocks in the l2 sync block storage process. This can be set through the cli arg `--flush-every-n-blocks` and defaults to 1000.

- Added the option to increase sync parallelism. This is used by warp sync and is available under the `--sync-parallelism` cli flag, which defaults to 10 (meaning 10 blocks will be fetched in parallel during l2 sync).

### Some important changes have also been made to the rpc admin endpoints:

- Added a new admin endpoint, `stop_node`, which can be used to stop madara at a distance. This is used by warp update to stop the warp update sender once the receiver has finished synchronizing.

- Added a new admin endpoint, `ping`, which just returns the unix time at which it was received. This can be used to check if a node is active or calculate network latency.

- Added various new admin endpoints which allow to control the capabilities of the node (for example stopping / restarting sync). This is quite basic atm and has only been implemented for sync and user-facing rpc services. It is sadly not possible to start services which were not available on startup atm.

> [!NOTE]
> There are currently two main issues I have with the current service architecture:
> 1. It is not possible to start a service later on in the execution of the node
> 2. Interruption with graceful shutdown can occur at any time in the execution. Other than that, we have no hook to simply pause the execution of a service.
> 
> I am thinking of how to best approach this. Currently, I am favoring a monitor solution where `ServiceGroupd` also keeps in memory the necessary information to start (and restart) a service. There also needs to be some mechanism for the `ServiceGroup` to be able to cleanly pause the execution of a service but I am not sure of the best way to go about this yet.

### On the rpc front:

- Updated the rpc versioning middleware to accept rpc methods sent with jsonrpsee. This is because (due to our versioning scheme) jsonrpsee will send requests of the form `namespace_version_method`, whereas our version middleware only accepts request of the form `namespace_method`, with `version` being specified in the header. This has been changed to allow for both formats. This is only useful for us to be able to make rpc calls from one node to the other, but the alternative would have been to do this manually. I think this is a better solution as jsonrpsee provides type checking and function signatures for each of our methods, which should reduce errors.

### Refactor

- Finally, the various methods in the l2 sync section of madara have been refactored to be a bit more readable.

No.


crates/client/sync/src/fetch/mod.rs
```diff
@@ -13,21 +14,50 @@ use crate::fetch::fetchers::fetch_block_and_updates;
...
+        sync_polling_interval,
+        n_blocks_to_sync,
+        stop_on_sync,
+        sync_parallelism,
+        warp_update,
+    } = config;
+
+    if warp_update {
+        let ping = jsonrpsee::http_client::HttpClientBuilder::default()
+            .build("http://localhost:9943")
```
antiyro, line 50: what happens if this port is not available? users cant migrate?

antiyro: what happens if this port is not available? users cant migrate?

antiyro (commented):
Tested locally and it works as expected. Just wondering why are we disabling root? for perfs? Also could we minimize downtime here?

antiyro:
You must force commitment computation or it will result in an invalid state root on sequencer b since the bonsai db is not migrated

antiyro:
Minimizing downtime, I mean:

1. We kill node A to restart it with `--warp-update-sender` -> downtime (I don't see how to avoid this).  
2. We stop the sync of node A -> downtime (maybe we can continue syncing, and once node B catches up, stop syncing A and finish syncing B).  
3. We kill node B after the migration -> downtime (perhaps we can continue syncing with the additional arguments, including `--warp-update-receiver`).

Trantorian1:
> Minimizing downtime, I mean:
> 
> 1. We kill node A to restart it with `--warp-update-sender` -> downtime (I don't see how to avoid this).
> 2. We stop the sync of node A -> downtime (maybe we can continue syncing, and once node B catches up, stop syncing A and finish syncing B).
> 3. We kill node B after the migration -> downtime (perhaps we can continue syncing with the additional arguments, including `--warp-update-receiver`).

Going to be working on a few fixes to this, moving the pr to draft in the meantime.

crates/client/sync/src/fetch/mod.rs
Trantorian1: This has been fixed by allowing the user to specify a fgw and admin rpc port during the warp update

crates/client/db/src/lib.rs
```diff
@@ -384,30 +384,21 @@ impl MadaraBackend {
...
+        let mut opts = FlushOptions::default();
+        opts.set_wait(true);
+        // we have to collect twice here :/
+        let columns = Column::ALL.iter().map(|e| self.db.get_column(*e)).collect::<Vec<_>>();
+        let columns = columns.iter().collect::<Vec<_>>();
 
-        Ok(will_flush)
+        self.db.flush_cfs_opt(&columns, &opts).context("Flushing database")?;
+
+        Ok(())
```
antiyro, line 397: Not mandatory now but we should add metrics to follow the flush frequency and behavior

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -36,12 +36,12 @@ pub async fn gas_price_worker(
     eth_client: &EthereumClient,
     l1_gas_provider: GasPriceProvider,
     gas_price_poll_ms: Duration,
-    cancellation_token: tokio_util::sync::CancellationToken,
+    ctx: ServiceContext,
```
antiyro, line 39: what does ctx stands for? could we use another name to make it clearer

antiyro:
Capture d’écran 2024-11-27 à 10 02 13

I get infinite freeze when running with:  `--full --network mainnet --l1-endpoint <> --base-path /tmp/madara-b --warp-update-receiver``

crates/client/eth/src/l1_gas_price.rs
Trantorian1: ctx -> context. Idk, seems pretty clear and is easier to handle than having to write `context` everywhere. Lemme know what you think.

Trantorian1:
> Capture d’écran 2024-11-27 à 10 02 13
> I get infinite freeze when running with: `--full --network mainnet --l1-endpoint <> --base-path /tmp/madara-b --warp-update-receiver``

Will be checking that out

crates/client/eth/src/l1_gas_price.rs
antiyro: yes that's fine saw its usage later on

jbcaron (commented):
Why switch Admin methods to release?

You can simplify the fetch task handling by using the run_until_cancelled method provided by tokio_util::sync::CancellationToken. This will cleanly stop the tasks without the need for additional logic.

antiyro:
Capture d’écran 2024-11-27 à 14 20 03

problems when sending sigint to the sync. sometimes not detected or no gracefull shutdown with segfault

jbcaron (approved):

Trantorian1:
> Capture d’écran 2024-11-27 à 14 20 03
> problems when sending sigint to the sync. sometimes not detected or no gracefull shutdown with segfault

This has been fixed

cchudant:
> I get infinite freeze when running with: `--full --network mainnet --l1-endpoint <> --base-path /tmp/madara-b --warp-update-receiver``

was this caught by the ci? if not, what was wrong here and why did the e2e pass? the idea of such a big problem not getting caught by the ci makes me worried

Trantorian1:
> > I get infinite freeze when running with: `--full --network mainnet --l1-endpoint <> --base-path /tmp/madara-b --warp-update-receiver``
> 
> was this caught by the ci? if not, what was wrong here and why did the e2e pass? the idea of such a big problem not getting caught by the ci makes me worried

This was in fact not an infinite freeze but just tracing being disabled after #401

README.md
```diff
@@ -486,7 +487,50 @@ Madara comes packed with OTEL integration, supporting export of traces, metrics
 - Signoz Docker Standalone can be setup following this [guide](https://signoz.io/docs/install/docker/).
 - Ensure to configure the correct service_name after importing the json to the dashboard.
 
-## ✔ Supported Features
+## 🔄 Migration
+
+When migration to a newer version of Madara which introduces breaking changes,
```
cchudant: When __migrating__ to a newer version of Madara [that introduces breaking changes]
i would probably simply remove the later part of the sentence tbh, people are not expected to migrate to new versions that dont require upgrading the db?

crates/client/block_import/src/lib.rs
```diff
@@ -176,11 +170,6 @@ impl BlockImporter {
         validation: BlockValidationContext,
     ) -> Result<BlockImportResult, BlockImportError> {
         let result = self.verify_apply.verify_apply(block, validation).await?;
-        // Flush step.
-        let force = self.always_force_flush;
```
cchudant, line 180: aaah!! no, we really need to flush every block, **always**, in block_production mode

crates/client/gateway/server/src/handler.rs
```diff
@@ -238,7 +239,13 @@ pub async fn handle_get_block_traces(
         traces: Vec<TransactionTraceWithHash>,
     }
 
-    let traces = v0_7_1_trace_block_transactions(&Starknet::new(backend, add_transaction_provider), block_id).await?;
+    // TODO: we should probably use the actual service context here instead of
+    // creating a new one!
```
cchudant, line 243: we should probably not implement trace block like this at all actually, i dont see why gateway would depend on rpc

cchudant: this was a hack afaik

crates/client/mempool/src/block_production.rs
```diff
@@ -418,9 +419,7 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         // todo, prefer using the block import pipeline?
         self.backend.store_block(self.block.clone().into(), state_diff, self.declared_classes.clone())?;
         // do not forget to flush :)
-        self.backend
-            .maybe_flush(true)
-            .map_err(|err| BlockImportError::Internal(format!("DB flushing error: {err:#}").into()))?;
+        self.backend.flush().map_err(|err| BlockImportError::Internal(format!("DB flushing error: {err:#}").into()))?;
```
cchudant, line 422: so, just a note here
flushing is the responsability of the block import crate, this flushing here is a hack for pending blocks.
i'd like the flushing to be moved back into block import and be removed from sync l2.rs once again

cchudant: you can keep this hack here of course since it was already present

crates/client/rpc/src/lib.rs
```diff
@@ -26,15 +27,29 @@ use providers::AddTransactionProvider;
...
     pub(crate) add_transaction_provider: Arc<dyn AddTransactionProvider>,
+    pub ctx: ServiceContext,
+}
+
+impl Clone for Starknet {
+    fn clone(&self) -> Self {
+        Self {
+            backend: Arc::clone(&self.backend),
+            add_transaction_provider: Arc::clone(&self.add_transaction_provider),
+            ctx: self.ctx.branch(),
```
cchudant: does that mean that cloning a starknet instance will make it a parent to a newly created child?
this feels weird.. i wouldnt expect Clone to do stuff like that

how about just wrapping ServiceContext into an Arc so that cloning a Starknet actually shares the cancellation token instead of forking it

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
```diff
@@ -14,3 +14,32 @@ pub trait MadaraWriteRpcApi {
...
+#[versioned_rpc("V0_1_0", "madara")]
+pub trait MadaraStatusRpcApi {
+    #[method(name = "ping")]
+    async fn ping(&self) -> RpcResult<u64>;
+
+    #[method(name = "shutdown")]
+    async fn shutdown(&self) -> RpcResult<u64>;
+
+    #[subscription(name = "pulse", unsubscribe = "unsubscribe", item = u64)]
+    async fn pulse(&self) -> jsonrpsee::core::SubscriptionResult;
```
cchudant, line 42: can you add some doc about what that is

crates/primitives/utils/src/service.rs
```diff
@@ -1,21 +1,297 @@
...
+            || !self.capabilities.is_active(self.id as u8)
+            || self.state() == MadaraState::Shutdown
+    }
+
+    pub fn id(&self) -> MadaraCapability {
+        self.id
+    }
+
+    /// Copies the context, maintaining its scope.
+    pub fn branch(&self) -> Self {
```
cchudant: ah why is it called branch, could you make it a Clone impl then?

crates/primitives/utils/src/service.rs
```diff
@@ -1,21 +1,297 @@
...
+            token_local: self.token_local.clone(),
+            capabilities: Arc::clone(&self.capabilities),
+            capabilities_notify: Arc::clone(&self.capabilities_notify),
+            state: Arc::clone(&self.state),
+            id: self.id,
+        }
+    }
+
+    /// Copies the context, maintaining its scope but with a new id.
+    pub fn branch_id(&self, id: MadaraCapability) -> Self {
```
cchudant: could then be .clone().with_id(id: xx)

crates/primitives/utils/src/service.rs
```diff
@@ -1,21 +1,297 @@
...
+pub struct ServiceContext {
+    token_global: tokio_util::sync::CancellationToken,
+    token_local: Option<tokio_util::sync::CancellationToken>,
+    capabilities: Arc<MadaraCapabilitiesMask>,
+    capabilities_notify: Arc<tokio::sync::Notify>,
+    state: Arc<std::sync::atomic::AtomicU8>,
+    id: MadaraCapability,
+}
+
+impl ServiceContext {
```
cchudant, line 140: i like this struct

crates/client/db/src/lib.rs
```diff
@@ -305,7 +303,11 @@ impl DatabaseService {
     }
 }
 
-impl Service for DatabaseService {}
+impl Service for DatabaseService {
+    fn id(&self) -> MadaraCapability {
+        MadaraCapability::Database
```
cchudant: cancellation needs to affect the backup thread if backups are enabled

crates/primitives/utils/src/service.rs
```diff
@@ -1,21 +1,297 @@
...
+    #[inline(always)]
+    pub fn capabilities_add(&self, cap: MadaraCapability) -> bool {
+        let res = self.capabilities.activate(cap);
+        self.capabilities_notify.notify_waiters();
+
+        res
+    }
+
+    #[inline(always)]
+    pub fn capabilities_remove(&self, cap: MadaraCapability) -> bool {
```
cchudant: i dont like that name, it doesnt convey that this disables a service

crates/primitives/utils/src/service.rs
```diff
@@ -1,21 +1,297 @@
...
 
 use anyhow::Context;
-use std::panic;
+use futures::Future;
+use std::{fmt::Display, panic, sync::Arc};
 use tokio::task::JoinSet;
 
+#[repr(u8)]
+#[derive(Clone, Copy, PartialEq, Eq, Default, Debug)]
+pub enum MadaraCapability {
```
cchudant: i kinda dont like the name capability here
it's a just a running service right? why capability?

telemetry doesn't *provide* anything for example, i wouldn't call that a capability

crates/client/rpc/src/versions/admin/v0_1_0/mod.rs
```diff
@@ -1,4 +1,10 @@
 pub mod api;
 pub mod methods;
 
+use std::time::SystemTime;
+
 pub use api::*;
+
+fn unix_now() -> u64 {
```
cchudant: why is it here

crates/client/sync/src/fetch/fetchers.rs
```diff
@@ -40,15 +41,25 @@ pub struct FetchConfig {
     pub sync_polling_interval: Option<Duration>,
     /// Number of blocks to sync (for testing purposes).
     pub n_blocks_to_sync: Option<u64>,
+    /// Number of blocks between db flushes
+    pub flush_every_n_blocks: u16,
```
cchudant: why u16? block numbers are u64 and so are the spans between blocks
doesnt this add a lot of unnecessary conversions?

crates/client/sync/src/fetch/fetchers.rs
```diff
@@ -40,15 +41,25 @@ pub struct FetchConfig {
...
     /// Stops the node once all blocks have been synced (for testing purposes)
     pub stop_on_sync: bool,
+    /// Number of blocks to fetch in parallel during the sync process
+    pub sync_parallelism: u8,
+    /// True if the node is called with `--warp-update-receiver`
+    pub warp_update: bool,
+    /// The port used for nodes to make rpc calls during a warp update.
+    pub warp_update_port_rpc: u16,
```
cchudant, line 55: for info, ports are u16 because at the os level they are u16 - the os accepts an u16

crates/client/sync/src/fetch/mod.rs
```diff
@@ -1,96 +1,111 @@
...
+use mp_utils::{channel_wait_or_graceful_shutdown, service::ServiceContext, wait_or_graceful_shutdown};
 use tokio::sync::{mpsc, oneshot};
+use url::Url;
 
 use crate::fetch::fetchers::fetch_block_and_updates;
 
 pub mod fetchers;
 
-#[allow(clippy::too_many_arguments)]
+pub struct L2FetchConfig {
```
cchudant, line 18: thankyouverymuch :)

crates/client/sync/src/fetch/mod.rs
```diff
@@ -1,96 +1,111 @@
...
-                }
-                val => {
-                    if fetch_stream_sender.send(val?).await.is_err() {
-                        // join error
-                        break;
-                    }
-                }
-            }
+        if client.ping().await.is_err() {
+            tracing::error!("❗ Failed to connect to warp update sender on http://localhost:{warp_update_port_rpc}");
```
cchudant, line 48: shouldn't this be an anyhow::bail?

crates/client/sync/src/fetch/mod.rs
```diff
@@ -1,96 +1,111 @@
...
+            .unwrap_or(NonZeroUsize::new(1usize).expect("1 should always be in usize bound"));
+        config.sync_parallelism = Into::<usize>::into(available_parallelism) * 2;
+
+        let next_block = match sync_blocks(backend.as_ref(), &provider, &ctx, &config).await? {
+            SyncStatus::Full(next_block) => next_block,
+            SyncStatus::UpTo(next_block) => next_block,
+        };
+
+        if client.shutdown().await.is_err() {
+            tracing::error!("❗ Failed to shutdown warp update sender");
```
cchudant, line 71: isn't this an unrecoverable error?

crates/client/sync/src/fetch/mod.rs
```diff
@@ -112,6 +127,57 @@ pub async fn l2_fetch_task(
     Ok(())
 }
 
+enum SyncStatus {
+    Full(u64),
+    UpTo(u64),
+}
+
+async fn sync_blocks(
```
cchudant, line 154: can you add some docs here?

cchudant: im unsure where the parallel catching up vs polling phase are here

cchudant: i think it's SyncStatus that i don't understand

crates/client/sync/src/l2.rs
```diff
@@ -44,22 +46,42 @@ pub struct L2StateUpdate {
...
+        stop_on_sync,
+        telemetry,
+        validation,
+        mut block_conv_receiver,
+    } = config;
+
+    while let Some(block) = channel_wait_or_graceful_shutdown(pin!(block_conv_receiver.recv()), &ctx).await {
         let BlockImportResult { header, block_hash } = block_import.verify_apply(block, validation.clone()).await?;
 
+        if header.block_number - last_block_n >= flush_every_n_blocks as u64 {
```
cchudant: re: i really think this should not be here, flushing should def not be the concern of l2.rs

crates/node/src/cli/mod.rs
```diff
@@ -24,6 +24,60 @@ use mp_chain_config::ChainConfig;
...
 
+#[derive(Clone, Debug, clap::Parser)]
+#[clap(
+    group(
+        ArgGroup::new("args-preset")
+            .args(&["warp_update_sender", "warp_update_receiver", "gateway", "rpc"])
+            .multiple(false)
+    )
+)]
+pub struct ArgsPresetParams {
```
cchudant, line 70: what's this preset thing and how does it relate to the older presets feature

crates/node/src/main.rs
```diff
@@ -201,7 +197,7 @@ async fn main() -> anyhow::Result<()> {
 
     app.start_and_drive_to_end().await?;
 
-    tracing::info!("Shutting down analytics");
+    tracing::info!("🔌 Shutting down analytics...");
```
cchudant: can you remove this log entirely, it's actually not useful tbh

cchudant (changes requested):
so, I'm not sure i understand where the difference between what's checked during regular block import and what's checked when importing a block in warp updates is

are the block hashes and stuff checked in warp mode too?

crates/client/block_import/src/lib.rs
Trantorian1: This has no importance anymore as `maybe_flush` has been removed and will _always_ flush anyways, so `always_force_flush` is no longer needed. This is after I noticed the only part of the code we were performing this check was in the block import, so instead we just do the check there while other parts of the code just call `flush` directly.

crates/client/rpc/src/lib.rs
Trantorian1: Hum. `branch` only clones the `CancellationToken` though, which is itself behind an `Arc`, so I don't think this is necessary.

```rust
pub struct CancellationToken {
    inner: Arc<tree_node::TreeNode>,
}
```

Trantorian1: Also `ServiceContext::branch` does not create any child `CancellationToken`, that is handled by `ServiceContext::branch_id`

crates/primitives/utils/src/service.rs
Trantorian1: Yea this seems like it would make more sense, will do :+1:

crates/primitives/utils/src/service.rs
Trantorian1: This seems like a good idea, though we are creating an extra `CancellationToken` which will be dropped immediately.

crates/primitives/utils/src/service.rs
Trantorian1: :D

crates/client/db/src/lib.rs
Trantorian1: Gotcha!

crates/primitives/utils/src/service.rs
Trantorian1: `service_disable`?

crates/client/rpc/src/versions/admin/v0_1_0/mod.rs
Trantorian1: I think I hoisted this after some changes to the admin rpcs to make it available across multiple files, but I don't think that's necessary anymore.

crates/client/sync/src/fetch/fetchers.rs
Trantorian1: I kind of wanted to restrict the max interval a user could set but I think there is a better way to do this by setting this to a `u64` and updating the clap parser which is used to retrieve this value.

crates/client/sync/src/fetch/mod.rs
Trantorian1: :D

crates/client/sync/src/fetch/mod.rs
Trantorian1: I personally prefer this as it more clearly conveys the error to the node operator.

Trantorian1: Also, if I remember correctly, having this error will eventually cause the entire node to error, and I don't think that is very clean.

crates/client/sync/src/fetch/mod.rs
Trantorian1: Yep, will make it one :+1:

crates/client/sync/src/fetch/mod.rs
Trantorian1: `SyncStatus` is just used to express if we have reached the tip of the chain. This is significant since we are actually synchronizing from _two_ chains during warp updates:

1. The warp update sender
2. Whatever feeder gateway we have set with `--gateway-url`

The issue is that `--n-blocks-to-sync` will cause sync to finish early, so we need to be able to know if we have reached the tip of the chain or else we do not display '🥳 The sync process has caught up with the tip of the chain'.

Trantorian1: `sync_n_blocks` is actually _only_ used during parallel fetching: this includes warp updates and normal parallel sync, and does not include post-sync catch up. So `sync_n_blocks` is not called for the polling phase.

crates/client/sync/src/l2.rs
Trantorian1: Where do you suggest we move this instead? This was already called indirectly in the `verify_apply` method if I remember correctly. The issue is that if we are running a full node with only sync services enabled we still need to have  a point where new blocks are flushed. Also I don't quite see how we can share information across services in an elegant way. Imo it makes most sense to call `flush` were information about the block store is directly available, be that `sync`, `block_production` or the `mempool`.

crates/node/src/cli/mod.rs
Trantorian1: So. Basically I realized some configs were starting to take up a lot of arguments. Initially, warp updates were just set up as a series of predefined arguments (this is still the case for `--warp-update-sender`), and it seemed really tedious and error-prone for the user to have to remember so many options. All this does is that it sets various options in the `RunCmd` struct _after_ parsing has occurred. This has its downsides, most notably the user cannot override args presets this way as they are evaluated _after_ user inputs. The advantage is that this streamlines more complex cli setups with many options and flags (this mostly applies to `--warp-update-sender`).

Trantorian1: Will be adding this in a doc comment

Trantorian1:
> so, I'm not sure i understand where the difference between what's checked during regular block import and what's checked when importing a block in warp updates is
> 
> are the block hashes and stuff checked in warp mode too?

Yes, the block hashes and state root are checked. This is a bit of a shame, as there clearly are lots of performance gains to be had in performing a single state root computation since we know we are synchronizing from a trusted source. This was not implemented as it felt like this was going a bit out of scope (at that point I would be implementing warp sync as well). Still, we are nonetheless cutting down on network latency and increasing the number of block synchronized by maximizing CPU parallelism. Some initial benchmarks have shown a 33% percent improvement in sync time through this alone, so I think this is good enough for now.

In the future, we have some low hanging fruits left which could make this a _lot_ faster, mainly computing the state root in a single go from genesis to full sync. But for now I think it'd be better to move on to other features, most importantly RPC v0.13.3, especially since this feature has already been divided between this and another upcoming pr.

crates/client/rpc/src/lib.rs
cchudant: yes i did not realize branch was clone sorry

crates/client/sync/src/fetch/mod.rs
cchudant: isnt that what we want here?

crates/client/sync/src/l2.rs
cchudant: i think it makes more sense to call it in block_import? this seems way too low level to be a concern of l2.rs

cchudant: what do you mean sharing info in this case?

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
Trantorian1: This has been done. Also took the time to move all the other admin method docs from their implementation to their definition.

crates/primitives/utils/src/service.rs
Trantorian1: This has been done

crates/primitives/utils/src/service.rs
Trantorian1: This has been done. Also took the time to update `SerciceContext::child` to work the same way.

crates/client/rpc/src/lib.rs
Trantorian1: No worries, I've update it to `clone` to make this more obvious :)

crates/client/db/src/lib.rs
Trantorian1: This is actually not necessary as `spawn_backup_db_task` will run as long as ` while let Some(BackupRequest { callback, db }) = recv.blocking_recv()` is true, where `recv` is the receiver to the `backup_handle` which is stored in `MadaraBackend`, meaning the channel will automatically close when `MadaraBackend` is dropped, resulting in the backup thread being driven to completion. 

This makes sense since even before adding `ServiceContext` we were not checking for graceful shutdown in the backup thread so there had to be another for this to occurr. Also, it would be quite complicated to get this to work with the current service architecture, as `ServiceContext` has to be retrieved from `Service::start` to be shared across services. However, `DatabaseService` does not follow this rule and starts its service on `DatabaseService::new`, which is a bit of a problem here.

crates/primitives/utils/src/service.rs
Trantorian1: This has been fixed

crates/primitives/utils/src/service.rs
Trantorian1: This has been fixed. Renamed this to `MadaraService` and made related changes across the codebase.

crates/client/rpc/src/versions/admin/v0_1_0/mod.rs
Trantorian1: This has been fixed. Pulled down to the relevant file.

crates/client/sync/src/fetch/mod.rs
Trantorian1: I thought it would be cleaner for the end user to just have an error message pop up and node gracefully shutdown. From a purely UX standpoint I think this is clearer than being met with a trace log in this case, as all this means is that the user need set up their node with `--warp-update-sender`

crates/client/sync/src/fetch/fetchers.rs
Trantorian1: This has been fixed. Set the value to `u64` and the legal value as 10 000

crates/client/sync/src/fetch/mod.rs
Trantorian1: This has been updated to behave the same as `if client.ping().await.is_err()`.

crates/client/sync/src/fetch/mod.rs
Trantorian1: This has been fixed. Added documentation for `sync_n_blocks` and `SyncStatus`

crates/node/src/main.rs
Trantorian1: This has been fixed.

README.md
Trantorian1: This has been fixed.

crates/client/sync/src/l2.rs
Trantorian1: Tangentially related but I have added a timer inside the l2 `verify_apply_task` which will flush blocks after a set interval of time has elapsed as you had suggested orally. This can be controlled via the `--sync-every-n-seconds` cli arg. So now we can be certain blocks will be flushed to the db either every n seconds or every n blocks, whichever one is faster.

crates/node/src/cli/mod.rs
Trantorian1: This has been fixed. Added some documentation to `ArgsPresetParams` which detail this, as well as some of the issues and limitation with the current cli setup.

crates/client/sync/src/fetch/mod.rs
```diff
@@ -112,6 +129,74 @@ pub async fn l2_fetch_task(
     Ok(())
 }
 
+/// Whether a chain has been caught up to the tip or only a certain block number
+///
+/// This is mostly relevant in the context of the `--n-blocks-to-sync` cli
+/// argument which states that a node might stop synchronizing before it has
+/// caught up with the tip of the chain.
+enum SyncStatus {
```
cchudant, line 137: mmh.. maybe rename it to SyncedStatus or SyncStopReason or idk?

crates/node/src/cli/mod.rs
cchudant: yea i this seems weird, but i think it's fine to merge that as is for now

crates/node/src/cli/sync.rs
```diff
@@ -66,16 +77,73 @@ pub struct SyncParams {
     /// blocks. This can either be once the node has caught up with the head of
     /// the chain or when it has synced as many blocks as specified by
     /// --n-blocks-to-sync.
-    #[clap(env = "MADARA_STOP_ON_SYNC", long, default_value_t = false)]
+    #[clap(env = "MADARA_STOP_ON_SYNC", long, value_name = "STOP ON SYNC", default_value_t = false)]
```
cchudant: adding value_name to a bool seems weird?

crates/node/src/cli/sync.rs
```diff
@@ -66,16 +77,73 @@ pub struct SyncParams {
...
+    /// does not affect the node once it has reached the tip of the chain.
+    /// Increasing this can lead to lower sync times at the cost of higher cpu
+    /// and ram utilization.
+    #[clap(
+        env = "MADARA_SYNC_PARALLELISM",
+        long, value_name = "SYNC PARALLELISM",
+        default_value_t = 10,
+        value_parser = clap::value_parser!(u8).range(1..)
+    )]
+    pub sync_parallelism: u8,
```
cchudant, line 137: why u8?

cchudant (changes requested):
you need to add a db flush in close_block.rs just after verify_apply

crates/node/src/cli/sync.rs
Trantorian1: This seemed like a reasonable bound for parallelism, I don't think anyone will be running Madara with `sync_parallelism` > 255 anytime soon.

crates/node/src/cli/sync.rs
Trantorian1: Yep this makes sense. Removed.

Trantorian1:
> you need to add a db flush in close_block.rs just after verify_apply

This has been fixed. Added a call to `backend.flus()` in `BlockProductionTask::<Mempool>::on_block_time`, which is the only place where `close_block` from `close_block.rs` is called. This mirrors the `flush` in `on_pending_time_tick`.

cchudant (approved):

PR #394, opened 2024-11-22
fix: fetch class

jbcaron:
## Pull Request type

- Bugfix

the class 0x371b5f7c5517d84205365a87f02dcef230efa7b4dd91a9e4ba7e04c5b69d69b has an empty ABI, which causes a deserialization error when fetching from the sequencer.

Resolves: #NA

## What is the new behavior?

- ensured compatibility with the deserialization process when fetching data from the sequencer
- added necessary tests to verify the fix and prevent regression

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix(fgw): fetch class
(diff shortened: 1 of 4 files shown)


Trantorian1 (approved):
All good :+1:

Mohiiit (approved):
lgtm

PR #397, opened 2024-11-26
Fix/rust log filtering

heemankv:
Filtering based on RUST_LOG stopped working after adding log-level as a clap argument.
removed log_level from clap to get it and added code to pick `RUST_LOG` from env directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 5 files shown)


heemankv (commented):
Self review completed, looks fine to me

Trantorian1 (approved):
:+1:

antiyro (approved):
lgtm thanks

PR #401, opened 2024-11-28
fix(log): define RUST_LOG=info by default

antiyro:
logs were not displaying by default


Trantorian1 (approved):
Tested in local :+1:

jbcaron (approved):

heemankv:
✅

Issue #387, opened 2024-11-19
refactor(tx): move AddTransactionProvider to its own crate


cchudant:
the AddTransactionProvider trait is the interface that abstracts the receiving of transactions in madara.

It is implemented by:
- mc-gateway-client, to forward all the transactions to another gateway. This is used in full-node mode, so that all invoke/declare/deployaccount transactions are forwarded to the starkware (or any other) public gateway.
- mc-mempool. When this implementation is used, the transactions are added to the mempool for block production.

It is used, as an `Arc<dyn AddTransactionProvider>` by:
- mc-gateway, our implementation of the gateway/feeder gateway
- mc-rpc, our implementation of the rpc specs
- the node operator rpcs for declare v0 txs (very soon it'll be in its own crate, see #386.)
- very soon, in mc-p2p to implement the adding of transactions by p2p pubsub.
- mc-eth (and long-term, its starknet equivalent for L3s) uses it to add L1 transactions into the mempool.

This abstraction is very useful to allow the "frontend" madara services like mc-gateway/mc-rpc to not care about where the transactions actually end up, whether forwarded to other p2p nodes, or forwarded to a gateway, or added to the mempool for block production.
Currently however it resides in mc-rpc, which limits its usefulness. We should put it into its own crate.

We also need to change the error types in that trait to a new error type that would also live in the same crate, because relying on starknet-types-rpc error types is a leaky abstraction that makes sense for mc-rpc but does not suit mc-p2p and mc-gateway. The consumer crates would just have to add a From implementation from that error type to the rpc/p2p/gateway server error type.

zkjamesfish:
Can I work on this?

heemankv:
Closing this as fixed.

The old transaction submission boundary has already been split out into the current crate structure. The original in-place `AddTransactionProvider` organization targeted by this issue is no longer the active design.

PR #405, opened 2024-12-03
feat(services): reworked Madara services for better cancellation control

Trantorian1:
# Pull Request type

- Bugfix
- Feature

Current Madara service architecture is limited and only allows us to start services on node startup. This is especially a problem if we wish to achieve zero downtime on warp updates.

## What is the new behavior?

Significantly overhauls the madara service architecture. Not much will be said in this pr as this has already been extensively documented in [service.rs](https://github.com/madara-alliance/madara/blob/feat/services/crates/primitives/utils/src/service.rs). You can also find more information about how this impacts warp update in the [Readme](https://github.com/madara-alliance/madara/tree/feat/services?tab=readme-ov-file#-database-migration-with-zero-downtime).

Some services also had to be updated to allow them to be compatible with the new architecture. This mostly meant supporting service restarts. You can check out the [RpcService](https://github.com/madara-alliance/madara/blob/c71a08c1d1b924790a3820d85625739eb9dac84f/crates/node/src/service/rpc/mod.rs#L53) for example as it has been quite significantly refactored.

Yes. Some changes were made to cli args for the sake of cleanup: `--sync-disabled` is now called `--l1-sync-disabled` to more accurately reflect its function.

## Other information

I am not sure how much I like my handling of [SERVICE_GRACE_PERIOD](https://github.com/madara-alliance/madara/blob/c71a08c1d1b924790a3820d85625739eb9dac84f/crates/primitives/utils/src/service.rs#L254). On the one hand this is handy to avoid node stalls in case of mistakes on our behalf and guarantee `SIGINT` and `SIGTERM` _always_ work, on the other though it can still lead to forceful cancellations (this is by design). Let me know if you think this trade is worth it.


crates/client/eth/src/l1_gas_price.rs
```diff
@@ -41,13 +42,16 @@ pub async fn gas_price_worker(
     l1_gas_provider.update_last_update_timestamp();
     let mut interval = tokio::time::interval(gas_price_poll_ms);
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-    while wait_or_graceful_shutdown(interval.tick(), &ctx).await.is_some() {
-        gas_price_worker_once(eth_client, l1_gas_provider.clone(), gas_price_poll_ms).await?;
+
+    while !ctx.is_cancelled() {
+        interval.tick().await;
+        gas_price_worker_once(eth_client, &l1_gas_provider, gas_price_poll_ms).await?;
     }
```
jbcaron: Currently, cancellation is only checked before waiting for the interval tick. If `ctx.is_cancelled()` becomes true during `interval.tick().await`, the worker cannot be canceled immediately, potentially delaying the shutdown. It would be better to modify this code to allow cancellation even during the tick wait.

README.md
```diff
@@ -426,16 +429,25 @@ are exposed on a separate port **9943** unless specified otherwise with
...
+| `madara_blockProductionDisable` | Disables block production (does not work on a full node)            |
+| `madara_blockProductionEnable`  | Enables block production (does not work on a full node)             |
+| `madara_blockProductionRestart` | Restarts block production (does not work on a full node)            |
+| `madara_rpcDisable`             | Disables user-facing rpc                                            |
+| `madara_rpcEnable`              | Enables user-facing rpc                                             |
+| `madara_rpcRestart`             | Restarts user-facing rpc                                            |
+| `madara_gatewayDisable`         | Disables the feeder gateway                                         |
+| `madara_gatewayEnable`          | Enables the feeder gateway                                          |
+| `madara_gatewayRestart`         | Restarts the feeder gateway                                         |
+| `madara_telemetryDisable`       | Disables node telemetry                                             |
```
cchudant: what is this used for?

README.md
```diff
@@ -544,67 +556,53 @@ into the subscription stream:
...
 When migration to a newer version of Madara you might need to update your
 database. Instead of re-synchronizing the entirety of your chain's state from
 genesis, you can use Madara's **warp update** feature.
 
-> [!NOTE]
-> Warp update requires an already synchronized _local_ node with a working
-> database.
+### Warp Update
+
+Warp update requires an already synchronized _local_ node with a working
```
cchudant: we may want to be a bit clearer that this is the source database that we're setting up here

README.md
```diff
@@ -544,67 +556,53 @@ into the subscription stream:
...
+database. If you do not already have one, you can use the following command to
+generate a sample database:
+
+```bash
+cargo run --release --      \
```
cchudant, line 560: at some point we'll have to replace all of the `cargo run --release` with `madara` , im not sure if its worth doing it now though

README.md
```diff
@@ -544,67 +556,53 @@ into the subscription stream:
...
-  --name Sender         \
-  --full                \ # This also works with other types of nodes
-  --network mainnet     \
-  --warp-update-sender
+cargo run --release -- \
+  --name Sender        \
+  --full               `# This also works with other types of nodes` \
+  --network mainnet    \
+  --warp-update-sender \
+  --l1-sync-disabled   `# We disable sync, for testing purposes` \
```
cchudant, line 581: "replace this argument with your --l1-endpoint <ethereum rpc url> parameter"

README.md
```diff
@@ -615,12 +613,120 @@ This will start generating a new up-to-date database under `/tmp/madara_new`.
...
+> You might already have noticed this line which appears at the end of the sync:
+> `📱 Running JSON-RPC server at 127.0.0.1:9944 ...`. More about this in the
+> next section
+
+### Achieving Zero Downtime
+
+Suppose your are an RPC service provider and your node is also running an RPC
+server and exposing it to your clients: if you have to shut it down or restart
+it for the duration of a migration this will result in downtime for your service
+and added complexity in setting up redundancies.
```
cchudant: well you can just resync a new node, i think it makes more sense to talk about sequencers here

README.md
```diff
@@ -615,12 +613,120 @@ This will start generating a new up-to-date database under `/tmp/madara_new`.
...
+
+The main issue is that it is not possible for multiple nodes to expose their
+services on the same port, so our receiver cannot start its rpc service if the
+sender node already has it active. Madara fixes this issue thanks to its
+microservice architecture which allows for deferred starts: when the sender has
```
cchudant: this is kubernetes' concern, that's not microservices here

rolling updates in kubernetes don't work like this, the nodes dont expose their ports to the same one in k8s
in proper k8s apps you would do their rolling update thing or even canary where you slowly switch your reverse proxy from one version of the app to another

that's not the problem we're trying to solve here right?

cchudant: also the way you're doing it isnt technically "no downtime" i believe

README.md
```diff
@@ -615,12 +613,120 @@ This will start generating a new up-to-date database under `/tmp/madara_new`.
...
+  --full                      \
+  --network mainnet           \
+  --l1-endpoint https://***   \
+  --warp-update-port-rpc 9943 `# Same as set with --rpc-admin-port on the sender` \
+  --warp-update-port-fgw 8080 `# Same as set with --gateway-port on the sender` \
+  --feeder-gateway-enable     \
+  --warp-update-receiver
+```
+
+Using this setup and adding any other arguments you need to the warp update
```
cchudant: but what about when you have to close the node to remove that warp-update-receiver argument

crates/client/eth/src/state_update.rs
```diff
@@ -30,85 +26,65 @@ pub async fn get_initial_state(client: &EthereumClient) -> anyhow::Result<L1Stat
...
+    {
+        let initial_state = get_initial_state(eth_client).await.context("Getting initial ethereum state")?;
+        update_l1(backend, initial_state, &eth_client.l1_block_metrics)?;
+    }
 
     // Listen to LogStateUpdate (0x77552641) update and send changes continusly
-    listen_and_update_state(eth_client, backend, &eth_client.l1_block_metrics, chain_id, ctx)
+    let event_filter = eth_client.l1_core_contract.event_filter::<StarknetCoreContract::LogStateUpdate>();
+
+    let mut event_stream = event_filter
```
jbcaron: same

README.md
cchudant: i dont understand why we would want no downtime updates for full nodes, this is definitely not the right layer to do it
i thought """no downtime""" upgrades would be something for sequencers, so that they can keep producing blocks after a migration - that's not a problem you can easily solve using existing stuff right?

README.md
```diff
@@ -615,12 +613,120 @@ This will start generating a new up-to-date database under `/tmp/madara_new`.
...
+> some discrepancies between official feeder gateway endpoints and our own
+> implementation. Please let us know about if you encounter this by
+> [raising an issue](https://github.com/madara-alliance/madara/issues/new/choose)
+
+### State Commitment Computation
+
+Madara supports merkelized state verification through its own implementation of
+Besu Bonsai Merkle Tries. See the [bonsai lib](https://github.com/madara-alliance/bonsai-trie).
+You can read more about Starknet Block structure and how it affects state
+commitment [here](https://docs.starknet.io/architecture-and-concepts/network-architecture/block-structure/).
```
cchudant, line 684: state commitments

crates/client/eth/src/sync.rs
```diff
@@ -12,25 +12,27 @@ use mc_db::MadaraBackend;
...
+            state_update_worker(&backend, &eth_client, ctx.clone()),
+            gas_price_worker(&eth_client, l1_gas_provider, gas_price_poll_ms, ctx.clone()),
+            sync(&backend, &eth_client, &chain_id, mempool, ctx.clone())
+        )?;
+    } else {
+        tokio::try_join!(
+            state_update_worker(&backend, &eth_client, ctx.clone()),
+            sync(&backend, &eth_client, &chain_id, mempool, ctx.clone())
+        )?;
+    }
```
jbcaron:
```suggestion
    let mut join_set = JoinSet::new();
    join_set.spawn(state_update_worker(&backend, &eth_client, ctx.clone()));
    join_set.spawn(sync(&backend, &eth_client, &chain_id, mempool, ctx.clone()));
    if !gas_price_sync_disabled {
      join_set.spawn(gas_price_worker(&eth_client, l1_gas_provider, gas_price_poll_ms, ctx.clone()));
    }
    while let Some(result) = join_set.join_next().await {
        result??;
    }
```

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -30,7 +31,7 @@ pub async fn gas_price_worker_once(
         );
     }
 
-    Ok(())
+    anyhow::Ok(())
```
cchudant, line 38: why?

crates/client/eth/src/l1_gas_price.rs
```diff
@@ -41,13 +42,16 @@ pub async fn gas_price_worker(
...
-    while wait_or_graceful_shutdown(interval.tick(), &ctx).await.is_some() {
-        gas_price_worker_once(eth_client, l1_gas_provider.clone(), gas_price_poll_ms).await?;
+
+    while !ctx.is_cancelled() {
+        interval.tick().await;
+        gas_price_worker_once(eth_client, &l1_gas_provider, gas_price_poll_ms).await?;
     }
-    Ok(())
+
+    anyhow::Ok(())
```
cchudant, line 54: same

crates/client/eth/src/state_update.rs
```diff
@@ -30,85 +26,65 @@ pub async fn get_initial_state(client: &EthereumClient) -> anyhow::Result<L1Stat
...
     backend.clear_last_confirmed_block().context("Clearing l1 last confirmed block number")?;
     tracing::debug!("update_l1: cleared confirmed block number");
 
     tracing::info!("🚀 Subscribed to L1 state verification");
-    // ideally here there would be one service which will update the l1 gas prices and another one for messages and one that's already present is state update
-    // Get and store the latest verified state
-    let initial_state = get_initial_state(eth_client).await.context("Getting initial ethereum state")?;
-    update_l1(backend, initial_state, &eth_client.l1_block_metrics, chain_id.clone())?;
+    // This does not seem to play well with anvil
+    #[cfg(not(test))]
```
cchudant, line 65: that's a hack? why is it here?

crates/client/gateway/server/src/service.rs
```diff
@@ -36,46 +36,32 @@ pub async fn start_server(
...
-
-    {
-        let shutdown_notify = Arc::clone(&shutdown_notify);
-        tokio::spawn(async move {
-            graceful_shutdown(&ctx).await;
-            shutdown_notify.notify_waiters();
-        });
-    }
-
     loop {
```
cchudant: (not related) huh i thought all of this was using hyper, why are we listener.accept() stuff by hand here?

crates/client/rpc/src/versions/admin/v0_1_0/methods/services.rs
```diff
@@ -10,61 +10,122 @@ const RESTART_INTERVAL: Duration = Duration::from_secs(5);
...
+    #[tracing::instrument(skip(self), fields(module = "Admin"))]
+    async fn service_block_production_restart(&self) -> RpcResult<MadaraServiceStatus> {
+        services_restart(&self.ctx, &[MadaraServiceId::BlockProduction], "Block production").await
     }
 
     #[tracing::instrument(skip(self), fields(module = "Admin"))]
-    async fn service_sync_disable(&self) -> RpcResult<bool> {
-        tracing::info!("🔌 Stopping Sync service...");
+    async fn service_rpc_disable(&self) -> RpcResult<MadaraServiceStatus> {
+        services_disable(&self.ctx, &[MadaraServiceId::RpcUser], "Rpc")
```
cchudant: at this point why not just have an endpoint that takes the service to disable as argument

crates/client/sync/src/fetch/fetchers.rs
```diff
@@ -143,12 +145,8 @@ pub async fn fetch_block_and_updates(
     Ok(converted)
 }
 
-async fn retry<F, Fut, T>(
-    mut f: F,
-    max_retries: u32,
-    base_delay: Duration,
-    ctx: &ServiceContext,
-) -> Result<T, SequencerError>
+// TODO: should we be checking for cancellation here? This might take a while
```
cchudant, line 154: i think this is cancellation safe and we can just drop the future here, it's just a fetch right?

cchudant: caller has to be aware of that

crates/client/sync/src/l2.rs
```diff
@@ -52,15 +53,15 @@ pub struct L2VerifyApplyConfig {
     flush_every_n_blocks: u64,
     flush_every_n_seconds: u64,
     stop_on_sync: bool,
-    telemetry: TelemetryHandle,
+    telemetry: Arc<TelemetryHandle>,
```
cchudant, line 56: telemetryhandle is already an arc

crates/client/sync/src/l2.rs
```diff
@@ -233,18 +243,16 @@ pub struct L2SyncConfig {
     pub flush_every_n_seconds: u64,
     pub pending_block_poll_interval: Duration,
     pub ignore_block_order: bool,
-    pub warp_update: bool,
-    pub warp_update_port_rpc: u16,
-    pub warp_update_port_fgw: u16,
     pub chain_id: ChainId,
-    pub telemetry: TelemetryHandle,
+    pub telemetry: Arc<TelemetryHandle>,
```
cchudant, line 240: ditto

crates/client/sync/src/lib.rs
```diff
@@ -21,13 +21,13 @@ pub struct SyncConfig {
     pub block_importer: Arc<BlockImporter>,
     pub starting_block: Option<u64>,
     pub backup_every_n_blocks: Option<u64>,
-    pub telemetry: TelemetryHandle,
+    pub telemetry: Arc<TelemetryHandle>,
```
cchudant, line 24: ditto

crates/client/telemetry/src/lib.rs
```diff
@@ -18,42 +14,33 @@ pub enum VerbosityLevel {
...
+#[derive(Debug, Clone)]
 pub struct TelemetryEvent {
     verbosity: VerbosityLevel,
     message: serde_json::Value,
 }
 
 #[derive(Debug, Clone)]
-pub struct TelemetryHandle(Option<Arc<mpsc::Sender<TelemetryEvent>>>);
+#[repr(transparent)]
+pub struct TelemetryHandle(tokio::sync::broadcast::Sender<TelemetryEvent>);
```
cchudant, line 25: ah! why the change?

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
```diff
@@ -44,69 +45,141 @@ pub trait MadaraStatusRpcApi {
 
 #[versioned_rpc("V0_1_0", "madara")]
 pub trait MadaraServicesRpcApi {
```
jbcaron, line 56: I would prefer grouping services together by adding parameters

crates/primitives/utils/src/service.rs
```diff
@@ -1,44 +1,441 @@
...
+use serde::{Deserialize, Serialize};
+use std::{
+    fmt::{Debug, Display},
+    panic,
+    sync::Arc,
+    time::Duration,
+};
 use tokio::task::JoinSet;
 
+pub const SERVICE_COUNT: usize = 8;
```
cchudant: i think this shouldnt be here btw, it should be in main.rs

if an appchain wants to add a service by importing all of our crates and just have their own main.rs which starts and hook services differently (/ using other services), this wont work for them

crates/primitives/utils/src/service.rs
```diff
@@ -1,44 +1,441 @@
...
+pub const SERVICE_GRACE_PERIOD: Duration = Duration::from_secs(10);
+
+/// Represents a [Service] as a unique power of 2.
+///
+/// Note that 0 represents [MadaraService::Monitor] as [ServiceMonitor] is
+/// always running and therefore is the genesis state of all other services.
 #[repr(u8)]
 #[derive(Clone, Copy, PartialEq, Eq, Default, Debug)]
-pub enum MadaraService {
+pub enum MadaraServiceId {
```
cchudant, line 365: same btw

crates/primitives/utils/src/service.rs
```diff
@@ -130,142 +528,241 @@ impl Clone for ServiceContext {
...
+        let mut rx = self.service_update_receiver.take().expect("Receiver was set above");
+        let token_global = &self.token_global;
+        let token_local = self.token_local.as_ref().unwrap_or(&self.token_global);
+
+        loop {
+            // We keep checking for service status updates until a token has
+            // been canceled or this service was deactivated
+            let res = tokio::select! {
+                svc = rx.recv() => svc.ok(),
+                _ = token_global.cancelled() => break,
```
cchudant, line 734: why isnt token global a parent of token local? you shouldnt have to await both

crates/primitives/utils/src/service.rs
```diff
@@ -274,110 +771,244 @@ impl ServiceContext {
...
-    pub fn state_advance(&mut self) -> MadaraState {
-        let state = self.state.load(std::sync::atomic::Ordering::SeqCst).saturating_add(1);
-        self.state.store(state, std::sync::atomic::Ordering::SeqCst);
-        state.into()
-    }
+/// Provides info about a [Service]'s status.
+///
+/// Used as part of [ServiceContext::service_subscribe].
+#[derive(Clone, Copy)]
+pub struct ServiceTransport {
```
cchudant, line 934: why is this called transport?

README.md
Trantorian1: These are supposed to be endpoint node admin can use in case they need to disable a specific service, for example for maintenance reasons. I am going to refactor this into a single endpoint though, as per @jbcaron's suggestion

README.md
Trantorian1: Yep, will add some clarifying information!

README.md
Trantorian1: I think this is fine atm since this still remains the main way to get madara off and running. But I'm in agreement with you this will have to be updated once we have an actual cli people can install (I guess I could also add info on how to do that with `cargo install --git https://github.com/madara-alliance/madara --locked madara `, wdyt?)

README.md
Trantorian1: Will add more info about this. The goal for this demonstration though _is_ to stop on sync though, else the sender node would keep synchronizing and the update would take forever

README.md
Trantorian1: Yea fair, though this is still a bit faster than that due to being on a local network. It would also be quite easy to make this a lot faster "in the future :tm: " by computing the state root once the migration has completed, instead of at each block which is what we do on a normal sync. Something to keep in mind.

README.md
Trantorian1: Zero downtime in this sense refers to the fact the receiver seamlessly takes the place of the sender, hence no downtime relating to having to restart the node. I'm not very familiar with v8.

README.md
Trantorian1: I'm confused. The node does not have to be closed if it is called with `--warp-update-receiver`: this is a cli flag which does not set any other arguments, so once the warp update has completed your node will run normally as per the other arguments that were passed to it. There is no need close the node to remove the `--warp-update-receiver` argument.

crates/client/eth/src/l1_gas_price.rs
Trantorian1: I like it :D

crates/client/eth/src/state_update.rs
Trantorian1: I refactored this from the previous version which separated `state_update_worker` into `listen_and_update_state`. I did not see the reason for that and so included it directly into `state_update_worker`. For some reason I don't quite understand though, getting the initial l1 state does not work, which might could either be an issue in the way in which we are mocking the core contract or with how we are using anvil.

crates/client/gateway/server/src/service.rs
Trantorian1: Linking @jbcaron here.

crates/client/rpc/src/versions/admin/v0_1_0/methods/services.rs
Trantorian1: Yep, planning on doing that, will be refactoring that now :+1:

crates/client/sync/src/fetch/fetchers.rs
Trantorian1: I'll be looking into that and adding something like

```rust
tokio::select! {
    res = retry(...) => res,
    _ = ctx.cancelled() => return anyhow::Ok(),
}
```

crates/client/telemetry/src/lib.rs
Trantorian1: So! We need services to be able to be potentially restarted. This does not work if we store `TelemetryHandle` as an `Option` and just call `std::mem::take` on it. Similarly, we cannot use a `tokio::sync::mpsc` as we need to be able to re-create the receiver on each service start (the service future takes ownership of it).

crates/primitives/utils/src/service.rs
Trantorian1: Well they'll have bigger problems due to the limit in size of using an `std::sync::atomic::AtomicU8`. I assume you mean having the service id behind a trait so appchains can create their own services?

crates/primitives/utils/src/service.rs
Trantorian1: We have to await both in case the `token_local` is cancelled but the `token_global` is not.

crates/primitives/utils/src/service.rs
Trantorian1: Just a name I came up with.

crates/client/sync/src/fetch/mod.rs
```diff
@@ -105,17 +107,27 @@ pub async fn l2_fetch_task(
 
         let mut interval = tokio::time::interval(sync_polling_interval);
         interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-        while wait_or_graceful_shutdown(interval.tick(), &ctx).await.is_some() {
+        while !ctx.is_cancelled() {
+            interval.tick().await;
```
jbcaron: same

crates/client/sync/src/l2.rs
```diff
@@ -190,17 +197,20 @@ async fn l2_pending_block_task(
 
     let mut interval = tokio::time::interval(pending_block_poll_interval);
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
-    while wait_or_graceful_shutdown(interval.tick(), &ctx).await.is_some() {
+    while !ctx.is_cancelled() {
+        interval.tick().await;
```
jbcaron: same

crates/client/gateway/server/src/service.rs
jbcaron: I think we don’t have a choice with the new version of Hyper.

crates/client/eth/src/l1_gas_price.rs
Trantorian1: Good catch, will be adding that!

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
Trantorian1: Yep, I've change this locally and will be pushing it once I have resolved the other comments as well

crates/client/sync/src/fetch/fetchers.rs
Trantorian1: Actually this is not needed as we are already checking for cancellation when we are polling the block fetch stream. This means the block fetching might be cancelled, but this should be safe as we are not mutating the state of the node until those blocks are sent to the verify and apply task.

crates/client/sync/src/l2.rs
Trantorian1: Not anymore, this had to be updated to make it so the service can be restarted.

crates/primitives/utils/src/service.rs
Trantorian1: Well, this was a bit of a refactor but `MadaraServiceId` is now only a blanket implementation of `Serviceid` for our services, with `PowerOfTwo` being the actual backing type used to identify services. Appchains can create and register their own services by implementing the `Service` and `ServiceId` traits on their own structs, and they are free to use any service id enum they want, all relevant methods now only require an `impl ServiceId`.

As for `SERVICE_COUNT`, it has been renamed to `SERVICE_COUNT_MAX` to better reflect what it actually represents and has been increased to 64.

crates/client/eth/src/l1_gas_price.rs
Trantorian1: This has been fixed!

README.md
Trantorian1: Also sequencer migrations are supported as well now (mempool is still not transferred though)

crates/client/eth/src/state_update.rs
Trantorian1: This has been fixed.

crates/client/eth/src/sync.rs
Trantorian1: This has been fixed.

crates/client/rpc/src/versions/admin/v0_1_0/api.rs
Trantorian1: This has been fixed.

crates/client/sync/src/fetch/mod.rs
Trantorian1: this has been fixed.

crates/client/sync/src/l2.rs
Trantorian1: This has been fixed.

crates/client/sync/src/fetch/mod.rs
Trantorian1: This has been fixed

crates/client/sync/src/l2.rs
Trantorian1: This has been fixed

crates/client/rpc/src/providers/mod.rs
jbcaron, line 1: You could have resolved #387 at the same time

jbcaron (approved):

antiyro (approved):
works extremely fast locally

cchudant (approved):
okay

PR #409, opened 2024-12-09
Feat: Replace Class Hash issue resolved, gas fees bug resolved, cap removed for pending tick

Mohiiit:
- Gas fees bug resolved.
- Replace class hash bug resolved
- Cap removed for pending tick
# Pull Request type

- Bugfix
- Feature

- In gas fees bug, it was disabling fee setting of eth when strk prices are defined in the arguments.
- Replace class issue occurs because we are saving the class hash in the db when a contract is declared, and while checking whether the class hash of the deployed contract already exist, it returns true even tho it's the first time the contract is being deployed. 
- Currently there is a cap on number of steps/gas a tick can consume, in ideal case the block should be close either when the limit is reached or when the block time is over.

Resolves: #NA

## What is the new behavior?

- gas fees is resolved by disabling the strk prices when it's defined in the args.
- for replace class hash issue, now we check whether the class hash of the deployed contract matches with the one stored in the db, if it matches then it means it's not replaced.
- cap is removed, blockifier returns whether the limit for the block is reached or not, we are using that flag to close the block early.

No


configs/presets/devnet.yaml
```diff
@@ -5,7 +5,7 @@ gateway_url: "http://localhost:8080/gateway/"
 native_fee_token_address: "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
 parent_fee_token_address: "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
 latest_protocol_version: "0.13.2"
-block_time: "10s"
+block_time: "30s"
```
Trantorian1: Why this change? Testing purposes?

configs/presets/devnet.yaml
```diff
@@ -27,8 +27,9 @@ bouncer_config:
     n_events: 18446744073709551615
     state_diff_size: 131072
 sequencer_address: "0x123"
-eth_core_contract_address: "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512"
-eth_gps_statement_verifier: "0xf294781D719D2F4169cE54469C28908E6FA752C1"
+eth_core_contract_address: "0xdf13b98ea1c1f8b104d63b9021be731f6e91c75b"
+eth_gps_statement_verifier: "0x07ec0D28e50322Eb0C159B9090ecF3aeA8346DFe"
+private_key: "0x1234"
```
Trantorian1: Ditto

crates/client/block_production/src/lib.rs
```diff
@@ -201,8 +201,9 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
     fn continue_block(
         &mut self,
         bouncer_cap: BouncerWeights,
-    ) -> Result<(StateDiff, VisitedSegments, BouncerWeights, ContinueBlockStats), Error> {
+    ) -> Result<(StateDiff, VisitedSegments, BouncerWeights, ContinueBlockStats, bool), Error> {
```
Trantorian1: At this point consider putting the return type behind a struct.

crates/client/devnet/src/lib.rs
```diff
@@ -388,8 +388,11 @@ mod tests {
 
         assert_eq!(res.class_hash, calculated_class_hash);
 
-        chain.block_production.set_current_pending_tick(1);
-        chain.block_production.on_pending_time_tick().unwrap();
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        rt.block_on(async {
+            chain.block_production.set_current_pending_tick(1);
+            chain.block_production.on_pending_time_tick().await.unwrap();
+        });
```
Trantorian1, line 391: Why are you spawning a tokio runtime instead of using `#[tokio::test]`?

crates/client/devnet/src/lib.rs
```diff
@@ -457,8 +460,11 @@ mod tests {
             .unwrap();
         tracing::debug!("tx hash: {:#x}", transfer_txn.transaction_hash);
 
-        chain.block_production.set_current_pending_tick(chain.backend.chain_config().n_pending_ticks_per_block());
-        chain.block_production.on_pending_time_tick().unwrap();
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        rt.block_on(async {
+            chain.block_production.set_current_pending_tick(chain.backend.chain_config().n_pending_ticks_per_block());
+            chain.block_production.on_pending_time_tick().await.unwrap();
+        });
```
Trantorian1, line 463: ditto

crates/client/devnet/src/lib.rs
```diff
@@ -489,8 +495,11 @@ mod tests {
 
         let res = chain.sign_and_add_deploy_account_tx(deploy_account_txn, &account).unwrap();
 
-        chain.block_production.set_current_pending_tick(chain.backend.chain_config().n_pending_ticks_per_block());
-        chain.block_production.on_pending_time_tick().unwrap();
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        rt.block_on(async {
+            chain.block_production.set_current_pending_tick(chain.backend.chain_config().n_pending_ticks_per_block());
+            chain.block_production.on_pending_time_tick().await.unwrap();
+        });
```
Trantorian1, line 498: ditto

crates/client/devnet/src/lib.rs
```diff
@@ -550,8 +559,11 @@ mod tests {
 
         tracing::info!("tx hash: {:#x}", result.transaction_hash);
 
-        chain.block_production.set_current_pending_tick(1);
-        chain.block_production.on_pending_time_tick().unwrap();
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        rt.block_on(async {
+            chain.block_production.set_current_pending_tick(1);
+            chain.block_production.on_pending_time_tick().await.unwrap();
+        });
```
Trantorian1, line 562: ditto

crates/client/devnet/src/lib.rs
```diff
@@ -753,8 +765,11 @@ mod tests {
             .unwrap();
 
         std::thread::sleep(max_age); // max age reached
-        chain.block_production.set_current_pending_tick(1);
-        chain.block_production.on_pending_time_tick().unwrap();
+        let rt = tokio::runtime::Runtime::new().unwrap();
+        rt.block_on(async {
+            chain.block_production.set_current_pending_tick(1);
+            chain.block_production.on_pending_time_tick().await.unwrap();
+        });
```
Trantorian1, line 771: ditto

crates/client/mempool/src/l1.rs
```diff
@@ -42,11 +42,11 @@ impl GasPriceProvider {
     }
 
     pub fn set_strk_gas_price_sync_enabled(&self, enabled: bool) {
-        self.gas_price_sync_enabled.store(enabled, Ordering::Relaxed);
+        self.strk_gas_price_sync_enabled.store(enabled, Ordering::Relaxed);
```
Trantorian1: I know this was not changed in this pr, but why are we using `Ordering::Relaxed` here?

Trantorian1 (changes requested):
Some aspects of this pr seem a bit strange to me and might need some tidying up.

crates/client/devnet/src/lib.rs
Mohiiit: this is because we are already using a fixture which creates it's own runtime and when we do tokio::test it's creating another runtime inside a runtime, which is a issue

configs/presets/devnet.yaml
Mohiiit: yes, will revert this

crates/client/mempool/src/l1.rs
Mohiiit: Because it provides the least overhead and also given we are updating gas price every few ms, even if the other process won't get the latest price immediately, this shouldn't be an issue. 

Now that I think of it, there might be an edge case over here, and we can change the ordering to make it `AcqRel` maybe, but it has a lot of overhead other options like `Acquire` and `Release` doesn't make sense because the operation of these would be visible in the same thread.

wdyt?

crates/client/block_production/src/lib.rs
Mohiiit: sure thing, what's the standard for this though? more than 4 should be struct? just curious

Trantorian1: Not really any standard tbf. I guess maybe clippy might start complaining at a certain point but mostly just for readability purposes, as its easy to get confused regarding the meaning of each field in a return type if it is just a tupple.

crates/client/devnet/src/lib.rs
Trantorian1: Which fixture is responsible for this? This still feels strange.

crates/client/mempool/src/l1.rs
Trantorian1: Makes sense! Might be worth adding a quick comment with what you just explained here to make things clear.

Trantorian1 (changes requested):

crates/client/block_production/src/lib.rs
Mohiiit: fair enough, will do

crates/client/devnet/src/lib.rs
Mohiiit: if I am not wrong, it was because of this: https://github.com/madara-alliance/madara/pull/409/files#diff-f184a5cffa283d57310a47700e3aa54889ed30d27fafe9eba1fe7b9c12e44624R317

but if I think about it, it shouldn't be the case because fixture runs before the test right, hmm let me re-check this once again

Mohiiit: my bad actually, so here is the thought process I followed in order to reach to the current scenario:

1. code was changed in the `on_pending_block_time` in order to do dynamic block closing, so as soon as the limit is reached, we close the close, so in the code the `close_block_time` function was called which is async function.
2. now the `pending_block_time` has to async as well, so I updated the tests with `tokio::test` as well
3. when I did so, now within the test, we have this fixture which creates its own runtime, which is not allowed because `tokio::test` creates it's own runtime. 
4. so to solve this, I removed the `tokio::test` and created runtime for the specific part of the code

crates/client/eth/src/client.rs
```diff
@@ -181,7 +181,7 @@ pub mod eth_client_getter_test {
     const L2_STATE_ROOT: &str = "1456190284387746219409791261254265303744585499659352223397867295223408682130";
 
     lazy_static::lazy_static! {
-        static ref FORK_URL: String = std::env::var("ETH_FORK_URL").expect("ETH_FORK_URL not set");
+        static ref FORK_URL: String = std::env::var("ETH_FORK_URL").unwrap_or("https://eth-mainnet.g.alchemy.com/v2/svwTUdkIFaUU8uh3Uutn-0-UzsM9ee6q".to_string());
```
jbcaron: It looks like the API key was committed by mistake

crates/client/block_production/src/lib.rs
```diff
@@ -18,7 +18,8 @@
 use crate::close_block::close_block;
 use crate::metrics::BlockProductionMetrics;
 use blockifier::blockifier::transaction_executor::{TransactionExecutor, BLOCK_STATE_ACCESS_ERR};
-use blockifier::bouncer::{BouncerWeights, BuiltinCount};
+use blockifier::bouncer::BouncerWeights;
+use blockifier::state::state_api::UpdatableState;
```
jbcaron:
```suggestion
```

crates/client/block_production/src/lib.rs
```diff
@@ -77,6 +77,15 @@ pub enum Error {
     #[error("State diff error when continuing the pending block: {0:#}")]
     PendingStateDiff(#[from] StateDiffToStateMapError),
 }
+
+struct ContinueBlockResult {
```
antiyro, line 83: could you document this struct for rust doc?

crates/client/eth/src/client.rs
Mohiiit: thanks, removed

crates/client/block_production/src/lib.rs
Mohiiit: added

antiyro (approved):
lgtm

crates/client/block_production/src/lib.rs
```diff
@@ -353,8 +351,13 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
```
cchudant, line 412: i think you should split the on_block_time function in two: on_block_time and close_block and on_pending_block_tick should call close_block instead of on_block_time

cchudant (changes requested):
i think you should also reset the intervals in block_production_task when a block is closed,
```rs
interval_pending_block_update.reset_at(instant + interval_pending_block_update.period());
interval_block_time.reset_at(instant + interval_block_time.period());
````

File: crates/client/block_production/src/lib.rs (hunk 13 above)
Comment (Mohiiit) [reply crates/client/block_production/src/lib.rs:412]: did something very similar, lmk wdyt?

File: crates/client/block_production/src/lib.rs [hunk 14]
```diff
@@ -473,10 +489,19 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
 
-                    if let Err(err) = self.on_pending_time_tick() {
-                        tracing::error!("Pending block update task has errored: {err:#}");
+                    match self.on_pending_time_tick().await {
+                        Ok(block_closed) => {
+                            if block_closed {
+                                interval_pending_block_update.reset_at(instant + interval_pending_block_update.period());
+                                interval_block_time.reset_at(instant + interval_block_time.period());
+                            }
+                            self.current_pending_tick += 1;
```
cchudant: could you reset the current_pending_tick to 0 as well?
(i believe the current pending tick variable is now useless, is that the case?

cchudant (changes requested):

crates/client/block_production/src/lib.rs
Mohiiit: yeah I thought about it as well, pushing the changes that I think made sense, lmk wdyt

cchudant (approved):
yes :)

PR #411, opened 2024-12-10
fix(block_production): continue pending block now reexecutes the previous transactions

cchudant:
# Pull Request type

- Bugfix

We had a hacky way of continuing the pending block state on restart.

## What is the new behavior?

Now, we re-add all the transactions from the pending block into the mempool when restarting with a priority.

I also did some refactoring here and there:
- added a backend get_converted_class method - BlockifierStateAdaptor, rpc to_blockifier_transaction & cie use that now instead of the duplicate code we used to have
- some helper functions to get from converted classes to blockifier class info format, remove some duplicate code
- removed the rpc to_blockifier_transaction logic (used for trace txs). It is now inside the primitives & shared with other places in the codebase.
- mempool_tx_max_age is now optional & defaults to no limit (this limit doesn't really make sense for FCFS-style sequencing anyway)
- moved devnet e2e tests inside their own file
- block timestamp in block headers have now a wrapper type, because a plain u64 was confusing :)

No, db remains unchanged.


cchudant:
note: the file change count is 45 which seems like a lot, but most files have a 1-line change. I don't believe this is a large PR

crates/client/block_production/src/lib.rs
```diff
@@ -104,85 +103,54 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         self.current_pending_tick = n;
     }
 
+    /// Continue the pending block state by re-adding all of its transactions back into the mempool.
+    /// This function will always clear the pending block in db, even if the transactions could not be added to the mempool.
+    pub fn re_add_pending_block_txs_to_mempool(
+        backend: &MadaraBackend,
+        mempool: &Mempool,
+    ) -> Result<(), Cow<'static, str>> {
```
Trantorian1, line 110: Why not use `anyhow::Result` here?

crates/client/db/src/class_db.rs
```diff
@@ -110,6 +112,44 @@ impl MadaraBackend {
         Ok(Some(compiled))
     }
 
+    /// Get class info + sierra compiled when it's a sierra class.
+    // Note/TODO: "ConvertedClass" is the name of the type that has info + sierra compiled, and it is used for blockifier
+    // convertion & storage. We should rename it, as this feels like undecipherable madara-specific jargon at this point.
```
jbcaron, line 117: We could rename it to `ResolvedClass` or `CanonicalClass`

crates/client/exec/src/transaction.rs
```diff
@@ -26,53 +26,26 @@ pub enum Error {
 ///
 /// **note:** this function does not support deploy transaction
 /// because it is not supported by blockifier
-pub fn to_blockifier_transactions(
+pub fn to_blockifier_transaction(
```
jbcaron, line 29: 👍

crates/client/rpc/src/utils/transaction.rs
jbcaron, line 1: 👍

jbcaron (approved):

crates/client/block_production/src/lib.rs
```diff
@@ -104,85 +103,54 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
...
+        let Some(current_pending_block) =
+            backend.get_block(&DbBlockId::Pending).map_err(|err| format!("Getting pending block: {err:#}"))?
+        else {
+            // No pending block
+            return Ok(());
+        };
+        backend.clear_pending_block().map_err(|err| format!("Clearing pending block: {err:#}"))?;
```
Trantorian1, line 117: Should we perhaps not clear the pending block only _after_ all transactions have been added back into the mempool?

crates/client/block_production/src/re_add_finalized_to_blockifier.rs
```diff
@@ -0,0 +1,82 @@
...
+            } else {
+                None
+            };
+
+            let tx = TransactionWithHash::new(tx, tx_hash)
+                .into_blockifier(converted_class.as_ref())
+                .map_err(|err| ReAddTxsToMempoolError::ToBlockifierError { tx_hash, err })?;
+
+            // HACK: we hack the order a little bit - this is because we don't have the arrived_at timestamp for the
+            // transaction. This hack ensures these trasactions should have priority and should be reexecuted
```
Trantorian1, line 61: Would it maybe be worthwhile to store the tx timestamps for the pending block to make this a little less hacky? I guess this method should work so that should be fine too, just wondering.

crates/client/mempool/src/inner/limits.rs
```diff
@@ -139,9 +137,11 @@ impl MempoolLimiter {
 
     pub fn mark_removed(&mut self, to_update: &TransactionCheckedLimits) {
         // These should not overflow unless block prod marks transactions as consumed even though they have not been popped.
-        self.current_transactions -= 1;
+        debug_assert!(self.current_transactions > 0);
+        self.current_transactions = self.current_transactions.saturating_sub(1);
         if to_update.check_declare_limit {
-            self.current_declare_transactions -= 1;
+            debug_assert!(self.current_declare_transactions > 0);
+            self.current_declare_transactions = self.current_declare_transactions.saturating_sub(1);
```
Trantorian1, line 144: Nice!

crates/client/mempool/src/inner/mod.rs
```diff
@@ -258,10 +267,23 @@ impl MempoolInner {
...
+        &mut self,
+        txs: impl IntoIterator<Item = MempoolTransaction>,
+        force: bool,
+    ) -> Result<(), TxInsersionError> {
+        for tx in txs {
+            self.insert_tx(tx, force, true)?;
+        }
+        Ok(())
+    }
+
```
Trantorian1, line 286: Consider renaming this function to avoid confusion.

crates/client/mempool/src/lib.rs
```diff
@@ -305,13 +307,31 @@ impl MempoolProvider for Mempool {
     /// This is called by the block production after a batch of transaction is executed.
     /// Mark the consumed txs as consumed, and re-add the transactions that are not consumed in the mempool.
     #[tracing::instrument(skip(self, txs, consumed_txs), fields(module = "Mempool"))]
-    fn re_add_txs<I: IntoIterator<Item = MempoolTransaction>, CI: IntoIterator<Item = MempoolTransaction>>(
+    fn re_add_txs<I: IntoIterator<Item = MempoolTransaction>>(
         &self,
         txs: I,
-        consumed_txs: CI,
-    ) {
+        consumed_txs: Vec<MempoolTransaction>,
```
Trantorian1, line 313: Should we not be using a slice instead?

crates/primitives/block/src/header.rs
```diff
@@ -29,14 +33,31 @@ impl From<BlockStatus> for starknet_types_rpc::BlockStatus {
     }
 }
 
-#[derive(Clone, Debug, Default, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[derive(Clone, Copy, Default, Debug, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct BlockTimestamp(pub u64);
```
Trantorian1, line 38: Consider adding `#[repr(transparent)]` (I don't know if `serde(transparent)` already does this or not)

crates/primitives/transactions/src/to_blockifier.rs
```diff
@@ -130,47 +178,38 @@ pub enum BroadcastedToBlockifierError {
...
+        ConvertedClass::Legacy(LegacyConvertedClass { class_hash, info: LegacyClassInfo { contract_class } });
+    Ok((Some(converted_class.to_blockifier_class_info()?), Some(converted_class), Some(class_hash)))
 }
 
 #[allow(clippy::type_complexity)]
 fn handle_class_sierra(
     contract_class: Arc<FlattenedSierraClass>,
     expected_compiled_class_hash: Felt,
-) -> Result<(Option<BClassInfo>, Option<ConvertedClass>, Option<Felt>), BroadcastedToBlockifierError> {
+) -> Result<(Option<BClassInfo>, Option<ConvertedClass>, Option<Felt>), ToBlockifierError> {
```
Trantorian1, line 200: Consider making a struct of this return type?

crates/tests/src/lib.rs
```diff
@@ -168,6 +165,11 @@ pub fn get_port() -> MadaraPortNum {
     MadaraPortNum(port)
 }
 
+/// Note: the builder is [`Clone`]able. When cloned, it will keep the same tempdir.
+/// This is useful for tests that need to restart the node using the same DB: they
+/// can just make a builder, clone() it and call [`MadaraCmdBuilder::run`] to launch
+/// the node. They can then [`drop`] the [`MadaraCmd`] instance to kill the node, and
+/// restart the node using the same db by reusing the earlier builder.
```
Trantorian1, line 172:
```suggestion
/// Note: the builder is [Clone]able. When cloned, it will keep the same tempdir.
/// This is useful for tests that need to restart the node using the same DB: they
/// can just make a builder, clone() it and call [MadaraCmdBuilder::run] to launch
/// the node. They can then [Drop] the [MadaraCmd] instance to kill the node, and
/// restart the node using the same db by reusing the earlier builder.
```

Trantorian1 (changes requested):

crates/client/db/src/class_db.rs
cchudant: why canonical?

crates/client/block_production/src/lib.rs
cchudant: it's not flushed anyway at this point, but i'll do it yes just in case, why not

crates/client/block_production/src/re_add_finalized_to_blockifier.rs
cchudant: we don't always have that info, eg. if we're continuing a block generated using full-node mode

antiyro (approved):

PR #429, opened 2024-12-18
fix(gateway): path

Uacias:
# Pull Request type

- Bugfix

The gateway_router function incorrectly matches paths for the `add_transaction` endpoint. Requests to `/gateway/add_transaction` result in a Not Found response because the matching logic does not handle paths correctly.

Resolves: #NA

## What is the new behavior?

- Fixed the path matching logic in the `gateway_router` fn

No

## Other information

Would also work using `req.uri().path().trim_start_matches('/')`, but `path` approach is cleaner imho.


antiyro:
Could you fix CI on that? Adding changelog should do the job.

jbcaron (approved):

antiyro (approved):

jbcaron:
close by #440

PR #435, opened 2024-12-20
fix: minor bug while trimming hash, test added

Mohiiit:
# Pull Request type

- Bugfix

Resolves: #NA

## What is the new behavior?

- it was failing when hash is 0x0 which is possible in case of the appchains

No


jbcaron (changes requested):
It would be better to directly use the function `mc-sync::utils::trim_hash` and move it directly into `mp-utils` or `mp-convert`

jbcaron (approved):

antiyro (approved):

PR #438, opened 2024-12-23
fix(rpc): call, simulate, estimate rpcs executed on top of the block, not at the start of it 

cchudant:
# Pull Request type
- Bugfix

Call, simulate, estimate rpcs are now executed on top of the block they request instead of at the start of it.

## What is the new behavior?

State changes that happen during the block requested in these rpc calls are now visible to the execution.
Trace block and trance transaction rpcs are unchanged - state changed during the block shouldn't be visible to these as they reexecute the requested block.

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix(rpc): call, simulate, estimate rpcs executed on top of the block, not at the start of it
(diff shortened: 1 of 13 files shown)


jbcaron (approved):

jbcaron:
Could you fix the CI @cchudant ?

antiyro (approved):

PR #446, opened 2024-12-30
fix(ci): migrated changelog action on releases for conventional commits only

antiyro:
## fix(ci): migrated changelog action on releases for conventional commits only

# Pull Request type

- Bugfix
- Feature
- Build-related changes

The CI is extremely intensive and takes a lot of time, largely due to rebases on changelogs. We remove this check from each PR and instead stick to the conventional commit format feat(scope): description when we want it to be included in the changelogs during releases. This will prevent changelogs with multiple useless entries like in our 0.7.0 release and add clarity for users.

Resolves: #NA

## What is the new behavior?

1. Removed the changelog enforcement workflow (`changelog.yml`)
2. Removed the changelog check from the PR workflow
3. Updated the release workflow to automatically generate changelogs from conventional commits

For commits to be included in changelogs, they should follow the conventional commit format:
```
feat(scope): description
fix(scope): description
chore(scope): description
docs(scope): description
```

This change will:
- Speed up CI by removing redundant changelog checks
- Keep changelogs focused on meaningful changes
- Reduce noise in the release notes
- Make it clearer for users what changed in each release

NO


.github/workflows/release.yml
```diff
@@ -16,6 +16,14 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Generate Release Notes
+        id: generate_notes
+        uses: conventional-changelog/releaser-action@v1
```
Trantorian1, line 24: Nice! Just linking this issue for completeness sake, touch I don't think it should affect us: https://github.com/TriPSs/conventional-changelog-action/issues/239

Trantorian1 (approved):
A very welcome change!

jbcaron (approved):

PR #447, opened 2024-12-30
fix(block_time): block time can no longer be less than pending block update time

### LINKED ISSUE #277: bug(mempool): tick time <= block time

### Description of bug

Currently, it's possible to have a tick time greater than the block time (it happened with Pragma as tick time was default 2s and block time was changed to 1s). We should ideally panic on node start if this happens (or maybe warn)

### Steps to reproduce

_No response_

Trantorian1:
# Pull Request type

- Bugfix

Block time can be less than pending block update time (also known as tick time). This is an issue as it means only empty blocks will be produced.

Resolves: #277 

## What is the new behavior?

Forbids the node from starting as a sequencer if ever its block time is less than its tick time.

No


crates/madara/node/src/cli/chain_config_overrides.rs
Mohiiit, line 1: this was required, thanks

crates/madara/node/src/main.rs
```diff
@@ -50,15 +50,32 @@ async fn main() -> anyhow::Result<()> {
...
 
+    // If block time is inferior to the tick time, then only empty blocks will
+    // be produced as we will never update the pending block before storing it.
+    if run_cmd.is_sequencer() && chain_config.block_time < chain_config.pending_block_update_time {
+        tracing::error!(
+            "Block time ({}s) cannot be less than the pending block update time ({}s), as this will yield only empty blocks",
+            chain_config.block_time.as_secs(),
+            chain_config.pending_block_update_time.as_secs()
+        );
+        anyhow::bail!("Block time");
```
Mohiiit: if we just do panic!() that would be enough right? I mean the tracing::error is covering the message precisely, having Block time here doesn't add any value, wdyt? or we can do: 
```suggestion
        anyhow::bail!(
            "invalid block configuration: block time {} is less than pending update time {}", 
            config.block_time.as_secs(),
            config.pending_block_update_time.as_secs()
        );
```

crates/madara/node/src/main.rs
```diff
@@ -50,15 +50,32 @@ async fn main() -> anyhow::Result<()> {
...
     // Check if the devnet is running with the correct chain id.
     if run_cmd.devnet && chain_config.chain_id != NetworkType::Devnet.chain_id() {
         if !run_cmd.block_production_params.override_devnet_chain_id {
-            tracing::error!("You're running a devnet with the network config of {:?}. This means that devnet transactions can be replayed on the actual network. Use `--network=devnet` instead. Or if this is the expected behavior please pass `--override-devnet-chain-id`", chain_config.chain_name);
-            panic!();
+            tracing::error!(
+                "You're running a devnet with the network config of {:?}. This means that devnet transactions can be replayed on the actual network. Use `--network=devnet` instead. Or if this is the expected behavior please pass `--override-devnet-chain-id`",
+                chain_config.chain_name
+            );
+            anyhow::bail!("devnet");
```
Mohiiit: same as above

Mohiiit (commented):
lgtm, minor doubts

crates/madara/node/src/cli/mod.rs
```diff
@@ -200,7 +200,7 @@ pub struct RunCmd {
     #[clap(env = "MADARA_PRESET", long, value_name = "PRESET NAME", group = "chain_config")]
     pub preset: Option<ChainPreset>,
 
-    /// Overrides parameters from the Chain Config.
+    #[allow(missing_docs)]
```
notlesh, line 207: Was the previous comment wrong? It seems more useful than nothing...

CHANGELOG.md
```diff
@@ -2,8 +2,7 @@
 
 ## Next release
 
-- chore: Move crates under a madara subdir
-- chore(nix): resolve flake and direnv compatibility issues
```
notlesh: Is this intentional?

crates/madara/node/src/cli/chain_config_overrides.rs
```diff
@@ -24,6 +24,10 @@ use url::Url;
 /// Format: "--chain-config-override chain_id=SN_MADARA,chain_name=MADARA,block_time=1500ms,bouncer_config.block_max_capacity.n_steps=100000000"
 #[derive(Parser, Clone, Debug)]
 pub struct ChainConfigOverrideParams {
+    /// Overrides parameters from the chain config.
+    ///
+    /// Use the following syntax:
+    /// --chain-config-override=block_time=30s,pending_block_update_time=2s...
```
notlesh, line 30: It would be good to mention here what the possible keys are, is there something you can reference about that?

notlesh (changes requested):
Left a couple nits/questions, but also wanted to ask if `block time` and `bending block update time` are documented anywhere. I don't see them in `--help`.

`Tick time` was also confusing to me while reading through the PR/issue, is that a user-facing term? If not, no big deal.

crates/madara/node/src/main.rs
Trantorian1: I don't like the use of `panic` as this feels like a correct execution path for the program and not something we should violently abort. The [rust docs](https://doc.rust-lang.org/std/macro.panic.html#when-to-use-panic-vs-result) are a bit unclear regarding this but I feel we shouldn't use it here. Switched the error message to just be `anyhow::bail` since our configuration of `tracing` should print to `stderr` anyway.

crates/madara/node/src/main.rs
Trantorian1: This is being address in #448

crates/madara/node/src/cli/mod.rs
Trantorian1: These docs do not appear in clap due to the use of `#[clap(flatten)]`. Instead, the docs on `ChainConfigOverrideParams` are used which are more extensive. Still, adding those back.

crates/madara/node/src/main.rs
Mohiiit: I see, so you mean developers will be allowed to keep pending tick time > block time?

and yes I agree that we have the same error in the `tracing`, we don't need it in the `anyhow::bail`

crates/madara/node/src/cli/chain_config_overrides.rs
Trantorian1: Added some more docs

crates/madara/node/src/main.rs
Trantorian1: Yes, well more specifically developers will be _required_ to keep tick time >= block time?

Mohiiit: you mean block time >= tick time?

Trantorian1: Oh yes sorry XD

notlesh (approved):

Mohiiit (approved):

PR #448, opened 2024-12-30
fix(devnet): loosened devnet chain id restrictions

### LINKED ISSUE #302: bug(devnet): incorrect error message when chain id isn't devnet

### Description of bug

Currently in the check [here](https://github.com/madara-alliance/madara/blob/1284dc23a5b48e66b1c6010e82b903c0c774a237/crates/node/src/main.rs#L206) we throw an error if the user is using `--devnet` but the chain id isn't devnet. We should instead only throw an error if the chain id is SN_SEPOLIA or SN_MAINNET. Or we can also consider removing the error completely but changing the default chain id to not SN_SEPOLIA or SN_MAINNET.

### Steps to reproduce

_No response_

Trantorian1:
# Pull Request type

- Bugfix

Current restrictions around devnet chain id to avoid replay attacks do not make entire sense and are a bit too strict.

Resolves: #302  

## What is the new behavior?

Devnet chain id restrictions have been loosed to only forbid devnets with `STARKNET_MAINNET` or `STARKNET_SEPOLIA` chain ids. An additinal warning has been added to the `README` to warn agains such potential replay attacks

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix(devnet): loosened devnet chain id restrictions
 - chore: Move crates under a madara subdir
 - chore(nix): resolve flake and direnv compatibility issues
 - fix: Gateway path fix
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ You can choose between different build modes:
 
 Start the Madara client with a basic set of arguments depending on your chosen mode:
 
+> [!NOTE]
+> Head to the [Configuration](#%EF%B8%8F-configuration) section to learn more about
+> customizing your node.
+
 #### Full Node
 
 Synchronizes the state of the chain from genesis.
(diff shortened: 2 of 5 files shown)


crates/node/src/main.rs
```diff
@@ -50,16 +51,13 @@ async fn main() -> anyhow::Result<()> {
...
-        } else {
-            // This log is immediately flooded with devnet accounts and so this can be missed.
-            // Should we add a delay here to make this clearly visisble?
-            tracing::warn!("You're running a devnet with the network config of {:?}. This means that devnet transactions can be replayed on the actual network.", run_cmd.network);
-        }
+    // Check if the devnet is running with the correct chain id. This is purely
+    // to avoid accidental setups which would allow for replay attacks. This is
+    // possible if the devnet has the same chain id as another popular chain,
+    // allowing txs which occur on it to also be replayed on that other chain.
+    if run_cmd.devnet && (chain_config.chain_id == ChainId::Mainnet || chain_config.chain_id == ChainId::Sepolia) {
```
apoorvsadana: do we not want to give an option to force this? like if I really need to run a devnet with sepolia mode just so I can test some script or something else that I have?

apoorvsadana (approved):

crates/node/src/main.rs
Trantorian1: Idk, this seems liek something which would eventually make its way to prod :sweat_smile:. Any situation in particular where you think this could be useful?

apoorvsadana: mostly when I have existing scripts that I use for SN sepolia/mainnet and I want to test the same thing here (given the script has some chain id hardcoding), not sure how big of a use case that is but I think not being able to run a specific chain id at all might be an extreme but wdyt?

heemankv (approved):

notlesh (commented):
This seems fine to me, but I'm wondering why this is a big deal.

If I spin up a dev chain and use some private keys, I should know not to reuse private keys that I'm using on some other chain, and this should be sufficient to prevent any replay attacks of any value, right? Is there more to it?

crates/node/src/main.rs
Trantorian1: Yep, this has been done!

PR #450, opened 2025-01-01
fix(mempool): nonce ordering

### LINKED ISSUE #442: bug(mempool): correctly handle tx from same account

### Description of bug

When the mempool receives multiple transactions from the same account, it currently fails to handle cases where a transaction with nonce (x+1) arrives before the prerequisite transaction with nonce (x). This results in an "invalid nonce" error for the (x+1) transaction.

Expected Behavior:
The transaction with nonce (x+1) should remain in the mempool in a pending state until the transaction with nonce (x) is received and successfully executed. Once the (x) transaction completes, the system should automatically process the waiting (x+1) transaction in the correct sequence.

### Steps to reproduce

1. Identify current nonce (n) for a account (maybe a devnet account)
2. Submit a transaction with nonce (n+1)
3. Observe immediate rejection instead of queuing

Trantorian1:
# Pull Request type

- Bugfix

Transactions added to the mempool are not ordered by Nonce and will not wait for previous nonces. This is an issue as it will lead to _valid_ dependant transactions from the same account being rejected if ever they are received out of order, which can happen due to network latency for example.

Resolves: #442

## What is the new behavior?

Transactions in the mempool are now ordered by readiness. A transaction is ready if its nonce directly follows that of the previous nonce for that account, as stored in db. Otherwise, a transaction is marked as pending. Pending transactions are still subject to removal due to their age.

Yes. The database schema has been updated to store mempool transaction readiness. This does not affect full nodes but will pose problems if migrating a sequencer or devnet node.


crates/madara/client/mempool/src/inner/intent.rs
```diff
@@ -0,0 +1,138 @@
...
+impl<K> Eq for TransactionIntent<K> {}
+
+impl<K> Ord for TransactionIntent<K> {
+    fn cmp(&self, other: &Self) -> cmp::Ordering {
+        // Important: Fallback on contract addr here.
+        // There can be timestamp collisions.
+        self.timestamp
+            .cmp(&other.timestamp)
+            .then_with(|| self.contract_address.cmp(&other.contract_address))
+            .then_with(|| self.nonce.cmp(&other.nonce))
```
notlesh: What is the use of `Ord` here? Sorting by `timestamp` first scares me a bit since it can easily be influenced by users.

notlesh (commented):
This looks good so far, but it deserves a more thorough review than I gave it.

One thing I wanted to point out is that tests for re-submitting with same or lower nonce would be good. The former is used frequently to alter txns in other blockchains (e.g. submit txn with same nonce but higher tip allows optimistic overriding of a txn).

crates/madara/client/mempool/src/inner/intent.rs
Trantorian1: So, a `TransactionIntent` represent the ordering of transactions in the mempool as well as extra information to retrieve the actual transaction.  You can find some docs on the matter in [`intent.rs`](https://github.com/madara-alliance/madara/blob/9dc71855a8878ce460c75de4e596459fbea293e4/crates/madara/client/mempool/src/inner/intent.rs#L1C1-L20C37) and [`mempool/inner/mod.rs`](https://github.com/madara-alliance/madara/blob/9dc71855a8878ce460c75de4e596459fbea293e4/crates/madara/client/mempool/src/inner/mod.rs#L31C1-L133C47). `timestamp` cannot be influenced by the user and only represents the "time at which the transaction was received by the mempool". I'm a bit confused how we should be sorting transactions in the mempool if not by timestamp.

Trantorian1: I would suggest reading the docs for [MempoolInner](https://github.com/madara-alliance/madara/blob/9dc71855a8878ce460c75de4e596459fbea293e4/crates/madara/client/mempool/src/inner/mod.rs#L31C1-L133C47) first as they provide a more general overview of the Mempool. If you IDE supports displaying rustdoc in-line I also suggest performing the review there instead of on github as you should find the code much easier to navigate through the docs.

Trantorian1:
> This looks good so far, but it deserves a more thorough review than I gave it.
> 
> One thing I wanted to point out is that tests for re-submitting with same or lower nonce would be good. The former is used frequently to alter txns in other blockchains (e.g. submit txn with same nonce but higher tip allows optimistic overriding of a txn).

Submitting a lower nonce is allowed, however submitting a tx with the same nonce is not as it results in a nonce conflict. The current behavior is to reject this in the case of user-submitted transactions, but the logic for replacing transactions with the same nonce is already in place and is used by the block production so we could easily update this to be default.

Regarding tests on this matter, [I have already added some](https://github.com/madara-alliance/madara/blob/9dc71855a8878ce460c75de4e596459fbea293e4/crates/madara/client/mempool/src/lib.rs#L1017C5-L1116C2). Generally speaking, you will find all tests for the mempool (including the inner mempool) [here](https://github.com/madara-alliance/madara/blob/9dc71855a8878ce460c75de4e596459fbea293e4/crates/madara/client/mempool/src/lib.rs#L522).

crates/madara/client/mempool/src/inner/intent.rs
notlesh: > I'm a bit confused how we should be sorting transactions in the mempool if not by timestamp.

I think my confusion is that this is shared for both `ready` and `pending`, which are used in very different ways (a "mixed account queue" and "per-account queue", respectively).

For `tx_intent_queue_ready: BTreeSet<TransactionIntentReady>` I think this sort criteria is fine, although you'd eventually want to replace txn inclusion priority with something based on economics instead of arrival time.

But for `tx_intent_queue_pending: HashMap<Felt, BTreeSet<TransactionIntentPending>>` I would imagine `nonce` should be the only sorting you care about. Sorting by timestamp could be wrong (if you allowed replacing txns, for example).

Big621 (commented):
Thank you

crates/madara/client/mempool/src/inner/intent.rs
Trantorian1: Yep, you're right, thanks for catching that one!

jbcaron (approved):

crates/madara/client/db/src/mempool_db.rs
```diff
@@ -3,10 +3,57 @@ use crate::{Column, MadaraBackend, MadaraStorageError};
 use mp_class::ConvertedClass;
 use rocksdb::IteratorMode;
 use serde::{Deserialize, Serialize};
+use starknet_api::core::Nonce;
 use starknet_types_core::felt::Felt;
 
 type Result<T, E = MadaraStorageError> = std::result::Result<T, E>;
 
+/// A nonce is deemed ready when it directly follows the previous nonce in db
+/// for a contract  address. This guarantees that dependent transactions are not
```
cchudant:
```suggestion
/// for a contract address. This guarantees that dependent transactions are not
```

crates/madara/client/db/src/mempool_db.rs
```diff
@@ -17,28 +64,33 @@ pub struct SavedTransaction {
...
     converted_class: &'a Option<ConvertedClass>,
+    nonce_info: &'a NonceInfo,
 }
-#[derive(Serialize, Deserialize)]
-struct TransactionWithConvertedClass {
-    tx: SavedTransaction,
-    converted_class: Option<ConvertedClass>,
+
+#[derive(Deserialize)]
+/// This struct is used as a templace to deserialize Mempool transactions from
```
cchudant:
```suggestion
/// This struct is used as a template to deserialize Mempool transactions from
```

crates/madara/client/mempool/src/inner/intent.rs
```diff
@@ -0,0 +1,233 @@
...
+impl PartialOrd for TransactionIntentPendingByTimestamp {
+    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// An [intent] to be consumed by the [Mempool].
+///
+/// This data struct will check [timestamp], [contract_address] and [nonce]
+/// (in that oreder) for equality. [nonce_next] is not considered as it should
```
cchudant:
```suggestion
/// (in that order) for equality. [nonce_next] is not considered as it should
```

crates/madara/client/mempool/src/inner/intent.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// The contract responsible for sending the transaction.
+    pub(crate) contract_address: Felt,
+    /// Time at which the transaction was received by the mempool.
+    pub(crate) timestamp: ArrivedAtTimestamp,
+    /// The [Nonce] of the transaction associated to this intent. We use this
+    /// for retrieval purposes later on.
+    pub(crate) nonce: Nonce,
+    /// This is the [Nonce] of the transaction right after this one. We
+    /// precompute this to avoid making calculations on a [Felt] in the hot
+    /// loop, as this can be expensive.
```
cchudant, line 213: > as this can be expensive

is that true? this is essentially a Felt increment right?
mmmh

I think i'm fine with nonce_next for now but i would love to switch the mempool nonces u64 at some point.

crates/madara/client/mempool/src/inner/mod.rs
```diff
@@ -3,62 +3,191 @@
...
+///
+/// ## [Ready]
+///
+/// FIFO queue. We use a [BTreeSet] to maintain logarithmic complexity and high
+/// performance with low reordering of the memory even in the case of very high
+/// transaction throughput.
+///
+/// ## [Pending]
+///
+/// FIFO queue. The queue is distributed across all current contract addresses
```
cchudant: queue is distributed?
i think we should rephrase, like, "the queue has an entry per contract address, not transaction hash"

crates/madara/client/mempool/src/inner/mod.rs
```diff
@@ -3,62 +3,191 @@
...
+/// in the mempool, with each contract address having a [BTreeMap] queue mapped
+/// to it. We do this to have access to [BTreeMap::entry] which avoids a double
+/// lookup in [pop_next] when moving pending intents to the ready queue. Intents
+/// in this queue are ordered by their [Nonce].
+///
+/// While this is handy to retrieve the tx with the next nonce for a particular
+/// contract, it is a performance bottleneck when removing age exceeded pending
+/// transaction. For this reason, we keep a [separate ordering] of all pending
+/// transactions, sorted by their time of arrival.
```
cchudant, line 66: right, okay that makes more sense

crates/madara/client/mempool/src/inner/mod.rs
```diff
@@ -3,62 +3,191 @@
...
-    nonce_chains: HashMap<Felt, NonceChain>,
-    /// FCFS queue.
-    tx_queue: BTreeSet<AccountOrderedByTimestamp>,
-    deployed_contracts: DeployedContracts,
+#[cfg_attr(any(test, feature = "testing"), derive(Clone))]
+pub struct MempoolInner {
+    /// We have one [Nonce] to  [MempoolTransaction] mapping per contract
+    /// address.
+    ///
+    /// [Nonce]: starknet_api::core::Nonce
```
cchudant, line 153: i thought we didnt want to use the starknet api types wherever possible, cc @jbcaron ?

crates/madara/client/mempool/src/inner/mod.rs
```diff
@@ -3,62 +3,191 @@
...
+    /// This is required for the rapid removal of age-exceeded txs in
+    /// [remove_age_exceeded_txs] and must be kept in sync with
+    /// [tx_intent_queue_pending_by_nonce].
+    ///
+    /// [pending]: TransactionIntentPendingByTimestamp
+    /// [remove_age_exceeded_txs]: Self::remove_age_exceeded_txs
+    /// [tx_intent_queue_pending_by_nonce]: Self::tx_intent_queue_pending_by_nonce
+    // TODO: can remove nonce_next from TransactionIntentPendingByTimestamp
+    pub(crate) tx_intent_queue_pending_by_timestamp: BTreeSet<TransactionIntentPendingByTimestamp>,
+    /// A count of all deployed contract declared so far.
```
cchudant:
```suggestion
    /// List of all new deployed contracts currently in the mempool.
```

crates/madara/client/mempool/src/inner/mod.rs
```diff
@@ -121,53 +339,167 @@ impl MempoolInner {
...
+                        if let ReplacedState::Replaced { previous } = replaced {
+                            let removed = self.tx_intent_queue_ready.remove(&TransactionIntentReady {
+                                contract_address,
+                                timestamp: previous.arrived_at,
+                                nonce: nonce_info.nonce,
+                                nonce_next: nonce_info.nonce_next,
+                                phantom: std::marker::PhantomData,
+                            });
+                            debug_assert!(removed);
+                            self.limiter.mark_removed(&TransactionCheckedLimits::limits_for(&previous));
```
cchudant, line 383: we need to decrement self.deployed_contracts here if it's replaced?
this is kinda sus we need to proptest that

cchudant: To be clear, the DeployedContracts thing is because of a very specific and weird corner case:
We need to support cases where clients send a DeployedAccount transaction followed by Invoke transactions, and in that case we cannot run the __validate__ of the outer mempool. We have to skip that validation step it in that case, for better UX.
The Argent wallet notably does that when sending a transaction from a wallet that's not yet deployed.
I am thinking of reconsidering how we deal with that corner case, especially if the invoke tx arrives before the deploy account? (could it be possible in p2p? idk?)

cchudant: (we should add a comment about why it's there)

cchudant (changes requested):
I will review this more in depth tomorrow morning, but it looks very good.

Big621 (approved):

crates/madara/client/mempool/src/inner/intent.rs
Trantorian1: [Here](https://github.com/lambdaclass/lambdaworks/blob/6fb54a22f74f13c216a96b59ca209ed28a8d55af/math/src/field/fields/montgomery_backed_prime_fields.rs#L121C5-L135C6) is the implementation of `add` for `MontgomeryBackendPrimeField`, which is what [`Stark252PrimeField`](https://github.com/lambdaclass/lambdaworks/blob/6fb54a22f74f13c216a96b59ca209ed28a8d55af/math/src/field/fields/fft_friendly/stark_252_prime_field.rs#L17) relies on (`Stark252PrimeField` relies on [`U256PrimeField`](https://github.com/lambdaclass/lambdaworks/blob/6fb54a22f74f13c216a96b59ca209ed28a8d55af/math/src/field/fields/montgomery_backed_prime_fields.rs#L16) which is a type alias to [`MontgomeryBackendPrimeField`](https://github.com/lambdaclass/lambdaworks/blob/6fb54a22f74f13c216a96b59ca209ed28a8d55af/math/src/field/fields/montgomery_backed_prime_fields.rs#L34)):

```rust
fn add(a: &Self::BaseType, b: &Self::BaseType) -> Self::BaseType {
    let (sum, overflow) = UnsignedInteger::add(a, b);
    if Self::MODULUS_HAS_ONE_SPARE_BIT {
        if sum >= M::MODULUS {
            sum - M::MODULUS
        } else {
            sum
        }
    } else if overflow || sum >= M::MODULUS {
        let (diff, _) = UnsignedInteger::sub(&sum, &M::MODULUS);
        diff
    } else {
        sum
    }
}
```

And [here](https://github.com/lambdaclass/lambdaworks/blob/6fb54a22f74f13c216a96b59ca209ed28a8d55af/math/src/unsigned_integer/element.rs#L588C5-L603C6) is the implementation of `add` for `UnsignedInteger`:

```rust
pub const fn add(
    a: &UnsignedInteger<NUM_LIMBS>,
    b: &UnsignedInteger<NUM_LIMBS>,
) -> (UnsignedInteger<NUM_LIMBS>, bool) {
    let mut limbs = [0u64; NUM_LIMBS];
    let mut carry = 0u64;
    let mut i = NUM_LIMBS;
    while i > 0 {
        let (x, cb) = a.limbs[i - 1].overflowing_add(b.limbs[i - 1]);
        let (x, cc) = x.overflowing_add(carry);
        limbs[i - 1] = x;
        carry = (cb | cc) as u64;
        i -= 1;
    }
    (UnsignedInteger { limbs }, carry > 0)
}
```

Supposing this is inlined (hopefully this is the case, as both impls are part of the same crate), we are still looking at the following for a simple add:

```rust
fn add(a: &Self::BaseType, b: &Self::BaseType) -> Self::BaseType {
    let mut limbs = [0u64; NUM_LIMBS];
    let mut carry = 0u64;
    let mut i = NUM_LIMBS;
    while i > 0 {
        let (x, cb) = a.limbs[i - 1].overflowing_add(b.limbs[i - 1]);
        let (x, cc) = x.overflowing_add(carry);
        limbs[i - 1] = x;
        carry = (cb | cc) as u64;
        i -= 1;
    }
    let (sum, overflow) = (UnsignedInteger { limbs }, carry > 0);
    if Self::MODULUS_HAS_ONE_SPARE_BIT {
        if sum >= M::MODULUS {
            sum - M::MODULUS
        } else {
            sum
        }
    } else if overflow || sum >= M::MODULUS {
        let (diff, _) = UnsignedInteger::sub(&sum, &M::MODULUS);
        diff
    } else {
        sum
    }
}
```

My worry was that performing this over a very large number of transactions each insertion or removal might end up being expensive.

crates/madara/client/mempool/src/inner/mod.rs
Trantorian1: Tbf I was just getting tired of having to switch between `Nonce` and `Felt` and just ended wrapping everything in `Nonce` :sweat_smile:. Tbf this is a bit of a shame though as for some reason `Nonce` is not `#[repr(transparent)]`

crates/madara/client/mempool/src/inner/mod.rs
Trantorian1: This is already being proptested :D `deployed_contracts` is checked in the impl of `CheckInvariants` for `MempoolInner`, [here](https://github.com/madara-alliance/madara/blob/77fc88f145cb1a0edff8b8bfff9884d84766ff8b/crates/madara/client/mempool/src/inner/mod.rs#L219C13-L225C13) and [here](https://github.com/madara-alliance/madara/blob/77fc88f145cb1a0edff8b8bfff9884d84766ff8b/crates/madara/client/mempool/src/inner/mod.rs#L263C17-L269C18), and this is [called by the proptest](https://github.com/madara-alliance/madara/blob/77fc88f145cb1a0edff8b8bfff9884d84766ff8b/crates/madara/client/mempool/src/inner/proptest.rs#L345) each time a state transition is applied to the System Under Test. The conditional incrementation of `deployed_contracts` is actually the fix to a bug which was found by proptesting :) You can find a [unit test](https://github.com/madara-alliance/madara/blob/77fc88f145cb1a0edff8b8bfff9884d84766ff8b/crates/madara/client/mempool/src/lib.rs#L811C5-L815C64) for this specific case which might provide more context.

crates/madara/client/mempool/src/inner/intent.rs
cchudant: sure okay

cchudant (changes requested):
Fix deployed contracts invariant checking, see this commit
https://github.com/madara-alliance/madara/commit/f801aa025e8321be9a2ea28a5d147c01c8539632

which makes the proptest fail

cchudant:
I think other than that it's fine, but the complexity of all the invariants now makes me a bit worried..

Trantorian1:
> Fix deployed contracts invariant checking, see this commit [f801aa0](https://github.com/madara-alliance/madara/commit/f801aa025e8321be9a2ea28a5d147c01c8539632)
> 
> which makes the proptest fail

Just pushed a fix to this!

Trantorian1:
> I think other than that it's fine, but the complexity of all the invariants now makes me a bit worried..

Ehh I don't like it too much either but this also just an inherently complex part of the codebase :/

cchudant (approved):
:+1:

PR #456, opened 2025-01-08
fix(block_production): dynamic block closing

Mohiiit:
## Summary
This PR ensures that madara correctly adds a special address to the state diff when dynamically closing a block with a number greater than 10. The special address uses the current block number minus 10 as the key and the hash of (current block - 10) as the value. This is required for the SNOS.

## Pull Request Type
- **Bugfix**

## Current Behavior
When the bouncer configuration reaches its limit, Madara dynamically closes the block. However, during this process, the required special address is not added to the state diff for blocks numbered greater than 10.

## New Behavior
When closing a block dynamically, if the block number is greater than 10, a special address is now correctly added to the state diff with:
- **Key:** Current block number - 10
- **Value:** Hash of (current block - 10)

## Breaking Changes
No breaking changes introduced.

## Additional Notes
This fix ensures consistency in state diffs when dynamic block closing occurs, maintaining expected behavior for blocks greater than 10.

### DIFF
--- a/crates/madara/client/block_production/src/lib.rs
+++ b/crates/madara/client/block_production/src/lib.rs
@@ -358,7 +358,28 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         Ok(())
     }
 
-    fn maybe_add_prev_block_hash(&self, state_diff: &mut StateDiff, block_n: u64) -> Result<(), Error> {
+    /// Updates the state diff to store a block hash at the special address 0x1, which serves as
+    /// Starknet's block hash registry.
+    ///
+    /// # Purpose
+    /// Address 0x1 in Starknet is a special contract address that maintains a mapping of block numbers
+    /// to their corresponding block hashes. This storage is used by the `get_block_hash` system call
+    /// and is essential for block hash verification within the Starknet protocol.
+    ///
+    /// # Storage Structure at Address 0x1
+    /// - Keys: Block numbers
+    /// - Values: Corresponding block hashes
+    /// - Default: 0 for all other block numbers
+    ///
+    /// # Implementation Details
+    /// For each block N ≥ 10, this function stores the hash of block (N-10) at address 0x1
+    /// with the block number as the key.
+    ///
+    /// For more details, see the [official Starknet documentation on special addresses]
+    /// (https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-state/#address_0x1)
+    ///
+    /// It is also required by SNOS for PIEs creation of the block.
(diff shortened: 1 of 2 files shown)


crates/madara/client/block_production/src/lib.rs
```diff
@@ -358,20 +358,45 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         Ok(())
     }
 
+    fn maybe_add_prev_block_hash(&self, state_diff: &mut StateDiff, block_n: u64) -> Result<(), Error> {
```
Trantorian1: Maybe add a doc comment to this function explaining why this is needed and what it does.

crates/madara/client/block_production/src/lib.rs
```diff
@@ -384,16 +409,19 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
 
         // Check if block is full
         if block_now_full {
+            let block_n = self.block_n();
+            self.maybe_add_prev_block_hash(&mut new_state_diff, block_n)?;
+
             tracing::info!("Resource limits reached, closing block early");
-            self.close_and_prepare_next_block(state_diff, visited_segments, start_time).await?;
+            self.close_and_prepare_next_block(new_state_diff, visited_segments, start_time).await?;
```
Trantorian1: This means the block might at any time contain on more transaction than the block limit. Are we okay with this behavior or is it worth updating that part of the code as well?

crates/madara/client/block_production/src/lib.rs
jbcaron: The current name `maybe_add_prev_block_hash` is somewhat ambiguous and doesn't fully capture the function's role in the state management process.

crates/madara/client/block_production/src/lib.rs
Mohiiit: sorry didn't understand the comment.

crates/madara/client/block_production/src/lib.rs
Mohiiit: does it make more sense now? or any suggestion what can be improved?

Mohiiit: added

Trantorian1 (approved):
Seems good! :+1:

crates/madara/client/block_production/src/lib.rs
jbcaron: perfect with documentation 👍

jbcaron (approved):

PR #459, opened 2025-01-09
fix(primitives): limit legacy class sizes

cchudant:
## Pull Request type

- Bugfix

Sending a legacy (cairo0) class with a zip-bomb in it could potentially DoS and OOM the node.

## What is the new behavior?

Uncompressed size is now limited

No

### DIFF
--- a/crates/madara/primitives/class/src/convert.rs
+++ b/crates/madara/primitives/class/src/convert.rs
@@ -6,12 +6,10 @@ use starknet_core::types::{
     },
     CompressedLegacyContractClass,
 };
-use std::io::Read;
+use std::io::{self, Read};
 
 #[derive(Debug, thiserror::Error)]
 pub enum ParseCompressedLegacyClassError {
-    #[error("I/O error: {0}")]
-    IoError(#[from] std::io::Error),
     #[error("JSON parse error: {0}")]
     JsonError(#[from] serde_json::Error),
     #[error("Unexpected legacy compiler version string")]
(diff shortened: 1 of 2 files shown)


crates/madara/primitives/class/src/convert.rs
```diff
@@ -84,3 +84,49 @@ fn parse_legacy_entrypoint(entrypoint: &LegacyContractEntryPoint, pre_0_11_0: bo
...
+impl<R: Read> Read for ReadSizeLimiter<R> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if self.limit == 0 {
+            // check if the inner read still has data for us
+            if self.inner.read(&mut [0])? > 0 {
+                return Err(io::Error::new(io::ErrorKind::Other, InputTooLarge));
+            }
+        }
+
+        let max = u64::min(buf.len() as u64, self.limit) as usize;
```
Trantorian1, line 112: The use of `max` along with the `min` function is a bit confusing here. I get what this does but maybe a name `to_take` would be clearer?

Trantorian1 (approved):

crates/madara/primitives/class/src/convert.rs
cchudant: i stole that code from the standard library, but i agree with you

Mohiiit (approved):

antiyro (approved):

PR #462, opened 2025-01-14
chore(docker): fix build and signal

jbcaron:
## Pull Request type

Please add the labels corresponding to the type of changes your PR introduces:

- Bugfix
- Build-related changes

## What is the current behavior?

Resolves: #NA

## What is the new behavior?

-

## Does this introduce a breaking change?

No

PR #467, opened 2025-01-16
Missing .db-versions.yml in Dockerfile

ftheirs:
The `.db-versions.yml` file is missing in the docker image and the build process will fail without it

## Pull Request type
- Bugfix

To reproduce the issue, just clone Madara repository and run: 
`docker build -t madara .`
The process will stop at the end of stage 1 since the compilation fails.


Mohiiit:
has been fixed here: https://github.com/madara-alliance/madara/pull/462

ftheirs:
Thanks @Mohiiit I haven't seen it 💪

PR #468, opened 2025-01-16
feat(pending): pending block sealed instead of re-added to mempool

### LINKED ISSUE #458: feat(blockprod): Revised Approach for Madara Block Production Shutdown
## Description  

This issue outlines a revised approach to handle block production shutdowns in Madara, improving on our previous proposal. The main goal is to eliminate transaction re-execution to preserve receipt consistency and ensure efficient handling of both graceful and ungraceful shutdown scenarios.  

The revised approach shifts block sealing logic to the restart phase and introduces configuration-based decision-making, allowing for seamless block production handling even when configurations change.  

Key changes in this proposal include:  
1. **No Block Sealing During Shutdown** (whether graceful or ungraceful).  
2. **On Restart:**  
   - seal the block irrespective of the config, because closing the block doesn't depend on the configs.  

---

### Key Changes  
- Avoid sealing blocks during both graceful and ungraceful shutdowns.  
- On restart, seal the block and make sure we have updated the db for the pending txns and classes etc.  

### Data Persistence Requirements  
- Persist pending transactions and their receipts in the database.  
- Do not clear the database on restart to avoid unnecessary transaction re-execution.  

### Pseudo-Code Overview  

```rust
async fn handle_restart() {
    seal_the_last_block();
    start_an_empty_block();
}

### PR DESCRIPTION
**Pull request by Trantorian1**
> [!IMPORTANT]
> Most of the changes in this pr are just me adding some tests. **Actual logic changes are pretty small.**

## Pull Request type

- Feature

Pending block transactions are re-added back to the mempool and re-executed upon startup. This happens in case of graceful or ungraceful shutdown in the middle of block production: the pending block will be saved to db to avoid loss of state.

This leads to several issues:

1. Transactions are re-executed, and might provided different outputs or hashes.
2. Transactions from the pending block are treated like any other and could be dropped by the mempool under certain edge cases.

Resolves: #458 

## What is the new behavior?

Pending block, alongside the pending state diff, visited segments and declared classes are retrieved from db and closed into a new finalized block.

This means that if a sequencer node is shutdown mid-production, the state contained in the pending block will automatically be persisted upon startup as it own _finalized_ block without re-executing transactions. This also guarantees that any changes to the config will not be applied to transactions executed under different parameters (these will only affect the next block).

No.

### CODE REVIEW

File: crates/madara/client/block_production/src/lib.rs [hunk 1]
```diff
@@ -693,4 +909,773 @@ mod tests {
...
+
+    /// This test makes sure that if a pending block is already present in db
+    /// at startup, then it is closed and stored in db.
+    ///
+    /// This happens if a full node is shutdown (gracefully or not) midway
+    /// during block production.
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn block_prod_pending_close_on_startup_pass(
```
Mohiiit, line 921: beautiful

crates/madara/client/block_production/src/lib.rs
```diff
@@ -693,4 +909,773 @@ mod tests {
...
+            .expect("Missing latest block");
+
+        assert_eq!(block.info.as_nonpending().unwrap().header.parent_block_hash, Felt::ZERO);
+        assert_eq!(block.inner, pending_inner);
+
+        let state_diff = backend
+            .get_block_state_diff(&mp_block::BlockId::Tag(mp_block::BlockTag::Latest))
+            .expect("Failed to retrieve latest state diff from db")
+            .expect("Missing latest state diff");
+        assert_eq!(state_diff, state_diff);
```
Mohiiit: we should equate with pending_state_diff here right?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -693,4 +909,773 @@ mod tests {
...
+        // ready blocks
+    }
+
+    /// This test makes sure that if a pending block is already present in db
+    /// at startup, then it is closed and stored in db on top of the latest
+    /// block.
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn block_prod_pending_close_on_startup_pass_on_top(
```
Mohiiit, line 1089: isn't this test very similar to the one above?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -693,4 +909,773 @@ mod tests {
...
+        // visited segments and bouncer weights are currently not stored for in
+        // ready blocks
+    }
+
+    /// This test makes sure that closing the pending block from db will fail if
+    /// the pending state diff references a non-existing class.
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn block_prod_pending_close_on_startup_fail_missing_class(
```
Mohiiit, line 1522: do we have test case which ensures that the state_diff in the pending state is getting updated with the class?

Mohiiit (commented):
Great work! The test cases are really clean and well-structured. I have a couple of thoughts:

 - Should we consider adding E2E tests for this? For example, we could simulate spawning Madara, shutting it down mid-process, and then restarting it to verify that everything works as expected.
- I found the stages a bit confusing. In most test cases, when we store the pending block in stage 2, it’s mentioned that this simulates stopping and restarting the node. However, isn’t that more related to the block production task? Maybe I’m misunderstanding, but this part could use some clarification.

Trantorian1:
> * I found the stages a bit confusing. In most test cases, when we store the pending block in stage 2, it’s mentioned that this simulates stopping and restarting the node. However, isn’t that more related to the block production task? Maybe I’m misunderstanding, but this part could use some clarification.

So, the block production functions by storing un-finalized blocks as pending. This is the only form of data we can recover without re-execution as everything else is stored in RAM (mempool transactions which have not yet been polled yet are also stored in db for retrieval, but these haven't been executed anyways). This means that if ever the node crashes, we will only be able to retrieve whatever data was stored in the pending block. This is done atomically so we never commit partial data to the database and only a full pending block can ever be stored.

We are therefore "simulate[ing] stopping and restarting the node", since:

1. This is the only pending data that can persist a node restart, and it cannot be partially valid (we still test failing cases though).
2. Upon restart, this is what the block production would be looking to seal.

I will be adding some if this in comments to the tests.

crates/madara/client/block_production/src/lib.rs
Trantorian1: Yep good catch :D

crates/madara/client/block_production/src/lib.rs
Trantorian1: Yep! These are slightly different edge cases though:

- in `block_prod_pending_close_on_startup_pass`, we are simulating storing the pending block with nothing else in db (ie: the pending block is our genesis block).
- in `block_prod_pending_close_on_startup_pass_on_top`, we are simulating storing the pending block with previous blocks in db.

This is significant, as it is possible to fudge the block number when storing the pending block and override existing blocks in db. I really don't want this to happen which is why this test is here.

crates/madara/client/block_production/src/lib.rs
Trantorian1: Yep, that would be `block_prod_pending_close_on_startup_pass` and `block_prod_pending_close_on_startup_pass_on_top`.

crates/madara/client/block_production/src/lib.rs
Mohiiit: ahh I see, make sense, in that case, we should test the state-diff of the latest block - 1 as well?

Trantorian1: Yep that's a good point, I will be adding this as well :+1:

Mohiiit (approved):

jbcaron (approved):

PR #470, opened 2025-01-20
fix(l1_receipt): trimmed hash

### LINKED ISSUE #469: bug(l1_receipt): trimmed hash

- [x] I have searched the existing issues

### Description of bug

Currently we convert H256 to a string when return the l1 receipt hash over [here](https://github.com/madara-alliance/madara/blob/429e3e71071c4b6115ce4bfbd9231168edba5ec7/crates/madara/primitives/receipt/src/to_starknet_types.rs#L62). However, the to string implementation from `primitive-types` removed the middle part of the hash over [here](https://github.com/paritytech/parity-common/blob/d54ac847cc5447941390b5e9e5e3baef436e1451/fixed-hash/src/hash.rs#L224). This caused the receipt to be broken.

### Steps to reproduce

1. Send L1 handler txn
2. Fetch receipt

Trantorian1:
## Pull Request type

- Bugfix

Resolves: #469

## What is the new behavior?

Correctly converts the `H256` message hash of an `L1HandlerTransactionReceipt` to a `String`.

No.

### DIFF
--- a/crates/madara/primitives/receipt/src/to_starknet_types.rs
+++ b/crates/madara/primitives/receipt/src/to_starknet_types.rs
@@ -59,7 +59,9 @@ impl L1HandlerTransactionReceipt {
         finality_status: starknet_types_rpc::TxnFinalityStatus,
     ) -> starknet_types_rpc::L1HandlerTxnReceipt<Felt> {
         starknet_types_rpc::L1HandlerTxnReceipt::<Felt> {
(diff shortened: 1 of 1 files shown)


jbcaron (approved):

apoorvsadana (commented):
Thanks for this quick PR! Should we add a test case/modify an existing one to be able to catch this?

apoorvsadana (approved):

PR #471, opened 2025-01-21
chore: Merge Orchestrator history

notlesh:
**This PR was closed in favor of #480**

This PR pulls in the entire [madara-orchestrator](https://github.com/madara-alliance/madara-orchestrator) git history into this project.

Because the entire history was pulled in (using `git merge --allow-unrelated-histories`) the PR is absolutely massive (and probably difficult to review). My process was basically this:

1. `git merge --allow-unrelated-histories` (that is, pull in the entire codebase history)
2. Move files (crates under ~~`crates/orchestrator`~~ `orchestrator/crates/` and project files to a temp dir)
3. Update `Cargo.toml` to build (note that `orchestrator` is not a member of `default-members`, disagree if you want!)
4. Fight dependency conflicts

### Status :construction: 

So far this builds and `madara` appears to be unaffected. The project files for `madara-orchestrator` were kept, but were moved (temporarily) to `old_orchestrator_files/`. This means that things like CI will not currently work.

**TODO:**
 - [ ] Migrate orchestrator workflows (#472)
 - [ ] Ensure minimal build and test in github workflows (filter by path, etc.)
 - [ ] Test (and fix) orchestrator build/docker/etc

**Some follow up work:**
* Update READMEs
* Clean up and remove `old_orchestrator_files/`

### Repository forks :fork_and_knife: 

This required forking several repositories in order to avoid version conflicts. Each fork uses the same `madara-monorepo` branch name. These repos were forked:

* [cairo-vm](https://github.com/Moonsong-Labs/cairo-vm/tree/madara-monorepo)
* [sequencer](https://github.com/Moonsong-Labs/sequencer/tree/madara-monorepo)
* [zaun](https://github.com/notlesh/zaun/tree/madara-monorepo)
* [snos](https://github.com/keep-starknet-strange/snos/tree/madara-monorepo)


.env.example
Trantorian1, line 1: Shouldn't this be moved to `crates/orchestrator` if it is related to that?

.env.test
Trantorian1, line 1: Same here

pathfinder
Trantorian1, line 1: Why is this here?

scripts/artifacts/README.md
Trantorian1, line 1: Are we sure we want to strore artefacts under scripts?

.github/workflows/rust-build.yml
Trantorian1, line 1: We probably want to figure out a way to compile only SNOS or Madara if we change only one of those no? At least I feel like testing Madara against SNOS should be a separate and last step in the CI if all other steps pass. And make it so that if changes are made to SNOS we do not run e2e on Madara.

.github/workflows/e2e-test.yml
Trantorian1, line 1: I feel like we should tests Madara and SNOS in different CI actions

bootstrap_input.json
Trantorian1, line 1: Imo this should also be under `crates/ortchestrator`. Given we are making this a monorepo we probably also want to move some Madara-specific files such as the `README` under `crates/madara`

Cargo.toml
```diff
@@ -139,7 +153,7 @@ mc-devnet = { path = "crates/madara/client/devnet" }
 m-cairo-test-contracts = { path = "crates/madara/cairo-test-contracts" }
 
 # Starknet dependencies
-cairo-vm = "=1.0.1"
+cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "madara-monorepo" }
```
Trantorian1, line 156: Instead of using forks I feel like we should be importing two separate versions of the same crate under an alias so that madara and the orchestrator can both have their own version. This feels more maintainable long term as far as supporting new features is concerned, and should also make it easier to merge both into a single dependency in the future.

Trantorian1 (changes requested):
Given we are moving to a monorepo, I feel like we need to rethink what files we have at the root of the project. We probably want to move any configuration files and `README`s to either `crates/madara` or `crates/orchestrator` and keep only common logic at the root. 

We also need to think about how we want to run the CI for each project so that this doesn't take for ever on simple changes.

PR #472, opened 2025-01-22
Orchestrator CI Merge

notlesh:
This PR merges in the `madara-orchestrator` github workflows. Part of #471.

**TODO:**
 - [ ] Fix all Madara CI
 - [ ] Fix all Orchestrator CI
 - [ ] Conditionally run CI based on what changed

### Conditionally running CI

Ideally, we can reliably detect when something changed that requires certain parts of CI to run. Orchestrator does not need to run if only Madara changed and vise-versa.

That's the theory, anyway. It could be argued that there is value in running e.g. Orchestrator when Madara changes, as changes in Madara could cause breakage in Orchestrator. The inverse is probably not true. This is a *lot* of CI workload, however, so a compromise might be to only do this when merging into `main` or when doing a release.


Trantorian1:
Is this pr here to test the CI? Wondering why we don't integrate this into #471 directly

notlesh:
> Is this pr here to test the CI? Wondering why we don't integrate this into #471 directly

Yes, it's to test CI without making a mess in the original PR.

notlesh:
Closing in favor of #480.

PR #473, opened 2025-01-23
feat: store event bloom filter in DB

jbcaron:
## Pull Request type

- Feature

Currently, the event pagination mechanism does not correctly handle the continuation token

Additionally, event filtering does not leverage a bloom filter, which could improve performance by reducing unnecessary block lookups.

Resolves: #NA

## What is the new behavior?

- **Integration of `mp-bloom-filter` for event filtering**:
  - Added the `mp-bloom-filter` crate to improve filtering efficiency.
  - Uses a bloom filter to quickly check if a block might contain relevant events before fetching it.
  - This optimization reduces the number of blocks read, improving query performance.

- **Optimized event retrieval**:
  - Introduced `take(chunk_size + 1)` in the filtering process to determine whether the current block has more events.
  - Ensures efficient pagination by accurately detecting when a block is fully processed.

Yes

- The continuation token format has changed.
- New DB column for bloom filter


crates/madara/client/db/src/lib.rs
```diff
@@ -241,6 +245,12 @@ impl Column {
     }
 }
 
+#[cfg(test)]
+#[test]
+fn test_column_all() {
+    assert_eq!(Column::ALL.len(), Column::NUM_COLUMNS);
```
Trantorian1: Is this really useful since `NUM_COLUMNS` is already defined as `Self::ALL.len();`?

crates/madara/client/rpc/src/versions/user/v0_7_1/methods/read/get_events.rs
```diff
@@ -94,7 +113,12 @@ pub async fn get_events(
 
             return Ok(EventsChunk { events: filtered_events, continuation_token: token });
         }
+        if current_block == to_block {
+            break;
+        }
     }
+    // TODO: handle the case where 'to_block' is pending
```
Trantorian1: Do we want this to be part of this pr?

crates/madara/primitives/bloom_filter/benches/bloom_benchmark.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    // group.bench_function("non atomic", |b| {
+    //     b.iter(|| {
+    //         let mut filter = BloomFilter::<H, BitStore>::optimal(TEST_DATA_SIZE, FALSE_POSITIVE_RATE)
+    //             .expect("Failed to create optimal filter");
+    //         for item in test_data {
+    //             filter.add(black_box(item));
+    //         }
+    //     });
+    // });
```
Trantorian1: Remove this?
```suggestion
```

crates/madara/primitives/bloom_filter/benches/bloom_benchmark.rs
```diff
@@ -0,0 +1,192 @@
...
+        get_hasher_benchmarks::<XxHash64>("XxHash64"),
+    ];
+
+    for hasher in &hashers {
+        (hasher.sequential_insertion)(c, &test_data, hasher.name);
+        (hasher.parallel_insertion)(c, &test_data, hasher.name);
+        (hasher.lookup)(c, &test_data, hasher.name);
+        (hasher.parallel_lookup)(c, &test_data, hasher.name);
+    }
+}
```
Trantorian1: Really cool testing structure!

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -0,0 +1,144 @@
...
+
+use crate::{AtomicBitStore, BitStore, BloomFilter, PreCalculatedHashes};
+
+/// Number of hash functions used in the Bloom filter.
+/// The value 7 is optimal for a false positive rate of 1% (0.01).
+/// Formula used: k = -ln(p)/ln(2) ≈ 7
+/// where:
+/// - k is the number of hash functions
+/// - p is the desired false positive rate (0.01)
+/// Reference: [Bloom, B. H. (1970). "Space/Time Trade-offs in Hash Coding with Allowable Errors"](https://dl.acm.org/doi/10.1145/362686.362692)
```
Trantorian1: Thanks!

crates/madara/primitives/bloom_filter/src/tests/utils.rs
```diff
@@ -0,0 +1,20 @@
...
+// Constants for Bloom filter configuration
+// Number of hash functions used in the Bloom filter.
+// The value 7 is optimal for a false positive rate of 1% (0.01).
+pub const HASH_COUNT: u8 = 7;
+
+// Number of bits per element in the Bloom filter.
+// The value 9.6 is optimal for a false positive rate of 1% (0.01).
+const BITS_PER_ELEMENT: f64 = 9.6;
+
+pub const FALSE_POSITIF_RATE: f64 = 0.01;
```
Trantorian1: These constants appear several times in the codebase, maybe we want to define them only once and re-use them?

Trantorian1 (approved):

crates/madara/client/db/src/lib.rs
jbcaron: just in case of update

crates/madara/primitives/bloom_filter/benches/bloom_benchmark.rs
jbcaron: done

crates/madara/primitives/bloom_filter/src/tests/utils.rs
jbcaron: I agree. I’d also like to keep testing and production separate. Each should have its own constants.

HermanObst:
Hey there! I made a first quick pass on the PR and I have 2 first questions:
1. Why do we need to introduce a bloom filter? Did we were facing low performance in get_events RPC? 
2. Why are we implementing the bloom filter from scratch and not using something like [this](https://crates.io/crates/bloomfilter)?

Trantorian1:
> Why do we need to introduce a bloom filter? Did we were facing low performance in get_events RPC?

Yes, `getEvents` was by far our lowest performing RPC method, and was lagging behind other clients' implementations by several orders of magnitude.

Here is a graph of our `getEvents` performance compared to other nodes (this was part of a perf test I made earlier in 2024, no other changes have been made to `getEvents` since then).
image

If ever you are interested, you can find the tool I developed for this bench [here](https://github.com/madara-alliance/madara-bench).

Trantorian1:
> Why are we implementing the bloom filter from scratch and not using something like [this](https://crates.io/crates/bloomfilter)?

Why not? This gives us much more control over our codebase and its performance characteristics. In particular, as I remember @jbcaron, the goal was to create a bloom filter with a constant error rate independant of the block size. That and other optimizations might not have been possible had this not been implemented by hand.

Trantorian1:
It is also worth noting that we would not be the only team to be developing our own bloom filter, this is already the case for [Pathfinder](https://github.com/eqlabs/pathfinder/blob/main/crates/storage/src/bloom.rs).

jbcaron:
> Hey there! I made a first quick pass on the PR and I have 2 first questions:
> 
> 1. Why do we need to introduce a bloom filter? Did we were facing low performance in get_events RPC?
> 2. Why are we implementing the bloom filter from scratch and not using something like [this](https://crates.io/crates/bloomfilter)?

That's exactly right. In some tests, I observed response times that were up to three orders of magnitude faster. The Bloom filter implementation allows the filter size to depend on the number of keys being inserted, ensuring a constant false positive rate regardless of block size.

Additionally, this implementation enables precomputation of hashes for a set of keys, so that when scanning the filters of each block, we only compute the filter indices to check for the keys, without needing to recompute the hashes every time.

Lastly, the key pattern is also specifically designed for event filtering, optimizing lookup efficiency further.

crates/madara/primitives/bloom_filter/src/filter.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// # Arguments
+    ///
+    /// * `size`: The desired size of the filter in bits
+    /// * `hash_count`: The number of hash functions to use
+    pub fn new(size: u64, hash_count: u8) -> Self {
+        let size_in_u64s = (size.max(1) + 63) >> 6;
+        let aligned_size = size_in_u64s << 6;
+
+        Self {
+            storage: AtomicBitStore::new(size_in_u64s as _),
```
HermanObst: Why don't be explicit with the type? `usize` 

Quick debate:
- Does it makes sense for `size` to be an u64?

If it makes sense, its also worth saying that this will panic if madara runs in a 32 bits architecture (actually almost all of this impl would fail)

ofc it could be argued that Madara will never run in 32 bits, but happy to hear what you guys think

crates/madara/primitives/bloom_filter/src/filter.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /// Adds an item to the Bloom filter.
+    ///
+    /// # Arguments
+    ///
+    /// * `item`: The item to add, which must implement the Hash trait
+    pub fn add<T: Hash>(&self, item: &T) {
+        calculate_hashes::<H, T>(self.hash_count, item)
+            .for_each(|hash| self.storage.set_bit((hash % self.bit_size) as usize));
```
HermanObst: ditto

crates/madara/primitives/bloom_filter/src/filter.rs
```diff
@@ -0,0 +1,280 @@
...
+fn calculate_hashes<H: Hasher + Default, T: Hash>(hash_count: u8, item: &T) -> impl Iterator<Item = u64> + '_ {
+    let mut hasher1 = H::default();
+    item.hash(&mut hasher1);
+    let h1 = hasher1.finish();
+
+    const PHI: u64 = 0x9e37_79b9_7f4a_7c15; // 2^64/φ
+    let h2 = h1.rotate_left(17).wrapping_mul(PHI);
+
+    (0..hash_count).map(move |i| h1.wrapping_add(h2.wrapping_mul(i as u64)))
+}
```
HermanObst: Nice stuff!

crates/madara/primitives/bloom_filter/src/tests/filter_tests.rs
```diff
@@ -0,0 +1,191 @@
+use crate::*;
```
HermanObst: Great test suite 🚀

crates/madara/primitives/bloom_filter/benches/bloom_benchmark.rs
```diff
@@ -0,0 +1,181 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
```
HermanObst: This is great as give us a tool to prevent us to introduce downgrades in the bloom filter perf (e.g we add this as part of CI and put alerts if the performance goes down)

Regarding the different hash functions used. There are 3 of them to decide which one to use? 

It would be even great if we can have something like that, but with RPC calls (basically the reason for this PR)

My understanding is that @Trantorian1 built some benchmarks that check that.

Would it makes sense to add them in CI? (and also I would love to see those benchs going down with this PR 🚀 )

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    fn event_to_bloom_keys(event: &Event) -> impl Iterator<Item = [u8; 33]> + '_ {
+        let from_address_key = create_bloom_key(0, &event.from_address);
+
+        let keys = event.keys.iter().enumerate().map(|(i, key)| create_bloom_key(i as u8 + 1, key));
+
+        std::iter::once(from_address_key).chain(keys)
+    }
+}
```
HermanObst: @jbcaron please add documentation on this logic (how we go from events to store them in bloom filter)

HermanObst: Also, it would be great to have some basic tests with events that also check `might` and `definitely not`
(would also serve as docs)

HermanObst (changes requested):
Incredible work here @jbcaron! 🚀 

I just requested some changes for:
- adding docs and test in bloom_filter/src/events.rs 
- Just general discussion on some topics.

Im also curious if we have some full e2e test that check `starknet_getEvent` rpc call

crates/madara/primitives/bloom_filter/benches/bloom_benchmark.rs
Trantorian1: The test suite was made more for local exploration and requires a full sync Juno, Pathfinder and Madara node running simultaneously. Because of this it is rather tricky to have it running in the CI as we would need to keep these nodes up to date with the tip of the chain and manage downtime :/

There probably is a better way to do this, I'm thinking a local devnet to which we publish some sample data for each node to sync from (we would essentially be mocking a chain), but I haven't explored this and to be honest I'm not very familiar with publishing state to Starknet (never done Cairo).

So basically while the benches are handy to detect broad performance trends between each nodes, they are pretty hard to run atm which makes it difficult to integrate them into the CI.

crates/madara/primitives/bloom_filter/src/filter.rs
Trantorian1: I mean if ever someone tries to run Madara on a 32 bit architecture they have bigger problems than just the bloom filter since half the codebase is going to explode/implode/both :sweat_smile:

HermanObst:
Awesome! @Trantorian1 could we run the benchmarks to check that we are indeed gaining performance?

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -137,9 +215,273 @@ impl EventBloomSearcher {
...
+    }
+
+    #[rstest]
+    fn test_single_event_key_matching(single_event: Event) {
+        let writer = EventBloomWriter::from_events(std::iter::once(&single_event));
+        let serialized = serde_json::to_string(&writer).unwrap();
+        let reader: EventBloomReader = serde_json::from_str(&serialized).unwrap();
+
+        // Should match either key
+        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![], vec![Felt::from(3)]]));
```
HermanObst: @jbcaron Whats the difference with line 282?

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -137,9 +215,273 @@ impl EventBloomSearcher {
...
+        let serialized = serde_json::to_string(&writer).unwrap();
+        let reader: EventBloomReader = serde_json::from_str(&serialized).unwrap();
+
+        // No patterns should match everything
+        let searcher = EventBloomSearcher::new(None, None);
+        assert!(searcher.search(&reader));
+    }
+
+    #[rstest]
+    fn test_single_event_exact_match(single_event: Event) {
```
HermanObst: shouldnt an exact match be:

`EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![Felt::from(2)], vec![Felt::from(3)]])); `

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -137,9 +215,273 @@ impl EventBloomSearcher {
...
+        let serialized = serde_json::to_string(&writer).unwrap();
+        let reader: EventBloomReader = serde_json::from_str(&serialized).unwrap();
+
+        // Empty vec array should be ignored
+        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![], vec![Felt::from(3)]]));
+        assert!(searcher.search(&reader), "Search should succeed when empty arrays are ignored");
+
+        // Only empty arrays should work
+        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![], vec![]]));
+        assert!(searcher.search(&reader), "Search with only empty arrays should work");
```
HermanObst: @jbcaron I dont understand the comments. 
This two cases should be `true`?

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -137,9 +215,273 @@ impl EventBloomSearcher {
...
+        assert!(searcher.search(&reader));
+    }
+
+    #[rstest]
+    fn test_single_event_key_matching(single_event: Event) {
+        let writer = EventBloomWriter::from_events(std::iter::once(&single_event));
+        let serialized = serde_json::to_string(&writer).unwrap();
+        let reader: EventBloomReader = serde_json::from_str(&serialized).unwrap();
+
+        // Should match either key
```
HermanObst: If you are trying to test the (OR) behavior should this be something like:
```
        // Should match either key
        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![], vec![Felt::from(3),Felt::from(4)]]));
        assert!(searcher.search(&reader));
    }
```

crates/madara/primitives/bloom_filter/src/events.rs
```diff
@@ -137,9 +215,273 @@ impl EventBloomSearcher {
...
+        // Should match only when all patterns match
+        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![Felt::from(2)], vec![Felt::from(3)]]));
+        assert!(searcher.search(&reader));
+    }
+
+    #[rstest]
+    fn test_multiple_events_negative_case(multiple_events: Vec<Event>) {
+        let writer = EventBloomWriter::from_events(multiple_events.iter());
+        let serialized = serde_json::to_string(&writer).unwrap();
+        let reader: EventBloomReader = serde_json::from_str(&serialized).unwrap();
```
HermanObst: Maybe we could add a case where some keys match, but one not:

```
        let searcher = EventBloomSearcher::new(Some(&Felt::from(1)), Some(&[vec![Felt::from(5)],vec![Felt::from(2)], vec![Felt::from(3)]]));
        assert!(!searcher.search(&reader));
```

HermanObst (commented):
@jbcaron thank you so much for this awesome work! 
I just leave some comments, but we are good to go after discussing them!

@Trantorian1 although not strictly needed for merging this PR, it would be COOL! if we can see those benchmarks going down in this PR. It would be awesome

HermanObst:
@jbcaron gentle ping

raynaudoe:
Hi @jbcaron , is this PR still relevant?

jbcaron:
> Hi @jbcaron , is this PR still relevant?

Yes, I just made the requested changes.

HermanObst:
@jbcaron just one final main merge will be needed to fix the docker rate limit 🙏

HermanObst (approved):

Cargo.toml
```diff
@@ -15,6 +15,8 @@ members = [
   "madara/crates/client/block_production",
   "madara/crates/node",
   "madara/crates/primitives/block",
+  "madara/crates/primitives/bloom_filter",
+  "madara/crates/primitives/bloom_filter",
```
cchudant: duplicate

Cargo.toml
```diff
@@ -68,6 +70,8 @@ default-members = [
   "madara/crates/client/analytics",
   "madara/crates/node",
   "madara/crates/primitives/block",
+  "madara/crates/primitives/bloom_filter",
+  "madara/crates/primitives/bloom_filter",
```
cchudant: duplicate

madara/crates/client/db/src/block_db.rs
```diff
@@ -424,4 +426,23 @@ impl MadaraBackend {
             }
         }
     }
+
```
cchudant, line 430: need a tiny doc comment saying something like
Warning! The returned iterator pins the rocksdb memtables during it's lifetime.
because if you hold that iterator struct through, eg. an await or a blocking part of the code, it could lead to serious memory issues

see: https://github.com/facebook/rocksdb/wiki/Iterator#resource-pinned-by-iterators-and-iterator-refreshing

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -216,6 +217,17 @@ impl MadaraBackend {
         Ok(())
     }
 
+    pub fn store_bloom(&self, block_n: u64, bloom: EventBloomWriter) -> Result<(), MadaraStorageError> {
```
cchudant: I think it should private imo, hidden as an impl detail of the backend, and used in store_events instead

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_events.rs
```diff
@@ -55,43 +57,72 @@ pub async fn get_events(starknet: &Starknet, filter: EventFilterWithPageRequest)
...
+        // - `latest_block`: Ensures we do not process beyond the latest finalized block.
+        // - `to_block`: Ensures we do not go beyond the user-specified range.
+        if current_block > to_block {
+            break;
+        }
 
-        let num_events = block_filtered_reduced_events.len();
+        // Use the bloom filter to quickly check if the block might contain relevant events.
+        // - This avoids unnecessary block retrieval if no matching events exist.
+        if !key_filter.search(&bloom_filter) {
```
cchudant: i think all of this bloom filter logic should be in the backend, because it's an implementation detail, wdyt?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_events.rs
```diff
@@ -33,7 +35,7 @@ pub async fn get_events(starknet: &Starknet, filter: EventFilterWithPageRequest)
     let chunk_size = filter.chunk_size;
 
     if let Some(keys) = &keys {
-        if keys.len() > MAX_EVENTS_KEYS {
+        if keys.iter().flatten().count() > MAX_EVENTS_KEYS {
```
cchudant: keys.map(|k| k.len()).sum()?

madara/crates/primitives/bloom_filter/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+description = "Madara primitive for block state update"
```
cchudant: description is wrong :)

cchudant (commented):
I need to look more in depth in the bloom filter implem (and the events encoding), but I really like that it's proptested and criterion'ed

Here are my notes:
- bloom filter logic should be in the backend, IMO - rpc endpoints should never have to know it's actually implemented using a bloom filter in the backend
  - I think there's not really any other way here than just putting almost the whole logic in backend - that would mean the rpc endpoint code would now be very very tiny but it's okay 
  - it would also reduce the risk of the rocksdb iterator being misused by downstream crates (since it would be a private function now)
- block prod currently does not set the bloom filters (solved if you move all of the bloom filter logic to backend)
- this endpoint is really lacking in tests - this was mostly fine because of how simple it is, but now it feels way more complicated and I don't know if we can just merge something like that without at least 1 or 2 tests
  - if you move the logic into the backend maybe the tests should be there instead, obviously

I was also thinking that maybe we want to make bloom filters optional, if you only care about block prod and don't plan on serving RPCs, wdyt? get_events rpc can still be served using the old slow way.

madara/crates/primitives/bloom_filter/src/events.rs
```diff
@@ -0,0 +1,487 @@
...
+/// Number of bits per element in the Bloom filter.
+/// The value 9.6 is derived from the formula for optimal size:
+/// m/n = -ln(p)/ln(2)² ≈ 9.6
+/// where:
+/// - m is the size of the filter in bits
+/// - n is the number of elements
+/// - p is the desired false positive rate (0.01)
+const BITS_PER_ELEMENT: f64 = 9.6;
+
+type BloomHasher = AHasher;
```
cchudant: cool! that makes a lot of sense

madara/crates/primitives/bloom_filter/src/filter.rs
```diff
@@ -0,0 +1,280 @@
...
+/// where:
+/// - h1 is the primary hash value of the item
+/// - h2 is a secondary hash derived from h1 using rotation and golden ratio multiplication
+/// - m is the bit array size (applied implicitly by the caller)
+/// - i iterates from 0 to k-1, with k being the number of desired hash functions
+///
+/// # Mathematical Justification
+///
+/// ## Double Hashing
+/// - Double hashing is used to minimize clustering in Bloom filters.
```
cchudant, line 228: Okay, I was thinking that we don't need that because the keys are always uniform
But in fact they're not - the from_address might be (if we forget about contract address upper bound) but event keys are user-made and arbitrary, they also usually encode ascii
So you still need that - got it

madara/crates/primitives/bloom_filter/src/storage.rs
cchudant, line 1: why Ordering::Relaxed?

madara/crates/primitives/bloom_filter/src/events.rs
```diff
@@ -0,0 +1,487 @@
...
+/// The searcher supports complex queries with both AND and OR semantics:
+/// - Different patterns are combined with AND logic
+/// - Keys within each pattern use OR logic
+///
+/// # Query Optimization
+///
+/// The searcher pre-calculates hashes for all patterns during construction,
+/// making subsequent searches more efficient.
+pub struct EventBloomSearcher {
+    patterns: Vec<Vec<PreCalculatedHashes>>,
```
cchudant: I think in theory you can OR the bits from the bloom keys to emulate the AND semantic right?
if your first key is 0100000 and your second key is 0001000, you can bitwise or the two filters to get a filter that tests for the first and second key at once

cchudant (commented):
bloom filter review

Okay, so, right now
BitStore is what you use when reading events (makes sense), but AtomicBitStore when writing
I think you're not actually using concurrency in the madara impl though, right?

jbcaron:
>>Okay, so, right now
BitStore is what you use when reading events (makes sense), but AtomicBitStore when writing
I think you're not actually using concurrency in the madara impl though, right?

I believe that even without concurrency, copying data is necessary to insert bits into the filter, which is less efficient than using atomics.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_events.rs
jbcaron: done

madara/crates/primitives/bloom_filter/Cargo.toml
jbcaron: fixed

madara/crates/client/db/src/block_db.rs
jbcaron: The refactored code now explicitly scopes the RocksDB iterator lifetime, ensuring it's only held open during the quick bloom filter search phase. 
The iterator is immediately dropped after finding a candidate block (or determining no candidates exist), before any potentially blocking operations like block retrieval or event processing. This prevents pinning RocksDB memtables for extended periods.

madara/crates/client/db/src/storage_updates.rs
jbcaron: The bloom-filter crate still exists, but all event-specific implementation is now private within the DB crate

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_events.rs
jbcaron: done

cchudant (approved):
cool!

PR #478, opened 2025-01-24
test(block_production): Add tests to block_production_task method

GMKrieger:
## Pull Request type

- Testing

Resolves: #NA

The block production tests currently only test block sealing when starting a new service.

## What is the new behavior?

This PR aims to add tests on the `block_production_task` method and related methods to thoroughly test the code.

No.

## Other information

The testing format is based on #468. All feedback is welcome!


crates/madara/client/block_production/src/lib.rs
```diff
@@ -667,6 +681,72 @@ mod tests {
...
+            });
+        }
+
+        let block = g.build(&chain_config).unwrap();
+        let backend = MadaraBackend::open_for_testing(Arc::clone(&chain_config));
+        let importer = Arc::new(BlockImporter::new(Arc::clone(&backend), None).unwrap());
+
+        let _ = importer
+            .add_block(block, BlockValidationContext::new(chain_config.chain_id.clone()).trust_class_hashes(true))
+            .await;
```
Trantorian1: Probably should `expect` this here no?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -758,6 +865,66 @@ mod tests {
...
+            ],
+            deployed_contracts: vec![DeployedContractItem { address: Felt::THREE, class_hash: Felt::THREE }],
+            replaced_classes: vec![ReplacedClassItem { contract_address: Felt::TWO, class_hash: Felt::TWO }],
+            nonces: vec![
+                NonceUpdate { contract_address: Felt::ONE, nonce: Felt::ONE },
+                NonceUpdate { contract_address: Felt::TWO, nonce: Felt::TWO },
+                NonceUpdate { contract_address: Felt::THREE, nonce: Felt::THREE },
+            ],
+        }
+    }
```
Trantorian1: I'm not sure its a good idea to have a `fixture` for the pending block state diff as this is very subject to change depending on the test we are running. Imo this is too specific and we should probably be defining this on a per-test basis. Wdyt?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -790,8 +957,108 @@ mod tests {
...
+
+        let tx_signature = match &mut invoke_txn {
+            BroadcastedInvokeTxn::V0(tx) => &mut tx.signature,
+            BroadcastedInvokeTxn::V1(tx) => &mut tx.signature,
+            BroadcastedInvokeTxn::V3(tx) => &mut tx.signature,
+            _ => unreachable!("the invoke tx is not query only"),
+        };
+        *tx_signature = vec![signature.r, signature.s];
+
+        let _ = mempool.tx_accept_invoke(invoke_txn);
```
Trantorian1: Should probably `expect` this too.

crates/madara/client/block_production/src/lib.rs
```diff
@@ -790,8 +957,108 @@ mod tests {
...
+
+        let tx_signature = match &mut declare_txn {
+            BroadcastedDeclareTxn::V1(tx) => &mut tx.signature,
+            BroadcastedDeclareTxn::V2(tx) => &mut tx.signature,
+            BroadcastedDeclareTxn::V3(tx) => &mut tx.signature,
+            _ => unreachable!("the declare tx is not query only"),
+        };
+        *tx_signature = vec![signature.r, signature.s];
+
+        let _ = mempool.tx_accept_declare(declare_txn);
```
Trantorian1: `expect`

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
         assert!(err.contains("Failed to retrieve pending declared class at hash"));
         assert!(err.contains("not found in db"));
     }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_on_pending_block_tick_block_still_pending(
```
Trantorian1, line 1905: Could you please add a doc comment with an overview of what this does?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+
+        assert!(mempool.is_empty());
+        assert_eq!(pending_block.inner.transactions.len(), 1);
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 0);
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_on_pending_block_tick_updates_correct_block(
```
Trantorian1, line 1965: Ditto

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 1);
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_on_block_time_fails_inconsistent_state(
```
Trantorian1, line 2263: Ditto

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+
+        assert!(result.is_err(), "Expected an error due to mismatched parent hash");
+        assert!(mempool.is_empty());
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 0);
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_start_block_production_task_normal_setup(
```
Trantorian1, line 2445: Ditto

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 1);
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_start_block_production_task_pending_tick_too_small(
```
Trantorian1, line 2513: Ditto

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 0);
+
+        // ================================================================== //
+        //                  PART 3: init block production task                //
+        // ================================================================== //
+
+        let result =
+            block_production_task.block_production_task(mp_utils::service::ServiceContext::new_for_testing()).await;
+
+        assert!(result.is_err(), "Expected an error due to very small pending block time");
```
Trantorian1: Is it possible to check the exact type of the error here?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+        assert!(!mempool.is_empty());
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_start_block_production_task_closes_block_right_after_pending(
```
Trantorian1, line 2570: Ditto

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1980,619 @@ mod tests {
...
+        let pending_block = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+    }
+
+    #[rstest::rstest]
+    #[tokio::test]
+    #[allow(clippy::too_many_arguments)]
+    async fn test_block_prod_start_block_production_task_ungracious_shutdown_and_restart(
```
Trantorian1, line 2636: Ditto

crates/madara/client/block_production/src/lib.rs
GMKrieger: I created this on a point of testing that many more tests used this pending block setup and the file was getting ridiculously long. In hindsight, it shouldn't have been a fixture anyway because of what you said, but maybe it could be a method that creates a simple pending block from the fixtures if need be and more tests are created.

crates/madara/client/block_production/src/lib.rs
```diff
@@ -667,6 +680,71 @@ mod tests {
...
+        };
+
+        let block = g.build(&chain_config).unwrap();
+        let backend = MadaraBackend::open_for_testing(Arc::clone(&chain_config));
+        let importer = Arc::new(BlockImporter::new(Arc::clone(&backend), None).unwrap());
+
+        let _ = importer
+            .add_block(block, BlockValidationContext::new(chain_config.chain_id.clone()).trust_class_hashes(true))
+            .await
+            .expect("Should add block");
```
Trantorian1:
```suggestion
        importer
            .add_block(block, BlockValidationContext::new(chain_config.chain_id.clone()).trust_class_hashes(true))
            .await
            .expect("Should add block");
```

crates/madara/client/block_production/src/lib.rs
```diff
@@ -771,27 +849,127 @@ mod tests {
...
+
+        let tx_signature = match &mut declare_txn {
+            BroadcastedDeclareTxn::V1(tx) => &mut tx.signature,
+            BroadcastedDeclareTxn::V2(tx) => &mut tx.signature,
+            BroadcastedDeclareTxn::V3(tx) => &mut tx.signature,
+            _ => unreachable!("the declare tx is not query only"),
+        };
+        *tx_signature = vec![signature.r, signature.s];
+
+        let _ = mempool.tx_accept_declare(declare_txn).expect("Should accept the transaction");
```
Trantorian1:
```suggestion
        mempool.tx_accept_declare(declare_txn).expect("Should accept the transaction");
```

crates/madara/client/block_production/src/lib.rs
```diff
@@ -771,27 +849,127 @@ mod tests {
...
+
+        let tx_signature = match &mut invoke_txn {
+            BroadcastedInvokeTxn::V0(tx) => &mut tx.signature,
+            BroadcastedInvokeTxn::V1(tx) => &mut tx.signature,
+            BroadcastedInvokeTxn::V3(tx) => &mut tx.signature,
+            _ => unreachable!("the invoke tx is not query only"),
+        };
+        *tx_signature = vec![signature.r, signature.s];
+
+        let _ = mempool.tx_accept_invoke(invoke_txn).expect("Should accept the transaction");
```
Trantorian1:
```suggestion
        mempool.tx_accept_invoke(invoke_txn).expect("Should accept the transaction");
```

crates/madara/client/block_production/src/lib.rs
Trantorian1: @GMKrieger

Trantorian1 (commented):
Left a few nitpicks, otherwise all seems good.

Trantorian1:
@GMKrieger Could you please avoid force-pushing changes to a PR after a review has occurred? This makes it harder to review new changes since github cannot display "changes since last review".

Trantorian1 (approved):
All good!

crates/madara/client/block_production/src/lib.rs
```diff
@@ -667,6 +680,71 @@ mod tests {
...
+        #[default(false)] use_bouncer_weights: bool,
+    ) -> (
+        Arc<MadaraBackend>,
+        Arc<BlockImporter>,
+        Arc<BlockProductionMetrics>,
+        Arc<MockL1DataProvider>,
+        Arc<Mempool>,
+        DevnetKeys,
+    ) {
+        let mut g = ChainGenesisDescription::base_config().unwrap();
```
HermanObst: @GMKrieger lets update this variable name to something more descriptive.

crates/madara/client/block_production/src/lib.rs
```diff
@@ -771,27 +849,127 @@ mod tests {
...
+        let flattened_class: mp_class::FlattenedSierraClass = sierra_class.clone().flatten().unwrap().into();
+
+        // starkli class-hash target/dev/madara_contracts_TestContract.compiled_contract_class.json
+        let compiled_contract_class_hash =
+            Felt::from_hex("0x0138105ded3d2e4ea1939a0bc106fb80fd8774c9eb89c1890d4aeac88e6a1b27").unwrap();
+
+        let mut declare_txn: BroadcastedDeclareTxn = BroadcastedDeclareTxn::V3(BroadcastedDeclareTxnV3 {
+            sender_address: contract.address,
+            compiled_class_hash: compiled_contract_class_hash,
+            signature: vec![],
```
HermanObst, line 892: add a small comment above it saying something like:

"Field to be filled below"

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+    ) {
+        let (backend, importer, metrics, l1_data_provider, mempool, contracts) = devnet_setup.await;
+
+        // ================================================================== //
+        //               PART 1: add a transaction to the mempool             //
+        // ================================================================== //
+
+        // The transaction itself is meaningless, it's just to check
+        // if the task correctly reads it and process it
+        sign_and_add_declare_tx(&contracts.0[0], &backend, &mempool, Felt::ZERO);
```
HermanObst, line 1924: Just for completeness add before adding the txs to the mempool:
`assert!(mempool.is_empty());`

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+                        tx_hashes: vec![Felt::ONE, Felt::TWO, Felt::THREE],
+                    }),
+                    inner: pending_inner.clone(),
+                },
+                pending_state_diff.clone(),
+                converted_classes.clone(),
+                Some(visited_segments.clone()),
+                Some(bouncer_weights),
+            )
+            .expect("Failed to store pending block");
```
HermanObst, line 2117: What happen with this `PendingBlock` that we are storing? What should be the expected behavior.
Are we testing something that can even occur? 

cc @Trantorian1 @GMKrieger

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+
+        assert_eq!(pending_block.inner.transactions.len(), 0);
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 0);
+
+        // ================================================================== //
+        //                  PART 3: call on pending time tick                 //
+        // ================================================================== //
+
+        // The small bouncer weights should make the tick close
+        // the block when called
```
HermanObst: What is 'small' about the bouncer config? 
(lets explain it above)

crates/madara/client/block_production/src/lib.rs
```diff
@@ -771,27 +849,127 @@ mod tests {
...
-            message_segment_length: 11,
-            n_events: 12,
-            n_steps: 13,
-            state_diff_size: 14,
+            gas: 100000,
+            message_segment_length: 100000,
+            n_events: 100000,
+            n_steps: 100000,
+            state_diff_size: 100000,
         }
```
HermanObst, line 871: Can we add what is the commonly used config?
Is it max?

Add a comment explaining that this config is defined to be small and will result as the block marked as full instantly.

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+
+        // ================================================================== //
+        //                      PART 6: call on block time                    //
+        // ================================================================== //
+
+        // If the program ran correctly, the pending block should
+        // have no transactions on it after the method ran for at
+        // least block_time
+
+        let result = block_production_task.on_block_time().await.expect_err("This should fail");
```
HermanObst, line 2427: Out of curiosity, what happens if this situation is reach? 
Is there some recovery?
cc/ @Trantorian1 @GMKrieger

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+        // plus a little bit to guarantee
+        tokio::time::sleep(std::time::Duration::from_secs(31)).await;
+        task_handle.abort();
+
+        let pending_block = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 1);
+    }
```
HermanObst: Wouldn't be good to check that we have 1 transaction in the `NotPending` block?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+        });
+
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        task_handle.abort();
+
+        let pending_block: mp_block::MadaraMaybePendingBlock = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+    }
```
HermanObst: Check what happened with the tx?

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+        //                PART 3: init block production task                  //
+        // ================================================================== //
+
+        let ctx = mp_utils::service::ServiceContext::new_for_testing();
+
+        let task_handle = tokio::spawn(async move { block_production_task.block_production_task(ctx).await });
+        tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+        task_handle.abort();
+        assert!(mempool.is_empty());
+
```
HermanObst: Check that is in the pending block

crates/madara/client/block_production/src/lib.rs
```diff
@@ -1713,4 +1891,816 @@ mod tests {
...
+        tokio::time::sleep(std::time::Duration::from_secs(35)).await;
+        task_handle.abort();
+
+        let pending_block: mp_block::MadaraMaybePendingBlock = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+        assert!(mempool.is_empty());
+        assert!(pending_block.inner.transactions.is_empty());
+        assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 2);
+    }
```
HermanObst, line 2736: Would be better to only add 1 tx just to have a difference within the previous one

Also check for the latest block number

HermanObst (changes requested):
Hey @GMKrieger great job here! 
In genreal I really like these test. I ask some changes. I general I would like to be more thorough on checking the state (empty, num txs) of:
- mempool
- pending block
- latest block

crates/madara/client/block_production/src/lib.rs
GMKrieger: I don't understand what you mean by adding the commonly used config. Should I put the values commented above as comparison?

crates/madara/client/block_production/src/lib.rs
GMKrieger: Removed the extra tx, but I check the latest block number with `assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 2);`

crates/madara/client/block_production/src/lib.rs
GMKrieger: This pending block is overwriting the previous pending block, which is not the one the task is working on. Now the database has an inconsistent state.

The BlockProductionTask struct has the block it's working on an variable called `block` inside itself. This is the value we should work on and insert back on the database when finished.

This test currently tests a very edge case, let's say, someone tries to insert a pending block directly on the database and the block production task needs to keep working on the correct block.

But it's important to have a test like this because if there are future changes on the code that break this rule, this test will fail and warn the developer that something is off.

crates/madara/client/block_production/src/lib.rs
GMKrieger: Not that I know of. It just fails and quits.

HermanObst (approved):
Good job!

PR #483, opened 2025-01-31
fix: fix incorrect handling of contract_address as an array

brawncode:
## Pull Request type

I noticed that `contract_address` was being treated as an array in the code, which caused an issue when trying to access `contract_address[0]`.

Since `contract_address` is actually a string, I've updated the code to correctly check its truthiness.

The fix ensures the test works as intended.  

```javascript
expect(deployResult.contract_address).toBeTruthy();
```  

This change resolves the issue and aligns with the expected data type.

no


Trantorian1:
Hello! Please understand we are hesitant in merging a one-liner from an unknown contributor. You can have a look at the [issues](https://github.com/madara-alliance/madara/issues) for a list of know bugs which need fixing. For anything else, we would prefer if you could open an issue first before proposing a fix.

PR #484, opened 2025-01-31
feat(rpc): implement starknet_subscribePendingTransactions WebSocket endpoint

HermanObst:
## Pull Request type
- Feature

Currently starknetSubscribePendingTransaction is not supported

Resolves: #NA

## What is the new behavior?

Support for starknetSubscribePendingTransaction

No


jbcaron:
I believe that `SentTxsCache` should be handled directly at the channel producer level to efficiently manage duplicate transactions and prevent unnecessary overhead across all channels.

PR #486, opened 2025-02-03
fix: l1 transactions nonce issue fix

Mohiiit:
Comparing wrong nonce while saving the l1 transactions.
## Pull Request type

- Bugfix

Every l1 transaction was getting saved by pending nonce info and hence not a single l1 transaction were getting executed
 
Resolves: #NA

## What is the new behavior?

- it's fixed now

No


madara/crates/client/mempool/src/lib.rs
```diff
@@ -1693,6 +1698,62 @@ mod test {
         );
     }
 
+    /// This test makes sure that retrieve_nonce_info_l1_handler has proper
+    /// error handling and returns correct values based on the state of the db.
+    #[rstest::rstest]
+    #[timeout(Duration::from_millis(1_000))]
+    fn mempool_retrieve_nonce_info_l1_handler(
```
raynaudoe: These tests are failing, am I right?

Mohiiit: nope, none of the tests are failing as such in this scope of the PR atleast, it was a very cheeky bug, I think bugs are these would be caught once we have a e2e test of bootstrapper <> madara <> orchestrator

raynaudoe: ok! just asking because I just ran `cargo test mempool_retrieve_nonce_info_l1_handler` and it failed 🤔

Mohiiit: oh? hmm interesting, maybe I am missing something then

madara/crates/client/eth/src/l1_messaging.rs
```diff
@@ -153,10 +150,14 @@ async fn process_l1_message(
             tracing::error!("⟠ Unexpected DB error: {:?}", e);
             return Err(e.into());
         }
+        _ => {}
     };
 
     let res = mempool.tx_accept_l1_handler(transaction.into(), fees)?;
 
+    // HERMAN TODO: Actually this should be updated after the tx l1 handler is executed
```
HermanObst, line 158: @Trantorian1 @raynaudoe @Mohiiit wdyt?

Trantorian1: I'd say conceptually this seem ok, but I would highly suggest double-checking with @Mohiiit as I think I already tried out this fix and it was not working for them on the bootstrapper side. Other than that all seem good :+1:

Trantorian1 (approved):

madara/crates/client/eth/src/l1_messaging.rs
raynaudoe: I think is correct since otherwise this will cause an "off by one" (between `nonce` and `nonce_target`) kind of scenario?

raynaudoe (approved):

madara/crates/client/eth/src/l1_messaging.rs
Mohiiit: checked it, it's working fine an the nonce update is right as well

madara/crates/client/eth/src/l1_messaging.rs
```diff
@@ -286,7 +292,7 @@ mod l1_messaging_tests {
                 payload[4] = 4543560;
                 payload[5] = 1082959358903034162641917759097118582889062097851;
                 payload[6] = 221696535382753200248526706088340988821219073423817576256483558730535647368;
-                uint256 nonce = 10000000000000000;
+                uint256 nonce = 0;
```
Mohiiit, line 295: why have we changed this to 0 tho? is it because it will be moved to pending otherwise?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -274,6 +274,35 @@ impl Mempool {
...
+    pub fn resolve_nonce_info_l1_handler(&self, nonce: Felt) -> Result<NonceInfo, MempoolError> {
+        let nonce = Nonce(nonce);
+        let nonce_next = nonce.try_increment()?;
+
+        // TODO: This would break if the txs are not ordered --> l1 nonce latest should be updated only after execution
+        // Currently is updated after inclusion in mempool
+        let current_nonce = self.backend.get_l1_messaging_nonce_latest()?;
+        // first l1 handler tx, where get_l1_messaging_nonce_latest returns None
+        let target_nonce = match current_nonce {
+            Some(nonce) => nonce.try_increment()?,
```
Mohiiit, line 296:
```suggestion
            Some(nonce) => nonce_next,
```

Mohiiit (commented):
looks good ser!!!

madara/crates/client/eth/src/l1_messaging.rs
Mohiiit: @Trantorian1 I am not sure why it was failing earlier tho, I have tested it again multiple times and seems like it's working

madara/crates/client/mempool/src/lib.rs
Mohiiit: nvm ignore

madara/crates/client/eth/src/l1_messaging.rs
HermanObst: Exactly, now order is enforced :) 
we could do something nasty like pushing 10000000000000000 - 1 into the data base, but it really do not made a lot of sense IMO

raynaudoe:
> This PR is not yet ready to merge, it needs a few more fixes 🙏🏻 

edit: ready!

PR #494, opened 2025-02-13
feat(node): load args from file

### LINKED ISSUE #285: feat(cli): Add support for environment variable and configuration file parameters

### Motivation

The goal of this issue is to introduce 2 new methods for specifying parameters on Madara. Currently only cli arguments are supported.

### Request

We would like to implement:

1. Environment variables

Set up your node's environment variables using the `MADARA_` prefix. For example:

`MADARA_BASE_PATH=/path/to/data`
`MADARA_LOG=info`

These variables allow you to adjust the node's configuration without using command-line arguments.

2. Configuration files

You could use a JSON, TOML, or YAML file to structure your configuration settings. Specify your configuration file on startup with the -c option. Here's a basic example in JSON format:

```
{
  "name": "able-vps-2",
  "base_path": "../deoxys-db",
  "network": "devnet",
  "l1_endpoint": "key",
  "rpc_port": 9944,
  "rpc_cors": "*",
  "rpc_external": true,
  "prometheus_external": true
}
```

Configuration priority would be as follows: command-line arguments > environment variables > configuration files.
When the same setting is configured in multiple places, the source with the highest priority takes effect.

### Solution

as described above

Yes!

GMKrieger:
## Pull Request type

- Feature

Currently, the node doesn't allow arguments to be loaded from a file.

Resolves: #285 

## What is the new behavior?

- Tries to load arguments from a file if a path is provided on the cli
- Loads data from cli after that

No.

## Other information

I used figment, as it was the crate that worked best, but I know there are others, mainly [twelf](https://crates.io/crates/twelf) and [config](https://crates.io/crates/config). If someone has experience with them, please, let me know!


README.md
```diff
@@ -313,6 +313,12 @@ change its value. For example:
 These variables allow you to adjust the node's configuration without using
 command-line arguments, which can be useful in CI pipelines or with docker.
 
+### Configuration files
+
+You can load the arguments directly from a file for ease of use.
+The supported file formats are `json`, `toml` and `yaml`.
+You can find examples on `configs/args`.
```
Trantorian1: `configs/args` should be a hyperlink to the related folder on the repo.

crates/madara/node/src/cli/mod.rs
```diff
@@ -134,9 +134,13 @@ impl ArgsPresetParams {
     ),
 )]
 pub struct RunCmd {
+    /// A path to a config file
+    #[clap(env = "MADARA_CONFIG_FILE", long, value_name = "NAME")]
+    pub config_file: Option<PathBuf>,
```
Trantorian1: I'm not sure about the ergonomics of this. I think we should keep this here, but also try and auto-detect a config file in known locations (such as the current working directory), and then this option can be used for unusual configurations. Also this needs more CLI-facing documentation, such as which formats are allowed.

crates/madara/node/src/cli/mod.rs
```diff
@@ -245,6 +249,14 @@ impl RunCmd {
         self.name.as_ref().expect("Name was just set")
     }
 
+    pub fn check_mode(&self) -> anyhow::Result<()> {
+        if (self.sequencer | self.full) ^ (self.full | self.devnet) ^ (self.sequencer | self.devnet) {
```
Trantorian1: Why do this through binary operations? Doesn't this also mean that if `self.full == true` and no other this will return an error?
```suggestion
        if !self.sequencer && !self.full && !self.devnet {
```

crates/madara/node/src/main.rs
```diff
@@ -31,7 +35,28 @@ async fn main() -> anyhow::Result<()> {
...
+        config = match config_path.extension() {
+            None => bail!("Unsupported file type for config file."),
+            Some(os_str) => match os_str.to_str() {
+                Some("toml") => config.join(Toml::file(config_path)),
+                Some("json") => config.join(Json::file(config_path)),
+                Some("yaml") => config.join(Yaml::file(config_path)),
+                _ => bail!("Unsupported file type for config file."),
+            },
+        }
+    }
```
Trantorian1: As mentioned before, I feel like we should _first_ check for config files in some known location and only if it is not found _then_ refer to the cli args to potentially load an arg file. It feels a little un-ergonomic if I have to specify a cli arg in order to load an arg file in order not to use cli args :zany_face:

Trantorian1 (changes requested):
A welcome change! This has been in the talks for a while but was never properly implemented, thanks a lot for this!

A few points still need clarifying / changing, see comments below.

Trantorian1:
It might be worth adding a "default" config file to the repo as well, so users can just compile madara and run it with `cargo run` only, wdyt?

GMKrieger:
> It might be worth adding a "default" config file to the repo as well, so users can just compile madara and run it with `cargo run` only, wdyt?

Good idea! I'll work on one right away.

crates/madara/node/src/cli/mod.rs
GMKrieger: It works for all cases, I just did it this way because I thought about the logic output and not through short circuiting, but you solution is waaaaaay cleaner

GMKrieger:
@Trantorian1 I have run into an issue when doing the changes you requested.

I thought that [figment](https://docs.rs/figment/latest/figment/) was correctly setting up the arguments in the priority defined in the issue, being cli > env > file. However, when setting up the default file, I realized I couldn't overwrite the values in no way. Searching online, I found the underlying issue and solution, but it's not flawless; it basically requires to have to identical structures, one with all options and one with definitive struct, which would make the whole cli very verbose. If you're curious, I found the information [here](https://steezeburger.com/2023/03/rust-hierarchical-configuration/).

I tried another config layering crate [twelf](https://docs.rs/twelf/latest/twelf/), but the documentation is outdated, the examples with clap are incomplete, and overall the experience was very poor. From my tests, I gathered the problem is it doesn't really like nested structures. When trying to deserialize, it kept complaining about missing certain fields.

I tested [clap-serde-derive](https://crates.io/crates/clap-serde-derive), which is small and not very robust, and it also needs an additional structs and just didn't work as I expected, presenting several other problems I couldn't fix.

Finally, [config-rs](https://docs.rs/config/latest/config/) is one of the most downloaded and used crates dealing with layering of configs, but it doesn't explicitly support clap and examples are not clear on how I could merge both.

Any suggestions are deeply appreciated, as I don't see a clear and best way forward.

Cheers!

Trantorian1:
This is unfortunate :/ It's surprising that the ecosystem hasn't developed a crate with unified cli and file parsing yet. I guess this really just is an ergonomics problem, just having a config file is already a good starting point.

Just a thought, but could we not first check if a config file exists in `./config.toml` for example, and if it does just deserialize it into the cli struct without running the arg parse step? IE: we only call `clap::Parser::Parse` if we cannot find a config file. Shouldn't this resolve the prioritization issue or am I missing something?

Anyway, thanks a lot for the deep investigation into this topic :)

Trantorian1:
(Also this is more of a nitpick but please avoid force-pushing after a review since this makes it impossible to view "changes since last review" :sweat_smile:)

crates/madara/node/src/cli/mod.rs
```diff
@@ -134,9 +134,15 @@ impl ArgsPresetParams {
     ),
 )]
 pub struct RunCmd {
+    /// A path to a config file.
+    /// It builds the arguments from files, env and cli.
```
Trantorian1: The second line feels a bit confusing in that it seems to depict arg parsing generally? Maybe we should just remove it and keep the last line.

Trantorian1 (approved):

GMKrieger:
> This is unfortunate :/ It's surprising that the ecosystem hasn't developed a crate with unified cli and file parsing yet. I guess this really just is an ergonomics problem, just having a config file is already a good starting point.
> 
> Just a thought, but could we not first check if a config file exists in `./config.toml` for example, and if it does just deserialize it into the cli struct without running the arg parse step? IE: we only call `clap::Parser::Parse` if we cannot find a config file. Shouldn't this resolve the prioritization issue or am I missing something?
> 
> Anyway, thanks a lot for the deep investigation into this topic :)

The original idea is that if you had a file, but wanted to change one argument, you could do that through the cli; mix and match between different sources. Your solution works, but we make it exclusive, either we get all information from the file or through the cli.

I'm going do your suggestion, but a little different: if there's any argument on the cli, we use all of the cli. This is so we can have the default `cargo run` execution working without any input or needing to put the file in the correct spot.

So it'll be like this:
- cargo run with no arguments uses a default config file present on `configs/args`
- cargo run pointing to a file will use that file
- cargo run with cli will ignore the files and use cli arguments

Trantorian1:
Okay got it, so we get file configuration but without CLI override. This makes sense.

HermanObst (commented):
@GMKrieger looking good! 
I would love to add something that test this in CI.
Maybe we can replicate the [JS E2E test](https://github.com/madara-alliance/madara/blob/main/.github/workflows/starknet-js-test.yml#L27-L35) (they take very little to run) and re-execute with this? 

cc @Trantorian1

Trantorian1:
I would generally be in agreement. We have already had regressions in the past where certain argument configurations no longer worked as expected after a merge since there was no test on this. Then again, this also feels like adding scope onto this PR

HermanObst (approved):
Lets fix CI, the rest LGTM!

HermanObst (changes requested):
Actually Ill leave this with RFC until we fix CI 
@GMKrieger

HermanObst (approved):

notlesh (approved):

PR #463, opened 2025-01-14
feat(p2p): implement p2p sync

cchudant:
supersedes and closes #400 

Pull request description will be added later

PR #495, opened 2025-02-14
feat(sync): Sync refactor

cchudant:
Sync is the service that is responsible for importing blocks into the node. It is currently only run in full-node mode, as we do not need to sync any blocks when running in sequencer (block production) mode.

### Peer-to-peer

Peer-to-peer block sync is a feature that will be required for all starknet full nodes to implement. We are already very late to the party, as pathfinder, juno and papyrus have had p2p sync capabilities for more than 6 months now.
This is also of great interest to us at KasarLabs for our upcoming launch of the [Quaza](https://github.com/KasarLabs/quaza), where we want to showcase peer-to-peer.

Our peer-to-peer implementation is still experimental, which is why we are not yet considering getting it merged upstream into Madara. Since it touches a big part of the core, we need to merge its architectural changes in a separate PR to keep the P2P feature merge-ready for the Madara Core codebase at any time. With that in mind, the Madara alliance has agreed on a flow for upstreaming stuff into the main branch: if there are any core modifications, we’ll try to merge them upstream. But this is a huge job because we want to minimize breakage, which means the new refactored code should work just like the old one and keep or even improve all the test coverage.

The details of that flow are still being debated, but this pull request could be the experiment that tell us how well that goes :)

**As such**, this pull request contains
- removal of the mc-block-import crate and mc-sync in its current form.
- a new mc-sync crate
- some changes around the database as to allow storing of partially downloaded blocks.
- some plumbing around block-production has been lightly changed, and new types in primitives are added to make the code cleaner.

Moreover, the other node teams (including us!) are currently also working on consensus (tendermint) integration, which has a lot of overlap with this work. In madara's case at least, we are integrating [InformalSystems's Malachite](https://github.com/informalsystems/malachite) implementation.

### Reorg handling

This refactor is a very big step forward in terms of reorg handling, and has been designed for that from the start. The biggest win here is that the pipeline is now centrally controllable from a single point. This means that reorg-handling can be implemented on top of this pull request without madara becoming a big spaghetti mess. More on that in the Details section of this pull request description.

### Performance

The current sync service has okay performance but it can be improved a lot more. One of the biggest slowdowns during sync is the fact that class compilation slows the whole pipeline. Class definitions don't happen at every block, but when they do, they make the pipeline wait on them to finish and you cannot parallelize the compilation of a single class. This results in big spikes in performance and a lot of cpu-underutilization.

This new sync implementation does a much better job at amortizing the cost of class compilation on a lot more blocks - it does this by splitting classes sync on a new different pipeline, which is now completely free to lag behind (up to a point) compared to the other sync pipelines.

I have mesured on my Apple M4 pro, **sync to 60k blocks on Sepolia now takes 15m35 = 64.17 blocks/s**. That is an immense speedup. We are now limited by the rate-limiting of the starkware gateway (we have a gateway key, but we still hit some sort of limit as syncing from a local gateway can hit much higher blocks/s).
We are also limited on some ranges of Sepolia blocks by the bonsai-trie global trie insertion, which I really want to optimize a lot more at some point. We are still not hitting 100% cpu on every core all of the time yet.

## In details

The current sync service works by having 3 tasks:
Untitled-2025-02-04-1118(2)

You can see this in action in [mc-sync l2.rs](https://github.com/madara-alliance/madara/blob/main/crates/madara/client/sync/src/l2.rs#L247-L332).

This simplistic approach has correct performance, but it is not suitable for peer-to-peer sync. P2p sync has a number of unique challenges:
- Any peer can be lying to you. You need to check everything, maintain peer scores, and punish/disconnect the peers accordingly.
- The [p2p specs](https://github.com/starknet-io/starknet-p2p-specs/tree/main) have a single-request multi-response design, where you send a request, and you get a stream of responses for a range of blocks.
- The p2p request/stream response "endpoints" (let's call them that), are separated into: headers, transactions and receipts, state diff, classes, events.

Moreover, you cannot verify that eg. an event is correct unless you have already received the header for that block. Classes are different; they cannot be checked using the class commitment in the header, as there is no such field. Instead, you have to check them using the fields in the block state diff, which in turn you can check using the state diff commitment in the header.

The dependency diagram for verification in the p2p specs looks like this

Untitled-2025-02-04-1118(3)

*Note: Backward sync is needed as it is not guaranteed, as per the specs, that the signature field will be populated for old blocks that are not yet on L1. This has to do with the fact that the starkware feeder-gateway does not have a static pubkey.*

*Second note: When tendermint consensus will be integrated, this signature verification step (forward sync) will check consensus signatures instead*

As you can see, our requirements for p2p sync are quite a bit more complex than the the current madara sync, and it cannot fit into the existing structure. Thus, we decided to refactor all of it into nice abstractions that supports both peer-to-peer sync as well as gateway sync.

In this refactor, a [Pipeline](https://github.com/madara-alliance/madara/blob/sync-refactor/crates/madara/client/sync/src/pipeline.rs#L24-L33) refers to a **Parallel-over-N-blocks step** + **Sequential one block at a time step** combination.
Our peer-to-peer forward sync implementation (not present in this pull request) has a pipeline for each of the "p2p stream endpoints" (mentioned earlier), and a pipeline for Apply State that only has a sequential step, which is when we update the global tries.

This PipelineController abstraction supports retries in the sequential step, which, while it is not used by the Gateway sync implementation present in this pull request, is very important for peer to peer as there are cases (particularily in the headers pipeline) where we cannot know whether a peer has returned wrong data until the sequential step. In that case, we want to reschedule a parallel step for that block and start over with a new peer. This is the main reason this abstraction looks the way it does.

On top of that, we have a [ForwardSync](https://github.com/madara-alliance/madara/blob/sync-refactor/crates/madara/client/sync/src/sync.rs) abstraction, which is responsible for responsible for following L1, and also running a "probe" - a small future that returns the highest known block on the network/gateway.

[This](https://github.com/madara-alliance/madara/blob/sync-refactor/crates/madara/client/sync/src/gateway/mod.rs#L157-L188) is how ForwardSync is used by the new Gateway sync. As you can see, the abstractions fit together neatly and it becomes easy to drive all of the pipelines. I believe this is a big deal, as **this makes the Reorg support easy to support in the future**: we have a single centralized place that has direct control over all of pipelines, and we can abort any fetch or task cleanly there at will.

Other details include:
- The backend now has different sync-tip (latest block_n counter) for every pipelines.
- The deletion of the block-import implies that the global tries are updated elsewhere now. They're updated in mc-db - the code is cleaner now as more of the logic is shared with block-production.
- In addition, closing a block (going from pending -> closed block) is now implemented in primitives. This makes block-production a lot cleaner as well. Moreover, we use PendingBlocks a lot more as a general construct for a "non-closed block".

Here is the implementation of [ForwardSync](https://github.com/madara-alliance/madara/blob/p2p_4/crates/madara/client/sync/src/p2p/forward_sync.rs#L220) in our p2p sync (not in this PR) implementation, if you are curious.

## TODO

Few things that I am still missing before this pull request can be merged, but it can still be reviewed.
- [x] More testing: re-add the feeder gateway mocking test.
- [ ] Fix pre-v0.12.3 block hash verification.
- [ ] Add the --n-blocks-to-sync -> --sync-stop-at breaking change in the CHANGELOG.md file


cchudant:
Breaking changes:
- DB schema is different.
- I have removed the `--n-blocks-to-sync <number of blocks>` cli arg, because I believe this arg is better expressed as `--sync-stop-at <block_n>`. I believe sync-stop-at is the objectively better version of that cli argument. What do you think? I do not see any use-case for n-blocks-to-sync, but if anyone really wants it I can add it back and we can have both.
- I have removed --no-sync-polling because it doesn't quite make sense in this new architecture.

crates/madara/client/db/src/block_db.rs
```diff
@@ -123,7 +110,8 @@ impl MadaraBackend {
 
     #[tracing::instrument(skip(self), fields(module = "BlockDB"))]
     pub fn get_latest_block_n(&self) -> Result<Option<u64>> {
-        get_latest_block_n(&self.db)
+        Ok(self.head_status().latest_full_block_n())
+        // get_latest_block_n(&self.db)
```
Trantorian1:
```suggestion
```

crates/madara/client/db/src/db_block_id.rs
```diff
@@ -1,8 +1,36 @@
...
-use mp_block::BlockId;
+/// Raw DB Block Id. Queries using these block ids will be able to index past the latest current block.
+/// The sync process has not yet marked these blocks as fully imported. You should use [`DbBlockId`] instead,
+/// as queries using it will not return anything past the current latest block.
+/// The sync process may use [`RawDbBlockId`]s to get the blocks it has saved past the latest full block.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub enum RawDbBlockId {
+    Pending,
+    Number(u64),
+}
```
Trantorian1: Why not implement this as a wrapper or alias around `DbBlockId`?

crates/madara/client/db/src/lib.rs
```diff
@@ -409,29 +412,29 @@ impl EventChannels {
...
     chain_config: Arc<ChainConfig>,
     db_metrics: DbMetrics,
     snapshots: Arc<Snapshots>,
-    trie_log_config: TrieLogConfig,
     sender_block_info: tokio::sync::broadcast::Sender<mp_block::MadaraBlockInfo>,
+    head_status: ChainHead,
     sender_event: EventChannels,
-    write_opt_no_wal: WriteOptions,
-    #[cfg(any(test, feature = "testing"))]
-    _temp_dir: Option<tempfile::TempDir>,
```
Trantorian1: No more tempdir?

crates/madara/client/db/src/lib.rs
```diff
@@ -506,56 +500,122 @@ impl Drop for MadaraBackend {
...
+#[derive(Debug)]
+pub struct MadaraBackendConfig {
+    pub base_path: PathBuf,
+    pub backup_dir: Option<PathBuf>,
+    pub restore_from_latest_backup: bool,
+    pub trie_log: TrieLogConfig,
+    pub backup_every_n_blocks: Option<u64>,
+    pub flush_every_n_blocks: Option<u64>,
+    #[cfg(any(test, feature = "testing"))]
+    pub _temp_dir: Option<tempfile::TempDir>,
```
Trantorian1: Ohh ok. Neat!

crates/madara/client/db/src/lib.rs
```diff
@@ -568,30 +628,34 @@ impl MadaraBackend {
...
-        });
+        let mut backend = Self::new(backup_handle, db, chain_config, config)?;
         backend.check_configuration()?;
+        backend.load_head_status_from_db()?;
         backend.update_metrics();
-        Ok(backend)
+        Ok(Arc::new(backend))
+    }
+
+    pub async fn on_block(&self, block_n: u64) -> anyhow::Result<()> {
```
Trantorian1: Maybe add a doc comemnt explaining what this function does in more details? Seems pretty vague atm.

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+    for receipt in receipts.iter_mut() {
+        let events_mut = match receipt {
+            TransactionReceipt::Invoke(receipt) => &mut receipt.events,
+            TransactionReceipt::L1Handler(receipt) => &mut receipt.events,
+            TransactionReceipt::Declare(receipt) => &mut receipt.events,
+            TransactionReceipt::Deploy(receipt) => &mut receipt.events,
+            TransactionReceipt::DeployAccount(receipt) => &mut receipt.events,
+        };
+        // just in case we stored them with receipt earlier, overwrite them
+        events_mut.clear()
```
Trantorian1: Does this happen in production or is this just a safety check?

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+            }
+            let _item = inner_m.next();
+        };
+
+        let events_mut = match receipt_mut {
+            TransactionReceipt::Invoke(receipt) => &mut receipt.events,
+            TransactionReceipt::L1Handler(receipt) => &mut receipt.events,
+            TransactionReceipt::Declare(receipt) => &mut receipt.events,
+            TransactionReceipt::Deploy(receipt) => &mut receipt.events,
+            TransactionReceipt::DeployAccount(receipt) => &mut receipt.events,
```
Trantorian1: Do we want to put this behind a function in `TransactionReceipt`? Seems like this could be handy.

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+        self.store_block_header(BlockHeaderWithSignatures {
+            header: block.header,
+            block_hash: block.block_hash,
+            consensus_signatures: vec![],
+        })?;
+        self.store_transactions(block_n, block.transactions)?;
+        self.store_state_diff(block_n, block.state_diff)?;
+        self.store_events(block_n, block.events)?;
+        Ok(())
+    }
```
Trantorian1: That will change for the p2p no? We will still be storing block headers, transactions, state diffs and events but at different heights, is that correct?

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+
+        self.block_db_store_pending(&MadaraPendingBlock { info, inner }, &block.state_diff)?;
+        self.contract_db_store_pending(ContractDbBlockUpdate::from_state_diff(block.state_diff))?;
+        Ok(())
+    }
+
+    pub fn store_block_header(&self, header: BlockHeaderWithSignatures) -> Result<(), MadaraStorageError> {
+        // Clear pending block when storing a new block header. This is the best place to do it IMO since
+        // it would make no sense to be able to store a block header if there is also a pending block, and
+        // we want to be sure to clear the pending block if we restart the sync pipeline.
```
Trantorian1: What are some cases where we might be storing a block header but still not have removed the pending block? Or is this the only place where we remove the pending block from the db?

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+
+        let block_n_encoded = bincode::serialize(&block_n)?;
+        tx.put_cf(&block_n_to_block, block_n.to_be_bytes(), bincode::serialize(&info)?);
+        tx.put_cf(&block_hash_to_block_n, &bincode::serialize(&header.block_hash)?, &block_n_encoded);
+
+        self.db.write_opt(tx, &self.writeopts_no_wal)?;
+        Ok(())
+    }
+
+    pub fn store_transactions(
```
Trantorian1: Maybe we want to add some doc comments for these function referencing `add_full_block_with_classes` so new maintainers have a rapid point of reference to it (not that I think new maintainers should be touching the db, but I think discovery is important)

crates/madara/client/db/src/storage_updates.rs
```diff
@@ -1,77 +1,265 @@
...
+
+        store_events_to_receipts(&mut inner.receipts, value)?;
+
+        batch.put_cf(&block_n_to_block_inner, &block_n_encoded, &bincode::serialize(&inner)?);
+        self.db.write_opt(batch, &self.writeopts_no_wal)?;
+
+        Ok(())
+    }
+
     /// NB: This functions needs to run on the rayon thread pool
```
Trantorian1: I really wonder: is there any way to enforce this at the type level? This is a wild tangent but Kotlin has a neat language construct known as [scope functions](https://kotlinlang.org/docs/scope-functions.html) which allows to transparently operate on a class instance within a scope (this is primarily used in frameworks such as Jetpack Compose to enforce certain functions are only available in a certain scope). This is not available in Rust of course but I wonder if there are some ways we can make calls to functions like this a bit safer, maybe taking inspiration from something like scope functions.

crates/madara/client/db/src/stream.rs
```diff
@@ -0,0 +1,822 @@
...
+        // This implementation solves this problem by checking the latest block in db in a loop and only once it really looks
+        //  like we caught up with the db, we subscribe to the new blocks channel. But hold on a minute, this subscribe is
+        //  done after getting the latest block number! There's a split second where it could have been possible to miss a
+        //  block. Because of this rare case, there are two supplementary things to note: we *also* get the latest block_n
+        //  *after* subscribing, so that we can check that we did not miss anything during subscription - and just in case,
+        //  we also handle the case when the subscription returns a block that's futher into the future than the one we
+        //  would expect.
+        // All in all, this implementation tries its *very best* not to subscribe to the channel when it does not have to.
+        // In addition, because rust does not have `yield` syntax (yet? I'm losing hope..) - this is implemented as a
+        //  funky looking state machine. yay!
```
Trantorian1: Move this to a doc comment so it can seen from the context of autocomplete.

crates/madara/client/eth/src/state_update.rs
```diff
@@ -20,18 +20,52 @@ pub struct L1StateUpdate {
...
+    // Listen to LogStateUpdate (0x77552641) update and send changes continuously
+    let mut event_stream = event_filter.watch().await.context(ERR_ARCHIVE)?.into_stream();
+
+    // // This does not seem to play well with anvil
+    // #[cfg(not(test))]
+    // {
+    let state_update = get_initial_state(&eth_client).await.context("Getting initial ethereum state")?;
+    update_l1(&backend, &state_update, &eth_client.l1_block_metrics)?;
+    l1_head_sender.send_modify(|s| *s = Some(state_update.clone()));
+    // }
```
Trantorian1:
```suggestion
    let state_update = get_initial_state(&eth_client).await.context("Getting initial ethereum state")?;
    update_l1(&backend, &state_update, &eth_client.l1_block_metrics)?;
    l1_head_sender.send_modify(|s| *s = Some(state_update.clone()));
```

crates/madara/client/gateway/client/src/methods.rs
```diff
@@ -36,6 +38,16 @@ impl GatewayProvider {
...
+    pub async fn get_header(&self, block_id: BlockId) -> Result<ProviderBlockHeader, SequencerError> {
+        let request = RequestBuilder::new(&self.client, self.feeder_gateway_url.clone(), self.headers.clone())
+            .add_uri_segment("get_block")
+            .expect("Failed to add URI segment. This should not fail in prod.")
+            .with_block_id(&block_id)
+            .add_param("headerOnly", "true");
+
+        request.send_get::<ProviderBlockHeader>().await
+    }
+
```
Trantorian1: Do we have some tests for this?

crates/madara/client/sync/src/counter.rs
Trantorian1, line 1: We probably want to add tests for this no?

crates/madara/client/sync/src/gateway/blocks.rs
```diff
@@ -0,0 +1,214 @@
...
+    type InputItem = ();
+    type SequentialStepInput = Vec<StateDiff>;
+    type Output = Vec<StateDiff>;
+
+    async fn parallel_step(
+        self: Arc<Self>,
+        block_range: Range<u64>,
+        _input: Vec<Self::InputItem>,
+    ) -> anyhow::Result<Self::SequentialStepInput> {
+        AbortOnDrop::spawn(async move {
```
Trantorian1: Why the AbortOnDrop if we are awaiting this anyway? Is this in case we cancel with the service context?

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+    }
+    pub fn all_verifications_disabled(self, no_check: bool) -> Self {
+        Self { no_check, ..self }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum BlockImportError {
+    // #[error("Transaction hash mismatch for index #{index}: expected {expected:#x}, got {got:#x}")]
+    // TransactionHash { index: usize, got: Felt, expected: Felt },
```
Trantorian1:
```suggestion
```

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+    ClassCount { got: u64, expected: u64 },
+    #[error("Compiled class hash mismatch for class hash {class_hash:#x}: expected {expected:#x}, got {got:#x}")]
+    CompiledClassHash { class_hash: Felt, got: Felt, expected: Felt },
+    #[error("Class with hash {class_hash:#x} failed to compile: {error}")]
+    CompilationClassError { class_hash: Felt, error: ClassCompilationError },
+    #[error("Failed to compute class hash {class_hash:#x}: {error}")]
+    ComputeClassHash { class_hash: Felt, error: ComputeClassHashError },
+
+    // #[error("Block hash mismatch: expected {expected:#x}, got {got:#x}")]
+    // BlockHash { got: Felt, expected: Felt },
```
Trantorian1:
```suggestion
```

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+    #[error("Block number mismatch: expected {expected:#x}, got {got:#x}")]
+    BlockNumber { got: u64, expected: u64 },
+
+    // #[error("Block order mismatch: database expects to import block #{expected}, trying to import #{got}. To import a block out of order, use the `ignore_block_order` flag.")]
+    // LatestBlockN { expected: u64, got: u64 },
+    // #[error("Parent hash mismatch: expected {expected:#x}, got {got:#x}")]
+    // ParentHash { got: Felt, expected: Felt },
```
Trantorian1:
```suggestion
```

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// Internal error, see [`BlockImportError::is_internal`].
+    #[error("Internal database error while {context}: {error:#}")]
+    InternalDb { context: Cow<'static, str>, error: MadaraStorageError },
+    /// Internal error, see [`BlockImportError::is_internal`].
+    #[error("Internal error: {0}")]
+    Internal(Cow<'static, str>),
+}
+impl BlockImportError {
+    /// Unrecoverable errors.
+    pub fn is_internal(&self) -> bool {
```
Trantorian1: perhaps we should rename this to `is_unrecoverable` then?

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+        // TODO: pre_v0_13_2_override
+        let _block_hash = signed_header
+            .header
+            .compute_hash(self.db.chain_config().chain_id.to_felt(), /* pre_v0_13_2_override */ true);
+        // if signed_header.block_hash != block_hash {
+        //     return Err(P2pError::peer_error(format!(
+        //         "Mismatched block_hash: {:#x}, expected {:#x}",
+        //         signed_header.block_hash, block_hash
+        //     )));
+        // }
```
Trantorian1: This section is still WIP? I assume we shouldn't merge until then no? Putting the review in RFC just in case.

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+                let got = tx.transaction.compute_hash(
+                    self.db.chain_config().chain_id.to_felt(),
+                    starknet_version,
+                    /* is_query */ false,
+                );
+                // For pre-v0.13.2, our tx hash is only used for commitment computation.
+                // let expected = tx.receipt.transaction_hash();
+                // // if expected != got {
+                // //     return Err(BlockImportError::TransactionHash { index, got, expected });
+                // // }
```
Trantorian1:
```suggestion
```

crates/madara/client/sync/src/import.rs
```diff
@@ -0,0 +1,577 @@
...
+            }
+
+            Ok(())
+        })
+        .await
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Trantorian1: Wdyt, is it worth adding some test cases for the other verify and update functions?

crates/madara/client/sync/src/pipeline.rs
```diff
@@ -0,0 +1,198 @@
...
+        input: Vec<Self::InputItem>,
+    ) -> impl Future<Output = anyhow::Result<Self::SequentialStepInput>> + Send;
+    fn sequential_step(
+        self: Arc<Self>,
+        block_range: Range<u64>,
+        input: Self::SequentialStepInput,
+    ) -> impl Future<Output = anyhow::Result<ApplyOutcome<Self::Output>>> + Send;
+
+    fn starting_block_n(&self) -> Option<u64>;
+}
```
Trantorian1: Real cool stuff!

Trantorian1: Maybe its worth adding some small docs here? The code is mostly self explanatory but at least personally the fact that the result of `parallel_step` was being driven into `sequential_step` and that the pipeline was essentially `InputItem` -> `SequentialStepInput` -> `Output` was a bit unclear until I read more of the code. Maybe this part can be made more explicit in a few lines of doc.

crates/madara/client/sync/src/pipeline.rs
```diff
@@ -0,0 +1,198 @@
...
+
+pub struct PipelineController<S: PipelineSteps> {
+    steps: Arc<S>,
+    queue: FuturesOrdered<ParallelStepFuture<S>>,
```
Trantorian1: Same here, it is still a bit unclear to me what `queue` does: am I right in assuming it is just a FIFO queue of futures to be executed by the `PipelineController`?

crates/madara/client/sync/src/pipeline.rs
Trantorian1, line 1: This section definitely feels like it could benefit from more docs. The code is very clear, but only in the context of the rest of the PR. It feels like someone who is discovering this part of the codebase for the first time without the context of having reviewed this PR would be rather confused. Perhaps a module-level doc with a brief overview of the pipeline architecture as well as some docs for `PipelineController` and its public-facing methods would be a good idea.

crates/madara/client/sync/src/probe.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+        let fut = self.future.get_or_insert_with(|| (self.make_future)(self.last_val.clone()));
+        let res = fut.await;
+        self.future = None;
+        let res = res?;
+        self.wait = Some(Instant::now() + self.wait_duration);
+
+        self.last_val = res.clone();
+        Ok(res)
+    }
```
Trantorian1: If I get this correctly, this is a future which continuously drives itself with the result it produces, introducing a throttle delay in between calls. This is really cool! At the risk of sounding repetitive, I feel like just a few lines of doc comments on the struct explaining this to newcomers as well as WHY this exists and where it might be used would go a long way.

crates/madara/client/sync/src/tests/gateway_mock.rs
```diff
@@ -233,120 +243,26 @@ impl TestContext {
...
+    //         when.method("GET").path_contains("get_signature");
+    //         then.status(200).header("content_type", "application/json").json_body(json!({
+    //                 "block_hash": "0x541112d5d5937a66ff09425a0256e53ac5c4f554be7e24917fc21a71aa3cf32",
+    //                 "signature": [
+    //                     "0x315b1d77f8b1fc85657725639e88d4e1bfe846b4a866ddeb2e74cd91ccff9ca",
+    //                     "0x3cbd913e55ca0c9ab107a5988dd4c54d56dd3700948a2b96c19d4728a5864de"
+    //                 ]
+    //         }));
+    //     });
+    // }
```
Trantorian1: Yeet?
```suggestion
```

crates/madara/client/sync/src/tests/pipeline.rs
```diff
@@ -0,0 +1,396 @@
...
+            .unwrap()
+            .unwrap()
+            .as_pending()
+            .unwrap()
+            .header
+            .parent_block_hash,
+        felt!("0x13")
+    );
+
+    task.await // task returned.
```
Trantorian1: Should add a timeout to this test in that case.

crates/madara/client/sync/src/tests/pipeline.rs
```diff
@@ -0,0 +1,396 @@
...
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::SyncingTo { target: 2 });
+    assert_eq!(ctx.service_state_recv.recv().await, None); // task ended
+
+    assert_eq!(ctx.backend.get_block_hash(&DbBlockId::Number(1)).unwrap().unwrap(), felt!("0x11"));
+    assert_eq!(ctx.backend.get_block_hash(&DbBlockId::Number(2)).unwrap().unwrap(), felt!("0x12"));
+    // third block should not be imported
+    assert_eq!(ctx.backend.get_block_hash(&DbBlockId::Number(3)).unwrap(), None);
+    assert_eq!(ctx.backend.has_pending_block().unwrap(), false);
+
+    task.await // task returned.
```
Trantorian1: Ditto

crates/madara/client/sync/src/tests/pipeline.rs
```diff
@@ -0,0 +1,396 @@
...
+    let task = AbortOnDrop::spawn(async move { sync.run(service_ctx_.child()).await });
+
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::Starting);
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::Idle);
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::SyncingTo { target: 1 });
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::Idle);
+    assert_eq!(ctx.service_state_recv.recv().await, None); // task ended
+
+    assert_eq!(ctx.backend.get_block_hash(&DbBlockId::Number(0)).unwrap().unwrap(), felt!("0x10"));
+    assert_eq!(ctx.backend.get_block_hash(&DbBlockId::Number(1)).unwrap().unwrap(), felt!("0x11"));
```
Trantorian1: SHould also make sure that block 2 has not been stored in the backend no?

crates/madara/client/sync/src/tests/pipeline.rs
Trantorian1, line 1: Noice! Really cool test suite and instrumentation :D

crates/madara/client/sync/src/util.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+pub struct DisplayFromFn<F: Fn(&mut fmt::Formatter<'_>) -> fmt::Result>(pub F);
+impl<F: Fn(&mut fmt::Formatter<'_>) -> fmt::Result> fmt::Display for DisplayFromFn<F> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        (self.0)(f)
+    }
+}
+
+pub struct AbortOnDrop<T>(JoinHandle<T>);
```
Trantorian1: I still don't quite understand in which scenario we need this: is this to handle panics/errors/service cancelling?

crates/madara/client/sync/src/util.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl<T> ServiceStateSender<T> {
+    pub fn send(&self, val: T) {
+        if let Some(sender) = &self.0 {
+            let _res = sender.send(val);
+        }
+    }
+}
+
+#[allow(unused)] // only used in tests
```
Trantorian1: Shouldn't we just `#[cfg(test)]` this then?

crates/madara/node/src/service/l2.rs
```diff
@@ -1,84 +1,171 @@
...
+                    tracing::error!(
+                        "❗ Failed to connect to warp update sender on http://localhost:{warp_update_port_rpc}"
+                    );
+                    ctx.cancel_global();
+                    return Ok(());
+                }
+
+                let gateway = Arc::new(GatewayProvider::new(
+                    Url::parse(&format!("http://localhost:{warp_update_port_fgw}/gateway/"))
+                        .expect("Failed to parse warp update sender gateway url. This should not fail in prod"),
```
Trantorian1: Now I think about i, `warp_update_port_fgw` is user-provided input, are we checking to make sure this is a valid port beforehand?

crates/madara/primitives/block/src/commitments.rs
```diff
@@ -0,0 +1,242 @@
...
+            parent_block_hash: Felt::ZERO,
+            sequencer_address: Felt::ZERO,
+            block_timestamp: BlockTimestamp(0),
+            protocol_version: StarknetVersion::V0_13_2,
+            l1_gas_price: GasPrices::default(),
+            l1_da_mode: L1DataAvailabilityMode::Blob,
+        }
+    }
+
+    // TODO: need more tests here.
```
Trantorian1: Would agree, we should probably be testing the other commitment methods.

crates/madara/primitives/block/src/lib.rs
```diff
@@ -294,6 +328,45 @@ pub struct VisitedSegmentEntry {
...
 
+#[derive(Clone)]
+pub struct FullBlock {
+    pub block_hash: Felt,
+    pub header: Header,
+    pub state_diff: StateDiff,
+    pub transactions: Vec<TransactionWithReceipt>,
+    pub events: Vec<EventWithTransactionHash>,
+}
+
```
Trantorian1: Nice! Does this mean our handling of `pending` and `full` has fully separate types at last?

crates/madara/primitives/class/src/compile.rs
```diff
@@ -25,25 +25,21 @@ pub enum ClassCompilationError {
...
-            serde_json::from_reader(crate::convert::gz_decompress_stream(self.program.as_slice()))?;
-
-        let program_object = program.as_object_mut().ok_or(ClassCompilationError::ProgramIsNotAnObject)?;
-
-        if !program_object.contains_key("debug_info") {
-            program_object.insert("debug_info".to_owned(), serde_json::json!(""));
-        }
-
+    // Returns `impl serde::Serialize` because the fact that it returns a serde_json::Value is an impl detail
+    // we should actually change that, it would be better to have a concrete type here.
```
Trantorian1: Which one is it, implementation detail or not? :rofl:

crates/madara/primitives/class/src/lib.rs
```diff
@@ -63,6 +64,47 @@ pub struct SierraConvertedClass {
...
+}
+
+impl fmt::Display for ClassType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Sierra => write!(f, "Sierra"),
+            Self::Legacy => write!(f, "Legacy"),
+        }
+    }
+}
```
Trantorian1: Why not just rely on `Debug`?

crates/madara/primitives/convert/src/felt.rs
```diff
@@ -1,11 +1,76 @@
...
+
+        let mut unpacked = [0; 32];
+        for i in 0..slice.len() {
+            unpacked[32 - slice.len() + i] = slice[i]
+        }
+
+        Felt::from_bytes_checked(&unpacked)
+    }
+
+    fn from_bytes_checked(b: &[u8; 32]) -> Result<Felt, MalformatedFelt> {
```
Trantorian1:
```suggestion
    fn from_bytes_be_checked(b: &[u8; 32]) -> Result<Felt, MalformatedFelt> {
```

crates/madara/primitives/convert/src/felt.rs
Trantorian1, line 1: We should and tests for the slice and bytes conversions

crates/madara/primitives/receipt/src/from_blockifier.rs
```diff
@@ -172,14 +160,16 @@ pub fn from_blockifier_execution_info(res: &TransactionExecutionInfo, tx: &Trans
...
+        Transaction::L1HandlerTransaction(tx) => TransactionReceipt::L1Handler(L1HandlerTransactionReceipt {
             transaction_hash,
             actual_fee,
             messages_sent,
             events,
             execution_resources,
             execution_result,
-            message_hash: message_hash.unwrap(), // it's a safe unwrap because it would've panicked earlier if it was Err
+            // This should not panic unless blockifier gives a garbage receipt.
+            // TODO: we should have a soft error here just in case.
```
Trantorian1: We should probably make this a soft error as part of this PR then no?

crates/madara/primitives/utils/src/rayon.rs
```diff
@@ -22,7 +22,7 @@ impl Default for RayonPool {
 impl RayonPool {
     pub fn new() -> Self {
         let n_cores = thread::available_parallelism().expect("Getting the number of cores").get();
-        let max_tasks = n_cores * 2;
+        let max_tasks = n_cores * 128;
```
Trantorian1: Why the sudden change to this constant? It feels like there should be more info around this magic number.

crates/madara/tests/src/lib.rs
```diff
@@ -265,9 +265,8 @@ async fn madara_can_sync_a_few_blocks() {
         "--full",
         "--network",
         "sepolia",
-        "--no-sync-polling",
-        "--n-blocks-to-sync",
-        "20",
+        "--sync-stop-at",
+        "19",
```
Trantorian1: Does this mean we still sync block 20 with `--sync-stop-at 19` or that the previous version would actually stop at block 19?

Trantorian1 (dismissed):
I apologize in advance, most of this review is about documentation X) (among a few clarifying questions and some nitpicks).

My worry is that since this PR introduces some pretty large and sweeping architectural changes that it will be easy for new and existing contributors alike to get lost in the codebase unless they have done a full review of this PR themselves. This is a lot of context to require, and it feels like we can make this easier by documenting the new sync architecture as part of the code. IMO this issue's opening comment should be added to the codebase, or at least linked, as it is an excelled overview of these changes.

The code seems good to me, though it is hard to review logic at a very granular level due to the size of this PR, but changes seem consistent with what was already present in the previous sync model.

I definitely feel like this makes tests all the more important, and there are some very important areas of the codebase which I think need more of those (though I think we are testing all the sync pipelines? Still feels like some db methods around `verify` and `update` in particular should be tested in isolation + maybe test more failing cases?). Please let me know what you think. 

Overall, this is definitely an amazing improvement to the current sync pipeline and some very impressive work :)

Trantorian1:
Well, for some reason github running at 1fps decided to split my review into two, strange :man_shrugging:

crates/madara/client/db/src/db_block_id.rs
cchudant: so that you can write RawDbBlockId::Pending and DbBlockId::Pending

crates/madara/client/db/src/lib.rs
cchudant: it's inside the Config struct now

crates/madara/client/db/src/storage_updates.rs
cchudant: It's just a way to make sure store_events is indempotent, which is a property that I want for these store functions
Also, I think I don't clear out the events when storing the receipts in gateway sync - and in that case it makes sure that we don't end up with duplicated events.

crates/madara/client/db/src/storage_updates.rs
cchudant: store_full_block is not called in the p2p code
instead, store_block_header, store_transactions, store_state_diff, store_events are called by their respective pipelines instead
store full block is just a convenience function

crates/madara/client/db/src/storage_updates.rs
cchudant: this is the only place where we clear the pending block in db afaik.

crates/madara/client/db/src/storage_updates.rs
cchudant: Yes it can be done, this is kind of what I did with BlockImportCtx in sync import.rs IMO
That kind of pattern is quite frequent in rust, it looks something like this:

You just have to make a kind of new type that acts like a "token" that you have to provide to the functions that need rayon, and this type cannot be instanciated from outside. It can only be given to you through the argument of an FnOnce like
The token struct can actually just be a ZST then.
```rs
run_in_rayon(|rayon_token: RayonToken| {
  backend.store_block(block, &rayon_token);
})
```

In the case of the BlockImportCtx I made it also acts as the receiver type, but it can be a standalone ZST struct too.

I want to limit my modifications of the db crate as much as possible in this pr though, it can be added afterwards
I want to add better read/write structs kind of like transactions objects in general tbh, unsure how rayon will fit in this exactly though.

crates/madara/client/db/src/stream.rs
cchudant: why? this is an implementation detail.
Also, it is not really used in this pr I am thinking of splitting that off and only leave the Iterator type.

crates/madara/client/sync/src/gateway/blocks.rs
cchudant: oooh
that's an interesting question

Well, if the current function's future gets dropped, we really want to abort the inner tokio task.
All of the sync module is cancellation-safe, meaning any of the futures can be aborted by just dropping them.
This makes sense as:
- it's all just fetching anyway. Aborting a fetch should be always fine.
- all of the db updates 1) they don't have any await point inside of them, meaning dropping them at any await point should not leave the db in an invalid state, 2) db updates happen in the rayon pool anyway, where tasks cannot be aborted. They will run to completion even if the tokio caller is dropped.

If the AbortOnDrop future gets dropped, the cleanup code in its Drop implementation gets called, which aborts the task. Meaning tokio will stop polling it and in turn will call the task's future Drop implementation. Tokio tasks (and any future in rust in general) can only be aborted at an await point - because there is not concept of abort in rust, it's just "stop polling" and call Drop. 
CancellationTokens are usually very helpful if you want to run cleanup code that needs async/awaits - you can't do that in a Drop implementation. This however, may change in the future if the language finally decides on the design details and stabilizes the [async flavor of the Drop trait](https://doc.rust-lang.org/std/future/trait.AsyncDrop.html).

I also think that cancellation tokens offer a much cleaner way of expressing how a service should behave when dropped - you do not have to think about the whole hierarchy of .await points of your code, and what kind of cleanup you need for all of them; you can just continue polling your current tasks and return as soon as you are in a state that supports it.

However, I think that this is not something we particularily care abour here in the sync module. Instead it's just easier here to make sure that any future can be dropped at any await point in the whole module.

crates/madara/client/sync/src/import.rs
cchudant: i think we use the word "internal" in other parts of madara though
IMO internal is better, because in the p2p part in #463 we punish the peer if it's not internal, and we close down the node (bubble up the error as an anyhow error) if it is

crates/madara/client/sync/src/pipeline.rs
cchudant: do you have better names for them? I dont particularily like the current names

crates/madara/client/sync/src/pipeline.rs
cchudant: FuturesOrdered<F> is a collection of futures F that will poll every future you add to it all at once, like a big select!, but it will return the completed futures outputs in FIFO order.

If you're curious, futures-rs implements it as a
```rs
pub struct FuturesOrdered<T: Future> {
    in_progress_queue: FuturesUnordered<Fut>,
    queued_outputs: BinaryHeap<Fut>,
    next_incoming_index: i64,
    next_outgoing_index: i64,
}
```
https://docs.rs/futures-util/0.3.31/src/futures_util/stream/futures_ordered.rs.html#99

The reason I really like FuturesOrdered is because it supports push_back but also push_front too. Push front is really useful to handle retrying the whole pipeline in #463, begining from the parallel step if a verification fails because of peer bad data during the sequencial step. If the method was missing, I would have probably rolled out my own FuturesOrdered struct using FuturesUnordered in the same way.

FuturesOrdered is the building block of a ton of the stream combinators in futures-rs. When a function in futures-rs returns stuff in order but also polls stuff concurrently, it's most likely FuturesOrdered working underneath.
In turn, if you're more curious, the FuturesUnordered struct is the one basic building block for nearly all of the functions that provide concurrency in futures-rs.

crates/madara/client/sync/src/probe.rs
cchudant: I wonder if I'm reinventing the wheel here. Surely there exists a combinator like that somewhere, but i've never seen it.

crates/madara/client/sync/src/tests/pipeline.rs
cchudant: timeouts in tests are usually bad though

crates/madara/node/src/service/l2.rs
cchudant: i think it's an u16 in Clap, and clap will do the check for us if thats the case. I'll check

crates/madara/primitives/convert/src/felt.rs
cchudant: ah uh this is only going to be used for p2p, i'll remove it from this PR i think
but I am pretty sure I had tests for that - if it's not in that PR I don't know where they went :)

crates/madara/primitives/utils/src/rayon.rs
cchudant: mostly just vibes 🤷 
i do not know why I changed that, to be perfectly honest. I think I wanted to test the change and did not find any evidence that it's faster or slower, and forgot to remove it
the only important thing here is that it's bounded, i don't think the actual number of the bound matters that much

crates/madara/tests/src/lib.rs
cchudant: the previous version stops at 19, since there are 20 blocks between 0..=19, as we have to import block 0 too.

crates/madara/client/sync/src/probe.rs
cchudant: i also changed the name of that thing like 4 times, and i'm still not happy with that name please help me

crates/madara/client/db/src/storage_updates.rs
Trantorian1: Neat, I wasn't aware of this pattern!  I agree this is probably not the time to introduce this kind of change into the codebase, was more so just curious as to how this could be implemented in the future.

crates/madara/client/db/src/stream.rs
Trantorian1: I feel like this kind of implementation detail which goes into the edge cases of the function can be handy to have in the docs. Then again, just IMO.

crates/madara/client/sync/src/pipeline.rs
Trantorian1: No I think it makes sense in the context of the implementation. This is what they do after all

crates/madara/client/sync/src/pipeline.rs
Trantorian1: Interesting, will give it a look :eyes:

crates/madara/client/sync/src/probe.rs
Trantorian1: Idk, `ThrottledRepeatedFuture` is pretty much just what it does, can't really think of a better name than that

Trantorian1: The closest thing I could find (which might still be useful) is a combination of [`Throttle`](https://docs.rs/tokio-timer/latest/tokio_timer/throttle/struct.Throttle.html) from `tokio-timer` and [`repeat_with`](https://docs.rs/futures/0.3.31/futures/prelude/stream/fn.repeat_with.html) from `futures`. Maybe this is what you are looking for?

crates/madara/client/sync/src/tests/pipeline.rs
Trantorian1: (I use them a lot :sweat_smile:)

Trantorian1: I was mostly thinking in the case that this test would hand indefinitely in case a change was made to the pipeline in the future which made it never return.

crates/madara/primitives/utils/src/rayon.rs
Trantorian1: fair enough!

Trantorian1: I mean then again with that high a number of tasks performance might not have changed because no more than 2 * cores tasks were being processed at once so this was still under the bound of the previous limit.

It feels like 128 * cores leaves more for DOS though. Especially if this didn't result in any noticeable change in performance maybe we might want to revert this change. Wdyt?

crates/madara/tests/src/lib.rs
Trantorian1: Okay, this is much clearer then

crates/madara/client/gateway/client/src/methods.rs
cchudant: added :)

crates/madara/client/sync/src/counter.rs
cchudant: probably

crates/madara/node/src/service/l2.rs
cchudant: yes it is an u16

crates/madara/primitives/block/src/lib.rs
cchudant: there is still MadaraPendingBlock./ MadaraBlock and we should remove that at some point.

crates/madara/primitives/receipt/src/from_blockifier.rs
cchudant: let's not, or it's going to get painful when charpa updates blockifier.

crates/madara/primitives/utils/src/rayon.rs
cchudant: true

crates/madara/client/db/src/lib.rs
```diff
@@ -1,4 +1,39 @@
...
+//! - The database can store partial blocks. Adding headers can be done using [`MadaraBackend::store_block_header`],
+//!   transactions and receipts using [`MadaraBackend::store_transactions`], classes using [`MadaraBackend::class_db_store_block`],
+//!   state diffs using [`MadaraBackend::store_state_diff`], events using [`MadaraBackend::store_events`]. Furthermore,
+//!   [`MadaraBackend::apply_to_global_trie`] also needs to be called.
+//! - Each of those functions can be called in parallel, however, [`MadaraBackend::apply_to_global_trie`] needs to be called
+//!   sequentially. This is because we cannot support updating the global trie in an inter-block parallelism fashion. However,
+//!   parallelism is still used inside of that function - intra-block parallelism.
+//! - Each of these block parts have a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
+//!   which the block importer service can use however it wants. However, [`ChainHead::full_block`] is special,
+//!   as it is updated by this crate.
```
Trantorian1: The bit about `ChainHead` is a bit unclear to me. What does this refer to exactly?

crates/madara/client/db/src/lib.rs
Trantorian1, line 1: Great docs!

crates/madara/client/sync/src/counter.rs
Trantorian1: Pinging this again.

crates/madara/client/sync/src/import.rs
```diff
@@ -353,12 +366,22 @@ impl BlockImporter {
...
+                        if self.db.chain_config().chain_id == ChainId::Mainnet {
+                            // We do not actually implement class hash verification for some cairo 0 classes.
+                            // See [`mp_class::mainnet_legacy_class_hashes`] for more information about this; but this
+                            // only applies to a few classes on mainnet in total. We have decided to just hardcode them.
+                            expected = mp_class::mainnet_legacy_class_hashes::get_real_class_hash(block_n, expected)
+                        }
+                    }
+
                     if !self.config.no_check && class_hash != expected {
+                        // tracing::info!("Mismatched class_hash={class_hash:#x} expected={expected:#x} block_n={block_n:?}");
```
Trantorian1: Remove this?

crates/madara/client/sync/src/import.rs
```diff
@@ -466,36 +489,40 @@ impl BlockImporter {
...
+    pub fn apply_to_global_trie(
         &self,
         block_range: Range<u64>,
         state_diffs: Vec<StateDiff>,
-    ) -> anyhow::Result<()> {
+    ) -> Result<(), BlockImportError> {
         if block_range.is_empty() {
             return Ok(());
         }
+        let got = self.db.apply_to_global_trie(block_range.start, state_diffs.iter()).map_err(|error| {
```
Trantorian1: This is non-blocking?

crates/madara/primitives/class/src/compile.rs
```diff
@@ -40,6 +39,8 @@ impl CompressedLegacyContractClass {
         // NOTE: that the `type` field is already present in each ABI entry
         // struct so we do not need to add it manually.
 
+        // we should actually change that, it would be better to have a concrete type here.
```
Trantorian1: Moved comment or is this related to the above?

crates/madara/primitives/class/src/lib.rs
```diff
@@ -422,4 +427,23 @@ mod test {
         assert_eq!(missed_class_hashes.len(), 38);
         assert_eq!(missed_class_hashes.iter().map(|(_, v)| v.len()).sum::<usize>(), 57);
     }
+
+    #[test]
+    fn legacy_class_mainnet_block_20732_no_abi() {
```
Trantorian1: What is this test checking? Am I correct in my understanding that this is a special case on mainnet where a class wes deployed with no ABI and this was causing errors in the class hash computation?

crates/madara/primitives/class/src/mainnet_legacy_class_hashes.rs
```diff
@@ -0,0 +1,150 @@
...
+const BLOCK_N_RANGE: RangeInclusive<u64> = CLASS_HASHES[0].0..=CLASS_HASHES[CLASS_HASHES.len() - 1].0;
+
+pub fn get_real_class_hash(block_n: u64, computed: Felt) -> Felt {
+    if !BLOCK_N_RANGE.contains(&block_n) {
+        return computed;
+    }
+
+    // binary search
+    // the partition point is the first index in the array where n is >= to block_n
+    let partition_point = CLASS_HASHES.partition_point(|(n, _, _)| n < &block_n);
```
Trantorian1: Why not `binary_search_by_key`?

crates/madara/client/db/src/lib.rs
cchudant: backend.head_status().full_block() returns the latest block, I may need to make that clearer

crates/madara/client/sync/src/import.rs
cchudant: this should be called on the rayon pool

crates/madara/primitives/class/src/compile.rs
cchudant: I moved the comment to make it clearer, but I might need to make it clearer again

crates/madara/primitives/class/src/lib.rs
cchudant: yep

crates/madara/primitives/class/src/mainnet_legacy_class_hashes.rs
cchudant: partition point is binary search

you're right actually,
```rs
    let partition_point = CLASS_HASHES.partition_point(|(n, _, _)| n < &block_n);

    for index in partition_point.. {
        let (entry_block_n, entry_real_hash, entry_computed) = CLASS_HASHES[index];
        if entry_block_n != block_n {
            // we overshot the block_n
            break;
        }

        if entry_computed == computed {
            // found :) return our override hash
            return entry_real_hash;
        }
    }
```
=>
```rs
    let partition_point = CLASS_HASHES.binary_search_by_key(|(block_n, _, computed_hash)| (block_n, computed_hash));
```
then I would need to make sure the table is sorted wrt. computed hash after block_n.

raynaudoe:
@cchudant could you please integrate `main` into this branch and solve the remaining conflicts?

cchudant:
@raynaudoe it's done

I have also added a few more tests

crates/madara/client/sync/src/counter.rs
cchudant: i don't know if it's worth testing and i've tried to test it, it's a bit clunky to test
it's just for showing the blocks/s in the terminal, idk if i should spend more of my time on it

crates/madara/client/sync/src/import.rs
cchudant: done

crates/madara/primitives/block/src/commitments.rs
cchudant: done

crates/madara/client/sync/src/import.rs
cchudant: yes ive added some, but not for classes yet unfortunately

cchudant:
For the sake of reviewability, i'll explain how testing is done in this PR:


The testing is split in three like this, and I believe it's the best approach to test almost everything regarding syncing in madara.

1. End to end tests: https://github.com/madara-alliance/madara/blob/sync-refactor/madara/crates/tests/src/lib.rs#L286-L392
  - mc-e2e-test madara_can_sync_a_few_blocks is the most general test, which was already present before the pr
  - new test: madara_can_sync_and_restart. This test shuts down the node, and try to re-open it to see if it boots back up and can continue to sync. It also does the same thing but killing the node (ungraceful exit), to make sure the db can't get corrupted
  - E2e tests are hard to write for syncing - we can't really easily test all of the features without having access to at least some of the internal state. This is why my approach was to focus on integration tests here.
2. Integration tests. This is the most interesting part:
 - We have pipeline tests. https://github.com/madara-alliance/madara/blob/sync-refactor/madara/crates/client/sync/src/tests/pipeline.rs . They test every feature we support for syncing, and they use a channel to get notifications from the syncing task. This allows us to avoid timeouts in tests, and to check state that is difficult to track from outside the node.
 - This is where, for example, we test that when the latest block on l1 changes, the sync tracks it and follows it correctly.
 - We also have some realistic tests, mocking the gateway with real blocks from mainnet/sepolia. https://github.com/madara-alliance/madara/blob/sync-refactor/madara/crates/client/sync/src/tests/realistic.rs
3. Block import unit tests. https://github.com/madara-alliance/madara/blob/sync-refactor/madara/crates/client/sync/src/import.rs#L667-L844
 - They are mostly negative tests: they corrupt a valid block and see if we catch the error when importing it.
 - Receipt, events, transactions tests here are mostly done, but I want to add more tests in a later PR mostly regarding block hashes and classes.

madara/crates/client/sync/src/import.rs
```diff
@@ -600,4 +640,210 @@ mod tests {
...
+                ctx.block_n,
+                &ctx.block.state_diff,
+                &ctx.block.header,
+                ctx.allow_pre_v0_13_2
+            ),
+            Err(BlockImportError::StateDiffLength { got: 17, expected: 123123 })
+        );
+    }
+    #[rstest]
+    fn test_error_state_diff_commitment(mut ctx: Ctx) {
```
Trantorian1: Nitpick but in cases like this I find `test_error_state_diff_commitment1` to be a better option so running `cargo test test_error_state_diff_commitment1` will only run one test (atm we have no way to specify we only want to run this test because of the way in which it is named)

Trantorian1: The same goes for other numbered tests in this file.

madara/crates/client/sync/src/import.rs
Trantorian1, line 1: Great tests, especially the negative edge cases. Thanks a lot!

madara/crates/client/sync/src/metrics.rs
```diff
@@ -19,88 +22,89 @@ pub struct SyncMetrics {
...
 
     // L2 network metrics
-    pub l2_block_number: Gauge<u64>,
-    pub l2_sync_time: Gauge<f64>,
-    pub l2_avg_sync_time: Gauge<f64>,
-    pub l2_latest_sync_time: Gauge<f64>,
-    pub l2_state_size: Gauge<f64>, // TODO: remove this, as well as the return value from db_metrics update.
-    pub transaction_count: Gauge<u64>,
-    pub event_count: Gauge<u64>,
+    pub l2_block_number: Histogram<f64>,
```
Trantorian1, line 25: Why are we using a `Histogram` here? Is this based off main?

madara/crates/primitives/block_hashes/src/bin/create_hash_tables.rs
Trantorian1, line 1: Nice!

madara/crates/primitives/block_hashes/src/lib.rs
Trantorian1, line 1: What is this doing here?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -131,12 +129,10 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
     /// as was done before.
     pub async fn close_pending_block(
         backend: &MadaraBackend,
-        importer: &BlockImporter,
         metrics: &BlockProductionMetrics,
     ) -> Result<(), Cow<'static, str>> {
```
jbcaron, line 133: why not return an anyhow error?

Trantorian1 (approved):
Thanks for the changes! The new tests especially look neat :ok_hand:

madara/crates/primitives/block_hashes/src/bin/create_hash_tables.rs
cchudant: i am sad to tell you, this file wasn't meant to be included. Removed.

madara/crates/primitives/block_hashes/src/lib.rs
cchudant: removed

madara/crates/client/sync/src/import.rs
cchudant: done :)

madara/crates/client/sync/src/metrics.rs
cchudant: yes, it's from main
you're looking at the relative changes from the last review you did, which is why you see that. This line isn't changed in this pr

madara/crates/client/db/src/block_db.rs
```diff
@@ -106,7 +93,7 @@ impl MadaraBackend {
     #[tracing::instrument(skip(self), fields(module = "BlockDB"))]
     fn get_block_info_from_block_n(&self, block_n: u64) -> Result<Option<MadaraBlockInfo>> {
         let col = self.db.get_column(Column::BlockNToBlockInfo);
-        let res = self.db.get_cf(&col, bincode::serialize(&block_n)?)?;
+        let res = self.db.get_cf(&col, block_n.to_be_bytes())?;
```
jbcaron, line 96: Does this change the encoding in the DB? If so, the DB version needs to be updated with the db-migration label

madara/crates/client/db/src/contract_db.rs
```diff
@@ -1,16 +1,58 @@
...
+    contract_kv_updates: Vec<((Felt, Felt), Felt)>,
+}
+
+impl ContractDbBlockUpdate {
+    pub fn from_state_diff(state_diff: StateDiff) -> Self {
+        let nonces_from_updates =
+            state_diff.nonces.into_iter().map(|NonceUpdate { contract_address, nonce }| (contract_address, nonce));
+
+        // let nonce_map: HashMap<Felt, Felt> = nonces_from_deployed.chain(nonces_from_updates).collect(); // set nonce to zero when contract deployed
+        let nonce_map: HashMap<Felt, Felt> = nonces_from_updates.collect();
```
jbcaron: Can this be cleaned up?

madara/crates/client/db/src/lib.rs
```diff
@@ -1,9 +1,44 @@
...
+//! For the low-level API, there are a few responsibilities to follow:
+//!
+//! - The database can store partial blocks. Adding headers can be done using [`MadaraBackend::store_block_header`],
+//!   transactions and receipts using [`MadaraBackend::store_transactions`], classes using [`MadaraBackend::class_db_store_block`],
+//!   state diffs using [`MadaraBackend::store_state_diff`], events using [`MadaraBackend::store_events`]. Furthermore,
+//!   [`MadaraBackend::apply_to_global_trie`] also needs to be called.
+//! - Each of those functions can be called in parallel, however, [`MadaraBackend::apply_to_global_trie`] needs to be called
+//!   sequentially. This is because we cannot support updating the global trie in an inter-block parallelism fashion. However,
+//!   parallelism is still used inside of that function - intra-block parallelism.
+//! - Each of these block parts have a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
```
jbcaron:
```suggestion
//! - Each of these block parts has a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
```

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -1,90 +1,278 @@
...
+    /// Add a new block to the db, calling the `on_block` handler that handles flushes and backups when they are enabled,
+    /// and update all the statuses.
+    /// When using the MadaraBackend API to store new blocks, you either have the choice of using this simple function to store
+    /// an entire new block, or, you are responsible for calling the `store_xxx`, applying global state, and calling the on_block
+    /// function when the full block is stored in the backend.
+    ///
+    /// This function takes a PendingFullBlock because the commitments can only be computed once we applied the state to the global
+    /// tries. This function will close the block and compute the commitments.
+    ///
+    /// The function returnes the new computed block_hash.
```
jbcaron:
```suggestion
    /// The function returns the new computed block_hash.
```

madara/crates/client/db/src/stream.rs
```diff
@@ -0,0 +1,822 @@
...
+        //  connection with the client/peer that's calling the endpoint.
+        // So! it may very well be the case that once we caught up with the latest block_n in db as we saw at the beginning
+        //  of the call, and once we sent all the blocks that have been added within the time we sent all of those, we might
+        //  still not have caught up with the latest block in db - because new blocks could have come by then.
+        // This implementation solves this problem by checking the latest block in db in a loop and only once it really looks
+        //  like we caught up with the db, we subscribe to the new blocks channel. But hold on a minute, this subscribe is
+        //  done after getting the latest block number! There's a split second where it could have been possible to miss a
+        //  block. Because of this rare case, there are two supplementary things to note: we *also* get the latest block_n
+        //  *after* subscribing, so that we can check that we did not miss anything during subscription - and just in case,
+        //  we also handle the case when the subscription returns a block that's futher into the future than the one we
```
jbcaron:
```suggestion
        //  we also handle the case when the subscription returns a block that's further into the future than the one we
```

madara/crates/client/db/src/update_global_trie/contracts.rs
```diff
@@ -113,34 +113,40 @@ fn contract_state_leaf_hash(
...
     let nonce = contract_leaf.nonce.unwrap_or(
-        backend.get_contract_nonce_at(&BlockId::Tag(BlockTag::Latest), contract_address)?.unwrap_or(Felt::ZERO),
+        backend.get_contract_nonce_at(&RawDbBlockId::Number(block_number), contract_address)?.unwrap_or(Felt::ZERO),
     );
 
     let class_hash = contract_leaf.class_hash.unwrap_or(
-        backend.get_contract_class_hash_at(&BlockId::Tag(BlockTag::Latest), contract_address)?.unwrap_or(Felt::ZERO), // .ok_or(MadaraStorageError::InconsistentStorage("Class hash not found".into()))?
+        backend
+            .get_contract_class_hash_at(&RawDbBlockId::Number(block_number), contract_address)?
+            .unwrap_or(Felt::ZERO), // .ok_or(MadaraStorageError::InconsistentStorage("Class hash not found".into()))?
```
jbcaron, line 125: Why this behavior?

madara/crates/client/eth/src/state_update.rs
```diff
@@ -20,18 +20,53 @@ pub struct L1StateUpdate {
...
+        let log = event_result.context("listening for events")?;
+        let state_update = convert_log_state_update(log.0.clone()).context("formatting event into an L1StateUpdate")?;
+        update_l1(&backend, &state_update, &eth_client.l1_block_metrics)?;
+        l1_head_sender.send_modify(|s| *s = Some(state_update.clone()));
+    }
+
+    Ok(())
+}
+
 pub fn update_l1(
```
jbcaron: Is there a verification system for state root and block hash during updates from L1?

madara/crates/client/gateway/client/src/methods.rs
```diff
@@ -36,6 +35,16 @@ impl GatewayProvider {
         }
     }
 
+    pub async fn get_header(&self, block_id: BlockId) -> Result<ProviderBlockHeader, SequencerError> {
```
jbcaron, line 38: Great for the sync, it would be good to implement this method on the server side as well.

madara/crates/client/sync/src/gateway/classes.rs
```diff
@@ -0,0 +1,152 @@
...
+use mp_block::BlockId;
+use mp_class::{ClassInfo, ClassInfoWithHash, ConvertedClass, LegacyClassInfo, SierraClassInfo, MISSED_CLASS_HASHES};
+use mp_state_update::DeclaredClassCompiledClass;
+use starknet_api::core::ChainId;
+use starknet_core::types::Felt;
+use std::{collections::HashMap, ops::Range, sync::Arc};
+
+/// for blocks before 2597 on mainnet new classes are not declared in the state update
+/// https://github.com/madara-alliance/madara/issues/233
+fn fixup_missed_mainnet_classes(block_n: u64, classes_from_state_diff: &mut HashMap<Felt, DeclaredClassCompiledClass>) {
```
jbcaron, line 18: is there a test in the sync for this?

madara/crates/node/src/cli/l2.rs
```diff
@@ -52,108 +52,37 @@ pub struct L2SyncParams {
...
-    )]
-    pub sync_parallelism: u8,
+    /// Disable pending block sync.
+    #[clap(env = "MADARA_STOP_NO_PENDING_SYNC", long)]
+    pub no_pending_sync: bool,
+
+    /// Compute post-v0.13.2 hashes. This means that the feeder gateway will display different block commitments
+    /// for blocks that were created before v0.13.2. When p2p sync will be merged, this option will become the
+    /// default, as post-v0.13.2 commitments are mendatory for checking the integrity of these old blocks.
+    /// Madara cannot yet check whether the post-v0.13.2 commiments are correct. Thus, for now, enabling this setting
```
jbcaron:
```suggestion
    /// default, as post-v0.13.2 commitments are mandatory for checking the integrity of these old blocks.
    /// Madara cannot yet check whether the post-v0.13.2 commitments are correct. Thus, for now, enabling this setting
```

madara/crates/node/src/cli/l2.rs
```diff
@@ -162,21 +91,15 @@ impl L2SyncParams {
...
-            flush_every_n_blocks: self.flush_every_n_blocks,
-            flush_every_n_seconds: self.flush_every_n_seconds,
-            stop_on_sync: self.stop_on_sync,
-            sync_parallelism: self.sync_parallelism,
-            warp_update,
+        let mut client = GatewayProvider::new(gateway, feeder_gateway);
+
+        if let Some(api_key) = &self.gateway_key {
+            client.add_header(
+                HeaderName::from_static("x-throttling-bypass"),
```
jbcaron, line 98: Are we sure about the case of `x-throttling-bypass`

raynaudoe (approved):

madara/crates/client/block_production/src/lib.rs
cchudant: good question, it's changed in a later pr (i didnt want to change this crate too much)

madara/crates/client/db/src/update_global_trie/contracts.rs
cchudant: the behavior hasnt changed here, it's just rustfmt

madara/crates/client/eth/src/state_update.rs
cchudant: not yet, i wanted to leave that for reorg support in a later pr
although i could have made the node check the case and crash

madara/crates/client/gateway/client/src/methods.rs
cchudant: It's already implemented :)

madara/crates/client/sync/src/gateway/classes.rs
cchudant: the realistic integration test tests this yes :)

madara/crates/node/src/cli/l2.rs
cchudant: what do you mean? you want me to add a test for that?
i could add that to the mock test yea

jbcaron (approved):

madara/crates/node/src/cli/l2.rs
cchudant: http header names are case-insensitive

cchudant:
The CI fails on something related to the madara bootstrapper merge #557. I think we should just merge it, i have checked the tests pass locally on my machine

antiyro (approved):

PR #496, opened 2025-02-14
bugfix(rpc): Fix proof RPC

notlesh:
This fixes the proof RPC, specifically:

* Fix for edge nodes where they weren't converted to `felt`s properly
* Add contract storage root to proof

In addition, it adds a lot of test coverage to verify the proofs generated by the API.

Currently the RPC copies the bits from an edge path incorrectly into a felt. This fixes that and adds a couple trivial tests.

Also, the storage root for a contract was not accessible (other than as one of the random proof nodes), this adds it to the `ContractLeavesData`.

## What is the new behavior?

RPC is now providing valid proofs, which are passing `fn verify_proof()` in SNOS.

No


crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -178,25 +200,30 @@ pub fn get_storage_proof(
...
+        assert_eq!(path.len(), 2);
+        let felt = path_to_felt(path);
+        assert_eq!(felt, Felt::ZERO);
+
+        let path = bits![u8, Msb0; 1];
+        assert_eq!(path.len(), 1);
+        let felt = path_to_felt(path);
+        assert_eq!(felt, Felt::ONE);
+    }
+}
```
Trantorian1: Please add some tests verifying the format of the `getProof` endpoint so this kind of error does not happen again in the future. Perhaps there is a way to test this against SNOS directly? Anyway, even if we do not go that far we should definitely have tests for this case.

crates/madara/client/rpc/src/versions/user/v0_8_0/api.rs
```diff
@@ -30,6 +30,7 @@ pub struct NodeHashToNodeMappingItem {
 pub struct ContractLeavesDataItem {
     pub nonce: Felt,
     pub class_hash: Felt,
+    pub storage_root: Felt,
```
Trantorian1: I think we should add some doc comments which provide an overview of the `get_proof` endpoint format. Given that the `get_storage_proof` function can seem quite arcane I believe this would be a good idea and help with maintainability.

Trantorian1 (changes requested):
Please add tests so we can enforce the `getStorageProof` format across future changes.

It also seems like this section of the codebase could benefit from more documentation.

crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -150,6 +151,24 @@ pub fn get_storage_proof(
         bonsai_identifier::CLASS,
         class_hashes,
     )?;
+
+    let mut contract_root_hashes = std::collections::HashMap::new();
```
cchudant: ah yes, it looks like this was added into the specs
thanks!

crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -22,7 +22,8 @@ fn saturating_sum(iter: impl IntoIterator<Item = usize>) -> usize {
 fn path_to_felt(path: &BitSlice<u8, Msb0>) -> Felt {
     let mut arr = [0u8; 32];
     let slice = &mut BitSlice::from_slice_mut(&mut arr)[5..];
-    slice[..path.len()].copy_from_bitslice(path);
+    let slice_len = slice.len();
+    slice[slice_len - path.len()..].copy_from_bitslice(path);
```
cchudant: makes sense
maybe add a tiny comment saying that slice_len is always equal to 251 and this minus sign is fine because path cannot be larger than 251, it took me a minute to understand :)

crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
cchudant: i wish the specs came with test vectors :(

cchudant (approved):

crates/madara/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
notlesh: > Perhaps there is a way to test this against SNOS directly?

Testing this against SNOS was how we discovered this problem to begin with.

More testing would be great, but I could use some tips. Test vectors would be perfect (e.g. something like `(existing_mpt, nodes_requested) -> proof`). Is there anything close to this already (in other parts of Madara or in Bonsai Trie)?

If not, I imagine something like this:
* Start with empty MPT
* insert some leaves
* get proof
* assert exact structure of proof
* write a quick verify_proof fn, something like [what we have in SNOS](https://github.com/keep-starknet-strange/bonsai-trie)

notlesh: I could also argue that this is out of scope for this PR -- testing this could be very thorough. For example, we should also test non-existence proofs...

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -178,25 +200,84 @@ pub fn get_storage_proof(
...
+            &storage_key.to_bytes_be().as_bits()[5..].to_owned(),
+            &value,
+        ).unwrap();
+        
+        // create a dummy block to make get_storage_proof() happy
+        let header = create_dummy_header();
+        let pending_block = finalized_block_zero(header);
+        backend.store_block(
+            pending_block,
+            finalized_state_diff_zero(),
```
notlesh: I think passing in an empty state diff is causing this to fail, even though I'm inserting values manually into the `bonsai_trie`. Currently we get back an empty proof.

Trantorian1: @cchudant ?

notlesh: The problem was a missing call to `commit()` on the Bonsai trie, 7792592 has a fix.

It could still be useful to review this, though.

notlesh:
@Trantorian1 I added some very extensive tests, PTAL

notlesh:
@cchudant one of the failures in CI is related to some tests you added which also touch the proof: https://github.com/madara-alliance/madara/actions/runs/13739637755/job/38427368157?pr=496

The tests have hard-coded expected JSON values so I'm not sure how to proceed. Do you know where these values came from and how I can tell if I've introduced a regression? They came from #292 (unsurprisingly).

cchudant:
> @cchudant one of the failures in CI is related to some tests you added which also touch the proof: https://github.com/madara-alliance/madara/actions/runs/13739637755/job/38427368157?pr=496
> 
> The tests have hard-coded expected JSON values so I'm not sure how to proceed. Do you know where these values came from and how I can tell if I've introduced a regression? They came from #292 (unsurprisingly).

@notlesh 
Oh, update them I guess. I think I just ran the endpoint and verified by hand that the values looked fine, and I put that into a test.

cchudant (approved):
thank you! :)

madara/crates/client/rpc/Cargo.toml
```diff
@@ -20,6 +20,7 @@ rstest = { workspace = true }
 mc-db = { workspace = true, features = ["testing"] }
 mp-utils = { workspace = true, features = ["testing"] }
 mc-block-import = { workspace = true, features = ["testing"] }
+async-std = { workspace = true }
```
Trantorian1, line 23: Why are we using `async-std`?

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -230,54 +231,370 @@ mod tests {
...
+        ContractStorageTestInput::new(Felt::from(333), Felt::from(5), Felt::from(55)),
+    ])]
+    #[case::multi_mpt_duplicate_values(vec![
+        ContractStorageTestInput::new(Felt::from(222), Felt::from(5), Felt::from(55)),
+        ContractStorageTestInput::new(Felt::from(333), Felt::from(5), Felt::from(55)),
+        ContractStorageTestInput::new(Felt::from(444), Felt::from(5), Felt::from(55)),
+        ContractStorageTestInput::new(Felt::from(555), Felt::from(5), Felt::from(55)),
+    ])]
+    /// Tests `get_storage_proof()` as it relates to contract storage tries. This includes testing
+    /// multiple contract storage MPTs to ensure that it provides proofs for each.
```
Trantorian1, line 283: MPT = Merkle Patricia Tree here?

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
```diff
@@ -230,54 +231,370 @@ mod tests {
...
+                index += 1;
+                (contract_address, contract_leaves_data_item.storage_root)
+            })
+            .collect::<HashMap<_, _>>();
+
+        // collect all proof nodes into one big hash map. since the keys are hashes of the nodes
+        // themselves, there should be no collisions.
+        let mut proof_nodes = HashMap::new();
+        for node in storage_proof_result.contracts_storage_proofs.into_iter().flatten() {
+            proof_nodes.insert(node.node_hash, node.node);
```
Trantorian1: Maybe we should add an assert here to make sure there are no collisions?

Trantorian1 (approved):
Great tests, thanks for the changes!

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
notlesh: Correct :)

madara/crates/client/rpc/Cargo.toml
notlesh: For the test cases. It's under `[dev-dependencies]` so it should be isolated.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/read/get_storage_proof.rs
notlesh: I added a check for "collisions" in [941b00b](https://github.com/madara-alliance/madara/pull/496/commits/941b00b051ad01afbe270c853e1dbc467f496c07). I also made a distinction between "duplicate" (identical k:v pairs) and "collision" (same k, different v) in my comments. Hopefully that is pretty clear now :)

raynaudoe:
@notlesh can we merge this PR? are there any TODOs that need to be addressed?

notlesh:
> @notlesh can we merge this PR?

Yes, it's ready to go.

> are there any TODOs that need to be addressed?

Nope :)

PR #480, opened 2025-01-27
refactor(repo): madara orchestrator merge

Mohiiit:
_This PR was initially a continuation of #471 and #472 but ended up being where most of the CI-related work was done. CI needed to target `main` anyway, so we merged this and closed the others. The description below is from the original PR._

This PR pulls in the entire [madara-orchestrator](https://github.com/madara-alliance/madara-orchestrator) git history into this project.

Because the entire history was pulled in (using `git merge --allow-unrelated-histories`) the PR is absolutely massive (and probably difficult to review). My process was basically this:

1. `git merge --allow-unrelated-histories` (that is, pull in the entire codebase history)
2. Move files (crates under ~~`crates/orchestrator`~~ `orchestrator/crates/` and project files to a temp dir)
3. Update `Cargo.toml` to build (note that `orchestrator` is not a member of `default-members`, disagree if you want!)
4. Fight dependency conflicts
5. Esp. in this PR: introduce `orchestrator` CI

**Some follow up work:**
* #492
* #493
* #500 

### Repository forks :fork_and_knife: 

This required forking several repositories in order to avoid version conflicts. Each fork uses the same `madara-monorepo` branch name. _(Note that #500 was opened to track this.)_ These repos were forked:

* [cairo-vm](https://github.com/Moonsong-Labs/cairo-vm/tree/madara-monorepo)
* [sequencer](https://github.com/Moonsong-Labs/sequencer/tree/madara-monorepo)
* [zaun](https://github.com/notlesh/zaun/tree/madara-monorepo)
* [snos](https://github.com/keep-starknet-strange/snos/tree/madara-monorepo)

PR #499, opened 2025-02-20
refactor(repo): madara orchestrator merge - revert

notlesh:
Same as #480 but reverts to last known commit where CI was green (bc859529).


Trantorian1:
A bit confused: is this still just for CI testing or does this superceed #480 ?

notlesh:
> A bit confused: is this still just for CI testing or does this superceed #480 ?

This PR can go away (closing now). I had opened it to see if I could get CI green on #480 at an older commit. My concern at the time was some files cached by CI which might not work anymore, that doesn't appear to have been relevant.

PR #514, opened 2025-03-06
fix v0.13.4 gateway client deserialization

cchudant:
## Pull Request type

- Bugfix

serde deny_unknown_fields - the reason we put that is because it forced us to support any new fields when syncing from integration and sepolia testnets.

## What is the new behavior?

We don't support v0.13.4, and it seems the madara-alliance is not wanting to support it. The CI however tests the import of a few blocks on sepolia, and it is now failing because we have not implemented the new v0.13.4 changes.
This PR allows the tests to pass.

Note that however, the new v0.13.4 blocks probably still can't be imported.

No


HermanObst (approved):
lgtm!

notlesh (approved):

PR #524, opened 2025-03-08
Update Orchestrator to Latest Changes and Prepare for Future L3 Support 

heemankv:
## Description 
  
This PR brings the orchestrator in this repo **up to date** with its latest changes. Moving forward, all orchestrator-related PRs will be raised **directly in this repo**, allowing us to archive the `madara-orch` repo.  

Additionally, this update makes the orchestrator **more production-ready** by resolving a memory fabrication issue using `jammalloc`. We’ve also **refactored the logic** to better support future **L3 integrations**.  

##  PR Type  
- **Feature**  

## Other Information  
This update ensures a **more stable and scalable orchestrator**, setting the foundation for future improvements.


orchestrator/crates/orchestrator/src/cli/service.rs
```diff
@@ -3,10 +3,15 @@ use clap::Args;
 #[derive(Debug, Clone, Args)]
 pub struct ServiceCliArgs {
     /// The maximum block to process.
-    #[arg(env = "MADARA_ORCHESTRATOR_MAX_BLOCK_NO_TO_PROCESS", long)]
-    pub max_block_to_process: Option<String>,
+    /// The default value is u64::MAX.
+    #[arg(env = "MADARA_ORCHESTRATOR_MAX_BLOCK_NO_TO_PROCESS", long, default_value = Some("18446744073709551615"))]
```
notlesh: Is there a useful distinction between `None` and `Some(u64::MAX)`? If so, it should be documented

notlesh:
> This PR brings the orchestrator in this repo up to date with its latest changes.

For clarity, does this mean that it picks up from orchestrator PRs that were abandoned (because they were done while the monorepo was in progress)?

If so, it would be good to link to the relevant PRs.

Mohiiit:
Yes this includes all the changes that were done to the orch while monorepo was getting merged, links to the PR merged:

1. https://github.com/madara-alliance/madara-orchestrator/pull/198
2. https://github.com/madara-alliance/madara-orchestrator/pull/204
3. https://github.com/madara-alliance/madara-orchestrator/pull/205
4. https://github.com/madara-alliance/madara-orchestrator/pull/206
5. https://github.com/madara-alliance/madara-orchestrator/pull/207

cc: @notlesh

orchestrator/crates/orchestrator/src/cli/service.rs
Mohiiit: nope there isn't removed this

orchestrator/crates/orchestrator/src/cli/mod.rs
```diff
@@ -606,20 +606,9 @@ pub mod validate_params {
...
-            min_block_to_process: service_args.min_block_to_process.clone().and_then(|s| {
-                if s.is_empty() {
-                    None
-                } else {
-                    Some(s.parse::<u64>().expect("Failed to parse min block to process"))
-                }
-            }),
+            max_block_to_process: service_args.max_block_to_process,
+            min_block_to_process: service_args.min_block_to_process,
+            max_concurrent_snos_jobs: service_args.max_concurrent_snos_jobs,
```
raynaudoe, line 609: It's ok to have these params without any validation?

orchestrator/crates/orchestrator/src/config.rs
```diff
@@ -162,6 +166,10 @@ pub async fn init_config(run_cmd: &RunCmd) -> color_eyre::Result<Arc<Config>> {
     let queue_params = run_cmd.validate_queue_params().map_err(|e| eyre!("Failed to validate queue params: {e}"))?;
     let queue = build_queue_client(&queue_params, provider_config.clone()).await;
 
+    let snos_processing_lock =
+        JobProcessingState::new(orchestrator_params.service_config.max_concurrent_snos_jobs.unwrap_or(1));
```
raynaudoe, line 170: I'd suggest that the choosing of a default value for any param should be located in one place only and done only once. Maybe the `validate_params` fn could be where you default the params if they're not valid.

orchestrator/crates/orchestrator/src/cron/mod.rs
```diff
@@ -29,6 +32,7 @@ pub trait Cron {
     ) -> color_eyre::Result<()>;
     async fn setup(&self) -> color_eyre::Result<()> {
         let trigger_arns = self.create_cron().await?;
+        sleep(Duration::from_secs(15)).await;
```
raynaudoe, line 35: why this sleep is necessary?

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub semaphore: Semaphore,
+    pub active_jobs: Mutex<HashSet<Uuid>>,
+}
+impl JobProcessingState {
+    pub fn new(max_parallel_jobs: usize) -> Self {
+        JobProcessingState { semaphore: Semaphore::new(max_parallel_jobs), active_jobs: Mutex::new(HashSet::new()) }
+    }
+
+    pub async fn get_active_jobs(&self) -> HashSet<Uuid> {
+        self.active_jobs.lock().await.clone()
```
raynaudoe: nit: a read only lock might be a little more efficient

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+        &'a self,
+        job: &JobItem,
+        config: Arc<Config>,
+    ) -> Result<SemaphorePermit<'a>, JobError> {
+        // Trying to acquire permit with a timeout.
+        match tokio::time::timeout(Duration::from_millis(100), self.semaphore.acquire()).await {
+            Ok(Ok(permit)) => {
+                {
+                    let mut active_jobs = self.active_jobs.lock().await;
+                    active_jobs.insert(job.id);
```
raynaudoe: Maybe check here if `active_jobs` already has `job.id`

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -205,6 +246,13 @@ impl Job for DaJob {
     fn verification_polling_delay_seconds(&self) -> u64 {
         60
     }
+
+    fn job_processing_lock(
+        &self,
+        _config: Arc<Config>,
+    ) -> std::option::Option<std::sync::Arc<helpers::JobProcessingState>> {
+        None
+    }
```
raynaudoe, line 255: is this function complete?

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -477,13 +668,112 @@ pub async fn verify_job(id: Uuid, config: Arc<Config>) -> Result<(), JobError> {
...
+        tracing::error!(
+            job_id = ?id,
+            status = ?job.status,
+            "Cannot retry job: invalid status"
+        );
+        return Err(JobError::InvalidStatus { id, job_status: job.status });
+    }
+
+    // Increment the retry counter in common metadata
+    job.metadata.common.process_retry_attempt_no += 1;
```
raynaudoe, line 715: Shouldn't here check if `process_retry_attempt_no + 1 >= MAX_RETRY_ATTEMPTS` (don't know the name of that max value) and return if true?

orchestrator/crates/orchestrator/src/setup/mod.rs
```diff
@@ -62,7 +62,7 @@ pub async fn setup_cloud(setup_cmd: &SetupCmd) -> color_eyre::Result<()> {
     // Cron
     println!("Setting up cron. ⏳");
     // Sleeping for few seconds to let AWS index the newly created queues to be used for setting up cron
-    sleep(Duration::from_secs(100)).await;
+    sleep(Duration::from_secs(60)).await;
```
raynaudoe, line 65: I wonder if there's a way to query aws and ask the queues status. If that is so, this could be dynamic instead of sleeping for some time

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -252,6 +248,10 @@ impl TestConfigBuilder {
         // Creating the SNS ARN
         create_sns_arn(provider_config.clone(), &params.alert_params).await.expect("Unable to create the sns arn");
 
+        let snos_processing_lock =
+            JobProcessingState::new(params.orchestrator_params.service_config.max_concurrent_snos_jobs.unwrap_or(1));
```
raynaudoe, line 252: Same as a few comments above, would be better to have in one place the default value for this param.

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -544,7 +545,13 @@ fn get_env_params() -> EnvParams {
     let env = get_env_var_optional("MADARA_ORCHESTRATOR_MIN_BLOCK_NO_TO_PROCESS").expect("Couldn't get min block");
     let min_block: Option<u64> = env.and_then(|s| if s.is_empty() { None } else { Some(s.parse::<u64>().unwrap()) });
 
-    let service_config = ServiceParams { max_block_to_process: max_block, min_block_to_process: min_block };
+    let env = get_env_var_optional("MADARA_ORCHESTRATOR_MAX_CONCURRENT_SNOS_JOBS")
```
raynaudoe, line 548: shouldn't be using `orchestrator_params.service_config.max_concurrent_snos_jobs` instead ?

orchestrator/crates/orchestrator/src/workers/update_state.rs
```diff
@@ -96,26 +97,72 @@ impl Worker for UpdateStateWorker {
...
                 }
             }
         }
 
-        let mut blocks_to_process: Vec<u64> = find_successive_blocks_in_vector(blocks_to_process);
-
+        let mut blocks_to_process = find_successive_blocks_in_vector(blocks_to_process);
         if blocks_to_process.len() > 10 {
             blocks_to_process = blocks_to_process.into_iter().take(10).collect();
         }
```
raynaudoe, line 111: Just curious, what's the reasoning behind not taking more than 10 blocks at a time?

orchestrator/crates/orchestrator/src/workers/update_state.rs
```diff
@@ -96,26 +97,72 @@ impl Worker for UpdateStateWorker {
...
+                if blocks_to_process[0] != min_block_to_process {
                     log::warn!("DA job for the first block is not yet completed. Returning safely...");
                     return Ok(());
                 }
             }
         }
 
-        let mut blocks_to_process: Vec<u64> = find_successive_blocks_in_vector(blocks_to_process);
-
+        let mut blocks_to_process = find_successive_blocks_in_vector(blocks_to_process);
```
raynaudoe, line 108: What if this vector cannot find any successive blocks or there are dangling blocks that don't have their corresponding neighbours?  Those blocks would eventually get processed, right?

raynaudoe (commented):
uACK, left a few comments

orchestrator/crates/orchestrator/src/cli/mod.rs
heemankv: Yes, this works because we have added their type in clap directly, so clap automatically does the type validation. For rest of the values we have used string string input since they are more complicated.

orchestrator/crates/orchestrator/src/config.rs
heemankv: Yes, this is part of a bigger issue where in orchestrator we currently can't provide types in clap and use them wisely in the tests config as well, this [issue is mentioned here ](https://github.com/madara-alliance/madara/issues/543)and will be resolved in other PRs, as this PR is just an amalgamation of Orchestrator repo prs.

orchestrator/crates/orchestrator/src/cron/mod.rs
heemankv: While the setup runs to configure AWS resources, we have noticed that it takes some time for AWS to index the worker trigger policies created, so we have to delay the creation of the triggers a bit to let AWS index the policies, to be fetched while creating the triggers.

orchestrator/crates/orchestrator/src/helpers.rs
heemankv: would like to hear more on this, can you provide more context / we can also hop on a meet if you'd like

orchestrator/crates/orchestrator/src/helpers.rs
heemankv: makes sense, let's discuss this with your idea of a read only lock,

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
heemankv: Yes, we don't have any job_processing_lock for DA / Proving / State Update jobs right now, but we have added this fn in the job trait hence it needs to be adapted by all the Jobs

orchestrator/crates/orchestrator/src/jobs/mod.rs
heemankv: As far as I know, there's no config like : MAX_RETRY_ATTEMPTS, but we can add one tho, here's an issue to track this : https://github.com/madara-alliance/madara/issues/543

orchestrator/crates/orchestrator/src/setup/mod.rs
heemankv: https://github.com/madara-alliance/madara/issues/544

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: https://github.com/madara-alliance/madara/issues/543

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: https://github.com/madara-alliance/madara/issues/543

orchestrator/crates/orchestrator/src/workers/update_state.rs
heemankv: It's to do multiple in one state update together, besides that, I don’t think that number is fine tuned . just randomly selected to get started. We can later move it to a config setting, but I would not make it a priority

orchestrator/crates/orchestrator/src/cli/mod.rs
raynaudoe: Ok, makes sense, maybe the value of `max_concurrent_snos_jobs` could be a little more protected, not sure what happens if you put a `0` in it or if the user tries to set that to a value greater than `num_cpus`

orchestrator/crates/orchestrator/src/cron/mod.rs
raynaudoe: Ok 👍🏻 . Maybe this can be improved in a following PR to avoid sleeps

raynaudoe:
@heemankv can you please merge main so the branch has the latest fixes for CI ?

orchestrator/crates/orchestrator/src/jobs/snos_job/mod.rs
```diff
@@ -147,64 +189,72 @@ impl Job for SnosJob {
...
-            .parse()
-            .map_err(|_| SnosError::InvalidBlockNumber {
-                internal_id: job.internal_id.clone(),
-                block_number: job.metadata[JOB_METADATA_SNOS_BLOCK].clone(),
-            })?;
 
-        Ok(block_number)
+    fn job_processing_lock(&self, config: Arc<Config>) -> std::option::Option<Arc<helpers::JobProcessingState>> {
+        Some(config.processing_locks().snos_job_processing_lock.clone())
     }
```
notlesh, line 195: This could use some documentation. The return type and/or fn name are a bit vague.

orchestrator/crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -305,26 +390,20 @@ impl StateUpdateJob {
...
 
     /// Update the state for the corresponding block using the settlement layer.
+    #[allow(clippy::too_many_arguments)]
     async fn update_state_for_block(
         &self,
         config: Arc<Config>,
         block_no: u64,
         snos: StarknetOsOutput,
         nonce: u64,
+        program_output: Vec<[u8; 32]>,
```
notlesh, line 400: Why not use a `Felt` as the type here?

orchestrator/crates/orchestrator/src/tests/jobs/da_job/mod.rs
```diff
@@ -126,6 +132,17 @@ async fn test_da_job_process_job_failure_on_pending_block() {
         then.status(200).body(serde_json::to_vec(&response).unwrap());
     });
 
+    // Create proper metadata structure
+    let block_number = internal_id.parse::<u64>().unwrap_or(0);
```
notlesh: It's not clear why this would fail (and trigger `unwrap_or(0)`). If it's an expected condition, maybe it should have some documentation/helper fn, and if it's not then I would fail hard instead of silently converting to 0.

notlesh: Maybe this is related to handling the genesis block? If so, I would leave a log message so it would be more obvious if this happens accidentally.

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -485,45 +524,47 @@ async fn verify_job_with_rejected_status_adds_to_queue_works() {
...
 
     let database_client = services.config.database();
-    let mut job_handler = MockJob::new();
 
-    // creating job in database
+    // Create a job with proper metadata structure
+    let mut job_item = build_job_item(JobType::DataSubmission, JobStatus::PendingVerification, 1);
+
+    // Set process_attempt_no to 1 to simulate max attempts reached
+    job_item.metadata.common.process_attempt_no = 1;
```
notlesh, line 540: Is there a constant or some value you can query instead of hard coding this to 1?

notlesh: ...Or one you could define here in the code

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -739,14 +769,14 @@ async fn handle_job_failure_job_status_completed_works(#[case] job_type: JobType
     let database_client = services.config.database();
     let internal_id = 1;
 
-    // create a job
+    // Create a job
```
notlesh, line 772: These kind of changes create a lot of noise (and massive headaches / merge conflicts when merging code), are you doing this manually or does your editor do it for you? :sweat_smile:

orchestrator/crates/orchestrator/src/tests/jobs/snos_job/mod.rs
```diff
@@ -39,6 +51,17 @@ async fn test_create_job() {
 #[tokio::test]
 async fn test_verify_job(#[from(default_job_item)] mut job_item: JobItem) {
     let services = TestConfigBuilder::new().build().await;
+
+    // Update job_item to use the proper metadata structure for SNOS jobs
+    job_item.metadata.specific = JobSpecificMetadata::Snos(SnosMetadata {
+        block_number: 0,
```
notlesh, line 57: All of these tests so far have used block 0, which is a very special case -- it might be good to test with other cases

orchestrator/crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -327,9 +391,30 @@ async fn process_job_invalid_input_gap_panics() {
...
-    let mut metadata: HashMap<String, String> = HashMap::new();
-    metadata.insert(String::from(JOB_METADATA_STATE_UPDATE_BLOCKS_TO_SETTLE_KEY), String::from("6, 7, 8"));
-    metadata.insert(String::from(JOB_PROCESS_ATTEMPT_METADATA_KEY), String::from("0"));
+    // Create proper metadata structure with valid paths
+    let metadata = JobMetadata {
+        common: CommonMetadata { process_attempt_no: 0, ..CommonMetadata::default() },
+        specific: JobSpecificMetadata::StateUpdate(StateUpdateMetadata {
+            blocks_to_settle: vec![6, 7, 8], // Gap between 4 and 6
+            snos_output_paths: vec![
+                format!("{}/{}", 6, SNOS_OUTPUT_FILE_NAME),
```
notlesh, line 400: This pattern (`"{}/{}"`) is repeated a lot in this PR, it could be really helpful for readability/documentation/DRY to create a util fn for it

orchestrator/crates/orchestrator/src/tests/utils.rs
```diff
@@ -0,0 +1,69 @@
...
+        },
+        JobType::DataSubmission => JobMetadata {
+            common: CommonMetadata::default(),
+            specific: JobSpecificMetadata::Da(DaMetadata {
+                block_number: internal_id,
+                blob_data_path: Some(format!("{}/{}", internal_id, BLOB_DATA_FILE_NAME)),
+                tx_hash: None,
+            }),
+        },
+        _ => panic!("Invalid job type"),
```
notlesh, line 55: I'd leave this out so the compiler can help

orchestrator/crates/orchestrator/src/tests/workers/proving/mod.rs
```diff
@@ -34,55 +34,62 @@ async fn test_proving_worker(#[case] incomplete_runs: bool) -> Result<(), Box<dy
...
+        .withf(|job_type, job_status, successor_type| {
+            *job_type == JobType::SnosRun
+                && *job_status == JobStatus::Completed
+                && *successor_type == JobType::ProofCreation
+        })
+        .returning(move |_, _, _| Ok(snos_jobs.clone()));
+
+    // Set up expectations for each job
+    for i in 1..=num_jobs {
+        if incomplete_runs && i == 3 {
```
notlesh: needs an explanation

notlesh: or a constant, e.g. `const MAGIC_JOB_NUM: usize = 3;`

notlesh (changes requested):
Looks good for the most part, I left a lot of comments but I don't have too much context so not all of it is relevant.

raynaudoe:
Hi @heemankv, can you please check the failing tests?

e2e-tests/tests.rs
```diff
@@ -433,16 +467,32 @@ pub async fn mock_starknet_get_latest_block(starknet_client: &mut StarknetClient
 
 /// Puts after SNOS job state into the database
 pub async fn put_job_data_in_db_update_state(mongo_db: &MongoDbServer, l2_block_number: String) {
+    let block_number = l2_block_number.parse::<u64>().unwrap() - 1;
+
+    // Create the StateUpdate-specific metadata
+    let state_update_metadata = StateUpdateMetadata {
+        blocks_to_settle: vec![block_number],
```
HermanObst, line 474: For now this can only be 1 block, right?
I guess this is in preparation for Applicative recursion?

e2e-tests/tests.rs
```diff
@@ -357,13 +378,26 @@ pub async fn mock_proving_job_endpoint_output(sharp_client: &mut SharpClient) {
 
 /// Puts after SNOS job state into the database
 pub async fn put_job_data_in_db_da(mongo_db: &MongoDbServer, l2_block_number: String) {
+    // Create the DA-specific metadata
+    let da_metadata = DaMetadata {
+        block_number: l2_block_number.parse::<u64>().unwrap() - 1,
```
HermanObst, line 383: Why is this -1?

HermanObst: Also, it seems to be the same as the internal_id

orchestrator/crates/orchestrator/src/cli/mod.rs
heemankv: IMO keeping 0 is a valid way to let the orch know to not process any jobs .
Do you think we should follow separation of concern for that ?

orchestrator/crates/orchestrator/src/config.rs
heemankv: The PR #529 will hopefully resolve these concerns.

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::queue::job_queue::add_job_to_process_queue;
+
+pub struct ProcessingLocks {
+    pub snos_job_processing_lock: Arc<JobProcessingState>,
+}
+
+pub struct JobProcessingState {
+    pub semaphore: Semaphore,
+    pub active_jobs: Mutex<HashSet<Uuid>>,
+}
```
HermanObst: @heemankv do we need a Mutex to protect `HashSet<Uuid>`? 
This will enforce that only one thread can write or read. 

Maybe you just need `RwLock` that allows many threads to read the data (but only one can write)

orchestrator/crates/orchestrator/src/helpers.rs
HermanObst: I was trying to understand the use case (to understand if we really want only one reader), but seems that this function is never call

orchestrator/crates/orchestrator/src/jobs/snos_job/mod.rs
heemankv: I've added some description to the function in PR #559.

orchestrator/crates/orchestrator/src/tests/jobs/da_job/mod.rs
heemankv: You are right, there's no benefit of using `unwrap_or` here, have defaulted it to `unwrap`.
And no this is not related to handling the genesis block.

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
heemankv: I think this is just for testing purposes, 
we are putting in 1 here to overcome the value defined here : https://github.com/madara-alliance/madara/blob/041b6f1fbf18f0860987da2c99ddac0b3dee3ee6/orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs#L197-L199

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Haha, I understand, it's the editor doing it by itself, IMO it's correct only, I don't see any harm in it.
lemme know if there are many like these

orchestrator/crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
heemankv: Created an issue for this here :  https://github.com/madara-alliance/madara/issues/568

orchestrator/crates/orchestrator/src/tests/utils.rs
heemankv: Since `ProofRegistration` is added but yet not handled we have to keep this statement, It shall be removed after L3 support for orchestrator gets merge.

orchestrator/crates/orchestrator/src/tests/workers/proving/mod.rs
heemankv: So this is a test case that's trying to generate a random scenario and it's based off the test writer that they chose the Job_number 3 to be still incomplete, 
IMO this is scoped to this test case only, creating a constant for this seems overkill, 
WDYT ?

orchestrator/crates/orchestrator/src/tests/jobs/snos_job/mod.rs
heemankv: IMO that shouldn't matter for these tests, since its just verifying against mock DB server if it's able to interact with stuff or not.

orchestrator/crates/orchestrator/src/jobs/state_update_job/mod.rs
heemankv: This is because as of now this code-piece serves both Ethereum and Starknet,
IMO it would be very type intensive to change it just for Starknet.
Let me see what I can do

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
notlesh: Yeah, the entire PR was littered with changes like that.

orchestrator/crates/orchestrator/src/tests/workers/proving/mod.rs
notlesh: A simple comment would work, too. With no context at all, though, it becomes a maintenance issue -- the next person who comes along and needs to maintain this code is going to have a hard time knowing what `3` means and how to handle it.

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::jobs::types::JobItem;
+use crate::jobs::JobError;
+use crate::queue::job_queue::add_job_to_process_queue;
+
+pub struct ProcessingLocks {
+    pub snos_job_processing_lock: Arc<JobProcessingState>,
+}
+
+pub struct JobProcessingState {
+    pub semaphore: Semaphore,
```
HermanObst, line 16: What is the core reason we want a Semaphore?

HermanObst: Sorry, let me rephrase so we dont end up in the [XY communication problem](https://en.wikipedia.org/wiki/XY_problem) 

Which is the problem you faced, which later made you think of limiting concurrency?

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.semaphore.available_permits()
+    }
+
+    pub async fn try_acquire_lock<'a>(
+        &'a self,
+        job: &JobItem,
+        config: Arc<Config>,
+    ) -> Result<SemaphorePermit<'a>, JobError> {
+        // Trying to acquire permit with a timeout.
+        match tokio::time::timeout(Duration::from_millis(100), self.semaphore.acquire()).await {
```
HermanObst, line 33: Why the timeout is needed?

orchestrator/crates/orchestrator/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    pub async fn try_release_lock<'a>(&'a self, permit: SemaphorePermit<'a>, job_id: &Uuid) -> Result<(), JobError> {
+        let mut active_jobs = self.active_jobs.lock().await;
+        active_jobs.remove(job_id);
+        drop(active_jobs); // Explicitly drop the lock (optional but clear)
+        drop(permit); // Explicitly drop the permit (optional but clear)
+        Ok(())
+    }
+}
```
HermanObst, line 51: @heemankv I would really like seeing test that check good behavior:
- Happy path
- A bunch of treads trying to acquire this lock (and after reaching availability, the system continue working fine)
- Etc

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
HermanObst: @heemankv maybe im not seeing something, but if there is a function that only makes sense for 1 implementation of a trait, then it shouldnt be in the trait. It should be a specific function of the Struct Impl. 

wdyt?

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -344,14 +392,16 @@ pub async fn state_update_to_blob_data(
 }
 
 /// To store the blob data using the storage client with path <block_number>/blob_data.txt
```
HermanObst: Is this still true?

orchestrator/crates/orchestrator/src/jobs/metadata/common.rs
```diff
@@ -0,0 +1,38 @@
...
+    #[serde(with = "chrono::serde::ts_seconds_option")]
+    pub process_completed_at: Option<DateTime<Utc>>,
+    /// Timestamp when job verification started
+    #[serde(with = "chrono::serde::ts_seconds_option")]
+    pub verification_started_at: Option<DateTime<Utc>>,
+    /// Timestamp when job verification completed
+    #[serde(with = "chrono::serde::ts_seconds_option")]
+    pub verification_completed_at: Option<DateTime<Utc>>,
+    /// Reason for job failure if any
+    pub failure_reason: Option<String>,
```
HermanObst, line 37: Do we know the possible errors?
If we do, it always makes sense to include those types in the type system, rather than a concatenation of chars

orchestrator/crates/orchestrator/src/jobs/metadata/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+    };
+}
+
+// Implement TryInto for all metadata types
+impl_try_into_metadata!(Snos, SnosMetadata);
+impl_try_into_metadata!(Proving, ProvingMetadata);
+impl_try_into_metadata!(Da, DaMetadata);
+impl_try_into_metadata!(StateUpdate, StateUpdateMetadata);
```
HermanObst, line 60: I think implementing the try into for every type its much more clear, rather than a macro. Still its just my opinion.

HermanObst: Anyway, could we add unit test on this to showcase usage (and ensure correctness)?

orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs
```diff
@@ -0,0 +1,31 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Input type specification for proving jobs.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ProvingInputType {
+    /// Path to an existing proof
+    Proof(String),
+    /// Path to a Cairo PIE file
+    CairoPie(String),
+}
```
HermanObst: This should be named as `ProvingInputTypePath`, right? as this struct is a path and the type is still not created 
(Actually, we are "trusting" this path points to some data that then can be converted in a meaningful type)

orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Metadata specific to proving jobs.
+///
+/// # Field Management
+/// All fields are initialized by the worker during job creation.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ProvingMetadata {
+    /// Block number to prove
+    pub block_number: u64,
+    /// Path to the input file (proof or Cairo PIE)
+    pub input_path: Option<ProvingInputType>,
```
HermanObst: How this relates to `ProvingInputType`?

Could you add some docs explaining it? (e.g: when is a proof, when a Cairo PIE, do they change in the job lifetime?, when can be None, etc)

HermanObst (commented):
@heemankv  Some preliminary comments, Ill continue with the rest of the PR :)

e2e-tests/tests.rs
notlesh: I would make sure this can't underflow (do a `checked_sub()` or `saturating_sub()`).

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
Mohiiit: not as of now, paths are provided by the worker itself, so ideally it shouldn't be here now

orchestrator/crates/orchestrator/src/jobs/metadata/common.rs
Mohiiit: I agree, and in the refactor we have included error types, it would be part of the refactor effort

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -109,65 +132,127 @@ impl From<String> for OtherError {
...
     /// Should build a new job item and return it
+    ///
+    /// # Arguments
+    /// * `config` - Shared configuration for the job
+    /// * `internal_id` - Unique identifier for internal tracking
+    /// * `metadata` - Additional key-value pairs associated with the job
+    ///
+    /// # Returns
+    /// * `Result<JobItem, JobError>` - The created job item or an error
     async fn create_job(
```
HermanObst, line 162: I got confused with the `create_job` function that actually creates the job in db, etc.
Maybe this should be named: `create_job_item`? And add in docs that the different impls of this function are called by `create_job` depending the job_type we want to create?

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -109,65 +132,127 @@ impl From<String> for OtherError {
...
 pub async fn create_job(
     job_type: JobType,
     internal_id: String,
-    metadata: HashMap<String, String>,
+    metadata: JobMetadata,
     config: Arc<Config>,
 ) -> Result<(), JobError> {
     let start = Instant::now();
-    tracing::info!(log_type = "starting", category = "general", function_type = "create_job", job_type = ?job_type, block_no = %internal_id, "General create job started for block");
+    tracing::info!(
```
HermanObst, line 234: Do we need all this as info? 
I think that Orchestrator has the info level logs flooded. 
Maybe in a subsequent refactor we can think on being more granular between info, debug and trace 

wdyt?

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -194,50 +286,96 @@ pub async fn create_job(
...
 /// DB. It then adds the job to the verification queue.
+///
+/// # Arguments
+/// * `id` - UUID of the job to process
+/// * `config` - Shared configuration
+///
+/// # Returns
+/// * `Result<(), JobError>` - Success or an error
+///
+/// # State Transitions
```
HermanObst, line 297:
```suggestion
/// # State Transitions depending Job state while entering this function
```

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -317,6 +470,28 @@ pub async fn process_job(id: Uuid, config: Arc<Config>) -> Result<(), JobError>
...
 /// the job back to the queue.
+///
+/// # Arguments
+/// * `id` - UUID of the job to verify
+/// * `config` - Shared configuration
+///
+/// # Returns
+/// * `Result<(), JobError>` - Success or an error
+///
+/// # State Transitions
```
HermanObst, line 481:
```suggestion
/// # State Transitions depending Job state while entering this function
```

orchestrator/crates/orchestrator/src/jobs/mod.rs
HermanObst: Or something like that xD

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -194,50 +286,96 @@ pub async fn create_job(
...
         }
         _ => {
-            tracing::warn!(job_id = ?id, status = ?job.status, "Job status is Invalid. Cannot process.");
+            tracing::warn!(job_id = ?id, status = ?job.status, "Cannot process job with current status");
             return Err(JobError::InvalidStatus { id, job_status: job.status });
         }
     }
+
+    let job_handler = factory::get_job_handler(&job.job_type).await;
+    let job_processing_locks = job_handler.job_processing_lock(config.clone());
```
HermanObst, line 344: This will be always `None` except if the given job is SNOS, right?

If this is the case (not other Types need this `job_processing_lock`) think it should make sense to:
- Not include `job_processing_lock` in the Job trait (same comment I leave before), basically because in not a generic behavior of a Job, but only of a SNOSJob. 
- Here be explicit in the variable naming: `snos_job_processing_locks` (under a if, depending the job type, or something similar)

cc @notlesh @raynaudoe @heemankv @Mohiiit 

WDYT?

orchestrator/crates/orchestrator/src/jobs/metadata/common.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+pub struct CommonMetadata {
+    /// Number of times the job has been processed
+    pub process_attempt_no: u64,
+    /// Number of times the job has been retried after processing failures
+    pub process_retry_attempt_no: u64,
+    /// Number of times the job has been verified
+    pub verification_attempt_no: u64,
+    /// Number of times the job has been retried after verification failures
+    pub verification_retry_attempt_no: u64,
```
HermanObst: Just being annoying here: u64 do not make sense for this :P 
I would say `u16` is more than sufficient

orchestrator/crates/orchestrator/src/jobs/mod.rs
```diff
@@ -477,13 +668,112 @@ pub async fn verify_job(id: Uuid, config: Arc<Config>) -> Result<(), JobError> {
...
+            job_id = ?id,
+            status = ?job.status,
+            "Cannot retry job: invalid status"
+        );
+        return Err(JobError::InvalidStatus { id, job_status: job.status });
+    }
+
+    // Increment the retry counter in common metadata
+    job.metadata.common.process_retry_attempt_no += 1;
+    job.metadata.common.process_attempt_no = 0;
```
HermanObst, line 717: Why this is 0?

orchestrator/crates/orchestrator/src/jobs/metadata/common.rs
HermanObst, line 37: Maybe not for this PR, but maybe it makes sense for this to be a `Option<Vec<String>>`?
I dont know if its relevant for us, but basically that way we wont loose the failure reasons and would have a log of them. 

cc @Mohiiit @heemankv

HermanObst (commented):
second run

orchestrator/crates/orchestrator/src/helpers.rs
heemankv: Yeah, you are right, this was being used in a more complex implementation earlier, I guess we missed removing it, removing it now.
Great catch!

orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs
heemankv: Your thoughts are correct when we consider it being used only for 1 job.
But if you checkout #559 we have already integrated it for `proving job` based on the errors we faced while pushing 200+ jobs to Atlantic at once.
I would suggest keeping it as a part of the struct given we'll mp be adding it for DA and SU as well.

orchestrator/crates/orchestrator/src/jobs/state_update_job/mod.rs
heemankv: Is there any specific advantage that we achieve when we use the type `Felt` here (which we can only use for starknet) or is it to ensure type consistency ? 
I think we can create an issue for this and handle this later if this is for type consistency, since this isn't a part of this PR's scope and would be a really `Good First Issue`.
Wdyt ?

orchestrator/crates/orchestrator/src/tests/jobs/da_job/mod.rs
heemankv: Removed it now

orchestrator/crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Alright, I understand, 
We'll make a note of it for sure, 
Do you believe these changes should be reverted ?

orchestrator/crates/orchestrator/src/tests/workers/proving/mod.rs
heemankv: Noted, added.
```
    let num_jobs = 5;
    // Choosing a random incomplete job ID out of the total number of jobs
    let random_incomplete_job_id: u64 = 3;
 ```

File: e2e-tests/tests.rs (hunk 23 above)
Comment (heemankv) [reply e2e-tests/tests.rs:474]: This is a helper function that adds the `block -1` state update job to the db, where `block` is the one being tested in e2e. This is to ensure that the job is created by the orch's state update logic.

File: e2e-tests/tests.rs (hunk 24 above)
Comment (heemankv) [reply e2e-tests/tests.rs:383]: This is a helper function that adds the block -1 DA job to the db, where block is the one being tested in e2e. This is to brute force way to ensure that the job is created by the orch's DA logic.

File: e2e-tests/tests.rs (hunk 23 above)
Comment (heemankv) [reply e2e-tests/tests.rs:474]: > In preparation for Applicative recursion 

No this is not for AR, rather we decided to do 10 (arbitrary number) block's state update together just so save up on the process of sending a job to queue and waiting for it to be picked, Since there's no way for us to do state update in parallel (due to nonce issue), we manually manage the nonce while sending upto 10 SU together.

File: orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs (hunk 29 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/da_job/mod.rs]: Removing this comment since now paths are provided by the worker itself.

File: orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs (hunk 32 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs]: Valid, I've changed it to `ProvingInputTypePath`

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 35 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:234]: Yes, we are to work on a better tracing usage, will create an issue, unsure as to when it will be picked for now.

Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:234]: Tracking issue : #574

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 36 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:297]: I am unsure about what you are looking for here, would appreciate elaboration.
Thanks!

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 37 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:481]: Same as above.
> I am unsure about what you are looking for here, would appreciate elaboration.
Thanks!

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 38 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:344]: Same as mentioned [here](https://github.com/madara-alliance/madara/pull/524#discussion_r2039066017).


> Here be explicit in the variable naming: snos_job_processing_locks (under a if, depending the job type, or something similar)
- We tried this in earlier implementations of processing locks, but it wasn't sitting well with the fact that this `process_job` is a generalised function and it shouldn't handle a specific job's logic. 
- Secondly, It's not recommended to handle the processing locks inside each job's `process_job` fn since the generalised `process_job` already marks the job as `LockedForProcessing`.

File: orchestrator/crates/orchestrator/src/jobs/metadata/common.rs (hunk 39 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/metadata/common.rs]: Valid argument, changed.

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 40 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:717]: The current flow is if a job is retried we reset it's process_attempt_number to track the new attempts,
It's a dual counter flow, 
`process_retry_attempt_no` counts the number of time a job is tried.
`process_attempt_no` counts the number of times a job is processed within a try.

Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:717]: Added a comment documenting this.

File: orchestrator/crates/orchestrator/src/jobs/metadata/common.rs (hunk 30 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/metadata/common.rs:37]: This is a valid argument, adding an issue to track this : #575

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 34 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:162]: I believe we can change the db function from `create_job` to `create_job_item`.
Sounds fine ? WDYT ?

File: orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs (hunk 33 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/jobs/metadata/proving.rs]: I believe this impl is a little ahead of itself, `ProvingInputTypePath` is to be heavily used in `L3 Support`, given we'll send the proofs instead of the pie for the bridge proofs while sending Cairo-Pie for the first proof. 

Although I would like to follow `YAGNI` but here given it's already written and we know we are going to use it, we might as well keep it, TBH anything works for me, wdyt ?

File: orchestrator/crates/orchestrator/src/helpers.rs (hunk 26 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/helpers.rs:16]: We noticed that on a heavy load (multiple blocks to process) after all the jobs are created.
- Orchestrator picks up all the jobs concurrently, given each jobs takes cpu and memory to process, with limited resources the orchestrator hangs in processing any of them, thus reducing throughput to 0.
- This is noticed in SnosRun (prove_block) and ProvingJob (downloading lots of cairopie-s)

Hence limiting the concurrency using locks.

File: orchestrator/crates/orchestrator/src/helpers.rs (hunk 25 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/helpers.rs]: Update: in the latest code, we've removed `active_jobs` since it was not being used.

File: orchestrator/crates/orchestrator/src/helpers.rs (hunk 27 above)
Comment (heemankv) [reply orchestrator/crates/orchestrator/src/helpers.rs:33]: Great question, 
Context: Multiple orchestrators can run in parallel to pick jobs.
So the timeout ensures that the job is not waiting a single instance of the orchestrator. 
If a job can't find permit in one instance of the orchestrator, it is sent back for retry so that some other orchestrator can pick it up

File: orchestrator/crates/orchestrator/src/jobs/state_update_job/mod.rs (hunk 15 above)
Comment (notlesh) [reply orchestrator/crates/orchestrator/src/jobs/state_update_job/mod.rs:400]: If it's used for multiple types, then `Felt` doesn't make sense. Seems fine for now, a `Good First Issue` for someone with some inspiration would be good :)

File: orchestrator/crates/orchestrator/src/tests/jobs/mod.rs (hunk 18 above)
Comment (notlesh) [reply orchestrator/crates/orchestrator/src/tests/jobs/mod.rs:772]: It's probably fine unless others feel strongly about it. If you try to remove it, git's `-p` switch can be useful (e.g. `git checkout -p` or `git add -p` when preparing a commit).

File: orchestrator/crates/orchestrator/src/jobs/mod.rs (hunk 34 above)
Comment (HermanObst) [reply orchestrator/crates/orchestrator/src/jobs/mod.rs:162]: Agreed!

File: orchestrator/crates/orchestrator/src/tests/jobs/state_update_job/mod.rs [hunk 41]
```diff
@@ -259,24 +291,39 @@ async fn process_job_works_unit_test() {
...
 
 #[rstest]
-#[case(String::from("651052, 651054, 651051, 651056"), "numbers aren't sorted in increasing order")]
-#[case(String::from("651052, 651052, 651052, 651052"), "Duplicated block numbers")]
-#[case(String::from("a, 651054, b, 651056"), "settle list is not correctly formatted")]
-#[case(String::from("651052, 651052, 651053, 651053"), "Duplicated block numbers")]
-#[case(String::from(""), "settle list is not correctly formatted")]
+#[case(vec![651052, 651054, 651051, 651056], "numbers aren't sorted in increasing order")]
+#[case(vec![651052, 651052, 651052, 651052], "Duplicated block numbers")]
+#[case(vec![651052, 651052, 651053, 651053], "Duplicated block numbers")]
```
HermanObst, line 324: @heemankv so if the block numbers are not ordered the system fails?
Is a way to recover?

@raynaudoe

orchestrator/crates/orchestrator/src/workers/proving.rs
```diff
@@ -25,12 +28,43 @@ impl Worker for ProvingWorker {
...
+
+            // Create proving job metadata
+            let proving_metadata = JobMetadata {
+                common: CommonMetadata::default(),
+                specific: JobSpecificMetadata::Proving(ProvingMetadata {
+                    block_number: snos_metadata.block_number,
+                    // Set input path as CairoPie type
+                    input_path: snos_metadata.cairo_pie_path.map(ProvingInputTypePath::CairoPie),
+                    // Set download path if needed
+                    download_proof: None,
```
HermanObst, line 55: This will be always None?
@heemankv

orchestrator/crates/orchestrator/src/workers/update_state.rs
```diff
@@ -30,30 +31,31 @@ impl Worker for UpdateStateWorker {
...
                             JobStatus::Completed,
-                            last_block_processed_in_last_job.to_string(),
+                            last_block_processed.to_string(),
                         )
                         .await?,
-                    Some(last_block_processed_in_last_job),
+                    Some(last_block_processed),
                 )
             }
             None => {
```
HermanObst, line 61: @heemankv how can this happen?

HermanObst (approved):
As a general comment I think the current implementation has scattered state transition logic, making it hard to reason about the full lifecycle. We should consider refactoring into a centralized state machine to improve clarity and testability.

I left some more comments, please answer :) 

Moving from that, this changes LGTM

notlesh (approved):

orchestrator/crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
heemankv: Yeah, 
No there's no way to recover from it as of now, since we expect the values to be sorted, as they are processed in that fashion

orchestrator/crates/orchestrator/src/workers/update_state.rs
heemankv: Would need more context on what you are asking. 😅

orchestrator/crates/orchestrator/src/workers/proving.rs
heemankv: Seems like this change is also ahead of itself, it is being heavily used in the L3 PR we'll be raising soon.

PR #526, opened 2025-03-10
fix: syncing endpoint fixed as per spec

### LINKED ISSUE #443: feat(rpc): Sync endpoint

### Motivation

A dedicated syncing status endpoint is essential for production deployments for several reasons:

*Health Monitoring*

Provides real-time status of node synchronization
Enables monitoring systems to detect out-of-sync nodes quickly
Helps maintain overall system reliability

*Load Balancing & Orchestration*

Allows Kubernetes to make intelligent routing decisions
Facilitates automatic traffic routing to fully synced nodes
Supports graceful handling of nodes that fall behind

*Production Readiness*

While not critical for local development
Essential for production environments where uptime and data consistency are crucial
Enables automated failover mechanisms

### Request

When running Madara in full node mode, we need an endpoint that indicates whether the node is currently syncing or not. Similar functionality exists in Juno, where this signal is used for orchestration logic in Kubernetes environments.

### Solution

Add a new endpoint that returns the node's current syncing status, similar to Juno's implementation. This should be integrated with Madara's existing API infrastructure.

Yes!

Mohiiit:
## Pull Request type

- Bugfix

syncing rpc endpoint was not as per the spec

Resolves: #443 

## What is the new behavior?

- now it's as per the spec

No

## Other Information

you can find the spec here: https://github.com/starkware-libs/starknet-specs/blob/2030a650be4e40cfa34d5051a0334f375384a421/api/starknet_api_openrpc.json#L1088C11-L1097C12


jbcaron:
I don't think it's a good idea for an RPC read method to depend on an external HTTP request. Moreover, the node risks being rate-limited on the gateway. It's also important to consider the processing time of a new block, meaning that even when synchronized, the node might be one block behind. I believe it would be better to implement this logic in the sync service and store it in the database rather than on the RPC side.
wdyt?

Mohiiit:
Pathfinder maintains a buffer of 6 blocks, meaning if the current synced node + 6 >= highest block, they respond with `false`, else they respond with `status`.  

https://github.com/eqlabs/pathfinder/blob/da3ae6d3b1f87d78b162d043300334ffb9ef8a23/crates/rpc/src/method/syncing.rs#L8  

There are two ways to go about moving it to the `sync` crate:  
1. **In-Memory Tracking** – Maintain the synced state in memory using `Arc<RwLock<SyncState>>`, reducing DB queries while keeping it lightweight. (or Mutex with relaxed ordering maybe)
2. **Database Persistence** – Store the latest synced block height in the DB and read from it when needed for resilience across restarts.  

I don't have a strong opinions about this, wdyt?

jbcaron:
I prefer in-memory tracking without persistence across restarts

madara/crates/client/rpc/src/errors.rs
```diff
@@ -142,6 +144,7 @@ impl From<&StarknetRpcApiError> for i32 {
             StarknetRpcApiError::ErrUnexpectedError { .. } => 63,
             StarknetRpcApiError::InternalServerError => 500,
             StarknetRpcApiError::UnimplementedMethod => 501,
+            StarknetRpcApiError::SyncStatusNotAvailable => 502,
```
raynaudoe: Could you add an optional error message field for better context, similar to `FailedToReceiveTxn`?
Will be useful to know the underlying error source if there's any syncing issue.

madara/crates/primitives/sync/src/sync_status.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// Update the entire sync status
+    pub async fn update_sync_status(&self, new_sync_status: SyncStatus) {
+        let mut status = self.sync_status.lock().await;
+        *status = new_sync_status;
+    }
+
+    /// Get the current sync status
+    pub async fn get_sync_status(&self) -> SyncStatus {
+        self.sync_status.lock().await.clone()
```
raynaudoe: nit: maybe switching to `RwLock` for getter funcs

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
 /// * `Syncing` - An Enum that can either be a `mc_rpc_core::SyncStatus` struct representing the
 ///   sync status, or a `Boolean` (`false`) indicating that the node is not currently synchronizing.
 pub async fn syncing(starknet: &Starknet) -> StarknetRpcResult<SyncingStatus> {
```
raynaudoe: Can we change the fn's name for a more descriptive one (get_status/is_syncing) ? Also I think the description needs to be updated.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
...
+        // We need to fetch the starting block hash from the backend
+        let Some(starting_block_info) = starknet
+            .backend
+            .get_block_info(&BlockId::Number(starting_block_num))
+            .or_internal_server_error("Error getting starting block")?
+        else {
+            // If we can't get the starting block info, use a default hash
+            // This is a fallback and shouldn't normally happen
+            tracing::warn!("Could not get starting block info for block {}", starting_block_num);
+            return Ok(SyncingStatus::NotSyncing);
```
raynaudoe: Here, wrapping the error with the string message will be useful.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
 /// * `Syncing` - An Enum that can either be a `mc_rpc_core::SyncStatus` struct representing the
 ///   sync status, or a `Boolean` (`false`) indicating that the node is not currently synchronizing.
 pub async fn syncing(starknet: &Starknet) -> StarknetRpcResult<SyncingStatus> {
-    // obtain best seen (highest) block number
+    // Get the sync status from the provider
+    let sync_status = starknet.sync_status().await.or_internal_server_error("Error getting sync status")?;
```
raynaudoe: Consider adding a timeout and implementing exponential backoff retry logic

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
...
-        starting_block_info.as_nonpending().ok_or_internal_server_error("Block cannot be pending")?;
-    let starting_block_hash = starting_block_info.block_hash;
     let current_block_num = current_block_info.header.block_number;
     let current_block_hash = current_block_info.block_hash;
 
+    // Get the starting block number from sync status
+    let starting_block_num = sync_status.starting_block_num;
+
+    // Get the starting block hash - if it's not in sync status or is zero, fetch it from the backend
+    let starting_block_hash = if sync_status.starting_block_hash == BlockHash::default() {
```
raynaudoe: Can you explain the cases where `sync_status.starting_block_hash == BlockHash::default() ` ?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
...
+            // If we can't get the starting block info, use a default hash
+            // This is a fallback and shouldn't normally happen
+            tracing::warn!("Could not get starting block info for block {}", starting_block_num);
+            return Ok(SyncingStatus::NotSyncing);
+        };
+
+        let starting_block_info =
+            starting_block_info.as_nonpending().ok_or_internal_server_error("Starting block cannot be pending")?;
+
+        starting_block_info.block_hash
```
raynaudoe: Also here should update `starting_block_num`, just to be sure that `starting_block_hash` and `starting_block_num` refers to the same block.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -16,30 +16,63 @@ use crate::Starknet;
...
+    };
+
+    // Get highest block info from sync status
+    let highest_block_num = sync_status.highest_block_num;
+    let highest_block_hash = sync_status.highest_block_hash;
+
+    // If highest block number is 0 or less than current, we're not syncing
+    if highest_block_num == 0 || highest_block_num - 6 <= current_block_num {
+        return Ok(SyncingStatus::NotSyncing);
+    }
```
raynaudoe: Here it would be useful to catch the underlaying error (if possible) and return that

raynaudoe (changes requested):
Nice feature! 🚀 
Left a few comments and suggestions.

Also, could you add unit tests for syncing_status? would be nice to check if we're returning the correct struct and errors based on the content of the status.

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +226,51 @@ async fn l2_pending_block_task(
...
+
+                        // Update the sync status provider with the highest block info
+                        sync_status_provider.set_highest_block_num(block_number).await;
+                        sync_status_provider.set_highest_block_hash(block_hash).await;
+
+                        tracing::debug!("Updated highest block: number={}, hash={}", block_number, block_hash);
+                    }
+                    _ => {
+                        tracing::warn!("Got unexpected block type from provider");
+                        continue;
```
notlesh: This seems like it could be problematic. In a best case scenario it would skip blocks we don't understand, and in a worst case scenario I could see it creating an infinite loop.

I would suggest failing hard in this case or putting in some mechanism to prevent it from getting stuck.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Mohiiit: I think you misunderstood it, here not syncing is a good thing because that means it's ready and it's synced, I understand this might be confusing, these links might help: https://github.com/eqlabs/pathfinder/blob/da3ae6d3b1f87d78b162d043300334ffb9ef8a23/crates/rpc/src/method/syncing.rs#L8

https://github.com/starkware-libs/starknet-specs/blob/2030a650be4e40cfa34d5051a0334f375384a421/api/starknet_api_openrpc.json#L765

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Mohiiit: in case we have a starting block number provided through cli argument, we would just have the block number info

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Mohiiit: we are getting block info using the `starting_block_num` isn't this redundant?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
raynaudoe: yeah my bad ... In my head I was thinking on 3 different states like `Syncing`, `Synced` and `Error` but that's not how is defined in the spec 😓

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
raynaudoe: yep, you're right. No need to update it 😃

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
notlesh: Seems like the spec only allows for a true of false response, so maybe the RPC should throw a fatal error instead (since this isn't expected to happen).

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +15,112 @@ use crate::Starknet;
...
+    let sync_status = get_sync_status_with_retry(starknet)
+        .await
+        .or_internal_server_error("Error getting sync status after retries")?;
+
+    // Get current block info
     let Some(current_block_info) = starknet
         .backend
         .get_block_info(&BlockId::Tag(BlockTag::Latest))
         .or_internal_server_error("Error getting latest block")?
     else {
```
raynaudoe: Maybe check this condition before calling `get_sync_status_with_retry`?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
raynaudoe: @Mohiiit wdyt? 👆🏻

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +15,112 @@ use crate::Starknet;
...
     }))
 }
+
+// Helper function to get sync status with exponential backoff retry
+async fn get_sync_status_with_retry(starknet: &Starknet) -> Result<mp_rpc::SyncStatus, anyhow::Error> {
+    use std::time::Duration;
+    use tokio::time;
+
+    const MAX_RETRIES: u32 = 3;
+    const INITIAL_TIMEOUT_MS: u64 = 1000;
```
raynaudoe: nit, move these `const` to the top of the file

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +15,112 @@ use crate::Starknet;
...
+
+        // Check if we've reached the maximum number of retries
+        retries += 1;
+        if retries >= MAX_RETRIES {
+            return Err(anyhow::anyhow!("Failed to get sync status after {} retries", MAX_RETRIES));
+        }
+
+        // Exponential backoff
+        timeout_ms *= 2;
+        tracing::debug!("Retrying sync status (attempt {}/{}), timeout: {}ms", retries + 1, MAX_RETRIES, timeout_ms);
```
raynaudoe: Shouldn't be like:
```suggestion
        tracing::debug!("Retrying sync status (attempt {}/{}), timeout: {}ms", retries, MAX_RETRIES, timeout_ms);
```
since you're already incrementing `retries` above

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +226,73 @@ async fn l2_pending_block_task(
...
+    sync_status_provider: Arc<SyncStatusProvider>,
+) -> anyhow::Result<()> {
+    tracing::debug!("Start highest block poll");
+
+    let mut interval = tokio::time::interval(Duration::from_secs(10));
+    interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+
+    // Track consecutive failures to prevent endless loops
+    let mut consecutive_failures = 0;
+    const MAX_CONSECUTIVE_FAILURES: usize = 5;
```
raynaudoe: Same here, move this `const` to the top

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +226,73 @@ async fn l2_pending_block_task(
...
+            );
+            tracing::error!("{}", error_msg);
+            return Err(anyhow::anyhow!(error_msg));
+        }
+
+        // If we had failures but not enough to terminate, add a small delay
+        if consecutive_failures > 0 {
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    }
```
raynaudoe: You can embed this logic inside:
```rust
Err(e) => {
    tracing::error!("Error getting latest block: {:?}", e);
    consecutive_failures += 1;
}
```
since you know that in there `consecutive_failures` is > 0

raynaudoe (commented):
Left a few more comments. I'd say the only missing thing in this PR would be unit tests for `syncing_status`

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Mohiiit: hmm, I agree here

README.md
```diff
@@ -109,7 +109,7 @@ Start the Madara client with a basic set of arguments depending on your chosen m
 Synchronizes the state of the chain from genesis.
 
 ```bash
-cargo run --release --        \
+cargo run --bin madara --release --        \
```
ftheirs: The readme was fixed in this PR:
https://github.com/madara-alliance/madara/pull/564/files

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +232,69 @@ async fn l2_pending_block_task(
...
+                consecutive_failures += 1;
+                tokio::time::sleep(Duration::from_secs(HIGHEST_BLOCK_FETCH_ERROR_DELAY_SECS)).await;
+            }
+        }
+
+        // If we've had too many consecutive failures, break out of the loop
+        if consecutive_failures >= MAX_HIGHEST_BLOCK_FETCH_FAILURES {
+            let error_msg = format!(
+                "Received unexpected block types or errors for {MAX_HIGHEST_BLOCK_FETCH_FAILURES} consecutive attempts"
+            );
```
raynaudoe: Maybe you meant this?
```suggestion
            let error_msg = format!(
                "Received unexpected block types or errors for {} consecutive attempts", MAX_HIGHEST_BLOCK_FETCH_FAILURES
            );
```

notlesh: These are equivalent :)

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
raynaudoe: @Mohiiit wdyt?

madara/crates/client/rpc/src/errors.rs
jbcaron: We must be careful about the errors returned to the client, some internal errors should only be logged and not transmitted to the client

madara/crates/client/rpc/src/errors.rs
```diff
@@ -102,6 +102,8 @@ pub enum StarknetRpcApiError {
     InternalServerError,
     #[error("Unimplemented method")]
     UnimplementedMethod,
+    #[error("Sync status not available")]
```
jbcaron: According to the specs, we're not supposed to return any errors. Wouldn't this case simply be considered an internal server error?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +19,101 @@ use crate::Starknet;
...
-    let Some(current_block_info) = starknet
+    // Get current block info first
+    let current_block_info = starknet
         .backend
         .get_block_info(&BlockId::Tag(BlockTag::Latest))
         .or_internal_server_error("Error getting latest block")?
-    else {
-        return Ok(SyncingStatus::NotSyncing); // TODO: This doesn't really make sense? This can only happen when there are no block in the db at all.
-    };
+        .ok_or_internal_server_error("Latest block not found")?;
```
jbcaron: I think this case is only for the genesis block

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +232,69 @@ async fn l2_pending_block_task(
...
+
+    // Track consecutive failures to prevent endless loops
+    let mut consecutive_failures = 0;
+
+    // Keep running until cancelled
+    while ctx.run_until_cancelled(interval.tick()).await.is_some() {
+        tracing::debug!("Getting highest block...");
+
+        // Try to get the latest block
+        match provider.get_block(BlockId::Tag(BlockTag::Latest)).await {
```
jbcaron: It's better to use the get_header request, which will only return the block number and block hash

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +232,69 @@ async fn l2_pending_block_task(
     Ok(())
 }
 
+async fn l2_highest_block_fetch(
```
jbcaron: The retry and timeout logic is already handled on the client side in the `GatewayProvider`

madara/crates/primitives/sync/src/sync_status.rs
```diff
@@ -0,0 +1,289 @@
...
+
+impl Default for SyncStatusProvider {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl SyncStatusProvider {
+    /// Create a new SyncStatusProvider with default values
+    pub fn new() -> Self {
```
jbcaron: IMO, it would be better to wrap the SyncStatus in an Option rather than handling default values in the codebase (also useful in sequencer mode).

madara/crates/primitives/sync/src/sync_status.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub async fn set_current_block_hash(&self, block_hash: BlockHash) {
+        let mut status = self.sync_status.write().await;
+        status.current_block_hash = block_hash;
+    }
+
+    /// Set highest block hash
+    pub async fn set_highest_block_hash(&self, block_hash: BlockHash) {
+        let mut status = self.sync_status.write().await;
+        status.highest_block_hash = block_hash;
+    }
```
jbcaron: All these methods risk breaking the consistency between block number and block hash

madara/crates/client/rpc/src/lib.rs
```diff
@@ -49,6 +51,7 @@ pub struct Starknet {
     pub(crate) add_transaction_provider: Arc<dyn AddTransactionProvider>,
     storage_proof_config: StorageProofConfig,
     pub ctx: ServiceContext,
+    sync_status_provider: Option<Arc<SyncStatusProvider>>,
```
jbcaron: SyncStatusProvider is already an Arc, you should either remove the Arc from the structure or implement the Clone trait.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +19,101 @@ use crate::Starknet;
...
-        highest_block_hash: current_block_hash,
+        highest_block_num,
+        highest_block_hash,
         current_block_num,
         current_block_hash,
     }))
 }
+
+// Helper function to get sync status with exponential backoff retry
+async fn get_sync_status_with_retry(starknet: &Starknet) -> Result<mp_rpc::SyncStatus, anyhow::Error> {
```
jbcaron: I don't quite understand the purpose of this function,  sync_status is just a simple RwLock. Is this a leftover from a previous implementation?

jbcaron (dismissed):

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Trantorian1: I agree with this. If we cannot retrieve the genesis block than we should assume that we are syncing and return a `SyncStatus::default()` object.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +19,101 @@ use crate::Starknet;
...
+    let starting_block_num = sync_status.starting_block_num;
+
+    // Get the starting block hash - if it's not in sync status or is zero, fetch it from the backend
+    let starting_block_hash = if sync_status.starting_block_hash == BlockHash::default() {
+        // We need to fetch the starting block hash from the backend
+        let starting_block_info = starknet
+            .backend
+            .get_block_info(&BlockId::Number(starting_block_num))
+            .or_internal_server_error("Error getting starting block")?
+            .ok_or_internal_server_error(format!("Starting block not found: block number {}", starting_block_num))?;
```
Trantorian1: is there a case where this can happen? Given the checks above we can assume we are not in a genesis state, so the starting block should be in db.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +19,101 @@ use crate::Starknet;
...
+        starting_block_info.block_hash
+    } else {
+        sync_status.starting_block_hash
+    };
+
+    // Get highest block info from sync status
+    let highest_block_num = sync_status.highest_block_num;
+    let highest_block_hash = sync_status.highest_block_hash;
+
+    // If highest block number is 0 or less than current, we're not syncing
```
Trantorian1: This is not the case if we are in a genesis state. We could be calling this method at genesis with `highest_block_num == 0` but still be syncing afterwards.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
Trantorian1:
```suggestion
async fn get_sync_status_with_retry(starknet: &Starknet) -> anyhow::Result<mp_rpc::SyncStatus> {
```

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +232,69 @@ async fn l2_pending_block_task(
...
+                        tracing::debug!("Updated highest block: number={}, hash={}", block_number, block_hash);
+
+                        // Reset failure counter after successful processing
+                        consecutive_failures = 0;
+                    }
+                    _ => {
+                        tracing::error!("Got unexpected block type from provider that wasn't handled");
+                        consecutive_failures += 1;
+                        tokio::time::sleep(Duration::from_secs(HIGHEST_BLOCK_FETCH_ERROR_DELAY_SECS)).await;
+                    }
```
Trantorian1: This should not happen unless the gateway is buggy, shouldn't we be treating this as a failure case instead of retrying?

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +232,69 @@ async fn l2_pending_block_task(
...
+                }
+            }
+            Err(e) => {
+                tracing::error!("Error getting latest block: {:?}", e);
+                consecutive_failures += 1;
+                tokio::time::sleep(Duration::from_secs(HIGHEST_BLOCK_FETCH_ERROR_DELAY_SECS)).await;
+            }
```
Trantorian1: In what case would we be reaching this failure branch here?

madara/crates/node/src/main.rs
```diff
@@ -258,13 +259,16 @@ async fn main() -> anyhow::Result<()> {
         None
     };
 
+    let sync_status_provider = Arc::new(SyncStatusProvider::new());
+    // fix: create a mutex arc of sync status, that would have information like starting block, current synced block and block till what we have to sync
```
Trantorian1: Should this be implemented as part of this PR / is it still relevant?

madara/crates/node/src/service/l2.rs
```diff
@@ -56,9 +60,11 @@ impl Service for L2SyncService {
             pending_block_poll_interval,
             block_importer,
             telemetry,
+            sync_status_provider,
         } = self.clone();
         let telemetry = Arc::clone(&telemetry);
 
+        // fix: pass down the sync status to this worker
```
Trantorian1: Ditto

madara/crates/node/src/main.rs
Trantorian1: This seems strange since `SyncStatusProvider` is already an `Arc<RwLock>` internally, we should not need to add an extra layer of locking on top of this.

madara/crates/primitives/sync/src/sync_status.rs
Trantorian1: This is already the case in the `Starknet` struct anyway? I think instead of this we should encapsulate the initialization logic inside of `SyncStatusProvider`.

```rust
#[derive(Default)]
pub enum SyncStatusProvider {
    #[default]
    NotSyncing,
    Syncing(Arc<RwLock<SyncStatus>>)
}

impl SyncStatusProvider {
    pub fn init(self, sync_status: SyncStatus) -> Self {
        Self::Syncing(Arc::new(RwLock::new(sync_status))
    }
}
```

Trantorian1: This has the downside that all methods will need to check if we are in an initialized state. Another but more leaky solution would be to use generics:

```rust
pub struct SyncInactive;
pub type SyncActive = Arc<RwLock<SyncStatus>>;
pub struct SyncStatusProvider<S: SyncStatus>(S);

impl<S> SyncStatusProvider<S> {
    pub fn new() -> SyncStatusProvider<SyncInactive> {
        SyncStatusProvider(SyncInactive)
    }

    pub fn init(self, sync_status: SyncStatus) -> SyncStatusProvider<SyncActive> {
        SyncStatusProvider(Arc::new(RwLock::New(sync_status)))
    }
}

impl SyncStatusProvider<SyncActive> {
    pub async fn get_sync_status(&self) -> SyncStatus {
        self.0.read().await.clone()
    }
}
```

This has the disadvantage that the generics need to be propagated to the struct holding this or else we need to make this a `dyn`, so this might not be a great solution.

madara/crates/primitives/sync/src/sync_status.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub async fn get_sync_status(&self) -> SyncStatus {
+        self.sync_status.read().await.clone()
+    }
+
+    /// Set both number and hash for starting block
+    pub async fn set_starting_block(&self, block_num: BlockNumber, block_hash: BlockHash) {
+        let mut status = self.sync_status.write().await;
+        status.starting_block_num = block_num;
+        status.starting_block_hash = block_hash;
+    }
```
Trantorian1: In what context are we using these methods? I agree with @jbcaron that this level of interior mutability seems risky. There probably is some level of abstraction we can work into this to limit the number of mutations to a controllable few.

madara/crates/client/rpc/src/lib.rs
Trantorian1: Why are we wrapping this in an `Option` here? As far as I can see, anytime we initialize `Starknet` we are just passing it `Some(sync_status_provider)`.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +19,101 @@ use crate::Starknet;
...
+// Helper function to get sync status with exponential backoff retry
+async fn get_sync_status_with_retry(starknet: &Starknet) -> Result<mp_rpc::SyncStatus, anyhow::Error> {
+    use std::time::Duration;
+    use tokio::time;
+
+    let mut retries = 0;
+    let mut timeout_ms = INITIAL_TIMEOUT_MS;
+
+    loop {
+        match time::timeout(Duration::from_millis(timeout_ms), starknet.sync_status()).await {
```
Trantorian1: Wow, why are we doing this here? If I get this right, we are essentially doing:
```rust
loop {
    let sync = self
            .sync_status_provider
            .as_ref()
            .ok_or(StarknetRpcApiError::SyncStatusNotAvailable);
            
    let timeout = time::timeout(
        Duration::from_millis(timout_ms),
        map sync {
            OK(provider) => Ok(provider.get_sync_status().await),
            Err(e) => Err(e)
        }
    );
    
    match timeout {
        // ...
    }
}
```

But `get_sync_status` is just

```rust
pub async fn get_sync_status(&self) -> SyncStatus {
    self.sync_status.read().await.clone()
}
```

So we are timing out on an async lock acquisition??? This seems _very_ strange. Why not just rely on the runtime to acquire the lock and wake the task when we need it?

Trantorian1 (changes requested):
This PR feels quite a bit strange in the way in which it handles `SyncStatus`. Some things definitely need to be updated.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Will there be any scenario when the starting block is not zero? Asking because if there is no case like this, we can, by default, return 0

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: If the chain is in genesis state, the state should be `NotSyncing`, right? Or am I missing something?

madara/crates/client/sync/src/l2.rs
prkpndy: Can you pinpoint where exactly this is written so I can take a look? Thanks

madara/crates/client/sync/src/l2.rs
prkpndy: This can happen when the send_get request fails in GatewayProvider

madara/crates/client/sync/src/l2.rs
jbcaron: https://github.com/madara-alliance/madara/blob/main/madara/crates/client/gateway/client/src/builder.rs

madara/crates/client/sync/src/l2.rs
jbcaron: Actually, we should add this method here [methods.rs](https://github.com/madara-alliance/madara/blob/main/madara/crates/client/gateway/client/src/methods.rs)
it's simply a `get_block` with the query parameter `headerOnly=true`

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
jbcaron: No matter what we return during the construction of the genesis, it should never be an internal error.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Will remove this function

madara/crates/client/rpc/src/lib.rs
prkpndy: We are using None in a test case

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Updated the code

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Fixed

madara/crates/client/sync/src/l2.rs
prkpndy: Done

madara/crates/client/sync/src/l2.rs
prkpndy: Removed the retry logic

madara/crates/client/sync/src/l2.rs
prkpndy: Done

madara/crates/primitives/sync/src/sync_status.rs
```diff
@@ -0,0 +1,290 @@
+use mp_rpc::{BlockHash, BlockNumber, SyncStatus};
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+/// Provider for synchronization status
+#[derive(Clone)]
+pub struct SyncStatusProvider {
+    sync_status: Arc<RwLock<SyncStatus>>,
```
cchudant: Hi, sorry for coming so late to the review
I think this would be better as a tokio::sync::watch channel no?

cchudant: also, what do you think of putting this in the backend?
having it in primitives feels quite awkward to me

madara/crates/primitives/sync/Cargo.toml
cchudant, line 1: IMO, this crate can just be replaced as a whole with a tokio::sync::watch channel, wdyt?

madara/crates/client/gateway/client/src/methods.rs
```diff
@@ -36,6 +37,16 @@ impl GatewayProvider {
         }
     }
 
+    pub async fn get_block_with_header_only(&self) -> Result<ProviderBlockHeaderOnly, SequencerError> {
```
cchudant: this function needs some tests but
turns out you're in luck and you don't have to write them, because I also added this endpoint in #495 and wrote some tests.

cchudant (changes requested):
I think sync status would be better off in a tokio::sync::watch channel with something like this in it

using our types instead of the RPC types
```rs
#[derive(Clone, Debug)]
pub struct BlockNumberAndhash { pub number: u64, pub hash: Felt }

#[derive(Default, Clone, Debug)]
pub enum SyncStatus {
  /// Note: this either means we're fully synced, or the sync service is not running
  #[default] NotSyncing,
  SyncingTo(/* target */ BlockNumberAndhash)
}
```

```rs
let (sync_status_sender, sync_status_recv) = tokio::sync::watch::channel(SyncStatus::default());
// give the sender half to mc-sync and receiver half to mc-rpc or any other service that needs it 
```

the RPC endpoint would then get the latest block_n and hash in backend, and the sync status from that channel


and for the starting block, I think it makes sense to have a field in MadaraBackend that looks like
```rs
/// Starting block number. None means genesis was not imported when the MadaraBackend instance was created.
starting_block: Option<u64>,
```
this field would be set when the MadaraBackend opens the db - allowing it to work correctly for block prod as well

--

my reason is, I don't like the RPC types because they don't usually map well to inner state.
For example, the latest block_n in db is a `Option<u64>`, not a `u64`, since the db can be in the state where the genesis block is not imported yet.
If we're in this case, i think the correct behavior for the jsonrpc is to return `StarknetRpcApiError::NoBlocks` and not InternalServerError

wdyt?

madara/crates/client/sync/src/l2.rs
jbcaron: get_header is added to the client gateway in this PR https://github.com/madara-alliance/madara/pull/495

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,31 +16,65 @@ use crate::Starknet;
...
-    let starting_block_num = 0; // TODO(rpc): fix this // starknet.starting_block;
-    let starting_block_info = starknet.get_block_info(&BlockId::Number(starting_block_num))?;
-    let starting_block_info =
-        starting_block_info.as_nonpending().ok_or_internal_server_error("Block cannot be pending")?;
-    let starting_block_hash = starting_block_info.block_hash;
-    let current_block_num = current_block_info.header.block_number;
-    let current_block_hash = current_block_info.block_hash;
+    // Get the highest block info from sync status
+    let highest_block_num = sync_status.highest_block_num;
+    let highest_block_hash = sync_status.highest_block_hash;
```
raynaudoe: is it possible here to get a default hash value instead of the real one?

madara/crates/client/sync/src/l2.rs
```diff
@@ -224,6 +230,40 @@ async fn l2_pending_block_task(
...
 }
 
+async fn l2_highest_block_fetch(
+    provider: Arc<GatewayProvider>,
+    mut ctx: ServiceContext,
+    sync_status_provider: SyncStatusProvider,
+) -> anyhow::Result<()> {
+    tracing::debug!("Start highest block poll");
+
+    let mut interval = tokio::time::interval(Duration::from_secs(10));
```
raynaudoe: nit: add const value for this interval

madara/crates/client/rpc/src/errors.rs
prkpndy: Removed this error type

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Done

README.md
prkpndy: Removed the changes

madara/crates/client/rpc/src/errors.rs
prkpndy: Removed the error type

madara/crates/client/sync/src/l2.rs
prkpndy: This PR seems to be very big and might take some time to get merged. I can create an issue to delete on of the methods when both the PR are merged to main. Let me know if that works? cc @jbcaron

madara/crates/primitives/sync/src/sync_status.rs
prkpndy: I have removed all the functions except - new, get_sync_status, set_highest_block

madara/crates/client/rpc/src/lib.rs
prkpndy: Removed Arc

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Removed this function

madara/crates/node/src/main.rs
prkpndy: Updated

madara/crates/primitives/sync/src/sync_status.rs
prkpndy: Removed

madara/crates/primitives/sync/src/sync_status.rs
prkpndy: I have added a field for starting_block_info in MadaraBackend

madara/crates/client/db/src/block_db.rs
```diff
@@ -486,4 +486,8 @@ impl MadaraBackend {
             }
         }
     }
+
+    pub fn get_starting_block_info(&self) -> Result<StartingBlockInfo> {
```
jbcaron: Why return a result?

madara/crates/client/db/src/lib.rs
```diff
@@ -409,6 +409,12 @@ impl EventChannels {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct StartingBlockInfo {
+    pub starting_block_num: Option<u64>,
```
jbcaron: In what case would starting block be None?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,31 +16,69 @@ use crate::Starknet;
...
+    } else {
+        sync_status.starting_block_hash
+    };
+
+    // Get the highest block info from the sync status
+    let highest_block_num = sync_status.highest_block_num;
+    let highest_block_hash = sync_status.highest_block_hash;
+
+    // If the highest block number is 0 or less than current, we're not syncing
+    if highest_block_num == 0 || highest_block_num - 6 <= current_block_num {
```
jbcaron: Is it possible to put 6 in a constant?

madara/crates/client/sync/src/lib.rs
```diff
@@ -24,28 +24,17 @@ pub struct SyncConfig {
...
-                .context("getting sync tip")?
-                .map(|block_id| block_id + 1) // next block after the tip
-                .unwrap_or_default() as _, // or genesis
-            false,
-        )
-    };
-
-    tracing::info!("⛓️  Starting L2 sync from block {}", starting_block);
+    let starting_block_info = backend.get_starting_block_info()?;
+    let starting_block = starting_block_info.starting_block_num.unwrap_or_default();
```
jbcaron: Does None mean to sync from block 0?

madara/crates/client/db/src/block_db.rs
prkpndy: I just wanted to keep this method consistent with other methods for MadaraBackend. Since most of them returns Result, I think it would be better to return this as a Result as well.
(also, if we do any future modification which might have a chance of returning an error, we can do that without breaking the api contract)

madara/crates/client/db/src/lib.rs
prkpndy: None means genesis was not imported when the MadaraBackend instance was created.

madara/crates/client/sync/src/lib.rs
prkpndy: Umm.. I think so, but not sure about this. This was the implementation before as well in `madara/crates/client/sync/src/lib.rs`. Do you think I should panic in this case?

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
prkpndy: Done

raynaudoe (approved):

madara/crates/client/db/src/lib.rs
```diff
@@ -447,6 +447,23 @@ impl EventChannels {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct StartingBlockInfo {
```
cchudant: unused struct

madara/crates/client/db/src/lib.rs
```diff
@@ -608,6 +629,8 @@ impl MadaraBackend {
             sender_block_info: tokio::sync::broadcast::channel(100).0,
             sender_event: EventChannels::new(100),
             config,
+            starting_block: starting_block_n,
+            sync_status: RwLock::new(SyncStatus::default()),
```
cchudant: i prefer if this RwLock is hidden behind a private field inside a struct
this would enable us to guarantee that we don't take this lock improperly and cause deadlocks - encapsulate the RwLock

on another file, db/src/sync_status.rs for eg.
```rs
#[derive(Default)]
pub(crate) struct SyncStatusCell(RwLock<SyncStatus>);
impl SyncStatusCell {
  pub(crate) fn set(&self, status: SyncStatus) { ... }
  pub(crate) fn get(&self) -> SyncStatus { ... }
}
```

this is just to enforce that we can't take the lock in the rest of the crate

a deadlock is statically impossible here since the lock is a leaf inside the global mutex hierarchy ie. it's impossible to take another lock while this one is active.
We know that, but by limiting access we can ensure this is the case by only looking at 1 file vs the whole crate; we statically can't misuse it

madara/crates/client/eth/src/l1_messaging.rs
cchudant, line 1: empty file

madara/crates/client/eth/src/state_update.rs
cchudant, line 1: empty file

madara/crates/client/rpc/src/test_utils.rs
```diff
@@ -63,6 +63,7 @@ pub fn rpc_test_setup() -> (Arc<MadaraBackend>, Starknet) {
         Arc::new(TestTransactionProvider),
         Default::default(),
         ServiceContext::new_for_testing(),
+        None,
```
cchudant: what is this None?

madara/crates/client/rpc/src/lib.rs
```diff
@@ -108,6 +122,43 @@ impl Starknet {
             .or_internal_server_error("Error getting L1 last confirmed block")?
             .unwrap_or_default())
     }
+
+    pub async fn sync_status(&self) -> StarknetRpcResult<StarknetSyncStatus> {
```
cchudant: idk if it's worth having a func inside Starknet directly for that, the logic could be directly inside of the endpoint function
(StarknetSyncStatus would go away too in that case)

madara/crates/client/db/src/block_db.rs
```diff
@@ -424,4 +424,24 @@ impl MadaraBackend {
...
+
+    pub fn get_starting_block(&self) -> Option<u64> {
+        self.starting_block
+    }
+
+    pub async fn get_sync_status(&self) -> SyncStatus {
+        self.sync_status.read().await.clone()
+    }
+
+    pub async fn set_sync_status(&self, highest_block_n: Option<u64>, highest_block_hash: Option<Felt>) {
```
cchudant: why not sync_status: SyncStatus directly instead of 2 options?

this would be less confusing since it's unclear what passing None, Some, Some None or None, None would mean here

madara/crates/client/db/src/lib.rs
```diff
@@ -593,12 +612,14 @@ impl MadaraBackend {
         chain_config: Arc<ChainConfig>,
         config: MadaraBackendConfig,
     ) -> anyhow::Result<Self> {
+        let chain_head = ChainHead::load_from_db(&db).context("Getting latest block_n from database")?;
```
cchudant: this chain_head is already taken from db inside of the `open` function.
IMO - you should put you set `starting_block` inside of that function too and not here, for consistency.
You can just default `starting_block` to be None in this constructor, and replace the value in the `open` function

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
```diff
@@ -15,32 +18,46 @@ use crate::Starknet;
...
-        starting_block_hash,
-        highest_block_num: current_block_num, // TODO(merge): is this correct?
-        highest_block_hash: current_block_hash,
-        current_block_num,
-        current_block_hash,
-    }))
+    match sync_status {
+        StarknetSyncStatus::NotRunning => Ok(SyncingStatus::NotSyncing),
+        StarknetSyncStatus::Running { starting_block_n, starting_block_hash, highest_block_n, highest_block_hash } => {
+            if highest_block_n - SYNC_THRESHOLD_BLOCKS <= current_block_num {
```
cchudant: this will panic in debug if highest_block_n is <= 6
use
```rs
if current_block_num + SYNC_THRESHOLD_BLOCKS >= highest_block_n {
```

madara/crates/primitives/gateway/src/block.rs
```diff
@@ -296,6 +296,14 @@ impl ProviderBlockPending {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+// #[serde(deny_unknown_fields)] // TODO(v0.13.4): Re-add this attribute when v0.13.4 is supported.
+#[cfg_attr(test, derive(Eq))]
+pub struct ProviderBlockHeaderOnly {
```
cchudant: unused struct (the struct you want is already there)

madara/crates/primitives/rpc/src/v0_7_1/starknet_api_openrpc.rs
```diff
@@ -941,7 +941,7 @@ pub struct StructMember {
 }
 
 /// An object describing the node synchronization status
-#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize, Default)]
```
cchudant: nitpick: maybe remove the Default derive here since it doesn't make sense IMO

cchudant (changes requested):
There are very fey changes left to do, mainly nitpicks / empty files / unused structs
and a possible panic to fix
Great!

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/syncing.rs
cchudant: oh, your condition is reversed
you meant `>=` not `<=`, so replace it with
```rs
if current_block_num + SYNC_THRESHOLD_BLOCKS <= highest_block_n {
```
it would really help if there was an integration test for the endpoint but i would prefer to merge the pr asap so don't bother

cchudant (approved):
cool! let's merge this once the CI says its okay!

Trantorian1 (approved):
Looks good, thanks for all the changes! :fire:

madara/crates/client/rpc/src/lib.rs
```diff
@@ -42,6 +42,18 @@ impl Default for StorageProofConfig {
...
+pub enum StarknetSyncStatus {
+    #[default]
+    NotRunning,
+    Running {
+        starting_block_n: u64,
+        starting_block_hash: Felt,
+        highest_block_n: u64,
+        highest_block_hash: Felt,
+    },
+}
```
Mohiiit, line 55: this is not being used right?

madara/crates/client/sync/src/l2.rs
Mohiiit, line 1: empty file?

Mohiiit (commented):
lfgg!

antiyro (approved):

PR #530, opened 2025-03-13
fix: declare_v0

jbcaron:
## Pull Request type

- Bugfix

The RPC endpoint `madara_addDeclareV0Transaction` does not use RPC type CompressedLegacy, which causes issues with the deserialization of class ABIs, as a field should be untagged.

Resolves: #NA
- The `madara_addDeclareV0Transaction` endpoint now uses a new RPC type: `BroadcastedDeclareTxnV0`.
- This type is not perfectly aligned with other RPC transaction types in terms of version, query_only, and type.
- It is now easier to use this endpoint with starknet-rs class types or other external libraries.

Yes, for clients using the old RPC admin.

### DIFF
--- a/madara/crates/primitives/rpc/src/admin/mod.rs
+++ b/madara/crates/primitives/rpc/src/admin/mod.rs
@@ -1,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+use starknet_types_core::felt::Felt;
+
+use crate::{Address, DeprecatedContractClass, Signature};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub struct BroadcastedDeclareTxnV0 {
+    /// The class to be declared
+    pub contract_class: DeprecatedContractClass,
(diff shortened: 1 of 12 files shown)


madara/crates/primitives/rpc/src/admin/mod.rs
```diff
@@ -0,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+use starknet_types_core::felt::Felt;
+
+use crate::{Address, DeprecatedContractClass, Signature};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub struct BroadcastedDeclareTxnV0 {
```
Trantorian1, line 7: Consider adding a quick comment explaining why this structure is here for future reference? Idk.

Trantorian1 (approved):
Looks good to me!

cchudant (approved):

Mohiiit (approved):
lgtm!

PR #535, opened 2025-03-18
ci: ♻️ Refactor CI & Release Process

timbrinded:
The goal of this PR is the clean the CI in order to:
* Make it more readable
* Make it easier to maintain/update
* Consolidate versions and variables
* Optimize the build execution
* Add support for local testing of the workflows
* Improve the release process

Here are the important changes:
* Rename the workflow following [`.github/README.md`](diffhunk://#diff-438478e7713ffdb96be895cd2cba03144f30c04f1442ae246acdeb2f1c13efd5R1-R126)
* Removed unused workflows/tasks
* Merged (madara & orchestrator) CHANGELOG.md. This file is now updated on release only
* Removed check on push-main (since the checks are done on PR and should be mandatory)
* Adds Manual Process to perform release (specifying a tag version)
  * Adds madara-production.Dockerfile to build "production" ready madara/orchestrator
  * Adds PR for changelog and draft the release note (for those to be manually reviewed before publish)
  * Uses conventional changelog (Ex of the [new madara changelog](https://github.com/crystalin/madara/blob/main/CHANGELOG.md) )
* Updates Rust toolchain from `1.81` to `1.85`
  
`act` is now supported for all the `task-...` in the workflow. More details on [`.github/README.md`](diffhunk://#diff-438478e7713ffdb96be895cd2cba03144f30c04f1442ae246acdeb2f1c13efd5R1-R126)


notlesh (approved):

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.81"
+channel = "1.85"
```
jbcaron, line 2: Don't we also want to update to Rust 2024?  
Especially since I don't think Madara is used as a crate.

raynaudoe (approved):
LGTM!

crystalin (approved):

.devcontainer/Dockerfile
jbcaron, line 1: I think Scarab is missing, maybe also a clean of the APT cache

.devcontainer/devcontainer.json
```diff
@@ -13,12 +13,11 @@
...
-        "1YiB.rust-bundle",
-        "tamasfe.even-better-toml",
-        "serayuzgur.crates",
-        "vivaxy.vscode-conventional-commits",
-        "streetsidesoftware.code-spell-checker"
+        //        "1YiB.rust-bundle",
+        //        "tamasfe.even-better-toml",
+        //        "serayuzgur.crates",
+        //        "vivaxy.vscode-conventional-commits",
+        //        "streetsidesoftware.code-spell-checker"
```
jbcaron: was it forgotten?

.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,139 @@
...
+          if [[ "$RELEASE_EXISTS" != "false" ]]; then
+            echo "Release ${{ github.event.inputs.tag || github.ref_name }} already exists"
+            exit 1
+          else
+            echo "Release ${{ github.event.inputs.tag || github.ref_name }} does not exist"
+          fi
+
+      - name: Update CHANGELOG
+        id: changelog
+        uses: requarks/changelog-action@v1
```
jbcaron, line 107: It would be nice to exclude certain types of commits with `excludeTypes` imo

docker/madara-production.Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+COPY --from=builder --chown=madara /madara/target/production/madara /madara/madara
+COPY --from=builder --chown=madara /madara/target/production/orchestrator /madara/orchestrator
+
+RUN chmod uog+x /madara/madara /madara/orchestrator
+
+# 30333 for parachain p2p
+# 30334 for relaychain p2p
+# 9944 for Websocket & RPC call
+# 9615 for Prometheus (metrics)
+EXPOSE 30333 30334 9944 9615
```
jbcaron: where does this configuration come from?

jbcaron (commented):
Great work! This is clearer and more maintainable. Would it be possible to check the nix environment as well? I don't think it has been updated with the merge of orchestrator.

rust-toolchain.toml
timbrinded: This causes build issues and would require some refactor. Should save for a separate PR for closer inspection of changes required

docker/madara-production.Dockerfile
timbrinded: These are the default port numbers for any substrate node. Have made them ARGS now

.github/workflows/manual-release.yml
timbrinded: i have a follow up PR planned to look at this, will tackle that then if ok

.devcontainer/Dockerfile
timbrinded: good catch, fixed

.devcontainer/devcontainer.json
timbrinded: Choosing extensions feels like a very personal choice, so i removed for simplicity. If there's any non-contentious choices we can add if you like.

.github/workflows/task-lint-cargo.yml
```diff
@@ -0,0 +1,37 @@
...
+          install-mold: false
+          install-scarb: true
+          python-version: ${{ env.BUILD_PYTHON_VERSION }}
+          build-snos: true
+
+      - name: Madara Format and clippy
+        run: |
+          cargo fmt -- --check
+          cargo clippy --workspace --no-deps -- -D warnings
+          cargo clippy --workspace --tests --no-deps -- -D warnings
```
notlesh, line 32: I suggest running `fmt` after `clippy` (and all other lints, for that matter).

The reasoning is that lints like `clippy` will often introduce changes that affect `fmt`'s opinions, but rarely the other way around.

A concrete example is when a lint suggestion adds or removes some code, which causes `fmt` to want to break that line of code up into multiple lines or vise-versa.

If you run `fmt` first, you may have to do:

* fix fmt
* fix clippy
* fix fmt (again)

Whereas if you run `clippy` first you would only have to fix fmt once:

* fix clippy
* fix fmt

This brings up another issue I've had with Madara CI where `fmt` and/or `clippy` on my machine don't produce the exact results that I see in CI, preventing me from testing this locally. If I see this again after this CI refactor, I'll try to dig in to understand why.

timbrinded: if that is the case, should we not run `fmt` first? since the CI is to flag up issues and not to actually apply a fix

timbrinded:
100 commits later this is ready to review again 😅

There are improvements to caching in another PR i have open - but so far at least coverage and testing is back to working again.

docker/madara-production.Dockerfile
jbcaron: I believe that only the RPC port is necessary

.devcontainer/devcontainer.json
jbcaron: It's all good

.github/workflows/task-lint-cargo.yml
notlesh: Yeah, that's my suggestion. I would run it after all rust linters

.devcontainer/Dockerfile
```diff
@@ -1,30 +1,95 @@
...
-        protobuf-compiler \
-        build-essential \
-        clang \
-        zsh \
-        htop \
-        jq \
-        tree \
-    && sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)"
\ No newline at end of file
+    libasound2-dev \
```
notlesh, line 29: Why would a CI server need `libasound2` :eyes: 

This reminds me of a time a colleague was running a Mac Mini in his office as part of our CI pipeline. I had SSH access to it and learned that you could use the `say` command to make the device speak arbitrary text strings out loud with its built in speaker... I had a lot of fun with that.

jagerman: Luckily that mac mini is now in my basement so I can enable colleagues to have access to it again!

.github/workflows/task-coverage-orchestrator.yml
```diff
@@ -0,0 +1,116 @@
...
+      RPC_FOR_SNOS:
+        required: true
+
+jobs:
+  orchestrator_coverage:
+    runs-on: karnot-arc-runner-set
+
+    services:
+      localstack:
+        image: localstack/localstack@sha256:763947722c6c8d33d5fbf7e8d52b4bddec5be35274a0998fdc6176d733375314
```
notlesh, line 25: A comment about where this came from would be helpful

docker/madara-production.Dockerfile
```diff
@@ -0,0 +1,136 @@
...
+
+RUN echo "*** Building Madara ***"
+RUN cargo build --profile=production --all
+
+FROM debian:stable-slim
+LABEL description="Production Binary for Madara Nodes"
+
+# Define port arguments with defaults
+ARG PARACHAIN_P2P_PORT=30333
+ARG RELAYCHAIN_P2P_PORT=30334
```
notlesh, line 104: Madara used to be based on Substrate, but it's not anymore, so these aren't relevant.

notlesh: I think this applies to `PROMETHEUS_PORT` as well...

notlesh (approved):

.devcontainer/Dockerfile
timbrinded: 1. This is a `.devcontainer` file, so is meant to be run on a dev's local machine. 
2. This wasnt added by this PR, just changed the formatting. Actually in the master all the way back to original commit!
3. If i had to guess why its here is to allow for things like `echo -e "\a"` style notification dings, which would be useful for long running tasks like e2e tests and builds

.github/workflows/task-coverage-orchestrator.yml
timbrinded: This is pre-existing [from](https://github.com/madara-alliance/madara/blob/f42827cecedcd47aceb3a84074981ee5a62e5f8b/.github/workflows/e2e-test.yml#L15) [master](https://github.com/madara-alliance/madara/blob/f42827cecedcd47aceb3a84074981ee5a62e5f8b/orchestrator/README.md?plain=1#L215). i have no idea about the rational of choosing this godly sha.

PR #541, opened 2025-03-20
ci: 💪 Improve Test Performance

timbrinded:
## Pull Request type

- Build-related changes

Caching introduced in https://github.com/madara-alliance/madara/pull/535 only really restores when a dev pushes sequential commits to the same PR.

Additionally, a lot of the `async` logic is broken around sharing Anvil and Madara node resources in the test harness. 

## What is the new behavior?

- Fixes the failing tests introduced by new `l2_gas` added to blockheader
- Reworked async logic for `mc-eth` and `mc-e2e-tests` crates
- Have added caches to all length tasks
- When a PR is merged, associated caches are deleted.

No

## Other information

- Cache layering means there will always be a hit (with degrading accuracy). This may seem unsafe but given we always run a build command on top of it, the build system (be it cargo, foundry etc) will be able to save time by what it deems still relevant. 
  - This is the same as being able to run `cargo build` locally when changing branches
- GH gives you 10GB of caching, and when full will delete the oldest. Deleting caches of merged PRs mean we won't accidentally delete still useful caches on code that hasn't changed in a while.

> [!NOTE]  
> You can use `gh cache <cmd>` to manually inspect and delete the caches as required.
> More info: [here](https://cli.github.com/manual/gh_cache)


.github/workflows/task-test-rust.yml
```diff
@@ -0,0 +1,45 @@
+---
+name: Task - Test Rust
+
+on:
+  workflow_dispatch:
+  workflow_call:
```
HermanObst, line 6: Does this get triggered at all?

HermanObst: Does this get triggered at all? @timbrinded

madara/crates/tests/src/lib.rs
```diff
@@ -131,9 +131,15 @@ impl Drop for MadaraCmd {
 
         // Ensure process cleanup
         let _ = child.wait();
+        std::thread::sleep(Duration::from_millis(500));
     }
 }
 
+fn is_port_available(port: u16) -> bool {
+    use std::net::TcpListener;
+    TcpListener::bind(("127.0.0.1", port)).is_ok()
```
notlesh: I think you can do `let socket = TcpListener::bind("127.0.0.1:0").unwrap();` and let the OS find a port for you

madara/crates/tests/src/lib.rs
```diff
@@ -131,9 +131,15 @@ impl Drop for MadaraCmd {
 
         // Ensure process cleanup
         let _ = child.wait();
+        std::thread::sleep(Duration::from_millis(500));
```
notlesh: I think a `sleep()` inside a `Drop` impl is probably asking for trouble, although I see that we did it just above this as well...

It is inside a `tests` crate, though... :thinking:

.github/workflows/task-lint-cargo.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       # Setup the rust environment
       - uses: ./.github/actions/rust-setup
         with:
-          cache-key: ${{ env.CLIPPY_RUST_CACHE_KEY }}
+          cache-key: ${{ env.BUILD_RUST_CACHE_KEY }}
```
notlesh, line 20: What was the point of `CLIPPY_RUST_CACHE_KEY`? Does it still exist?

.github/workflows/task-test-madara.yml
```diff
@@ -40,16 +40,23 @@ jobs:
...
-      - name: Run madara tests
+      - name: Run madara tests (parallel)
+        run: |
+          cargo nextest run --release --package "mp-*" --package "m-proc-*" --no-fail-fast --tests
+
+      - name: Run madara tests (sequential)
         run: |
           export COVERAGE_BIN=$(realpath target/release/madara)
-          cargo test --release --package "mc-*" --package "mp-*"
+          cargo nextest run --release --package "mc-*" --no-fail-fast -j 1
```
raynaudoe: Shouldn't also be testing with `--package "mp-*"` ?

.github/workflows/task-test-rust.yml
```diff
@@ -0,0 +1,45 @@
...
+            target/nextest
+            target/nextest-*
+          key: ${{ runner.os }}-rust-test-${{ hashFiles('madara/**/*.rs','**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-rust-test-${{ hashFiles('**/Cargo.lock') }}
+            ${{ runner.os }}-rust-test-
+
+      - name: Run unit tests
+        run: |
+          cargo nextest run --release --workspace --test-threads=1
```
raynaudoe: Would be nice to have in the .env file the variable for the number of threads to use in nextest, just to able to bump it if we need to.

raynaudoe (approved):
LGTM!
Also, status badges could be added to the `README.md` 🤩

.github/workflows/task-test-rust.yml
timbrinded: good catch, thanks. This is deleted but reintroduced with rebase.

madara/crates/tests/src/lib.rs
timbrinded: fair enough, removed

.github/workflows/task-lint-cargo.yml
timbrinded: as far as i can tell it was so that it got its own cache entry, but actually you benefit from being able to reuse the build cache for running it (hence i changed it)

.github/workflows/task-test-madara.yml
timbrinded: the `mp-*` crates are tested in the parallel tests above.

madara/crates/tests/src/lib.rs
timbrinded: I did not know this, it's very useful. Completely rewrote the async logic here to take advantage of this fact.

.github/workflows/task-test-rust.yml
timbrinded: should be multithreaded now

notlesh (approved):

raynaudoe (approved):

HermanObst (approved):
Looking good!

PR #542, opened 2025-03-24
fix(cli): private key input

jbcaron:
## Pull Request type

- Bugfix

Currently, the private key handling in Madara has several issues:

- the private key is deserialized from the chain config file using a custom deserializer
- there's no way to provide a private key via environment variable or CLI
- when overriding the chain config, the private key is set to default and cannot be overridden

Resolves: #NA

## What is the new behavior?

- added support for specifying the private key via environment variable (MADARA_PRIVATE_KEY) or CLI parameter
- improved private key handling by implementing TryFrom<String> for ZeroingPrivateKey
- made the chain config mutable in the appropriate places to support setting the private key
- fixed issue where chain config overrides would reset the private key to default

No


Trantorian1 (approved):
All looks good!

raynaudoe (approved):
LGTM!

PR #546, opened 2025-03-26
test(CI): Avoid shared anvil

notlesh:
## Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/madara/crates/client/eth/src/client.rs
+++ b/madara/crates/client/eth/src/client.rs
@@ -168,7 +168,7 @@ pub mod eth_client_getter_test {
     };
 
     use tokio;
-    use tokio::sync::OnceCell;
+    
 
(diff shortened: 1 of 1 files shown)


Trantorian1:
Is there a reason why we are removing the shared anvil? If I remember correctly this was implemented to avoid some issues related to test performance / rate limiting. @jbcaron wdyt?

jbcaron:
> Is there a reason why we are removing the shared anvil? If I remember correctly this was implemented to avoid some issues related to test performance / rate limiting. @jbcaron wdyt?

The problem was that starting an Anvil instance takes quite a long time.

notlesh:
This didn't resolve anything, so I'll close it out

PR #547, opened 2025-03-27
ci: 🤝 Manage shared resources better

timbrinded:
## Pull Request type

- Bugfix

Latest changes to multi-thread the test execution has revealed some race conditions - which are only visible when the CI runners are going slowly.

## What is the new behavior?

More reliable CI:

- Added thread protection to opentelemetry connection
- Improve error reporting around `ETH_FORK_URL`
- More granular caching for CI jobs

No


Trantorian1:
Why do we need to make anvil tests serial? If I remember correctly @Mohiiit @jbcaron  these used to be serial before (using `serial-test`) but this was removed to achieve better test performance and lower CI times.

Trantorian1:
Also, it seems the CI has moved to using `cargo-nextest` which has a built-in mode to run tests serially. You can find out more about it [here](https://nexte.st/docs/configuration/test-groups/?h=serial)

timbrinded:
@Trantorian1 
Cargo nextest exclusion group means the test executes on a single thread, thus serially. We still want the shared anvil to be on its own thread, but we force the tests to execute serially so they don't time out. What we are seeing here is that when the CI runners slow down, the tests get scheduled and then wait for the anvil lock to be released - which when its above 480s it dies. Attempt here is to get the best of both worlds.

What would be best is that every test spawns an anvil - but then we get rate limited by the ETH fork rpc.

Trantorian1:
Thanks for the clarification, this makes sense!

.config/nextest.toml
```diff
@@ -0,0 +1,10 @@
+[test-groups]
+serial-integration = { max-threads = 1 }
+
+#[[profile.default.overrides]]
+#filter = 'package(mc-eth)'
+#test-group = 'serial-integration'
+#
+#[[profile.default.overrides]]
+#filter = 'package(mc-e2e-tests)'
+#test-group = 'serial-integration'
```
notlesh: Can this be removed?

notlesh (approved):

.config/nextest.toml
timbrinded: I thought be helpful to leave in case you wanted to run sequentially- but happy to kill.

Thoughts?

notlesh: I would document it, then

madara/crates/client/eth/src/harness.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+pub fn create_anvil_instance() -> AnvilInstance {
+    let fork_url = std::env::var("ETH_FORK_URL").expect("ETH_FORK_URL not set");
+    Anvil::new()
+        .fork(fork_url)
+        .fork_block_number(L1_BLOCK_NUMBER)
+        .port(0u16)
+        .timeout(600_000) // 10 minutes
```
notlesh: According to `anvil --help` this value applies to requests made to RPC -- I don't think 10 minutes is appropriate in that case.

```
      --timeout <timeout>
          Timeout in ms for requests sent to remote JSON-RPC server in forking mode.

          Default value 45000
```

notlesh: Actually, this seems to be a different value than what the Rust `Anvil` impl provides:

https://docs.rs/ethers-core/2.0.14/src/ethers_core/utils/anvil.rs.html#259

timbrinded: Both of those links are something else. What you're looking for is: [alloy::node_bindings::Anvil](https://docs.rs/alloy/latest/alloy/node_bindings/struct.Anvil.html#method.timeout)

But yes it's the start up timeout, not rpc timeout. 10 minutes is overkill, but needed for when we ran sequentially. Can be relaxed down now to a lower number.

notlesh: It's the same thing, you can see here the same code snippet from `alloy`: https://docs.rs/alloy-node-bindings/0.12.6/src/alloy_node_bindings/nodes/anvil.rs.html#379

timbrinded: this is alloy not ethers.rs but yes, they've appeared to reuse the same code (same team after all).

notlesh (approved):

raynaudoe (approved):
LGTM!

raynaudoe:
@timbrinded can you please merge main into this branch?

PR #551, opened 2025-03-28
CI/fix anvil instances

raynaudoe:
On this PR:
- Spawn an external anvil instance to use in tests.
- Minor fixes on CI setup files


Mohiiit (approved):
Shall we remove the task-test-madara.yml in this PR or do you think we shall move that to some other PR? Apart from that, LGTM

raynaudoe:
> Shall we remove the task-test-madara.yml in this PR or do you think we shall move that to some other PR? Apart from that, LGTM

you mean leaving just the task-coverage-madara.yml ?

Mohiiit:
yes, because I don't see merit in having both? wdyt? is there any?

raynaudoe:
> yes, because I don't see merit in having both? wdyt? is there any?

Yep, I agree, but AFAIK the ones with coverage have a different running config? like they don't fail fast. Anyways, I think we can remove it in a following PR so we avoid blocking other PRs that are waiting for a green ci 😃 , would that be ok?

Mohiiit:
yeah we should merge this for now and we can focus on it later on

notlesh (approved):

PR #552, opened 2025-03-28
Update deps & Address conversion function

ftheirs:
Update dependencies and unify all alloy and alloy-primitives to use the version from the general `Cargo.toml`
Some updates to adjust the code to a newer `alloy` version

Create a helper function to convert from hexstring to Address type. This function will pad with zeroes if the hexstring doesn't have all the 40-char that are needed for an Ethereum address. This will prevent some errors in the execution.
Use this helper function for Core contract address, verifier address, etc

- Bugfix
- Feature

No


raynaudoe (approved):

HermanObst (approved):
LGTM

Mohiiit (approved):

PR #557, opened 2025-04-02
Monorepo: Adding bootstrapper 

Mohiiit:
- Added bootstrapper to the monorepo
- Updated CI to include bootstrapper’s workflow
- Bridge is now tested in CI using the latest Madara binary via the bootstrapper


jbcaron:
Do we really want the contract artifacts? And not just their source code? The `build.rs` can handle compiling the contracts.

.github/workflows/task-build-madara.yml
```diff
@@ -45,6 +60,28 @@ jobs:
...
           build-snos: true
 
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Install Ganache
+        run: |
+          npm install -g ganache@7.9.0
```
ftheirs: Can you define the version using:
```
env:
  GANACHE_VERSION: 7.9.0
  SOLC_VERSION: 0.8.19
```

```
      - name: Install Solc
        run: |
          pip install solc-select
          solc-select install $SOLC_VERSION
          solc-select use $SOLC_VERSION
          solc --version
```

It will simplify future version upgrades

Mohiiit: yeah I am planning to follow the same, will push the commits soon

Makefile
```diff
@@ -167,3 +167,214 @@ snos:
 	git checkout $(CAIRO_LANG_COMMIT) && \
 	cd ../.. && \
 	cairo-compile orchestrator/cairo-lang/src/starkware/starknet/core/os/os.cairo --output orchestrator/build/os_latest.json --cairo_path orchestrator/cairo-lang/src
+
+# =============================================================================
```
ftheirs, line 172: I'd split this. I think that these lines should be part of a Bootstrapper Makefile and I'll keep here just a call to that file. Besides, it would awesome to have smaller Makefile, at least one for every contract 
`starkgate-contracts-legacy`, `starkgate-contracts-latest`, `braavos-account-cairo` and `argent-contracts-starknet`
 That would heavily improve the readability from each step!

Makefile
```diff
@@ -167,3 +167,214 @@ snos:
...
+	cp ./lib/argent-contracts-starknet/target/dev/argent_ArgentAccount.compiled_contract_class.json ./artifacts/ArgentAccount.casm.json
+
+# =============================================================================
+# Main Build Targets
+# =============================================================================
+
+# Target: build-contracts
+# Builds all contracts (legacy and latest versions)
+build-contracts:
+	make starkgate-contracts-legacy
```
ftheirs, line 366: I'm getting an error in macOS in this step. I'll try to add more details 🔍

ftheirs: `git submodule update --init --recursive` was not working correctly cause the folders from `lib/` were already there!
It's working fine for macOS but everything related to bootstrapper should be encapsulated under `bootstrapper` folder

bootstrapper/crates/bootstrapper/Cargo.toml
```diff
@@ -0,0 +1,69 @@
+[package]
+name = "madara-bootstrapper"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.80"
+base64 = "0.22"
+ethers = { git = "https://github.com/gakonst/ethers-rs", rev = "f0e5b194f09c533feb10d1a686ddb9e5946ec107" }
```
ftheirs, line 9: why this is pinned to an older version?

bootstrapper/crates/bootstrapper/src/contract_clients/legacy_class.rs
```diff
@@ -0,0 +1,550 @@
+// Note: This code has been taken from the madara code-base.
```
ftheirs, line 1: isn't any way to reuse madara's code? Too many duplicated lines

bootstrapper/crates/bootstrapper/Cargo.toml
```diff
@@ -0,0 +1,69 @@
+[package]
+name = "madara-bootstrapper"
```
ftheirs, line 2: not sure if the name should be `bootstrapper` or `madara-bootstrapper`

bootstrapper/crates/bootstrapper/src/tests/constants.rs
```diff
@@ -0,0 +1,19 @@
+pub const ETH_RPC: &str = "http://127.0.0.1:8545";
+pub const ETH_PRIV_KEY: &str = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+pub const ROLLUP_SEQ_URL: &str = "http://127.0.0.1:19944";
```
ftheirs, line 3: Is the port ok? I think that there is an extra `1`. `19944` --> `9944`

bootstrapper/crates/bootstrapper/src/tests/constants.rs
```diff
@@ -0,0 +1,19 @@
...
+pub const ROLLUP_DECLARE_V0_SEQ_URL: &str = "http://127.0.0.1:19943";
```
ftheirs, line 4: same as above

.env.test
```diff
@@ -115,3 +115,23 @@ MADARA_ORCHESTRATOR_MADARA_RPC_URL=http://81.16.176.130:9545
...
+ETH_PRIV_KEY=""
+ETH_RPC=""
+FEE_TOKEN_ADDRESS=""
+L1_DEPLOYER_ADDRESS=""
+L1_WAIT_TIME=""
+L2_DEPLOYER_ADDRESS=""
+ROLLUP_PRIV_KEY=""
+ROLLUP_SEQ_URL=""
+SN_OS_CONFIG_HASH_VERSION=""
+SN_OS_PROGRAM_HASH=""
```
raynaudoe, line 134: nit: rename to `SNOS_*` so it's the same as other envs

.github/actions/rust-setup/action.yml
```diff
@@ -22,12 +22,32 @@ inputs:
...
+  install-solc:
+    description: "Whether to install Solc"
+    required: false
+    default: "false"
   foundry-version:
     description: "Foundry version to install"
     required: false
   python-version:
     description: "Python version to use for Cairo 0/SNOS"
-    default: "3.9"
```
raynaudoe, line 30: IIRC this default value is used in another action?

.github/workflows/task-build-madara.yml
```diff
@@ -24,12 +30,21 @@ jobs:
...
         with:
-          submodules: true
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
raynaudoe, line 47: I think this is not needed in this task?

.github/workflows/task-build-madara.yml
```diff
@@ -42,9 +57,20 @@ jobs:
...
+          install-ganache: true
+          ganache-version: ${{ env.BUILD_GANACHE_VERSION }}
+          install-solc: true
+          solc-version: ${{ env.BUILD_SOLC_VERSION }}
           python-version: ${{ env.BUILD_PYTHON_VERSION }}
           build-snos: true
 
+      - name: Build the artifacts
+        run: |
+          git submodule update --init --recursive
```
raynaudoe: Maybe this step can be omitted since it's already getting submodules in the `action/checkout` step

.github/workflows/task-build-madara.yml
```diff
@@ -104,6 +144,21 @@ jobs:
           echo "$HASH" > cairo_artifacts_hash.txt
           echo "Hash of Cairo artifacts is: $HASH (with timestamp: $TIMESTAMP)"
 
+      - name: Generate Bootstrapper artifacts hash
+        id: generate_bootstrapper_artifacts_hash
+        run: |
+          # Get current Unix timestamp for uniqueness
+          TIMESTAMP=$(date +%s)
+          # Generate hash of all files in cairo/target directory
```
raynaudoe:
```suggestion
          # Generate hash of all files in artifacts directory
```

.github/workflows/task-build-madara.yml
```diff
@@ -104,6 +144,21 @@ jobs:
...
+          FILES_HASH=$(find artifacts -type f -exec sha256sum {} \; | sort | sha256sum | awk '{ print $1 }')
+          # Combine timestamp with files hash for unique identifier
+          HASH=$(echo "$TIMESTAMP $FILES_HASH" | sha256sum | awk '{ print $1 }')
+          # Set output for use in subsequent steps
+          echo "hash=$HASH" >> $GITHUB_OUTPUT
```
raynaudoe: not sure if this env will collide with the other one from the previous step with the same name

.github/workflows/task-build-madara.yml
```diff
@@ -125,14 +187,27 @@ jobs:
...
+        with:
+          path: artifacts
+          key: ${{ runner.os }}-${{ env.BUILD_ACTION_CACHE_PREFIX }}-bootstrapper-${{ steps.generate_bootstrapper_artifacts_hash.outputs.hash }}
+
       - name: Display hashes
         run: |
           echo "madara-binary-hash: ${{ steps.generate-binary-hash.outputs.madara-hash }}"
           echo "orchestrator-binary-hash: ${{ steps.generate-binary-hash.outputs.orchestrator-hash }}"
           echo "cairo-artifacts-hash: ${{ steps.generate_cairo_hash.outputs.hash }}"
+          echo "bootstrapper-artifacts-hash: ${{ steps.generate_bootstrapper_artifacts_hash.outputs.hash }}"
```
raynaudoe: also add here `generate-binary-hash.outputs.bootstrapper-hash`

.github/workflows/task-coverage-bootstrapper.yml
```diff
@@ -0,0 +1,129 @@
...
+            echo "Config file NOT found at specified path!"
+            ls -la $(pwd)/configs/presets/ || echo "Directory not found!"
+          fi
+      - name: Download Bootstrapper artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: bootstrapper-artifacts-${{ inputs.bootstrapper-artifacts-hash }}
+          path: artifacts
+
+      - name: Start Anvil with fork
```
raynaudoe, line 90: The final step `Kill Anvil` (for resources cleanup) is missing. You can copy it from other actions,

.github/workflows/task-lint-cargo.yml
```diff
@@ -22,8 +27,26 @@ jobs:
...
+        run: |
+          npm install -g ganache@7.9.0
+          ganache --version
+
+      - name: Install Solc
+        run: |
+          pip install solc-select
+          solc-select install 0.8.19
+          solc-select use 0.8.19
+          solc --version
```
raynaudoe: Use `install-solc `and `install-ganache` instead

raynaudoe (changes requested):
Looking good! Left a few comments and some change requests

Makefile
Mohiiit: I will like to tackle this separately, given we have the CI working, it would be super easy to test it as well, wdyt?

here is the issue: https://github.com/madara-alliance/madara/issues/572

Makefile
Mohiiit: you mean the submodule should be in `madara-boostrapper/lib`?

bootstrapper/crates/bootstrapper/Cargo.toml
Mohiiit: I am not sure, it was there before me xD, but anyway we have to remove ethers and use alloy, so this would be removed when we will be doing the boostrapper refactor

bootstrapper/crates/bootstrapper/src/contract_clients/legacy_class.rs
Mohiiit: there would be a way soon! for now, we can keep it separate, but ideally yes that's how we can leverage the monorepo's true power

bootstrapper/crates/bootstrapper/Cargo.toml
Mohiiit: I am open to any, I don't have a strong opinion here, let me know wdyt?

.github/workflows/task-build-madara.yml
Mohiiit: the build artifacts of bootstrapper uses some docker files, hence this is required

Makefile
ftheirs: yeap, that one! I don't know if it's better to add the git submodule update inside the makefile to fully automate the `make artifacts` command

bootstrapper/crates/bootstrapper/Cargo.toml
ftheirs: I think that it would be more concise if we use `bootstrapper` now that we have all the pieces in the monorepo. Besides, orchestrator binary  is just `orchestrator`
cc: @HermanObst

Makefile
Mohiiit: hmm, we can discuss about it, I don't have strong opinions about it, I will read about the best practices, but we can keep it in the same effort when we make the Makefile changes?

ftheirs (approved):
LGTM

raynaudoe (approved):

Makefile
HermanObst, line 1: @Mohiiit does it makes sense to have this Makefile under bootstrapper directory?

cc @raynaudoe @ftheirs

bootstrapper/crates/bootstrapper/Cargo.toml
HermanObst: Yup, agree with @ftheirs

lib/argent-contracts-starknet
HermanObst, line 1: @Mohiiit Shouldn't all the lib be under orchestrator?
cc @ftheirs @raynaudoe

HermanObst (commented):
looking great, just quick comments

lib/argent-contracts-starknet
Mohiiit: yeah, @ftheirs pointed out the same and I have created the issue regarding the same, it would a one time effort to move the submodules and the makefile to the respective folders

PR #559, opened 2025-04-03
Processing lock on Proving job

heemankv:
This PR introduces `Processing Locks` for `Proving Job` similar to locks that are available on `SNOS Job`.


orchestrator/crates/orchestrator/src/cli/service.rs
```diff
@@ -13,4 +13,8 @@ pub struct ServiceCliArgs {
     /// The maximum number of SNOS jobs to process concurrently.
     #[arg(env = "MADARA_ORCHESTRATOR_MAX_CONCURRENT_SNOS_JOBS", long, default_value = Some("1"))]
```
Mohiiit: let's not have the default values? by default orch shouldn't be bound? wdyt?

orchestrator/crates/orchestrator/src/cli/service.rs
```diff
@@ -13,4 +13,8 @@ pub struct ServiceCliArgs {
...
     pub max_concurrent_snos_jobs: Option<usize>,
+
+    /// The maximum number of proving jobs to process concurrently.
+    #[arg(env = "MADARA_ORCHESTRATOR_MAX_CONCURRENT_PROVING_JOBS", long, default_value = Some("1"))]
```
Mohiiit: same here

orchestrator/crates/orchestrator/src/config.rs
```diff
@@ -168,7 +169,12 @@ pub async fn init_config(run_cmd: &RunCmd) -> color_eyre::Result<Arc<Config>> {
 
     let snos_processing_lock =
         JobProcessingState::new(orchestrator_params.service_config.max_concurrent_snos_jobs.unwrap_or(1));
-    let processing_locks = ProcessingLocks { snos_job_processing_lock: Arc::new(snos_processing_lock) };
+    let proving_processing_lock =
+        JobProcessingState::new(orchestrator_params.service_config.max_concurrent_proving_jobs.unwrap_or(1));
```
Mohiiit: let's remove unwrap here and if there is none value, let's allow orch to have multiple jobs?

Mohiiit (commented):
let's change the default limits, ideally should be unbounded when given none

orchestrator/crates/orchestrator/src/config.rs
Trantorian1: Yes , it is also strange that there is no coherence between this and the default value specified using `Clap`. In fact, since the default is `Some(1)`, this means it will never be possible for this `unwrap` to execute.

We should either have `None` as the default, in which case we assume an infinite upper bound, or serialize this as a usize directly.

The same applies to `max_concurrent_snos_jobs`.

orchestrator/crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -230,8 +231,8 @@ impl Job for ProvingJob {
 
     fn job_processing_lock(
         &self,
-        _config: Arc<Config>,
+        config: Arc<Config>,
     ) -> std::option::Option<std::sync::Arc<helpers::JobProcessingState>> {
-        None
+        Some(config.processing_locks().proving_job_processing_lock.clone())
```
Trantorian1: Imo we should allow for some state where this function returns `None` if a processing lock has not been set.

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -250,7 +250,12 @@ impl TestConfigBuilder {
 
         let snos_processing_lock =
             JobProcessingState::new(params.orchestrator_params.service_config.max_concurrent_snos_jobs.unwrap_or(1));
-        let processing_locks = ProcessingLocks { snos_job_processing_lock: Arc::new(snos_processing_lock) };
+        let proving_processing_lock =
+            JobProcessingState::new(params.orchestrator_params.service_config.max_concurrent_proving_jobs.unwrap_or(1));
```
Trantorian1: Again, if we have a global default for the number of max concurrent jobs, it would be nice for that to be reflected in a global static variable which is shared by `clap` and unwrap logic like this. Might have to represent it as a macro though since I don't know if clap does direct text insertion of the default value or if it is ok to reference a variable.

Trantorian1 (commented):
Looks good but I agree with @Mohiiit that the default behavior probably shouldn't be to have a hard bound of the number of concurrent tasks. This goes for the existing options, which I think should be updated as well. lmk what you think.

orchestrator/crates/orchestrator/src/cli/service.rs
heemankv: Alright, changed to not have any value by default.

orchestrator/crates/orchestrator/src/config.rs
heemankv: Updated.

orchestrator/crates/orchestrator/src/jobs/proving_job/mod.rs
heemankv: So, there can not be a scenario where `processing_lock` is not set as it is a part of config. Although I have changed this code to return `config.processing_locks().proving_job_processing_lock.clone()` (without the Some() ) so this logic is abstracted from this function now.

This fn now is simply serving the purpose of selecting which processing lock this job needs to use.

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: updated.

orchestrator/crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use alloy::signers::k256::pkcs8::der::oid::db::rfc4519::SN;
```
Mohiiit: is this required?

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -250,7 +250,12 @@ impl TestConfigBuilder {
...
+        let processing_locks = ProcessingLocks {
+            snos_job_processing_lock: Arc::new(snos_processing_lock),
```
Mohiiit: isn't `snos_job_processing_lock` an `Option`?

orchestrator/crates/orchestrator/src/jobs/proving_job/mod.rs
Trantorian1: Thanks for the clarification!

orchestrator/crates/orchestrator/src/config.rs
```diff
@@ -166,9 +167,16 @@ pub async fn init_config(run_cmd: &RunCmd) -> color_eyre::Result<Arc<Config>> {
...
-        JobProcessingState::new(orchestrator_params.service_config.max_concurrent_snos_jobs.unwrap_or(1));
-    let processing_locks = ProcessingLocks { snos_job_processing_lock: Arc::new(snos_processing_lock) };
+    let mut processing_locks = ProcessingLocks::default();
+
+    if let Some(max_concurrent_snos_jobs) = orchestrator_params.service_config.max_concurrent_snos_jobs {
+        processing_locks.snos_job_processing_lock = Some(Arc::new(JobProcessingState::new(max_concurrent_snos_jobs)));
+    }
+
+    if let Some(max_concurrent_proving_jobs) = orchestrator_params.service_config.max_concurrent_proving_jobs {
+        processing_locks.proving_job_processing_lock =
```
Trantorian1, line 177: What value does this default to then?

Trantorian1 (approved):
Looks good. @Mohiiit has left a few comments which I think need addressing, in particular [this one](https://github.com/madara-alliance/madara/pull/559/files#r2029255545). Other than that I didn't notice anything else myself.

orchestrator/crates/orchestrator/src/jobs/proving_job/mod.rs
heemankv: My apologies, this shouldn't have been pushed.
have fixed now.

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: Yeah, it's resolved with the latest commit.

Mohiiit (approved):

0xvasanth (approved):

Mohiiit (approved):

PR #562, opened 2025-04-03
fix: Orchestrator test workflow

Mohiiit:
Fixing the orchestrator test workflow: the Madara binary was at the wrong path, causing test failures. Also ensuring errors don’t go unnoticed by stopping on failure.

example: https://github.com/madara-alliance/madara/actions/runs/14243653759/job/39920216960?pr=526

Screenshot 2025-04-03 at 7 47 26 PM


raynaudoe (approved):

orchestrator/crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -167,6 +168,7 @@ async fn test_settle(#[future] setup: (LocalWalletSignerMiddleware, MadaraCmd))
 
     let is_success = wait_for_tx(&account, deploy_tx_hash, Duration::from_secs(2)).await;
     assert!(is_success, "Deploy trasaction failed");
+    sleep(Duration::from_secs(10));
```
raynaudoe: wouldn't be the same effect increasing the duration above?

Mohiiit: I am taking some time after each transaction so that the block get mined, I am not sure this is the ideal behaviour tbh, but I assumed this might be the case hence I added a delay of 10 secs which is also the block_time and it worked, I am also investigating it as we speak

.env.test
```diff
@@ -80,8 +80,8 @@ MADARA_ORCHESTRATOR_STARKNET_OPERATOR_ADDRESS=0x5b98B836969A60FEC50Fa925905Dd1D3
 ## STARKNET ##
 
 MADARA_ORCHESTRATOR_STARKNET_SETTLEMENT_RPC_URL=https://starknet-sepolia.public.blastapi.io
-MADARA_ORCHESTRATOR_STARKNET_PRIVATE_KEY=0x76f2ccdb23f29bc7b69278e947c01c6160a31cf02c19d06d0f6e5ab1d768b86
-MADARA_ORCHESTRATOR_STARKNET_ACCOUNT_ADDRESS=0x3bb306a004034dba19e6cf7b161e7a4fef64bc1078419e8ad1876192f0b8cd1
+MADARA_ORCHESTRATOR_STARKNET_PRIVATE_KEY=0x077e56c6dc32d40a67f6f7e6625c8dc5e570abe49c0a24e9202e4ae906abcc07
```
ftheirs, line 84: I assume that the priv key is just a random value but if you can add a comment on how to get/compute the address and the private key would be awesome! 😃

ftheirs (approved):

.env.test
Mohiiit: will do, it's just the first key account pair when madara is running as devnet, but will add the comment over there

Mohiiit: added

orchestrator/crates/settlement-clients/starknet/src/tests/test.rs
Mohiiit: it's resolved now

ftheirs (approved):

raynaudoe (approved):

PR #567, opened 2025-04-08
Add resource readiness checks with polling mechanism

prkpndy:
Introduced methods to verify the existence of resources (queues, data storage, alerts, cron) and added a polling mechanism to ensure readiness during setup. This avoids redundant resource creation and enhances setup reliability by handling resource initialization status effectively.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, there are hardcoded delays before and after creating any resource on cloud

Resolves: #NA

## What is the new behavior?

Now, it polls the cloud services to check if the resource and its dependents are ready

NO

### DIFF
--- a/orchestrator/crates/orchestrator/src/alerts/aws_sns/mod.rs
+++ b/orchestrator/crates/orchestrator/src/alerts/aws_sns/mod.rs
@@ -3,6 +3,7 @@ use aws_config::SdkConfig;
 use aws_sdk_sns::Client;
 
 use crate::alerts::Alerts;
+use crate::setup::ResourceStatus;
 
 #[derive(Debug, Clone)]
 pub struct AWSSNSValidatedArgs {
@@ -38,5 +39,18 @@ impl Alerts for AWSSNS {
 
     async fn get_topic_name(&self) -> String {
         self.topic_arn.split(":").last().expect("Failed to get last part of topic ARN").to_string()
+    }
+
+    async fn exists(&self) -> color_eyre::Result<bool> {
+        let topic_name = self.get_topic_name().await;
+        Ok(self.client.get_topic_attributes().topic_arn(topic_name).send().await.is_ok())
+    }
+}
+
+#[allow(unreachable_patterns)]
+#[async_trait]
+impl ResourceStatus for AWSSNS {
+    async fn are_all_ready(&self) -> bool {
+        self.exists().await.is_ok()
     }
 }
--- a/orchestrator/crates/orchestrator/src/alerts/mod.rs
+++ b/orchestrator/crates/orchestrator/src/alerts/mod.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use mockall::automock;
+use color_eyre::Result;
 
 pub mod aws_sns;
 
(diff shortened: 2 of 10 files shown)


orchestrator/crates/orchestrator/src/alerts/mod.rs
```diff
@@ -1,18 +1,24 @@
 use async_trait::async_trait;
 use mockall::automock;
+use color_eyre::Result;
 
 pub mod aws_sns;
 
 #[automock]
 #[async_trait]
 pub trait Alerts: Send + Sync {
```
0xvasanth, line 9:
```suggestion
pub trait Alerts: ResourceStatus {
```

orchestrator/crates/orchestrator/src/setup/mod.rs
```diff
@@ -18,15 +18,69 @@ use crate::data_storage::aws_s3::AWSS3;
...
+    #[strum(serialize = "data_storage")]
+    DataStorage,
+    #[strum(serialize = "cron")]
+    Cron,
+    #[strum(serialize = "alerts")]
+    Alerts,
+}
+
+#[async_trait]
+pub trait ResourceStatus {
```
0xvasanth, line 41:
```suggestion
pub trait ResourceStatus: Send + Sync {
```

orchestrator/crates/orchestrator/src/setup/mod.rs
```diff
@@ -39,13 +93,14 @@ pub async fn setup_cloud(setup_cmd: &SetupCmd) -> color_eyre::Result<()> {
     match queue_params {
         QueueValidatedArgs::AWSSQS(aws_sqs_params) => {
             let sqs = Box::new(SqsQueue::new_with_args(aws_sqs_params, aws_config));
-            sqs.setup().await?
+            sqs.setup().await?;
+            if poll(&*sqs, ResourceType::Queues, setup_cmd.poll_interval.unwrap(), setup_cmd.timeout.unwrap()).await {
+                println!("Queues setup completed ✅");
+            } else {
+                println!("Queues setup failed ❌");
+            }
```
0xvasanth, line 101: Move this check before eventbridge / Cron so we can start the storage create parallel which the queue creation is in Progress

0xvasanth (changes requested):

orchestrator/crates/orchestrator/src/alerts/aws_sns/mod.rs
```diff
@@ -39,4 +40,17 @@ impl Alerts for AWSSNS {
...
         self.topic_arn.split(":").last().expect("Failed to get last part of topic ARN").to_string()
     }
+
+    async fn exists(&self) -> color_eyre::Result<bool> {
+        let topic_name = self.get_topic_name().await;
+        Ok(self.client.get_topic_attributes().topic_arn(topic_name).send().await.is_ok())
+    }
 }
+
+#[allow(unreachable_patterns)]
```
0xvasanth, line 50: The `exists` method in the `QueueProvider` trait implementation for `SqsQueue` returns a `bool`, but the `are_all_ready` method in the `ResourceStatus` trait implementation for `SqsQueue` calls `exists` and expects a `bool` directly. This can lead to a panic if `exists` returns an `Err`.

0xvasanth: same Comment will be applicable for other resource too

0xvasanth (changes requested):

orchestrator/crates/orchestrator/src/alerts/mod.rs
0xvasanth: Inheritance of the trait is completely broken as we spoken over the call

Mohiiit:
shall we close this? @prkpndy @itsparser

PR #569, opened 2025-04-08
fix: trace api match spec

apoorvsadana:
Fixes the `traceTransaction` RPC result to match the spec


Trantorian1 (approved):
Looks good, thanks for pointing this out! :+1:

Mohiiit (approved):

PR #585, opened 2025-04-18
 fix: nix development environment

jbcaron:
## Pull Request type

- Build-related changes

Resolves: #NA

## What is the new behavior?

-

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ clean-db:
 		[yY]*) true;; \
 		*) false;; \
 	esac
-	@make --silent clean
+	@$(MAKE) --silent clean
 	@echo -e "$(DIM)removing madara database on host$(RESET)"
 	@rm -rf $(DB_PATH);
 
@@ -149,11 +149,11 @@ fclean: clean-db
 
 .PHONY: restart
 restart: clean
-	@make --silent start
+	@$(MAKE) --silent start
 
 .PHONY: frestart
 frestart: fclean
-	@make --silent start
+	@$(MAKE) --silent start
 
 
 .PHONY: snos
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -137,7 +137,7 @@ clean-db:
 		[yY]*) true;; \
 		*) false;; \
 	esac
-	@make --silent clean
+	@$(MAKE) --silent clean
```
Trantorian1, line 140: What is the difference with this?

flake.nix
```diff
@@ -35,11 +35,16 @@
             openssl
             pkg-config
             protobuf
+            nodejs
```
Trantorian1, line 38: Why is `nodejs` a dependency now?

flake.nix
```diff
@@ -52,6 +57,33 @@
               darwin.apple_sdk.frameworks.Security
             ];
 
+          shellHook = ''
+            # Workaround: the starkgate-contracts setup script attempts to globally install npm packages
+            # (ganache, prettier, etc.), which fails in a Nix environment due to read-only /nix/store.
+            # To fix this, we redirect npm's global prefix to a writable local directory (.npm-global),
+            # making global installs succeed and allowing `npm list -g` checks to pass.
```
Trantorian1: Maybe its worth linking to [this section](https://nixos.wiki/wiki/Node.js#Install_to_your_home) of the NixOs wiki as well?

Trantorian1 (approved):

Makefile
jbcaron: The difference is that `$(MAKE)` is a special variable in Makefiles that expands to the exact make command used to invoke the current Makefile

flake.nix
Mohiiit: we use ganache for compiling the old solidity contracts and ganache uses nodejs

flake.nix
jbcaron: done

Mohiiit (approved):

PR #593, opened 2025-04-21
add: REST endpoint to check if the node is ready to serve traffic

prkpndy:
The endpoint will return a 200 status code if ready, otherwise it'll send 400. 500 will be returned for internal server error. This endpoint can be used for the readiness probe in K8s

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

No such endpoint

Resolves: #NA

## What is the new behavior?

A new endpoint - `/ready` - which can be used by K8s as a readiness probe

No


madara/crates/node/src/service/rpc/server.rs
```diff
@@ -129,6 +134,22 @@ pub async fn start_server(
                             .body(hyper::Body::from("GONE"))?)
                     } else if req.uri().path() == "/health" {
                         Ok(hyper::Response::builder().status(hyper::StatusCode::OK).body(hyper::Body::from("OK"))?)
+                    } else if req.uri().path() == "/ready" {
+                        let sync_status = syncing(&starknet).await;
+                        match sync_status {
+                            Ok(sync_status) => match sync_status {
+                                SyncingStatus::Syncing(_) => Ok(hyper::Response::builder()
+                                    .status(hyper::StatusCode::BAD_REQUEST)
```
Trantorian1: Why are we using `StatusCode::BAD_REQUEST` here?

Shouldn't we be using something like [`StatusCode::TOO_EARLY`](https://docs.rs/http/latest/http/status/struct.StatusCode.html#associatedconstant.TOO_EARLY)?

madara/crates/node/src/service/rpc/server.rs
```diff
@@ -129,6 +134,22 @@ pub async fn start_server(
...
+                                    .body(hyper::Body::from("SYNCING"))?),
+                                SyncingStatus::NotSyncing => Ok(hyper::Response::builder()
+                                    .status(hyper::StatusCode::OK)
+                                    .body(hyper::Body::from("OK"))?),
+                            },
+                            Err(_) => Ok(hyper::Response::builder()
+                                .status(hyper::StatusCode::INTERNAL_SERVER_ERROR)
+                                .body(hyper::Body::from("INTERNAL_SERVER_ERROR"))?),
+                        }
+                        // Ok(hyper::Response::builder().status(hyper::StatusCode::OK).body(hyper::Body::from("OK"))?)
```
Trantorian1:
```suggestion
```

Trantorian1 (approved):

madara/crates/node/src/service/rpc/server.rs
prkpndy: I have updated it to 503 (service unavailable). This is how it's implemented in pathfinder and juno. Will this be good?

Trantorian1: Sounds good :)

Mohiiit (approved):

PR #594, opened 2025-04-21
refactor: submit-tx crate, split gateway-sequencer configuration support

cchudant:
Note: This is step 2 of getting the quaza perf changes back into madara main. It's a lot of file changes, but this is a specific situation and should not be considered a standard-size PR.

After this PR is landed, there will be only a misc-changes PR and a PR that only touches the block-production crate. The former one will be relatively contained and no architecture change, the latter is only a rewrite the block production crate - which is very few file changes in comparison.

In addition, I will open more testing PRs :)

## Pull Request type

- Feature

Also: Refactor

One of the central traits in the madara node is the AddTransactionProvider trait. This trait allows the gateway and rpc write api to send transactions to either the validation + then mempool or redirect them to another gateway.

There is in fact no difference on whether the transaction comes from gateway or rpc write api: every transaction goes through this central trait, and is dispatched to mempool or gateway depending on the node configuration.

Untitled-2025-02-04-1118(10)

This allowed madara to support all of these configurations, with minimal code, by just mixing-and-matching which implementor of AddTransactionProvider you give to the rpc/gateway implementations:
- accept transactions from the gateway interface and redirect them to another gateway [full-node mode]
- accept transactions from the rpc interface and redirect them to another gateway [full-node mode]
- accept transactions from the gateway interface and add them to the mempool, to be executed [sequencer]
- accept transactions from the rpc interface and add them to the mempool, to be executed [sequencer]

### 1) An awkward, leaky abstraction

While this trait is central to madara, it has always been in an awkward situation because:
- it is currently in the mc-rpc crate. This was temporary, because at the time only mc-rpc used that trait - gateway interface was not a thing yet.
- when gateway interface was implemented, it became clear that using the RPC error types for this trait is a very bad idea. They don't map well to the error types that the gateway inteface is supposed to return. This resulted in some validation errors marked as internal server errors, some clients getting the wrong error codes, and a whole other mess of issues.

### 2) Improving sequencer performance: split gateway-sequencer configuration 

When sending a transaction, before going into the mempool, there is a validation step. This validation step is quite expensive. This leads to a huge loss of performance if it is done in the sequencer. In fact, when running benchmarks, we at Karnot and KasarLabs usually disable this validation step because we suppose it is going to be removed from the sequencer at some point, so it shouldn't count. The time has come!

For context, here is the Starknet architecture from the [official starknet docs](https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-architecture-overview/):
image

As you can see, validation is done in the gateway and not the sequencer. This is how they avoid this expensive transaction validation: they don't run it on the most important part and most bottlenecked part of the whole system (sequencer), they run it on another machine.

Notes on what validation is: (skip this section unless you want to hear me nerd out about starknet architecture)
> Validation here refers to doing the pre-checks (using blockifier) on transactions, which includes running `__validate__` on the account of the tx sender. It is done before insertion to the mempool, and is very important as it allows the sequencer to avoid being very easily DoS-able.
> Validation is where signature verification is done in Argent accounts, for example. It is very limited: low cairo step limit, and you cannot call functions on other contracts. This is because the validation is made to be run on a slightly outdated version of the chain - you do not need to be entirely up-to-date with the current chain to run validation. Which makes sense - this is a pre-check before the transaction enters the mempool. The transactions may get re-ordered once inside the mempool, and they will only be included in a block much later. You don't want them to get invalidated while they are in the mempool, otherwise they will either be reverted, or worse, rejected (when revert is not possible).
> Rejected transactions are a huge problem; as you are spending mempool space and compute time for a transaction that you can't include into a block, thus you can't charge fees for it. This is really bad, and leads to all sorts of DoS opportunities.
> Declare and deployaccount txs cannot be reverted -which, if you're curious about why, this is also an interesting topic- and invoke transactions also cannot be in some cases, one of which is related to when the account doesn't even have enough fees to pay the reverted tx fees.
> More info here: https://docs.starknet.io/architecture-and-concepts/accounts/account-functions/#limitations_of_validation

## What is the new behavior?

This pr moves this trait into its own crate, and finally fixes all of the error conversion between rpc/gateway/mempool - introducing a new error type to rule them all. Implementors of the submit-tx trait implement their conversion from their error types to this common error trait, and users of the trait implement their conversion from this common error type to their usecase-specific error type. 

The new trait looks like this:

https://github.com/madara-alliance/madara/blob/39dad83ae578242c4b3c70c8142f7a6354583be5/madara/crates/client/submit_tx/src/lib.rs#L14-L42

In addition, transaction validation is removed from the mempool: it is now in the submit-tx crate as well. TransactionValidator is a struct that implements `SubmitTransaction`: you can submit transaction to it. The gateway client also implements `SubmitTransaction`, allowing the redirection of transactions to another gateway, like before.

The RPC and Gateway http services thus accept a `Arc<dyn SubmitTransaction>`, and you can give them whichever of the two you want : either GatewayClient or TransactionValidator. They are completely abstract.

In addition, there is a new trait called

https://github.com/madara-alliance/madara/blob/39dad83ae578242c4b3c70c8142f7a6354583be5/madara/crates/client/submit_tx/src/lib.rs#L54-L59

When instanciating TransactionValidator, you have to give it a `Arc<dyn SubmitValidatedTransaction>`, telling it where to put the validated txs. In most configuration, this would be the Mempool, which implements that trait.

However, in order to support split gateway-sequencer configuration, sometimes you also want to have a node (the gateway) validate the transactions, but redirect them to an other node (the sequencer). In order to do that, the gateway-server now has a new /madara/trusted_add_validated_transaction endpoint. This endpoint accepts already-validated transactions. It needs to be enabled via the `--gateway-trusted-add-transaction-endpoint` cli argument.

All of this just works by having `GatewayClient` also implement `SubmitValidatedTransaction`, and the `GatewayServer` accept an `Option<Arc<dyn SubmitValidatedTransaction>>` - thanks to these 2 traits we can now mix-and-match implementors/trait users to make the split gateway config work!

no db schema change, no cli arg removed

## Other information

- Validation has been optimized. This is because we're not loading the pending block info from the database anymore for every transaction - we have a cached version of it in the MadaraBackend through a tokio watch channel now.
- Class compilation is done in the TransactionValidator, as part of validation. This means that in split gateway-sequencer configuration, it's run in the gateway, not the sequencer.
- Rocksdb configuration is now more configurable via CLI. (for profiling)

### Sequencer e2e-testing.

Testing is quite important, and I think I found a good way to do it.

Since submit-tx is mostly about inter-node and touches the whole madara node architecure, I think it's better to focus on e2e tests:

I am currently writing a battery of tests here https://github.com/madara-alliance/madara/blob/submit-tx/madara/crates/tests/src/transaction_flow.rs#L22-L42
They are made to run in all of those three configurations:
- Sequencer only.
- Full node + sequencer.
- Gateway (validates txs) + sequencer.

The way I wrote those tests, you only need to write the test once, and it can run in any of those configurations. They also test the gateway and RPC interfaces at the same time.

Here is a test list to be included in this pr, to be tested across all the different node configurations:

Test list:
- [x] working invoke (transfer, check events + resulting state diff)
- [x] working deploy_account
- [x] working declare
- [x] invalid nonce: nonce too low [validation]
- [x] invalid nonce: nonce too low - error caught at mempool and not gateway [mempool]
- [x] duplicate tx in mempool [mempool]
- [x] invoke tx arrived before deploy_account tx [validation]
- [x] test with more transfers, over multiple blocks (state should correctly roll over new blocks)

A more in depth list is present in #608 - the rest of the tests will be included in a later PR.


madara/crates/client/db/src/block_db.rs
```diff
@@ -282,7 +281,7 @@ impl MadaraBackend {
                     receipt.events().iter().map(move |event| (tx_hash, event))
                 })
                 .for_each(|(transaction_hash, event)| {
-                    if let Err(e) = self.sender_event.publish(EmittedEvent {
+                    if let Err(e) = self.events_watch.publish(EmittedEvent {
```
Trantorian1: The reason why a `tokio::sync::broadcast` channel was being used was to avoid v0.8 websocket RPC methods missing an event update as they are catching up with the sync tip as part of calls to `starknet_subscribeEvents`. Are we sure this is still guaranteed when using a `tokio::sync::watch` channel? I don't think this had any tests either so it would not be detected (this kind of thing is a pain to test as it can only happen in very specific cases which involve the db and RPCs getting out-of-sync mid-response).

madara/crates/client/db/src/block_db.rs
```diff
@@ -399,7 +390,7 @@ impl MadaraBackend {
                 Ok(Some((info.into(), TxIndex(tx_index as _))))
             }
             None => {
-                let info = self.get_pending_block_info()?;
```
Trantorian1, line 402: Nice! If I remember correctly you were doing some perf optimization which led to the decision of removing this: do we have any benchmarks regarding improvements? I seemed to remember this constant db fetch used to be a pretty important bottleneck.

madara/crates/client/db/src/lib.rs
```diff
@@ -312,150 +319,15 @@ impl Default for TrieLogConfig {
...
     backup_handle: Option<mpsc::Sender<BackupRequest>>,
     db: Arc<DB>,
     chain_config: Arc<ChainConfig>,
     db_metrics: DbMetrics,
     snapshots: Arc<Snapshots>,
-    sender_block_info: tokio::sync::broadcast::Sender<mp_block::MadaraBlockInfo>,
     head_status: ChainHead,
-    sender_event: EventChannels,
+    events_watch: EventChannels,
+    watch: BlockWatch,
```
Trantorian1: _(nitpick)_ Maybe we should rename this to `watch_blocks` and `events_watch` to `watch_events`?

madara/crates/client/db/src/lib.rs
```diff
@@ -543,8 +414,11 @@ pub struct MadaraBackendConfig {
     pub trie_log: TrieLogConfig,
     pub backup_every_n_blocks: Option<u64>,
     pub flush_every_n_blocks: Option<u64>,
+    pub rocksdb: RocksDBConfig,
     #[cfg(any(test, feature = "testing"))]
-    pub _temp_dir: Option<tempfile::TempDir>,
+    pub temp_dir: Option<tempfile::TempDir>,
+    #[cfg(not(any(test, feature = "testing")))]
+    pub temp_dir: Option<()>,
```
Trantorian1: Why are we still using a `temp_dir` outside of a testing context? I guess this is a zst so no big deal but still feels strange?

madara/crates/client/db/src/mempool_db.rs
```diff
@@ -54,30 +54,19 @@ impl Default for NonceInfo {
...
-    pub arrived_at: u128,
-}
-
 #[derive(Serialize)]
 /// This struct is used as a template to serialize Mempool transactions from the
 /// database without any further allocation.
 struct DbMempoolTxInfoEncoder<'a> {
-    saved_tx: &'a SavedTransaction,
-    converted_class: &'a Option<ConvertedClass>,
+    tx: &'a ValidatedMempoolTx,
```
Trantorian1, line 61: Nice!

madara/crates/client/db/src/mempool_db.rs
```diff
@@ -94,33 +83,33 @@ impl MadaraBackend {
...
 
-    #[tracing::instrument(skip(self), fields(module = "MempoolDB"))]
-    pub fn remove_mempool_transaction(&self, tx_hash: &Felt) -> Result<()> {
+    pub fn remove_mempool_transactions(&self, tx_hashes: impl IntoIterator<Item = Felt>) -> Result<()> {
         // Note: We do not use WAL here, as it will be flushed by saving the block. This is to
         // ensure saving the block and removing the tx from the saved mempool are both done at once
         // atomically.
 
         let col = self.db.get_column(Column::MempoolTransactions);
-        self.db.delete_cf_opt(&col, bincode::serialize(tx_hash)?, &self.writeopts_no_wal)?;
```
Trantorian1, line 104: Would this not have been transactional, even with the `writeopts_no_wal`? I guess it could have been possible for the db to be flushed with only part of the transactions, was that the issue here?

madara/crates/client/db/src/rocksdb_options.rs
```diff
@@ -9,7 +9,34 @@ const KiB: usize = 1024;
...
+#[derive(Debug, Clone)]
+pub struct RocksDBConfig {
+    pub enable_statistics: bool,
+    pub statistics_period_sec: u32,
+    pub statistics_level: StatsLevel,
+    pub memtable_blocks_budget_mib: usize,
+    pub memtable_contracts_budget_mib: usize,
+    pub memtable_other_budget_mib: usize,
+    pub memtable_prefix_bloom_filter_ratio: f64,
+}
```
Trantorian1: It might be a good idea to add some comments here to explain what each of these configuration options do.

madara/crates/client/db/src/rocksdb_options.rs
```diff
@@ -25,6 +52,14 @@ pub fn rocksdb_global_options() -> Result<Options> {
     options.set_keep_log_file_num(3);
     options.set_log_level(rocksdb::LogLevel::Warn);
 
+    // I think this one if a column-specific option, but I'm not entirely sure. So let's just also set it here.
+    options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
```
Trantorian1: It does seem like its a cf-specific option. Here is the reference file from the RocksDB 
- [`include/rocksdb/advanced_options.h`]
- [`include/rocksdb/options.h`] (not directly related but has some good docs for other options)

[`include/rocksdb/advanced_options.h`]: https://github.com/facebook/rocksdb/blob/c237022831aa129aa707bc28e0702a1617ef23b5/include/rocksdb/advanced_options.h#L325
[`include/rocksdb/options.h`]: https://github.com/facebook/rocksdb/blob/c237022831aa129aa707bc28e0702a1617ef23b5/include/rocksdb/options.h#L68

Trantorian1: You can find some more info on all the `DB:Open` options [here](https://github.com/facebook/rocksdb/blob/c237022831aa129aa707bc28e0702a1617ef23b5/include/rocksdb/options.h#L1653).

madara/crates/client/db/src/rocksdb_options.rs
```diff
@@ -37,35 +72,34 @@ pub fn rocksdb_global_options() -> Result<Options> {
...
+            options.set_prefix_extractor(SliceTransform::create_fixed_prefix(prefix_extractor_len));
+            options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
         }
 
         options.set_compression_type(DBCompressionType::Zstd);
         match self {
             Column::BlockNToBlockInfo | Column::BlockNToBlockInner => {
-                options.optimize_universal_style_compaction(1 * GiB);
+                options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
+                options.optimize_universal_style_compaction(config.memtable_blocks_budget_mib);
```
Trantorian1, line 103: Why was this changed to something dynamic? Just asking because I'm curious the cases in which this would have an impact. As far as I can remember, this was set to resolve some RAM consumption issues due to compaction being to sloow, but I think this is still not fixed?

madara/crates/client/db/src/rocksdb_options.rs
```diff
@@ -37,35 +72,34 @@ pub fn rocksdb_global_options() -> Result<Options> {
...
+            }
+            Column::ContractStorage | Column::ContractToClassHashes | Column::ContractToNonces => {
+                options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
+                options.optimize_universal_style_compaction(config.memtable_contracts_budget_mib);
             }
             _ => {
-                options.optimize_universal_style_compaction(100 * MiB);
```
Trantorian1, line 68: Same here

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -123,12 +134,14 @@ impl MadaraBackend {
         let mut inner = MadaraBlockInner { transactions, receipts };
         store_events_to_receipts(&mut inner.receipts, block.events)?;
 
-        self.block_db_store_pending(&MadaraPendingBlock { info, inner }, &block.state_diff)?;
+        self.block_db_store_pending(&MadaraPendingBlock { info: info.clone(), inner }, &block.state_diff)?;
         self.contract_db_store_pending(ContractDbBlockUpdate::from_state_diff(block.state_diff))?;
+
+        self.watch.update_pending(info.into());
```
Trantorian1: Same question here as with the even `tokio::sync::watch` channel: are we sure there is no scenario possible where RPC methods (in particular websocket methods) might be missing some updates because of this?

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -277,6 +292,19 @@ impl MadaraBackend {
     }
 
     pub fn clear_pending_block(&self) -> Result<(), MadaraStorageError> {
+        let parent_block = if let Some(block_n) = self.get_latest_block_n()? {
+            Some(
+                self.get_block_info(&DbBlockId::Number(block_n))?
+                    .ok_or(MadaraStorageError::InconsistentStorage("Can't find block info".into()))?
+                    .into_closed()
+                    .ok_or(MadaraStorageError::InconsistentStorage(
```
Trantorian1, line 300: I do wonder if there ever would be a way to enforce this at the type level. We sure have a lot of these checks throughout the codebase...

madara/crates/client/db/src/stream.rs
```diff
@@ -99,13 +99,13 @@ impl MadaraBackend {
         struct State {
             backend: Arc<MadaraBackend>,
             iteration: BlockStreamConfig,
-            buf: VecDeque<MadaraBlockInfo>,
+            buf: VecDeque<Arc<MadaraBlockInfo>>,
```
Trantorian1, line 102: Why the move to an `Arc`?

madara/crates/client/db/src/watch.rs
```diff
@@ -0,0 +1,83 @@
+use crate::{MadaraBackend, MadaraStorageError};
+use mp_block::{header::PendingHeader, MadaraBlockInfo, MadaraPendingBlockInfo};
+use std::sync::Arc;
+use tokio::sync::{broadcast, watch};
+
+pub type ClosedBlocksReceiver = broadcast::Receiver<Arc<MadaraBlockInfo>>;
+pub type PendingBlockReceiver = watch::Receiver<Arc<MadaraPendingBlockInfo>>;
+
+fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<MadaraPendingBlockInfo> {
```
Trantorian1, line 9: Petition to rename this to `pending_block_trenchcoat` X)

madara/crates/client/db/src/watch.rs
```diff
@@ -0,0 +1,83 @@
...
+            protocol_version: parent_block.header.protocol_version,
+            l1_gas_price: parent_block.header.l1_gas_price.clone(),
+            l1_da_mode: parent_block.header.l1_da_mode,
+        },
+        tx_hashes: vec![],
+    }
+    .into()
+}
+
+pub(crate) struct BlockWatch {
```
Trantorian1, line 27: So if I understand this correctly we use this struct to replace the _many_ calls we did to the pending block in db with a _single_ call on pending update, the result of which we store into a `tokio::sync::watch` channel. If no value is present we default to dummy pending block. Is that correct?

madara/crates/client/exec/src/execution.rs
```diff
@@ -124,6 +134,40 @@ impl TxInfo for Transaction {
         }
     }
 
+    // FIXME: fix this, this is wrong for L1HandlerTxs.
```
Trantorian1, line 137: How is this different for `L1HandlerTxs`?

madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -328,9 +329,27 @@ pub async fn handle_get_public_key(backend: Arc<MadaraBackend>) -> Result<Respon
     Ok(create_string_response(hyper::StatusCode::OK, format!("\"{:#x}\"", public_key)))
 }
 
+pub async fn handle_add_validated_transaction(
```
Trantorian1, line 335: Should this not be feature-gated behind `#[cfg(feature = "add_validated_transaction")]` as well?

madara/crates/client/gateway/server/src/service.rs
```diff
@@ -1,37 +1,53 @@
...
 use tokio::net::TcpListener;
 
-use super::router::main_router;
+#[derive(Debug, Clone)]
+pub struct GatewayServerConfig {
+    pub feeder_gateway_enable: bool,
+    pub gateway_enable: bool,
+    pub gateway_external: bool,
+    pub gateway_port: u16,
+    pub enable_trusted_add_validated_transaction: bool,
```
Trantorian1, line 20: I'm a bit confused now as to weather this is feature-gated or config-gated?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -133,78 +185,54 @@ impl Mempool {
...
-                .get_block_hash(&BlockId::Tag(BlockTag::Latest))?
-                .unwrap_or(/* genesis block's parent hash */ Felt::ZERO);
-            MadaraPendingBlockInfo::new(
-                make_pending_header(parent_block_hash, self.backend.chain_config(), self.l1_data_provider.as_ref()),
-                vec![],
-            )
-            .into()
-        };
+        // TODO: should we update this to store only if the mempool accepts
+        // this transaction?
```
Trantorian1, line 196: Yep! The main question is: do we have a way to remove rejected transactions from the db? Because as far as I can remember, we only remove transactions from the mempool column family during the block prod when they are polled, but they cannot be polled if they are rejected by the mempool! Are we sure we have a mechanism to remove them?

Trantorian1: (Well, this was my old comment, this is why I put it there, curious to know what you think haha)

madara/crates/client/mempool/src/lib.rs
```diff
@@ -445,18 +334,17 @@ impl MempoolProvider for Mempool {
         }
     }
 
-    #[tracing::instrument(skip(self, contract_address), fields(module = "Mempool"))]
+    // #[tracing::instrument(skip(self, contract_address), fields(module = "Mempool"))]
```
Trantorian1: ?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -268,155 +296,16 @@ impl Mempool {
...
-            Some(nonce) => nonce.try_increment()?,
-            None => Nonce(Felt::ZERO),
-        };
-
-        match nonce.cmp(&target_nonce) {
-            std::cmp::Ordering::Less => Err(MempoolError::StorageError(MadaraStorageError::InvalidNonce)),
-            std::cmp::Ordering::Equal => Ok(NonceInfo::ready(nonce, nonce_next)),
-            std::cmp::Ordering::Greater => Ok(NonceInfo::pending(nonce, nonce_next)),
-        }
-    }
```
Trantorian1: Does this mean L1 handler transactions are handled the same as other transaction or that because of the way in which we have separated validation from the mempool then we are never having to process l1 handler transactions here? If the latter is the case I think we should add a runtime `assert` to `accept_validated_tx` which makes sure this is the case.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -268,155 +296,16 @@ impl Mempool {
...
-}
-
-fn deployed_contract_address(tx: &Transaction) -> Option<Felt> {
-    match tx {
-        Transaction::AccountTransaction(AccountTransaction::DeployAccount(tx)) => Some(**tx.contract_address),
-        _ => None,
-    }
 }
 
+#[async_trait]
```
Trantorian1: Why the `async_trait` if we do not have any `async` methods?

madara/crates/client/mempool/src/lib.rs
Trantorian1, line 1: Really nice changes, the outer mempool code feels much clearner and easier to follow now that validation logic has been hoisted out :)

madara/crates/client/rpc/src/errors.rs
```diff
@@ -198,42 +268,107 @@ impl From<StarknetError> for StarknetRpcApiError {
...
+
+            E::EntryPointNotFound
+            | E::TransactionFailed
+            | E::OutOfRangeTransactionHash
+            | E::UnsupportedSelectorForFee
+            | E::TransactionLimitExceeded
+            | E::OutOfRangeFee
+            | E::OutOfRangeContractAddress
+            | E::InvalidSignature
+            | E::ValidateFailure // this might be a ContractError? TxnExecutionError?
```
Trantorian1, line 337: What do you mean by "this might be a ContractError? TxnExecutionError?"?

madara/crates/client/rpc/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 
 mod constants;
 mod errors;
-pub mod providers;
+// pub mod providers;
```
Trantorian1:
```suggestion
```

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -302,9 +302,10 @@ async fn process_message(
...
         _ => {}
     };
-    let res = mempool
-        .tx_accept_l1_handler(transaction.into(), fees.unwrap_or(0))
-        .map_err(|e| SettlementClientError::Mempool(format!("Failed to accept transaction in mempool: {}", e)))?;
+    let res = submit_tx
+        .submit_l1_handler_transaction(transaction.into(), fees.unwrap_or(0))
+        .await
+        .map_err(|e| SettlementClientError::SubmitTx(format!("Failed to accept transaction in mempool: {e:#}")))?;
     // HERMAN TODO: Actually this should be updated after the tx l1 handler is executed
```
Trantorian1, line 309: This should no longer be a problem if we are not processing l1 handler transactions in the mempool.

madara/crates/client/submit_tx/src/error.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Currently only returned when trying to add a validated transaction to a gateway that doesn't support or allow it.
+    #[error("Unsupported operation")]
+    Unsupported,
+    /// Validation failed, or any other expected error.
+    #[error("Transaction rejected: {0}")]
+    Rejected(#[from] RejectedTransactionError),
+    /// Any internal error. Note that when redirecting transactions from one node to another,
+    /// any transport/connectivity/unexpected gateway errors will appear as internal here.
+    #[error("Internal error: {0:#}")]
+    Internal(anyhow::Error),
```
Trantorian1, line 14: Out of curiosity, why are we using `anyhow::Error` here vs a `String`?

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+impl From<TransactionPreValidationError> for SubmitTransactionError {
+    fn from(err: TransactionPreValidationError) -> Self {
+        use RejectedTransactionErrorKind::*;
+        use TransactionPreValidationError as E;
+
+        match err {
+            E::InvalidNonce { .. } => rejected(InvalidTransactionNonce, format!("{err:#}")),
+            // TODO: Some of those are internal server errors.
```
Trantorian1, line 39: In which case would these be internal server errors?

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+impl From<StatefulValidatorError> for SubmitTransactionError {
+    fn from(err: StatefulValidatorError) -> Self {
+        use RejectedTransactionErrorKind::*;
+        use StatefulValidatorError as E;
+
+        match err {
+            // TODO: Some of those are internal server errors.
```
Trantorian1, line 52: Ditto

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -0,0 +1,280 @@
...
+        converted_class: Option<ConvertedClass>,
+        arrived_at: TxTimestamp,
+    ) -> Result<(), SubmitTransactionError> {
+        if tx.is_only_query() {
+            return Err(RejectedTransactionError::new(
+                RejectedTransactionErrorKind::InvalidTransactionVersion,
+                "Cannot submit query-only transactions",
+            )
+            .into());
+        }
```
Trantorian1, line 175: OK, this makes sense. So we are indeed not processing l1 handler transactions.

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -0,0 +1,280 @@
...
+
+        // We have to skip part of the validation in a the very specific case where you send an invoke tx directly after a deploy account:
+        // the account is not deployed yet but the tx should be accepted.
```
Trantorian1:
```suggestion
        // We have to skip part of the validation in the very specific case where you send an invoke tx directly after a deploy account:
        // the account is not deployed yet but the tx should be accepted.
```

madara/crates/node/src/cli/db.rs
```diff
@@ -1,8 +1,42 @@
-use mc_db::{MadaraBackendConfig, TrieLogConfig};
+use mc_db::{MadaraBackendConfig, RocksDBConfig, TrieLogConfig};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
 
-#[derive(Clone, Debug, clap::Args, Serialize, Deserialize)]
+/// Starknet network types.
```
Trantorian1: This comment is outdated perhaps?

madara/crates/node/src/cli/db.rs
```diff
@@ -1,8 +1,42 @@
...
+    /// Collect all stats except the counters requiring to get time inside the
+    /// mutex lock.
+    ExceptTimeForMutex,
+    /// Collect all stats, including measuring duration of mutex operations.
+    /// If getting time is expensive on the platform to run, it can
+    /// reduce scalability to more threads, especially for writes.
+    All,
+}
+
+impl From<StatsLevel> for mc_db::StatsLevel {
```
Trantorian1, line 25: Why the duplication?

madara/crates/node/src/cli/db.rs
```diff
@@ -51,19 +85,61 @@ pub struct DbParams {
     /// might be an issue for chains which synchronize slowly.
     #[clap(env = "MADARA_FLUSH_EVERY_N_BLOCKS", long, value_name = "NUMBER OF BLOCKS")]
     pub flush_every_n_blocks: Option<u64>,
+
+    /// Enable rocksdb statistics. This has a small performance cost for every database operation.
+    #[clap(env = "MADARA_DB_ENABLE_STATISTICS", long)]
+    pub db_enable_statistics: bool,
+
+    /// If not zero, the rocksdb statistics will be dumped into the db LOG file with this frequency.
```
Trantorian1, line 93: Where is that db log file saved to by default? Seems like that would be worth mentioning here, as well as how this info can be used to debug slow queries.

madara/crates/node/src/cli/gateway.rs
```diff
@@ -19,7 +20,31 @@ pub struct GatewayParams {
...
 
-    /// The gateway port to listen at.
-    #[arg(env = "MADARA_GATEWAY_PORT", long, value_name = "GATEWAY PORT", default_value_t = FGW_DEFAULT_PORT)]
+    /// Enable the madara-specific add_validated_transaction. Beware that this endpoint should not be exposed, this
+    /// is for internal trusted madara-to-madara communication only.
+    /// This endpoint allows madara to accept pre-validated transactions, with already-compiled classes.
+    /// The version of the madara node consuming this HTTP interface must match the version of the gateway server, as they
+    /// use internal unstable types.
+    #[arg(env = "GATEWAY_MADARA_TRUSTED_ADD_TRANSACTION_ENDPOINT", long)]
+    pub gateway_trusted_add_transaction_endpoint: bool,
```
Trantorian1: Compilation flag or not?

madara/crates/node/src/cli/gateway.rs
```diff
@@ -19,7 +20,31 @@ pub struct GatewayParams {
...
     pub gateway_external: bool,
 
-    /// The gateway port to listen at.
-    #[arg(env = "MADARA_GATEWAY_PORT", long, value_name = "GATEWAY PORT", default_value_t = FGW_DEFAULT_PORT)]
+    /// Enable the madara-specific add_validated_transaction. Beware that this endpoint should not be exposed, this
+    /// is for internal trusted madara-to-madara communication only.
+    /// This endpoint allows madara to accept pre-validated transactions, with already-compiled classes.
+    /// The version of the madara node consuming this HTTP interface must match the version of the gateway server, as they
+    /// use internal unstable types.
+    #[arg(env = "GATEWAY_MADARA_TRUSTED_ADD_TRANSACTION_ENDPOINT", long)]
```
Trantorian1:
```suggestion
    #[arg(env = "MADARA_GATEWAY_TRUSTED_ADD_TRANSACTION_ENDPOINT", long)]
```

madara/crates/node/src/cli/validator.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[arg(env = "MADARA_VALIDATE_THEN_FORWARD_TXS", long)]
+    pub validate_then_forward_txs: bool,
+
+    /// Disable transaction validation: no prior validation will be made before inserting into the mempool.
+    /// See: Trasaction validation in [Starknet docs Transaction Validation](https://docs.starknet.io/architecture-and-concepts/network-architecture/transaction-life-cycle/)
+    #[arg(env = "MADARA_NO_TRANSACTION_VALIDATION", long)]
+    pub no_transaction_validation: bool,
+
+    // TODO: move this, idk where this arg would make sense.
+    /// Disable mempool saving. Mempool transactions will not be saved. This can increase performance quite a lot.
```
Trantorian1, line 20: Maybe this should be a db option?

madara/crates/primitives/gateway/Cargo.toml
```diff
@@ -14,6 +14,12 @@ workspace = true
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+# TODO(v0.13.4): re-add this flag.
+# default = ["deny_unknown_fields"]
```
Trantorian1, line 19: This can be removed-then?
```suggestion
```

madara/crates/tests/src/devnet.rs
```diff
@@ -6,18 +7,42 @@ use starknet_core::utils::starknet_keccak;
...
+    Felt::from_hex_unchecked("0x0514977443078cf1e0c36bc88b89ada9a46061a5cf728f40274caea21d76f174"),
+    Felt::from_hex_unchecked("0x00177100ae65c71074126963e695e17adf5b360146f960378b5cdfd9ed69870b"),
+    Felt::from_hex_unchecked("0x07ae55c8093920562c1cbab9edeb4eb52f788b93cac1d5721bda20c96100d743"),
+    Felt::from_hex_unchecked("0x02ce1754eb64b7899c64dcdd0cff138864be2514e70e7761c417b728f2bf7457"),
+    Felt::from_hex_unchecked("0x037a683c3969bf18044c9d2bbe0b1739897c89cf25420342d6dfc36c30fc519d"),
+    Felt::from_hex_unchecked("0x07b4a2263d9cc475816a03163df7efd58552f1720c8df0bd2a813663895ef022"),
+    Felt::from_hex_unchecked("0x064b37f84e667462b95dc56e3c5e93a703ef16d73de7b9c5bfd92b90f11f90e1"),
+    Felt::from_hex_unchecked("0x0213d0d77d5ff9ffbeabdde0af7513e89aafd5e36ae99b8401283f6f57c57696"),
+    Felt::from_hex_unchecked("0x0410c6eadd73918ea90b6658d24f5f2c828e39773819c1443d8602a3c72344c2"),
+];
```
Trantorian1: What do these constants represent? Are those the addresses of the contracts deployed by the devnet?

madara/crates/tests/src/lib.rs
```diff
@@ -49,10 +48,12 @@ async fn wait_for_cond<F: Future<Output = Result<(), anyhow::Error>>>(
...
     ready: bool,
-    json_rpc: JsonRpcClient<HttpTransport>,
-    rpc_url: Url,
+    json_rpc: Option<JsonRpcClient<HttpTransport>>,
+    rpc_url: Option<Url>,
+    gateway_root_url: Option<Url>,
     tempdir: Arc<TempDir>,
-    _port: u16,
+    _rpc_port: Option<MadaraPortNum>,
+    _gateway_port: Option<MadaraPortNum>,
```
Trantorian1: Are we using these fields?

madara/crates/tests/src/lib.rs
```diff
@@ -141,45 +154,35 @@ impl MadaraCmd {
...
-    }
-}
-
-pub fn extract_port_from_stderr(process: &mut Child) -> Result<u16, String> {
-    let stderr = process.stderr.take().ok_or_else(|| "Could not capture stderr from Madara process".to_string())?;
+// this really should use unix sockets, sad
 
-    let reader = BufReader::new(stderr);
-    let (tx, rx) = mpsc::channel();
+const PORT_RANGE: Range<u16> = 19944..20000;
```
Trantorian1: Why the hard-coded range? Why not use something like:

```rust
TcpListener::bind("127.0.0.1:0").unwrap().local_addr().unwrap().port();
```

(You might need to save the tcp connection and not just unwrap on it)

madara/crates/tests/src/rpc/read.rs
```diff
@@ -21,27 +21,55 @@ mod test_rpc_read_calls {
...
+        }
+    }
+    impl Drop for SharedMadaraInstance {
+        fn drop(&mut self) {
+            // let mut guard = MADARA_HANDLE_COUNT.lock().expect("poisoned lock");
+            // *guard -= 1;
+            // if *guard == 0 {
+            //     // :/
+            //     tokio::task::spawn_blocking(|| *MADARA.blocking_lock() = None);
+            // }
```
Trantorian1: ?

Trantorian1: We keep adding and removing this section of code XD Maybe we should add some info here so someone doesn't add it back in the next PR again X).

Also, this doesn't risk never decrementing the `MADARA_HANDLE_COUNT`? (ie: do we have a mechanism to shutdown the madara process we are running or do we just leave it hanging at the end of the tests)

madara/crates/tests/src/rpc/read.rs
```diff
@@ -21,27 +21,55 @@ mod test_rpc_read_calls {
...
+        } else {
+            SharedMadaraInstance(madara_guard.clone().unwrap())
+        };
 
-                madara.wait_for_ready().await;
-                madara.wait_for_sync_to(19).await;
-                madara
-            })
-            .await
+        *guard += 1;
```
Trantorian1: When is this decremented?

madara/crates/tests/src/transaction_flow.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// performed - both offer the same user-facing APIs, but the latter requires the use of the madara-specific add_validated_tx
+    /// gateway endpoint.
+    GatewayAndSequencer,
+}
+
+use TestSetup::*;
+
+impl TestSetup {
+    pub async fn run(self) -> RunningTestSetup {
+        let _ = tracing_subscriber::fmt().with_test_writer().try_init();
```
Trantorian1: Thank you for not using `traced_test` :pray:

Trantorian1 (commented):
Great changes! Still a bit confused whether trusted validation is behind a compilation flag or a cli arg.

madara/crates/client/db/src/storage_updates.rs
cchudant: yes the backend api sux
we need to fix it but right now it's not really a priority :/

madara/crates/client/db/src/stream.rs
cchudant: the new block receiver uses an arc, so that listeners dont need to copy the whole block info at every notification if they only care about 1 field, for example

madara/crates/client/db/src/watch.rs
cchudant: yes, and we can watch for changes in the pending block too
although i don't think i do it here

madara/crates/client/exec/src/execution.rs
cchudant: the nonce is not the account nonce in l1 handler txs, it's the nonce of the core contract on l1
meaning, the nonce of l1handlertxs is an increasing variable that is shared between all l1handlertxs, regardless of account

we really shouldn't order l1 handler txs by nonce IMO..

or actually, we shouldn't put l1 handler txs in the mempool at all imo, block prod batch maker should just pop them at the start of all batches
(this will become an issue when we use tip instead of time of arrival for ordering)

but anyway right now, if you make a l1handlertx to an account contract (which is weird, but whatever) the l1 handler tx will have a nonce that will not match the contract, and so it will be put into the nonce chain of the contract and it really should not
this can lead to bugs i think

madara/crates/client/gateway/server/src/handler.rs
cchudant: the idea was to gate the gateway client behind this cfg, so that it may be used by other people than us without having to pull-in serde_bincode and stuff like that
for the gateway server I don't really care

cchudant: we probably don't care, do you think i should remove it?
especially because the gateway client now depends on submit-tx, which depends on blockifier and mc-db ....

yea i think we should remove it

madara/crates/client/gateway/server/src/service.rs
cchudant: ditto

madara/crates/client/mempool/src/lib.rs
cchudant: i think starknet does not have a way in general to remove rejected txs - once starknet is decentralized, there is really no way to fit rejected txs into the mix
I think we should just let them rot in the mempool until mempool TTL properly disposes of their cadaver

madara/crates/client/rpc/src/errors.rs
cchudant: i am not sure about the error conversion here, but this will be fixed once we test for that case (see #608) and we'll find out whether this one is correct or not

madara/crates/client/settlement_client/src/messaging.rs
cchudant: we still are though :(

madara/crates/client/submit_tx/src/error.rs
cchudant: to keep traceback info i think? but you're right in that the {0:#} wouldn't show traceback anyway..

that's how i did it in the rest of the codebase though..

madara/crates/client/submit_tx/src/validation.rs
cchudant: we are though i think?

madara/crates/node/src/cli/db.rs
cchudant: this one derives clap::Args

madara/crates/primitives/gateway/Cargo.toml
cchudant: no, we want to uncomment that at some point :)

madara/crates/tests/src/devnet.rs
cchudant: yes, ill add a comment

madara/crates/tests/src/lib.rs
cchudant: they are here so that they don't get dropped, because in their Drop impl the port is given back to the global ports-to-reuse list
i'll add a comment

madara/crates/tests/src/lib.rs
cchudant: i htink it got fixed on main, i need to fix this yes

madara/crates/client/db/src/block_db.rs
cchudant: yes, exactly
i can dig back some profiler runs that show CPU usage with and without the change, but no "block/s" change number

madara/crates/client/db/src/block_db.rs
cchudant: Yes,
the new blocks channel is a broadcast channel, but the pending block channel is a watch channel
Reason is, you don't really ever need the previous values of a pending block, you only care about the latest one. Whereas for the new blocks channel you want to see every block as they arrive

The events channel is unchanged

madara/crates/client/db/src/mempool_db.rs
cchudant: i wanted to have the saving of a block and deleting a tx from the mempool to be atomic, but i realize now that we don't really care - we should add a check that we're not re-adding already processed txs instead

madara/crates/client/db/src/rocksdb_options.rs
cchudant: thanks!

madara/crates/client/db/src/rocksdb_options.rs
cchudant: I wanted to play around with these configs for some profiling. I think in general this config should be configurable from outside.

madara/crates/client/db/src/storage_updates.rs
cchudant: they may be missing some updates, but that doesn't matter since they really only want the latest state everytime

madara/crates/client/db/src/stream.rs
cchudant: also, i would like the backend to use more Arcs like this, if we ever implement a block cache

madara/crates/client/db/src/block_db.rs
Trantorian1: Thanks, this makes sense!

madara/crates/client/db/src/block_db.rs
Trantorian1: CPU usage would be great for future reference, but I'd say don't bother yourself if that's not some information that is easily accessible.

madara/crates/client/gateway/server/src/handler.rs
Trantorian1: I would second removing the cfg it in that case. Reading through the code it felt very confusing why some of the validation logic was being feature-gated and some not.

madara/crates/client/mempool/src/lib.rs
Trantorian1: Hmm, I mean we could be storing txs only after they have been accepted by the Mempool? Although that would pose problems in terms of replay if ever the node was shutdown after a tx had been added to and polled from the mempool but not yet stored in db. Also do we handle db removal when a transaction exceeds its TTL?

madara/crates/client/mempool/src/lib.rs
Trantorian1: Rephrasing: since l1 handler transactions should no longer be passed to the mempool, I think we should add a runtime assertion which makes sure we this is not the case. I am assuming this is _not_ enforced at the type level.

madara/crates/client/settlement_client/src/messaging.rs
Trantorian1: Ouch. Where does the insertion happen now, since the validation logic has been moved outside of the mempool?

madara/crates/client/submit_tx/src/error.rs
Trantorian1: :man_shrugging:

madara/crates/client/submit_tx/src/validation.rs
Trantorian1: Are we or are we not X)

madara/crates/node/src/cli/db.rs
Trantorian1: Would it be possible to combine those into a single `struct`?

madara/crates/client/mempool/src/lib.rs
cchudant: Unsure, I really need to look again at the mempool in another pr, I think bugs are still lurking there.

madara/crates/client/submit_tx/src/validation.rs
cchudant: StateError => StateReadError which has errors that our state adaptor can return
There are also a lot of other blockifier errors where it's unclear if they're supposed to be validation failure ou internal

madara/crates/node/src/cli/db.rs
cchudant: i wouldnt recommend, one is for the cli arguments, the other is from the backend api
the concerns are different

madara/crates/node/src/cli/validator.rs
cchudant: I.. dont know
db seems weird to me but I don't have a strong opinion

madara/crates/tests/src/rpc/read.rs
cchudant: fixed :)

madara/crates/tests/src/rpc/read.rs
cchudant: fixed :)

madara/crates/client/db/src/rocksdb_options.rs
cchudant: fixed

madara/crates/client/gateway/server/src/handler.rs
cchudant: fixed

madara/crates/client/mempool/src/lib.rs
cchudant: fixed

madara/crates/client/mempool/src/lib.rs
cchudant: fixed

madara/crates/client/rpc/src/lib.rs
cchudant: fixed

madara/crates/client/submit_tx/src/validation.rs
cchudant: ditto

madara/crates/client/submit_tx/src/validation.rs
cchudant: fixed

madara/crates/node/src/cli/db.rs
cchudant: fixed

madara/crates/client/submit_tx/src/validation.rs
cchudant: fixed

madara/crates/node/src/cli/db.rs
cchudant: fixed

madara/crates/node/src/cli/gateway.rs
cchudant: fixed

madara/crates/client/db/src/rocksdb_options.rs
```diff
@@ -52,16 +60,13 @@ pub fn rocksdb_global_options(config: &RocksDBConfig) -> Result<Options> {
     options.set_keep_log_file_num(3);
     options.set_log_level(rocksdb::LogLevel::Warn);
 
-    // I think this one if a column-specific option, but I'm not entirely sure. So let's just also set it here.
```
Trantorian1: Now I think of this we might want to add a quick comment linking to these pages for future reference no?

- https://github.com/facebook/rocksdb/blob/c237022831aa129aa707bc28e0702a1617ef23b5/include/rocksdb/advanced_options.h#L325
- https://github.com/facebook/rocksdb/blob/c237022831aa129aa707bc28e0702a1617ef23b5/include/rocksdb/options.h#L68

madara/crates/node/src/cli/db.rs
```diff
@@ -87,6 +86,7 @@ pub struct DbParams {
     pub flush_every_n_blocks: Option<u64>,
 
     /// Enable rocksdb statistics. This has a small performance cost for every database operation.
+    /// Statistics are dumped into the `LOG` file in the rocksdb database directory.
```
Trantorian1, line 89: Is this the same directory as specified by `--base-path`?

madara/crates/tests/src/lib.rs
```diff
@@ -38,7 +41,8 @@ async fn wait_for_cond<F: Future<Output = Result<R, anyhow::Error>>, R>(
 
         attempt += 1;
         if attempt >= max_attempts {
-            panic!("No answer from the node after {attempt} attempts: {:#}", err)
+            let elapsed = (attempt as f64 * duration.as_secs() as f64) / 60.0;
+            panic!("No answer from the node after {:.1} minutes: {:#}", elapsed, err);
```
Trantorian1: _(nitpick)_ Is there any way we can format this as a duration instead?

madara/crates/client/submit_tx/src/validation.rs
Trantorian1: image

madara/crates/tests/src/lib.rs
```diff
@@ -152,37 +165,78 @@ impl Drop for MadaraCmd {
...
+                rpc_port = rpc_port.or(get_port(&line, "Running JSON-RPC server at "));
+                gateway_port = gateway_port.or(get_port(&line, "Gateway endpoint started at "));
 
-struct AvailablePorts<I: Iterator<Item = u16>> {
-    to_reuse: Vec<u16>,
-    next: I,
-}
+                if (rpc == rpc_port.is_some()) && (gateway == gateway_port.is_some()) {
+                    let _ = tx.send((rpc_port, gateway_port));
+                }
```
Trantorian1: Should we perhaps panic if the `rpc_port.is_some() == true` but `rpc == false`? Seems like that would be a violation of expectations in the test and indicate some missbehavior.

madara/crates/tests/src/lib.rs
```diff
@@ -335,9 +389,86 @@ async fn madara_can_sync_a_few_blocks() {
...
+            // https://sepolia.voyager.online/block/19
             block_hash: Felt::from_hex_unchecked("0x4177d1ba942a4ab94f86a476c06f0f9e02363ad410cdf177c54064788c9bcb5"),
             block_number: 19
         }
     );
 }
+
+#[rstest]
+#[tokio::test]
+async fn madara_can_sync_and_restart() {
```
Trantorian1: Really liking this pipeline, nice changes!

Trantorian1 (approved):
Looking good!

madara/crates/client/db/src/lib.rs
```diff
@@ -23,7 +23,7 @@
 //! - Each of those functions can be called in parallel, however, [`MadaraBackend::apply_to_global_trie`] needs to be called
 //!   sequentially. This is because we cannot support updating the global trie in an inter-block parallelism fashion. However,
 //!   parallelism is still used inside of that function - intra-block parallelism.
-//! - Each of these block parts has a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
+//! - Each of these block parts have a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
```
jbcaron:
```suggestion
//! - Each of these block parts has a [`chain_head::BlockNStatus`] associated inside of [`MadaraBackend::head_status`],
```

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -50,18 +51,16 @@ fn store_events_to_receipts(
...
+    /// an entire new block, or, you are responsible for calling the `store_xxx`, applying global state, and calling the on_full_block_imported
     /// function when the full block is stored in the backend.
     ///
     /// This function takes a PendingFullBlock because the commitments can only be computed once we applied the state to the global
     /// tries. This function will close the block and compute the commitments.
     ///
-    /// The function returns the new computed block_hash.
-    ///
-    /// `pre_v0_13_2_hash_override`: when true, we will compute the post-v0.13.2 commitments when given a pre-v0.13.2 block.
+    /// The function returnes the new computed block_hash.
```
jbcaron:
```suggestion
    /// The function returns the new computed block_hash.
```

madara/crates/client/db/src/storage_updates.rs
```diff
@@ -50,18 +51,16 @@ fn store_events_to_receipts(
...
 impl MadaraBackend {
-    /// Add a new block to the db, calling the `on_block` handler that handles flushes and backups when they are enabled,
+    /// Add a new block to the db, calling the `on_full_block_imported` handler that handles flushes and backups when they are enabled,
     /// and update all the statuses.
     /// When using the MadaraBackend API to store new blocks, you either have the choice of using this simple function to store
-    /// an entire new block, or, you are responsible for calling the `store_xxx`, applying global state, and calling the on_block
+    /// an entire new block, or, you are responsible for calling the `store_xxx`, applying global state, and calling the on_full_block_imported
     /// function when the full block is stored in the backend.
     ///
     /// This function takes a PendingFullBlock because the commitments can only be computed once we applied the state to the global
```
jbcaron:
```suggestion
    /// This function takes a PendingFullBlock because the commitments can only be computed once we have applied the state to the global
```

madara/crates/client/db/src/stream.rs
```diff
@@ -211,7 +211,7 @@ impl MadaraBackend {
         //  done after getting the latest block number! There's a split second where it could have been possible to miss a
         //  block. Because of this rare case, there are two supplementary things to note: we *also* get the latest block_n
         //  *after* subscribing, so that we can check that we did not miss anything during subscription - and just in case,
-        //  we also handle the case when the subscription returns a block that is futher into the future than the one we
+        //  we also handle the case when the subscription returns a block that's futher into the future than the one we
```
jbcaron:
```suggestion
        //  we also handle the case when the subscription returns a block that's further into the future than the one we
```

madara/crates/client/gateway/client/src/request_builder.rs
```diff
@@ -81,13 +88,48 @@ impl<'a> RequestBuilder<'a> {
...
 
-        let req = req_builder.body(String::new())?;
+        let req = req_builder.body(Full::new(Bytes::from(String::new())))?;
 
         let response: Response<Incoming> =
             self.client.clone().call(req).await.map_err(SequencerError::HttpCallError)?;
         Ok(response)
     }
 
+    pub async fn send_post_bincode<T, D>(self, body: D) -> Result<T, SequencerError>
```
jbcaron, line 99: Consider configuring `bincode` with a fixed endianness

madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -328,9 +329,27 @@ pub async fn handle_get_public_key(backend: Arc<MadaraBackend>) -> Result<Respon
...
+    req: Request<Incoming>,
+    submit_validated: Option<Arc<dyn SubmitValidatedTransaction>>,
+) -> Result<Response<String>, GatewayError> {
+    let Some(submit_validated) = submit_validated else { return Ok(not_found_response()) };
+    let whole_body = req.collect().await.or_internal_server_error("Failed to read request body")?.aggregate();
+
+    let transaction: ValidatedMempoolTx = bincode::deserialize_from(whole_body.reader())
```
jbcaron: ditto bincode

madara/crates/node/src/cli/db.rs
cchudant: yes

madara/crates/tests/src/lib.rs
cchudant: true idk why it's like that

madara/crates/client/mempool/src/lib.rs
```diff
@@ -451,12 +372,11 @@ impl MempoolProvider for Mempool {
         debug_assert!(removed.is_some());
     }
 
-    /// Warning: A lock is taken while a user-supplied function (iterator stuff)
-    /// is run - Callers should be careful This is called by the block
+    /// This is called by the block
     /// production after a batch of transaction is executed. Mark the consumed
```
jbcaron:
```suggestion
    /// production after a batch of transactions is executed. Mark the consumed
```

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/estimate_fee.rs
```diff
@@ -51,13 +51,15 @@ pub async fn estimate_fee(
...
-                    error: result.execution_info.revert_error.clone().unwrap_or_default(),
+                    error: result.execution_info.revert_error.as_ref().map(|e| e.to_string()).unwrap_or_default(),
                 });
             }
             acc.push(exec_context.execution_result_to_fee_estimate(result));
             Ok(acc)
         },
     )?;
 
+    tracing::debug!("fee_estimates = {fee_estimates:?}");
```
jbcaron: why?

madara/crates/primitives/gateway/Cargo.toml
```diff
@@ -14,6 +14,12 @@ workspace = true
...
+default = []
+deny_unknown_fields = []
```
jbcaron, line 21: 👍

madara/crates/client/db/src/rocksdb_options.rs
cchudant: done :)

madara/crates/client/gateway/client/src/request_builder.rs
cchudant: fixed, thanks for pointing this out I didn't know that

madara/crates/client/gateway/server/src/handler.rs
cchudant: fixed, thanks for pointing this out I didn't know that

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/estimate_fee.rs
cchudant: removed :)

madara/crates/node/src/service/l2.rs
```diff
@@ -118,7 +118,7 @@ impl Service for SyncService {
                     this.db_backend.clone(),
                     importer.clone(),
                     gateway,
-                    SyncControllerConfig::default().stop_on_sync(true),
+                    SyncControllerConfig::default().stop_on_sync(true).no_pending_block(true),
```
Mohiiit: shouldn't these be configurable args?

madara/crates/node/src/submit_tx.rs
```diff
@@ -0,0 +1,145 @@
...
+/// sync transaction forwarding to local transaction execution if we are
+/// launching the sync on a local sequencer.
+#[derive(Clone)]
+pub struct SubmitTransactionSwitch {
+    redirect_to_gateway: Arc<dyn SubmitTransaction>,
+    mempool_with_validator: Arc<dyn SubmitTransaction>,
+    ctx: ServiceContext,
+}
+
+impl SubmitTransactionSwitch {
```
Mohiiit, line 34: interesting approach, although it comes with an overhead of having one function call extra on every call to the submitTransaction, is this specifically for the case where madara switches from full node to sequencer mode remotely? I can imagine using an api to down the l2 sync service and up the block production service

but what's your thought process on this?

madara/crates/primitives/gateway/src/block.rs
```diff
@@ -218,7 +219,7 @@ impl ProviderBlock {
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-// #[serde(deny_unknown_fields)] // TODO(v0.13.4): Re-add this attribute when v0.13.4 is supported.
+#[cfg_attr(feature = "deny_unknown_fields", serde(deny_unknown_fields))]
```
Mohiiit, line 222: without l2 gas? the full node would still break in case of sepolia or mainnet sync right?

Mohiiit (commented):
nice work!

jbcaron (approved):

madara/crates/node/src/service/l2.rs
cchudant: this is only for warp updates: this is the initial sync where you pull all of the blocks from the other sequencer before continuing the other stuff
however, I think i'll remove the no_pending_block thing, in hteory it should sync the pending block too imo

madara/crates/node/src/submit_tx.rs
cchudant: IMO this sucks and should go away :)
Warp updates is a feature that I kept with the sync refacto but I do not think anyone is actually using it, I don't think it's useful / is how things should be architectured.

madara/crates/primitives/gateway/src/block.rs
cchudant: if a transaction has l2 gas on sepolia (which, all of them do now) the field will be ignored
This means the node can still sync sepolia, however it will fail with tx hash mismatch on the first tx where the l2 gas is supposed to be kept and added to the tx hash

madara/crates/node/src/submit_tx.rs
Mohiiit: we were thinking to do a POC on warp tbh, haven't got the time but yeah we would like to see it's benefits as well

Mohiiit (approved):

Issue #608, opened 2025-04-22
Madara Node block production / validation tests


cchudant:
From #594 

The multi-setup transaction flow e2e tests (madara/crates/tests/src/transaction_flow.rs) should include:
- [x] working invoke
- [x] working deploy_account
- [x] working declare
- [x] invalid nonce: nonce too low [validation]
- [x] invalid nonce: nonce too low - error caught at mempool and not gateway [validation]
- [ ] gas checks - InsufficientMaxFee (todo: elaborate on tests here, would that include InsufficientAccountBalance? check blockifier's code)
- [ ] __validate__ failure (eg. invalid signature) [validation]
- [ ] execution failure [execution - tx is reverted]
- [ ] deploy_account: class hash not found [validation]
- [ ] declare: invalid program / contract class [validation]
- [ ] declare: class compilation error [validation]
- [ ] declare: compiled class hash mismatch [validation]
- [ ] declare: class too big [validation]
- [ ] declare: class too big but only after decompression (DoS vector) [validation]
- [ ] declare: class already declared [validation]
- [ ] declare: UnsupportedContractClassVersion [validation]
- [ ] invoke: contract not found [validation]
- [ ] invoke: contract is not an account [validation] StarknetError::NonAccount
- [ ] mempool tx limit [validation]
- [ ] deprecated tx [validation]

Block production tests : (sequencer only)
- [ ] Test different transaction types, including L1HandlerTransaction.
- [ ] Test that earlier state changes are taken into account in the current block -> block 2 should see state changes from block 1.
- [x] Test that new transactions roll over to a new block on block_time.
- [x] Test that new transactions roll over to a new block on bouncer config.
- [ ] Test that the pending block is updated correctly, including its state diff, receipts and events.
- [ ] Test that blocks are able to be closed correctly, including its state diff, receipts and events AND global state changes.
- [x] Test that state diffs are produced with contract 0x1 containing the block hash of block_n - 10.
- [ ] Test that in case of a graceful stop, we don't lose transactions nor do we lose pending blocks / full blocks / anything.
- [ ] Test that block time is respected even when mempool is very very full. (there is a bug in madara main around this)
- [x] Test that mempool notify is respected (how we do that? unsure atm)
- [ ] Test rejected transactions are removed from the mempool.
- [x] Test where you close more than a single block.
- [ ] Test that an empty block can be produced when reaching the block_time.

[...]

PR #623, opened 2025-04-29
Upstream block production rewrite

cchudant:
## Pull Request type

+: bugfix / feature

Block production (ie. block building) needs more optimisation. Here is what it looks like on my profiler: (optimistic parallel execution enabled)
image
image

In the first picture, you can clearly see very big gaps of CPU usage when we are not executing a batch, and CPU jumps to a very high utilization when executing a batch.

In addition, you can also see a single bigger gap of a few hundred ms in the picture where the CPU is very underutilized. This gap is when we are updating the pending block in database.
While this may indicate that our rocksdb option tuning is suboptimal (it shouldn't take that long!) and saving the pending block is also suboptimal right now, it points to a bigger issue in the block production: we should be offloading all of this other less-important stuff, and minimize the time before executing each tx chunk.

Also, there is also another big issue not shown in the picture: block closing (ie. updating the global tries and computing the global state root) will slow down the block building significantly; as you can't currently do both at the same time. In addition, saving the closed block in db may take time.

Lastly, there is also currently another issue regarding to block production not respecting block time when there are a lot of txs in the mempool.

This PR aims to address all of these issues with a complete rewrite of the crate, providing a new high performance architecture.

## What is the new behavior?

The main goal here is to minimize the time between batches, ie. the exeuctor is moved to its own thread and is just a loop that repeatedly just calls blockifier's `TransactionExecutor::execute_txs`. Processing the results of the execution, popping transactions from the mempool, saving blocks to the db, updating the global trie, closing the blocks, computing the block hash - all of that is moved away and done async during the execution of the next / previous batch.
This maximizes CPU usage: internally, blockifier's `TransactionExecutor::execute_txs` will spawn a worker thread per CPU core, thus will raise CPU usage to its maximum. Any time that is spent not inside that function is time wasted where you are only using a single CPU core instead of all of them.

To achieve that, we move the executor to its own thread. This is the new architecture:
Untitled-2025-02-04-1118(15)

The batcher task is in charge of popping txs from the mempool, and the block production "master" task is in charge of processing the execution results sent by the executor thread.
This means that in practice:
- transactions to be executed in the next batch are popped from the mempool during the execution of the previous batch
- saving of the pending block to the db is now done asynchronously while the next batch is being executed in the executor
- the closing of block is now done completely asynchronously to execution; meaning, block production can start executing the next block while closing is being done in the background.
  - block closing (global trie computation) can lag behind the executor, up to a maximum of 10 blocks - ensuring backpressure is handled correctly in the unlikely case that block closing lags so much behind.

With this new design, the previous benchmark sees a significant performance increase: we're closer to 90% cpu usage now. The perf increase when taking block closing into account is even more impressive.

There are a few ways we could further optimize block production, but unfortunately we're limited by the architecture of blockifier:
- Blockifier starts all of its worker OS threads everytime you start executing a batch of tx, and then shuts them down when it's finished. This results in a significant waiting time (i've seen up to 20ms!) at the start of every execution batch, wasted because of this unfortunate decision. It takes a really long time to reach 100% cpu usage across every core. 
  - (note: starting/stopping threads at every batch is a poor decision on their part, but it shouldn't take 20ms. 20ms is excessive, I think something bigger might be going on here)
- Every DB access takes a global lock (see their VersionedState struct), even when it shouldn't need to. This is significant: on the transfers benchmark, (which really is mainly a bench that tests nonce read/kv read/nonce write/kv write perf)
  - This is the main reason adding more CPUs to your machine does not increase the perf on that benchmark. The more CPUs you add, the the lock just gets more contended and you reach a plateau of performance around 16-32 CPU cores. That's sad :( this is currently the biggest bottleneck at least for that benchmark.

These require changing the blockifier code, so we'll have to deal with it (for now?). that's sad :(

No

## Other information

There is a new `no_empty_block: bool` chain config now! This new chain config option is really quite neat, it allows us to make appchains that would not cost anything to keep running when no one is using them. Including proving costs :)

However, there's an important detail to keep in mind: when the mempool goes from the empty state to the non-empty state, the executor will be woken up to execute a batch of tx. The gas prices used to execute this new block are the current gas prices when the executor is woken up, and not when the executor has gone to sleep. This is important to ensure the gas prices are correct even when it might have been days since the last block has been closed.

This causes a problem though: like pathfinder and juno, the madara RPC (and gateway for that matter) will display a fake pending block that has the previous block's gas prices when there is no current pending block. This means that doing estimate_fee on the pending block will use the previous block gas prices, and this is a big problem if the previous block is very very old - the gas prices may be all wrong at this point! to fix that, we should change the fake pending block behavior to reflect the current l1 gas prices in a next PR. Until that is fixed, we shouldn't really use no_empty_block is production :)

## Before merging

This pr is ready for review, but I am currently adding more tests! as per #608, I want to add a few of them in this PR, and the rest in a later PR.


Cargo.lock
```diff
@@ -9043,6 +9043,7 @@ dependencies = [
  "tracing-core",
  "tracing-opentelemetry",
  "tracing-subscriber",
+ "tracing-test",
```
Trantorian1: :smiling_face_with_tear:

Trantorian1: Do we really need this?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+    results
+        .into_iter()
+        .map(|r| match r {
+            TransactionExecutorResult::Ok(res) => anyhow::Ok(TransactionExecutorResult::Ok((
+                res,
+                if !added {
+                    added = true;
+                    state.to_state_diff().context("Converting to state diff")?
+                } else {
+                    StateMaps::default()
```
Trantorian1: I assume this results in an empty state map then?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+impl Executor {
+    /// Create the executor thread and returns a handle to it.
+    pub(crate) fn start(
+        backend: Arc<MadaraBackend>,
+        l1_data_provider: Arc<dyn L1DataProvider>,
+    ) -> anyhow::Result<ExecutorHandle> {
+        // buffer is 1.
+        let (sender, recv) = mpsc::channel(1);
```
Trantorian1: Why use this over a `tokio::sync::watch` instead?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+            Err(_timed_out) => {
+                tracing::debug!("Waiting for batch timed out.");
+                Some(Default::default())
+            }
+        }
+    }
+
+    /// We are making a new block - we need to put the hash of current_block_n-10 into the state diff.
+    /// current_block_n-10 however might not be saved into the database yet. In that case, we have to wait.
+    /// This shouldn't create a deadlock (cyclic wait) unless the database is in a weird state (?)
```
Trantorian1: What kind of 'weird state' would we have to be in for this to be the case?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+                .context("Getting block hash of block_n - 10")
+        };
+
+        // Optimistically get the hash from database without subscribing to the closed_blocks channel.
+        if let Some(block_hash) = get_hash_from_db()? {
+            Ok(Some((block_n_min_10, block_hash)))
+        } else {
+            tracing::debug!("Waiting on block_n={} to get closed. (current={})", block_n_min_10, block_n);
+            loop {
+                let mut receiver = self.backend.subscribe_closed_blocks();
```
Trantorian1: Shouldn't we be able to hoist this out of the `loop`?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        // Create the TransactionExecution, but reuse the layered_state_adaptor.
+        let mut executor =
+            self.backend.new_executor_for_block_production(state.state_adaptor, exec_ctx.to_blockifier()?)?;
+
+        // Prepare the block_n_min_10 state diff entry.
+        let mut state_maps_storages = HashMap::default();
+
+        if let Some((block_n_min_10, block_hash_n_min_10)) = self.wait_for_hash_of_block_min_10(exec_ctx.block_n)? {
+            let contract_address = ContractAddress(PatriciaKey::try_from(Felt::ONE).expect("Const conversion"));
```
Trantorian1: Can we make this `static`?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+            let block_full = blockifier_results.len() < to_exec.len();
+
+            let executed_txs = to_exec.remove_n_front(blockifier_results.len()); // Remove the used txs.
+
+            let mut stats = ExecutionStats::default();
+            stats.n_batches += 1;
+            stats.n_executed += executed_txs.len();
+            stats.exec_duration += exec_duration;
+
+            // We does not process the results, it just inspect them for logging stats, and figure out which classes were declared.
```
Trantorian1:
```suggestion
            // Doesn't process the results, it just inspects them for logging stats, and figures out which classes were declared.
```

madara/crates/client/exec/src/block_context.rs
```diff
@@ -1,17 +1,75 @@
...
-    context::{BlockContext, ChainInfo, FeeTokenAddresses},
+    context::BlockContext,
     state::cached_state::CachedState,
 };
 use mc_db::{db_block_id::DbBlockId, MadaraBackend};
 use mp_block::{header::L1DataAvailabilityMode, MadaraMaybePendingBlockInfo};
 use starknet_api::block::{BlockNumber, BlockTimestamp};
 use std::sync::Arc;
 
+/// Extension trait that provides execution capabilities on the madara backend.
```
Trantorian1, line 17: Why aren't we implementing this directly into `MadaraBackend`?

madara/crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -27,159 +26,141 @@ impl BlockifierStateAdapter {
...
-                    "Failed to retrieve storage value for contract {contract_address:#?} at key {:#x}: {err:#}",
-                    key.to_felt()
-                );
-                StateError::StateReadError(format!(
-                    "Failed to retrieve storage value for contract {contract_address:#?} at key {:#x}",
-                    key.to_felt()
-                ))
-            })?
-            .unwrap_or(Felt::ZERO);
+        let value = match self.on_top_of_block_id {
```
Trantorian1, line 33: How is the handling of address `0x1` done now?

madara/crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -27,159 +26,141 @@ impl BlockifierStateAdapter {
...
+        );
+
+        Ok(Nonce(value))
     }
 
     fn get_class_hash_at(&self, contract_address: ContractAddress) -> StateResult<ClassHash> {
-        tracing::debug!("get_class_hash_at for {}", contract_address);
-        let Some(on_top_of_block_id) = self.on_top_of_block_id else { return Ok(ClassHash::default()) };
-
-        // Note that blockifier is fine with us returning ZERO as a class_hash if it is not found, they do the check on their end after
```
Trantorian1, line 93: Maybe we should keep this comment?

madara/crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -27,159 +26,141 @@ impl BlockifierStateAdapter {
...
+                        "Failed to retrieve class_hash: on={:?}, class_hash={:#x}: {err:#}",
+                        self.on_top_of_block_id,
+                        class_hash.to_felt(),
+                    ))
+                })?
             }
-        }
-    }
-}
+            None => None,
```
Trantorian1, line 148: This seems like a use case for `map(...).transpose()?`. Something like:
```rust
let value = self
    .on_top_of_block_id
    .map(|on_top_of_block_id| {
        self.backend
            .get_class_info(&on_top_of_block_id, &class_hash.to_felt())
            .map_err(|err| {
                StateError::StateReadError(format!(
                    "Failed to retrieve class_hash: on={:?}, class_hash={:#x}: {err:#}",
                    self.on_top_of_block_id,
                    class_hash.to_felt(),
                ))
            })
    })
    .transpose()?;
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -718,6 +727,7 @@ impl MempoolInner {
         self.nonce_cache_inner.insert(tx_mempool.contract_address(), tx_mempool.nonce_next);
 
         // do not update mempool limits, block prod will update it with re-add txs.
```
Trantorian1: This comment appears to be out of data, consider removing it?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -206,7 +194,7 @@ impl Mempool {
         Ok(())
     }
 
-    fn accept_validated_tx(&self, tx: ValidatedMempoolTx) -> Result<(), MempoolError> {
+    fn accept_tx(&self, tx: ValidatedMempoolTx) -> Result<(), MempoolError> {
```
Trantorian1: This still does not perform any validation no?

madara/crates/client/mempool/src/notify.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+}
+
+// todo(deadlocks): document mutex ordering
+pub(crate) struct MempoolInnerWithNotify {
+    // TODO(perf): figure out if it's worth to have a tokio lock here instead of std lock.
+    // (ie. use tracing spans to log how long the write lock is taken)
+    inner: RwLock<MempoolInner>,
+    // TODO: should this be in inner mempool? I don't understand why it's here.
```
Trantorian1, line 49: This is used to mark transactions as `Pending` or `Ready` if they are received after the previous transaction from the same account has been added to the block _but_ this change is not yet reflected in db. For example:

1. Tx with nonce `0x0` is received.
2. Tx with nonce `0x0` is inserted into the mempool as `Ready`
3. Tx with nonce `0x0` is inserted into the latest block
4. Tx with nonce `0x1` is received.
5. Tx with bonce `0x1` is inserted into the mempool as `Pending` since `0x0` is the next expected nonce according to the db.
6. Db is updated (next nonce expected nonce is `0x1`).

Notice that at the end of step `6.` tx with nonce `0x1` is already in the mempool as `Pending`! We do not currently update the state of transactions in the mempool to deal with this. Instead, the nonce cache serves to keep track of the next expected nonce for each account between db updates.

madara/crates/client/sync/src/util.rs
```diff
@@ -1,6 +1,4 @@
-use futures::Future;
-use std::{fmt, pin::Pin, task};
-use tokio::task::JoinHandle;
+use std::fmt;
```
Trantorian1, line 1: Where are we using `std::fmt` in this file? It seems like this can be removed.

madara/crates/primitives/block/src/header.rs
```diff
@@ -44,6 +44,12 @@ impl BlockTimestamp {
     }
 }
 
+impl From<SystemTime> for BlockTimestamp {
+    fn from(value: SystemTime) -> Self {
+        Self(value.duration_since(SystemTime::UNIX_EPOCH).expect("SystemTime::now() < Unix epoch").as_secs())
```
Trantorian1, line 50: Shouldn't this be `"SystemTime::now() > Unix epoch"`?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1,101 +1,188 @@
...
+            header,
+            transactions: Default::default(),
+            events: Default::default(),
+            declared_classes: Default::default(),
+            state_diff: StateMaps {
+                nonces: Default::default(),
+                class_hashes: Default::default(),
+                storage: initial_state_diffs_storage,
+                compiled_class_hashes: Default::default(),
+                declared_contracts: Default::default(),
```
Trantorian1: Can some of this be replaced by `..Default::default()`?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -105,505 +192,305 @@ struct ContinueBlockResult {
...
 }
 
-impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
-    #[cfg(any(test, feature = "testing"))]
-    #[tracing::instrument(skip(self), fields(module = "BlockProductionTask"))]
-    pub fn set_current_pending_tick(&mut self, n: usize) {
-        self.current_pending_tick = n;
+impl BlockProductionTask {
+    /// This is a channel that helps the testing of the block production task. It is unused outside of tests.
+    pub fn subscribe_state_notifications(&mut self) -> mpsc::UnboundedReceiver<BlockProductionStateNotification> {
```
Trantorian1, line 243: This should probably be behind a `cfg(test)` then no?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -105,505 +192,305 @@ struct ContinueBlockResult {
...
+        self.backend.chain_config().precheck_block_production()?;
 
-        // Complete the block with full bouncer capacity
-        let start_time = Instant::now();
-        let ContinueBlockResult { state_diff: mut new_state_diff, .. } =
-            self.continue_block(self.backend.chain_config().bouncer_config.block_max_capacity)?;
+        if let Err(err) = self.close_pending_block_if_exists().await {
+            // This error should not stop block production from working. If it happens, that's too bad. We drop the pending state and start from
+            // a fresh one.
+            tracing::error!("Failed to continue the pending block state: {err:#}");
```
Trantorian1, line 417: Shouldn't we be preemptively returning here then?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -2362,51 +1658,224 @@ mod tests {
...
+
+    //     let pending_block = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+    //     assert!(mempool.is_empty());
+    //     assert_eq!(pending_block.inner.transactions.len(), 2);
+    //     assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 0);
+    // }
+
+    // This test makes sure that the pending tick closes the block
+    // if the bouncer capacity is reached
```
Trantorian1:
```suggestion
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -2476,58 +1960,231 @@ mod tests {
...
+    //     // least block_time
+
+    //     wait_block_produced(block_production_task).await.unwrap();
+
+    //     assert!(mempool.is_empty());
+    //     assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 1);
+    // }
+
+    // This test checks when the block production task starts on
+    // normal behaviour, it updates properly
```
Trantorian1:
```suggestion
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -2540,91 +2197,277 @@ mod tests {
...
+
+    //     tokio::time::sleep(std::time::Duration::from_secs(35)).await;
+    //     task_handle.abort();
+
+    //     let pending_block: mp_block::MadaraMaybePendingBlock = backend.get_block(&DbBlockId::Pending).unwrap().unwrap();
+
+    //     assert!(mempool.is_empty());
+    //     assert!(pending_block.inner.transactions.is_empty());
+    //     assert_eq!(backend.get_latest_block_n().unwrap().unwrap(), 2);
+    // }
```
Trantorian1:
```suggestion
```

madara/crates/client/block_production/src/util.rs
```diff
@@ -0,0 +1,435 @@
...
+        self.additional_info.push_back(additional_info);
+    }
+
+    pub fn remove_n_front(&mut self, n_to_remove: usize) -> BatchToExecute {
+        // we can't acutally use split_off because it doesnt leave the cap :/
+
+        if n_to_remove >= self.len() {
+            // if we take everything, just replace self with an empty batch and return self.
+            // unsure if this can actually help with perf, but hey this theorically saves a memcopy
+            // let's be frank, ideally blockifier just isn't dumb and takes an iterator and not a &[Transaction]..
```
Trantorian1: XD

Trantorian1 (commented):
Looks good! Really liking the new design for the block prod, the state machine code is quite clean as well. Overall this a nice improvement over the previous state of things, thanks a lot for all the work! :)

Trantorian1 (approved):

madara/crates/client/block_production/src/executor.rs
cchudant: i dont think a watch would make sense here? we're not interested in the latest state

madara/crates/client/block_production/src/executor.rs
cchudant: i feel like I had to put a comment about cyclic waits here since you have two tasks that can theorically wait for each other
I haven't thought about when it could be the case, i think it can never be the case ; except if the prod master task is waiting for a reply from the executor and the executor is waiting for block_n-10 to show up

i think this case cannot make sense, this is just me being pedantic about having to prove that deadlocks(dead-*waits*?) are not possible everytime there is a task that waits on something, and usually it is trivially true because the graph of tasks/threads waiting on each other is acyclic

madara/crates/client/block_production/src/executor.rs
cchudant: not hoisting it out is tiny bit more efficient but i'm splitting some hairs here
both are correct

madara/crates/client/exec/src/block_context.rs
cchudant: i think i like it better as a separate thing that exists on top of the backend, it doesn't need any of the internal details of the backend impl, it's just supplementing the backend api on top of the other public apis

madara/crates/client/exec/src/blockifier_state_adapter.rs
cchudant: not done here
we talked with @jbcaron and this doesn't really apply here, it should be in the state diffs of the generated blocks anyway.
I think we used to have that special case handling because some blocks on mainnet needed this special handling. I will recheck if that's still needed

madara/crates/client/mempool/src/lib.rs
cchudant: yes
do you think the argument is enough to understand or should i keep the old name?

madara/crates/client/sync/src/util.rs
cchudant: fn fmt_option (...) -> impl fmt::Display

madara/crates/primitives/block/src/header.rs
cchudant: UNIX_EPOCH is january the 1st 1970 0:00 utc
SystemTime::duration_since will return None if the resulting duration is negative,
thus, the only way to get None here is to have a systemtime before UNIX_EPOCH, before jan 1st 1970

However, uh, this shouldnt say SystemTime::now() < Unix epoch it should say value < Unix epoch as you can give it a systemtime before now
i might just remove this impl From tbh just to be sure it's never used with a user-supplied SystemTime (wouldn't make sense to happen tho but still.)

madara/crates/client/block_production/src/lib.rs
cchudant: i think it's fine to leave it, it's always a real pain to have cfg(test), especially if the thing is also used in other crates and so you have to make a "testing" feature flag and everything, and then you may end up with compile errors that only happen when this feature flag is not set and so rust-analyzer and your tests and rustfmt won't tell you but cargo build --release only will tell you that ....
it's a pain :(

.. okay sure i'll add the #[cfg(any(feature = "testing", test))]..

madara/crates/client/block_production/src/lib.rs
cchudant: This is from the old block production code
I don't know, really what we should do here
This part used to re-run the pending block as a whole so that you can continue it, and we were thinking about stuff like: what if the chain config changed? what if now the pending state can't be reexecuted for any other reason? we shouldn't softlock ourselves, we should just throw out the pending state and start a fresh one. if the user changed the versioned constants / chain config options / etc. it's their fault but we shouldn't keep crashing on start-up.

but then with Mohit we realized something: what if we just close the block?
closing the block doesn't require any of the chain config stuff, it's just commitment computation and global trie update
so this was changed to closing the pending block on start-up. that's definitely the best way to handle this case, by far.
we still kept the "if it fails, let's just throw it out and shrug it off" logic. even, I realize now that the error message still says "continue" instead of "closing".
You're right tho, I can't see how this case could happen and it might be an error we don't want to recover from now that we're just closing it. Throwing it out is a micro-reorg of only the pending block state, and if any of the other block close operation fails we indeed bubble the error and crash the node.
The only way to trigger this now is to have corrupted db state, and in that case i don't know if we really want to try to recover that, we should just crash

what do you think?

madara/crates/client/block_production/src/util.rs
cchudant: i'll remove that lol

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+    pub stop: StopErrorReceiver,
+    /// Channel with the replies from the executor thread.
+    pub replies: mpsc::Receiver<ExecutorMessage>,
+}
+
+// TODO(blockifier_update): When updating blockifier, remove the following lines! (TransactionExecutionOutput and blockifier_v0_8_0_fix_results).
+//
+// The reason is, in the new blockifier update, TransactionExecutor::execute_txs does not return `Vec<TransactionExecutorResult<TransactionExecutionInfo>>`
+// anymore. Instead, it returns `Vec<TransactionExecutorResult<TransactionExecutionOutput>>`, where they define `TransactionExecutionOutput` as the type alias
+// `type TransactionExecutionOutput = (TransactionExecutionInfo, StateMaps)`. This new StateMaps field is is the state diff produced for the corresponding TransactionExecutionInfo.
```
jbcaron:
```suggestion
// `type TransactionExecutionOutput = (TransactionExecutionInfo, StateMaps)`. This new StateMaps field is the state diff produced for the corresponding TransactionExecutionInfo.
```

madara/crates/client/block_production/src/executor.rs
jbcaron: There is no const constructor for `PatriciaKey`, but `From<u128>` is implemented.

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+        // Create the TransactionExecution, but reuse the layered_state_adaptor.
+        let mut executor =
+            self.backend.new_executor_for_block_production(state.state_adaptor, exec_ctx.to_blockifier()?)?;
+
+        // Prepare the block_n_min_10 state diff entry.
+        let mut state_maps_storages = HashMap::default();
+
+        if let Some((block_n_min_10, block_hash_n_min_10)) = self.wait_for_hash_of_block_min_10(exec_ctx.block_n)? {
+            let contract_address = ContractAddress(PatriciaKey::try_from(Felt::ONE).expect("Const conversion"));
+            let key = StorageKey::try_from(Felt::from(block_n_min_10)).context("Block_n overflows storage key")?;
```
jbcaron:
```suggestion
            let key = StorageKey::from(block_n_min_10);
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -105,505 +192,305 @@ struct ContinueBlockResult {
...
-        let new_empty_block = MadaraPendingBlock::new_empty(make_pending_header(
-            parent_block_hash,
-            self.backend.chain_config(),
-            self.l1_data_provider.as_ref(),
-        ));
-
-        let block_to_close = mem::replace(&mut self.block, new_empty_block);
-        let declared_classes = mem::take(&mut self.declared_classes);
+    /// Returns the block_hash.
+    async fn close_and_save_block(
```
jbcaron, line 281: Would adding #[instrument] here be a good idea?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+
+/// Note: The reason this exists is because we want to create the new block execution context (meaning, the block header) as late as possible, as to have
+/// the best gas prices. This is especially important when the no_empty_block configuration is enabled, as otherwise we would end up:
+/// - Creating a new execution context, using the current gas prices.
+/// - Waiting for a transaction to arrive.... potentially for a very, very long time..
+/// - Transaction arrives, we execute it and close the block, as the block_time is reached.
+///
+/// At that point, the gas prices would be all wrong! In order to support no_empty_block correctly, we have to delay execution context creation
+/// until the first transaction has arrived.
+enum ExecutorState {
```
Mohiiit: we have another struct with the same name, it was confusing for a while to get the right idea about which is which, is it possible to rename it or the other one?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+impl Executor {
+    /// Create the executor thread and returns a handle to it.
+    pub(crate) fn start(
+        backend: Arc<MadaraBackend>,
+        l1_data_provider: Arc<dyn L1DataProvider>,
+    ) -> anyhow::Result<ExecutorHandle> {
+        // buffer is 1.
+        let (sender, recv) = mpsc::channel(1);
+        let (replies_sender, replies_recv) = mpsc::channel(100);
+        let (stop_sender, stop_recv) = oneshot::channel();
```
Mohiiit: given we are creating these channels here and providing some handlers a level up, for me it makes more sense to have it a level up and sending the `recv`, `replies_sender` and `stop_sender` in the start itself (essentially sending the `Executor` to the `start` function, maybe adding a function `new` too? wdyt?

madara/crates/client/block_production/src/executor.rs
```diff
@@ -0,0 +1,451 @@
...
+
+            // TODO(blockifier_update): Remove the following line when updating blockifier.
+            let blockifier_results = blockifier_v0_8_0_fix_results(execution_state, blockifier_results)?;
+
+            let exec_duration = exec_start_time.elapsed();
+
+            // When the bouncer cap is reached, blockifier will return fewer results than what we asked for.
+            let block_full = blockifier_results.len() < to_exec.len();
+
+            let executed_txs = to_exec.remove_n_front(blockifier_results.len()); // Remove the used txs.
```
Mohiiit: `to_exec` would be at max of the `batch_size` size right? 

could it be the case that madara is executing one batch and while this one is getting executed another batch came in which updates the `to_exec`?

we can make `to_exec` empty here right?

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -0,0 +1,135 @@
...
+    state_diff: StateMaps,
+    classes: HashMap<ClassHash, ContractClass>,
+}
+
+/// Special cache that allows us to execute blocks past what the db currently has saved. Only used for
+/// block production.
+/// We need this because when a block is produced, saving it do the database is done asynchronously by another task. This means
+/// that we need to keep the state of the previous block around to execute the next one. We can only remove the cached state of the
+/// previous blocks once we know they are imported into the database.
+pub struct LayeredStateAdaptor {
```
Mohiiit, line 32: I'm a bit confused about how this works.

Let's say block x has been executed and is now being sent for closure. Meanwhile, another batch is being processed for execution. In this case, does Blockifier call the LayeredStateAdapter to get the necessary state updates, even though the last block hasn't been stored yet?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -105,505 +192,305 @@ struct ContinueBlockResult {
...
-    mempool: Arc<Mempool>,
-    block: MadaraPendingBlock,
-    declared_classes: Vec<ConvertedClass>,
-    pub(crate) executor: TransactionExecutor<BlockifierStateAdapter>,
     l1_data_provider: Arc<dyn L1DataProvider>,
-    current_pending_tick: usize,
+    mempool: Arc<Mempool>,
+    current_state: Option<ExecutorState>,
     metrics: Arc<BlockProductionMetrics>,
+    state_notifications: Option<mpsc::UnboundedSender<BlockProductionStateNotification>>,
```
Mohiiit, line 213: I am not sure about how this is being used.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -105,505 +192,305 @@ struct ContinueBlockResult {
...
-
-        let block_to_close = mem::replace(&mut self.block, new_empty_block);
-        let declared_classes = mem::take(&mut self.declared_classes);
+    /// Returns the block_hash.
+    async fn close_and_save_block(
+        &mut self,
+        block_n: u64,
+        block: PendingFullBlock,
+        classes: Vec<ConvertedClass>,
+        _txs_executed: Vec<Felt>,
```
Mohiiit, line 286: will there be a case in p2p where we would need this?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -636,23 +525,11 @@ mod tests {
...
-            },
-            gas: 100000,
-            message_segment_length: 100000,
-            n_events: 100000,
-            n_steps: 100000,
-            state_diff_size: 100000,
+            gas: 1000000,
+            message_segment_length: 10000,
+            n_events: 10000,
+            state_diff_size: 10000,
```
Mohiiit, line 567: I am not sure how relevant it is for testing but there was an update in the bouncerWeights in the madara recently, like state-update being 4000, n-events I think was 5k and message length was 3700 (this is as per the official starkware mainnet weights)

Mohiiit (commented):
gg!

I am a bit confused about the executor flow and it changes over time and also about how does the blockifier is made aware about the changes made in the last batch! I remember you mentioning something about blockier not being aware about the block/batch as such? but I am not sure!

madara/crates/client/exec/src/blockifier_state_adapter.rs
jbcaron: Even if there are specials on old blocks, it’s not an issue because we don’t support the execution of those block versions.

madara/crates/client/exec/src/blockifier_state_adapter.rs
```diff
@@ -27,159 +26,141 @@ impl BlockifierStateAdapter {
...
+                .map_err(|err| {
+                    StateError::StateReadError(format!(
+                        "Failed to retrieve storage value: on={:?}, contract_address={:#x} key={:#x}: {err:#}",
+                        self.on_top_of_block_id,
+                        contract_address.to_felt(),
+                        key.to_felt(),
+                    ))
+                })?
+                .unwrap_or(Felt::ZERO),
+            None => Felt::ZERO,
```
jbcaron, line 46: only for genesis?

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -0,0 +1,135 @@
...
+    core::{ClassHash, CompiledClassHash, ContractAddress, Nonce},
+    state::StorageKey,
+};
+use std::{
+    collections::{HashMap, VecDeque},
+    sync::Arc,
+};
+
+#[derive(Debug)]
+struct CacheByBlock {
```
jbcaron, line 21: 👍

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -29,59 +30,82 @@ fn rejected(kind: RejectedTransactionErrorKind, message: impl Into<Cow<'static,
...
         use StatefulValidatorError as E;
 
         match err {
-            // TODO: Some of those are internal server errors.
-            err @ E::StateError(_) => rejected(ValidateFailure, format!("{err:#}")),
+            E::StateError(err) => err.into(),
             E::TransactionExecutionError(err) => err.into(),
-            err @ E::TransactionExecutorError(_) => rejected(ValidateFailure, format!("{err:#}")),
+            // this is a weird one... why does this variant exist..?
+            E::TransactionExecutorError(err) => err.into(),
```
jbcaron, line 67: Wouldn't it be for the deploy tx?

madara/crates/node/src/cli/chain_config_overrides.rs
```diff
@@ -128,6 +128,8 @@ impl ChainConfigOverrideParams {
             mempool_tx_max_age: chain_config.mempool_tx_max_age,
             feeder_gateway_url: chain_config.feeder_gateway_url,
             gateway_url: chain_config.gateway_url,
+            no_empty_blocks: chain_config.no_empty_blocks,
+            block_production_concurrency: chain_config.block_production_concurrency,
```
jbcaron, line 134: The configuration files would need to be updated.

madara/crates/client/submit_tx/src/validation.rs
cchudant: this would be validationerror i think

cchudant: i think i said that because `TransactionExecutorError` is also a variant in the parent error enum so i was confused about why it's in there twice

madara/crates/client/block_production/src/lib.rs
cchudant: i think i'm gonna remove that, in p2p we will just let rejected tx during block building rot in the mempools until their time to live expires

cchudant: that's how SW plans on doing it anyway, since we're doing FCFS instead of tip-sorting maybe it feels weirder tho

Cargo.lock
cchudant: fixed :)

madara/crates/client/block_production/src/executor.rs
cchudant: fixed :)

madara/crates/client/block_production/src/executor.rs
cchudant: fixed :)

madara/crates/client/exec/src/blockifier_state_adapter.rs
cchudant: fixed :)

madara/crates/client/exec/src/blockifier_state_adapter.rs
cchudant: doesnt work, your code returns an Option<Option
you can make it work but i thinl it's more messy

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/notify.rs
cchudant: will see when i get back on the mempool, thanks :)

madara/crates/client/block_production/src/lib.rs
cchudant: fixed :)

madara/crates/client/block_production/src/lib.rs
cchudant: fixed :)

madara/crates/client/block_production/src/lib.rs
cchudant: fixed :)

madara/crates/client/block_production/src/lib.rs
cchudant: fixed :)

madara/crates/client/block_production/src/util.rs
cchudant: removed

madara/crates/client/block_production/src/executor.rs
cchudant: renamed :)

madara/crates/client/block_production/src/executor.rs
cchudant: reworked the hierarchy :)

madara/crates/client/block_production/src/executor.rs
cchudant: if the block cap has been reached the to_exec batch will not be empty here

madara/crates/client/exec/src/layered_state_adaptor.rs
cchudant: LayeredStateAdaptor allows the us to layer some state on top of the current db state, so that blockifier can see that state without having to update the db.

madara/crates/client/block_production/src/lib.rs
cchudant: this is for testing only :)

madara/crates/client/block_production/src/lib.rs
cchudant: will keep it until I get the confirmation that get_tx_status/get_msg_status status REJECTED is not going to be supported in v0.14.

madara/crates/client/block_production/src/lib.rs
cchudant: i think that should be in chain_config then?

madara/crates/client/block_production/src/executor.rs
cchudant: fixed :)

madara/crates/client/block_production/src/lib.rs
cchudant: added, but I still dont know how we should sprinkle #[instrument] around the codebase. right now it's very haphazard

madara/crates/client/exec/src/blockifier_state_adapter.rs
cchudant: yes, the blockifierstateadaptor always returns empty when executing the genesis block

madara/crates/client/submit_tx/src/validation.rs
cchudant: removed :)

madara/crates/node/src/cli/chain_config_overrides.rs
cchudant: will fix :)

antiyro (approved):

PR #624, opened 2025-05-02
feat: add `madara_close_block` admin RPC method

jbcaron:
## Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?

-

No

### DIFF
--- a/madara/crates/client/block_production/src/lib.rs
+++ b/madara/crates/client/block_production/src/lib.rs
@@ -35,16 +35,65 @@ use mp_state_update::{ContractStorageDiffItem, DeclaredClassItem, NonceUpdate, S
 use mp_transactions::TransactionWithHash;
 use mp_utils::service::ServiceContext;
 use opentelemetry::KeyValue;
+use serde::{Deserialize, Serialize};
 use starknet_types_core::felt::Felt;
 use std::borrow::Cow;
 use std::collections::VecDeque;
 use std::mem;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Instant;
 
 mod close_block;
 mod finalize_execution_state;
 pub mod metrics;
+
+/// Block production mode.
+///
+/// This enum defines the different modes for triggering block closure.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlockProductionMode {
+    /// Uses periodic time intervals to trigger block production.
+    ///
+    /// In this mode, blocks are produced at regular intervals defined by the chain configuration.
+    TimedTicks,
+
+    /// Relies exclusively on external triggers to produce blocks.
+    ///
+    /// In this mode, blocks are only produced when explicitly triggered through an external channel,
+    /// allowing for manual or event-driven block production.
+    ExternalTrigger,
+
+    /// Combines both timed intervals and external triggers.
+    ///
+    /// This mode allows blocks to be produced both at regular intervals and when explicitly
+    /// triggered through an external channel, providing maximum flexibility.
+    #[default]
+    Hybrid,
+}
+
+impl FromStr for BlockProductionMode {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "timed-ticks" => Ok(BlockProductionMode::TimedTicks),
+            "external-trigger" => Ok(BlockProductionMode::ExternalTrigger),
+            "hybrid" => Ok(BlockProductionMode::Hybrid),
+            _ => Err(format!("Invalid block production mode: {s}")),
+        }
+    }
+}
+
+impl ToString for BlockProductionMode {
+    fn to_string(&self) -> String {
+        match self {
+            BlockProductionMode::TimedTicks => "timed-ticks".to_string(),
+            BlockProductionMode::ExternalTrigger => "external-trigger".to_string(),
+            BlockProductionMode::Hybrid => "hybrid".to_string(),
+        }
+    }
+}
 
 #[derive(Default, Clone)]
 struct ContinueBlockStats {
@@ -114,6 +163,8 @@ pub struct BlockProductionTask<Mempool: MempoolProvider> {
     l1_data_provider: Arc<dyn L1DataProvider>,
     current_pending_tick: usize,
     metrics: Arc<BlockProductionMetrics>,
+    production_mode: BlockProductionMode,
+    external_trigger: Option<Arc<tokio::sync::Notify>>,
 }
 
 impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
@@ -210,6 +261,8 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
         mempool: Arc<Mempool>,
         metrics: Arc<BlockProductionMetrics>,
         l1_data_provider: Arc<dyn L1DataProvider>,
+        production_mode: BlockProductionMode,
+        external_trigger: Option<Arc<tokio::sync::Notify>>,
     ) -> Result<Self, Error> {
         if let Err(err) = Self::close_pending_block(&backend, &metrics).await {
             // This error should not stop block production from working. If it happens, that's too bad. We drop the pending state and start from
@@ -239,6 +292,8 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
             declared_classes: Default::default(),
             l1_data_provider,
             metrics,
+            production_mode,
+            external_trigger,
         })
     }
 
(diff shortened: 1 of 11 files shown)


madara/crates/client/block_production/src/lib.rs
```diff
@@ -35,17 +35,66 @@ use mp_state_update::{ContractStorageDiffItem, DeclaredClassItem, NonceUpdate, S
...
 
+/// Block production mode.
+///
+/// This enum defines the different modes for triggering block closure.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlockProductionMode {
+    /// Uses periodic time intervals to trigger block production.
+    ///
+    /// In this mode, blocks are produced at regular intervals defined by the chain configuration.
+    TimedTicks,
```
Trantorian1, line 59: I think this should be the default no? Imo we would probably want external block sealing to be opt-in.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -530,7 +585,7 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
 
         loop {
             tokio::select! {
-                instant = interval_block_time.tick() => {
+                instant = interval_block_time.tick(), if matches!(self.production_mode, BlockProductionMode::Hybrid | BlockProductionMode::TimedTicks) => {
```
Trantorian1, line 588: Maybe this kind of logic should be refactored into a method inside of `BlockProductionMode`, something like `is_production_tick`.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -572,6 +627,24 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
                         }
                     }
                 },
+                _ = async {
+                    if let Some(trigger) = self.external_trigger.as_ref() {
+                        trigger.notified().await;
+                    } else {
+                        std::future::pending::<()>().await;
+                    }
+                }, if matches!(self.production_mode, BlockProductionMode::Hybrid | BlockProductionMode::ExternalTrigger) => {
```
Trantorian1, line 636: Same here, something like `is_production_external`

madara/crates/client/block_production/src/lib.rs
```diff
@@ -572,6 +627,24 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
                         }
                     }
                 },
+                _ = async {
+                    if let Some(trigger) = self.external_trigger.as_ref() {
+                        trigger.notified().await;
+                    } else {
+                        std::future::pending::<()>().await;
+                    }
```
Trantorian1, line 635: Move this into a closure which is defined outside the `select`? It can be a bit of pain to write code here since Rust does not handle formatting or suggestions well in a `select!` branch so I would recommend this.

madara/crates/client/rpc/src/lib.rs
```diff
@@ -59,6 +59,7 @@ pub enum StarknetSyncStatus {
 pub struct Starknet {
     backend: Arc<MadaraBackend>,
     pub(crate) add_transaction_provider: Arc<dyn AddTransactionProvider>,
+    pub(crate) close_block_trigger: Option<Arc<tokio::sync::Notify>>,
```
Trantorian1, line 62: I fell like we should be moving this into `MadaraBackend` instead to avoid polluting the `Starknet` struct? To me this seems logical as it is more of a db-related operation anyway but I guess that's debatable. Lemme know what you think.

madara/crates/client/rpc/src/versions/admin/v0_1_0/api.rs
```diff
@@ -21,6 +22,9 @@ pub trait MadaraWriteRpcApi {
         &self,
         declare_v0_transaction: BroadcastedDeclareTxnV0,
     ) -> RpcResult<ClassAndTxnHash>;
+
+    #[method(name = "closeBlock")]
+    async fn close_block(&self) -> RpcResult<MadaraBlockInfo>;
```
Trantorian1, line 27: This should incur a new minor version of the admin rpc specs!

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -20,4 +21,20 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+    ///
+    /// * `block_info` - the block info of the closed block
+    async fn close_block(&self) -> RpcResult<MadaraBlockInfo> {
+        let mut rx = self.backend.subscribe_block_info();
+        match self.close_block_trigger.as_ref() {
+            Some(trigger) => {
+                trigger.notify_one();
+                rx.recv().await.map_err(|_| StarknetRpcApiError::InternalServerError.into())
+            }
+            None => Err(StarknetRpcApiError::UnimplementedMethod.into()),
```
Trantorian1, line 37: `StarknetRpcApiError::UnimplementedMethod` doesn't feel very accurate to the kind of error we are encountering.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -35,17 +35,66 @@ use mp_state_update::{ContractStorageDiffItem, DeclaredClassItem, NonceUpdate, S
...
+
+impl ToString for BlockProductionMode {
+    fn to_string(&self) -> String {
+        match self {
+            BlockProductionMode::TimedTicks => "timed-ticks".to_string(),
+            BlockProductionMode::ExternalTrigger => "external-trigger".to_string(),
+            BlockProductionMode::Hybrid => "hybrid".to_string(),
+        }
+    }
+}
```
Trantorian1, line 96: Is this needed for `clap`?

madara/crates/node/src/service/block_production.rs
```diff
@@ -13,6 +13,8 @@ pub struct BlockProductionService {
     metrics: Arc<BlockProductionMetrics>,
     l1_data_provider: Arc<dyn L1DataProvider>,
     n_devnet_contracts: u64,
+    production_mode: BlockProductionMode,
+    external_trigger: Option<Arc<tokio::sync::Notify>>,
```
Trantorian1, line 17: Same as for `Starknet`, I feel like moving this to the `MadaraBackend` and adding a `notify` and `notified` function would make this more ergonomic and setup cleaner.

madara/crates/node/src/service/block_production.rs
```diff
@@ -40,13 +55,15 @@ impl Service for BlockProductionService {
...
-        let Self { backend, l1_data_provider, mempool, metrics, .. } = self;
+        let Self { backend, l1_data_provider, mempool, metrics, production_mode, external_trigger, .. } = self;
 
         let block_production_task = BlockProductionTask::new(
             Arc::clone(backend),
             Arc::clone(mempool),
             Arc::clone(metrics),
             Arc::clone(l1_data_provider),
+            *production_mode,
+            external_trigger.take(),
```
Trantorian1, line 66: This can pose issues in terms of restarting a `Service` :sweat_smile: You can read the docs in `service.rs` for more info. Note that `MadaraBackend` is not meant to be restarted so it would be fine to do that there.

madara/crates/node/src/service/rpc/mod.rs
```diff
@@ -31,6 +31,7 @@ pub struct RpcService {
     backend: Arc<MadaraBackend>,
     add_txs_provider_l2_sync: Arc<dyn AddTransactionProvider>,
     add_txs_provider_mempool: Arc<dyn AddTransactionProvider>,
+    close_block_trigger: Option<Arc<tokio::sync::Notify>>,
```
Trantorian1, line 34: Wait so if we have that here why are we storing it in `Starknet` and `BlockProduction` as well? This should just be made accessible via `notify` and `notified` methods (you can probably find better names than those X)

madara/crates/client/block_production/src/lib.rs
```diff
@@ -114,6 +163,8 @@ pub struct BlockProductionTask<Mempool: MempoolProvider> {
     l1_data_provider: Arc<dyn L1DataProvider>,
     current_pending_tick: usize,
     metrics: Arc<BlockProductionMetrics>,
+    production_mode: BlockProductionMode,
+    external_trigger: Option<Arc<tokio::sync::Notify>>,
```
cchudant, line 167: i would prefer a
struct TriggerCloseBlock; // unit struct
`mpsc::Receiver<TriggerCloseBlock>`

i think it makes more sense Notify is too low level i don't like it here, wdyt

madara/crates/client/block_production/src/lib.rs
```diff
@@ -572,6 +627,24 @@ impl<Mempool: MempoolProvider> BlockProductionTask<Mempool> {
                         }
                     }
                 },
+                _ = async {
```
cchudant, line 630: i would have written this as
`Some(_) = OptionFuture::from(trigger.as_mut().map(|t| t.notified())`

cchudant (commented):
i dont know about `BlockProductionMode`, maybe just a bool is enough
but I'm fine with this enum, at least until #623 is merged - i'll probably also add stuff to pause/resume execution of pending blocks too 

@jbcaron do you want to have this PR merged before #623 ?
if so, i'm okay with reworking this implem to make it work with #623 in #623 directly.

jbcaron:
feature integrated into #623

PR #625, opened 2025-05-05
Add v0 batching logic

prkpndy:
This PR adds the logic for Batching worker for applicative recursion

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Right now, the flow is like this:
```
snos → proving → da → state-update
```

## What is the new behavior?

This PR introduces a batching worker that will squash the state update of all the blocks in it and at the end will generate an array of felts that can be sent in a blob on Ethereum

New flow (all changes are not in this PR)
```
snos → proving → aggregator-program → submit aggregator → state-update (+ a worker running which makes sure that we are pushing 6 blobs at a time)
```

No


orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,107 @@ impl DatabaseClient for MongoDbClient {
...
+        non_null_updates.insert("batch_size", Bson::Int64(update.end_block as i64 - batch.start_block as i64 + 1));
+        non_null_updates.insert("updated_at", Bson::DateTime(Utc::now().round_subsecs(0).into()));
+
+        let update = doc! {
+            "$set": non_null_updates
+        };
+
+        let result = self.get_batch_collection().find_one_and_update(filter, update, options).await?;
+        match result {
+            Some(b) => {
```
Mohiiit: lets make this more descriptive

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,107 @@ impl DatabaseClient for MongoDbClient {
...
+                let attributes = [KeyValue::new("db_operation_name", "update_job")];
+                let duration = start.elapsed();
+                ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+                Ok(b)
+            }
+            None => {
+                tracing::warn!(job_id = %batch.id, category = "db_call", "Failed to update job. Job version is likely outdated");
```
Mohiiit: this log is wrong

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,107 @@ impl DatabaseClient for MongoDbClient {
...
+        non_null_updates.insert("updated_at", Bson::DateTime(Utc::now().round_subsecs(0).into()));
+
+        let update = doc! {
+            "$set": non_null_updates
+        };
+
+        let result = self.get_batch_collection().find_one_and_update(filter, update, options).await?;
+        match result {
+            Some(b) => {
+                let attributes = [KeyValue::new("db_operation_name", "update_job")];
```
Mohiiit: this metrics is wrong

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,107 @@ impl DatabaseClient for MongoDbClient {
...
+
+    async fn create_batch(&self, batch: Batch) -> Result<Batch, DatabaseError> {
+        let start = Instant::now();
+
+        match self.get_batch_collection().insert_one(batch.clone(), InsertOneOptions::builder().build()).await {
+            Ok(_) => {
+                let duration = start.elapsed();
+                tracing::debug!(duration = %duration.as_millis(), "Batch created in MongoDB successfully");
+
+                let attributes = [KeyValue::new("db_operation_name", "create_job")];
```
Mohiiit: update this
```suggestion
                let attributes = [KeyValue::new("db_operation_name", "create_batch")];
```

orchestrator/crates/orchestrator/src/tests/workers/batching/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        server.mock(|when, then| {
+            when.path("/").body_includes(format!("starknet_getStateUpdate?blockId={}", block_num));
+            then.status(200).body(serde_json::to_vec(&state_update).unwrap());
+        });
+    }
+
+    crate::worker::event_handler::triggers::batching::BatchingTrigger.run_worker(services.config).await?;
+
+    rpc_block_call_mock.assert();
```
Mohiiit: let's also test the output

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,278 @@
...
+        // Calculating the last block number to for which a batch needs to be assigned
+        let last_block_to_assign_batch = config
+            .service_config()
+            .max_block_to_process
+            .map_or(block_number_provider, |max_block| min(max_block, block_number_provider));
+
+        tracing::debug!(latest_block_number = %last_block_to_assign_batch, "Fetched latest block number from starknet");
+
+        // Getting the latest batch in DB
+        let latest_batch = config.database().get_latest_batch().await?;
```
Mohiiit, line 42: this might return a None, let's handle that too

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    fn get_state_update_file_name(&self, batch_index: u64) -> String {
+        format!("{}/batch/{}.json", STATE_UPDATE_DIR, batch_index)
+    }
+
+    /// squash_state_updates merge all the StateUpdate into a single StateUpdate
+    fn squash_state_updates(&self, state_updates: Vec<StateUpdate>) -> Result<StateUpdate, JobError> {
+        if state_updates.is_empty() {
+            panic!("Cannot merge empty state updates");
```
Mohiiit: let's throw an error in case of this

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        // Take the last block hash and number from the last update as our "latest"
+        let last_update = state_updates.last().unwrap();
```
Mohiiit: let's remove unwrap and do a proper error handling

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,278 @@
...
+        let block_hash = last_update.block_hash;
+        let new_root = last_update.new_root;
+        let old_root = state_updates.first().unwrap().old_root;
```
Mohiiit: same

0xvasanth:
@prkpndy  this PR have not been closed by me it got closed automatically  since I have deleted the branch `restructure/orch`, kindly update the PR so it should be cool

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: This I am handling in the next line?

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,110 @@ impl DatabaseClient for MongoDbClient {
...
+        let mut non_null_updates = Document::new();
+        updates.iter_mut().for_each(|(k, v)| {
+            if v != &Bson::Null {
+                non_null_updates.insert(k, v);
+            }
+        });
+
+        // throw an error if there's no field to be updated
+        if non_null_updates.is_empty() {
+            return Err(DatabaseError::NoUpdateFound("No field to be updated, likely a false call".to_string()));
```
Mohiiit, line 654: do we need to throw an error when there is nothing to update? how about we throw a warning and just return the batch?

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +243,128 @@ async fn database_test_update_job() {
...
+        id: Uuid::new_v4(),
+        index: 3,
+        size: 100,
+        start_block: 301,
+        end_block: 400,
+        is_batch_ready: false,
+        squashed_state_updates_path: String::from("path/to/file.json"),
+        created_at: Utc::now(),
+        updated_at: Utc::now(),
+    };
```
Mohiiit: let's create a util function for this or maybe we can use fixtures here!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+
+        // Convert maps back to the required StateDiff format
+
+        // Storage diffs
+        for (contract_addr, storage_map) in storage_diffs_map {
+            let storage_entries = storage_map
+                .into_iter()
+                .filter(|(_, value)| *value != Felt::ZERO) // Filter out entries with value 0x0
```
Mohiiit, line 235: technically this is not right, it will fail in our test cases, we need to add the code which checks the storage slot at block x-1

orchestrator/crates/orchestrator/src/worker/mod.rs
```diff
@@ -20,6 +20,7 @@ pub async fn initialize_worker(config: Arc<Config>) -> OrchestratorResult<()> {
         Ok(_) => tracing::info!("Consumers initialized successfully"),
         Err(e) => {
             tracing::error!(error = %e, "Failed to initialize consumers");
+            panic!("Failed to init consumers: {}", e);
```
Mohiiit, line 23: hmm is this panic required? won't it just work the next time?

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
```diff
@@ -59,4 +61,10 @@ pub trait DatabaseClient: Send + Sync {
         status: Vec<JobStatus>,
         limit: Option<i64>,
     ) -> Result<Vec<JobItem>, DatabaseError>;
+    /// get_latest_batch - Get the latest batch from DB. Returns `None` if the DB is empty
+    async fn get_latest_batch(&self) -> Result<Option<Batch>, DatabaseError>;
+    /// update_batch - Update the bath
+    async fn update_batch(&self, batch: &Batch, update: &BatchUpdates) -> Result<Batch, DatabaseError>;
+    /// create_batch - Create a new batch
```
0xvasanth: Passing BatchUpdates by reference doesn’t make much sense here, especially if the function is expected to mutate its fields (e.g., updated_at for audit purposes). It would be better to pass it by value to allow internal modifications without borrowing issues.
```suggestion
    async fn update_batch(&self, batch: &Batch, update: BatchUpdates) -> Result<Batch, DatabaseError>;
    /// create_batch - Create a new batch
```

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,110 @@ impl DatabaseClient for MongoDbClient {
...
+                        let attributes = [KeyValue::new("db_operation_name", "get_latest_batch")];
+                        let duration = start.elapsed();
+                        ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+                        Ok(Some(batch))
+                    }
+                    Err(e) => {
+                        tracing::error!(
+                            error = %e,
+                            document = ?doc,
+                            "Failed to deserialize document into Batch"
```
0xvasanth, line 625: Would it be possible to extract a private function in the MongoDB Client, maybe something like find_one, to avoid repeating the same logic across multiple business functions that return either an item or None?

This could help us reuse the same logic in places like `get_latest_job_by_type`, and the function can be generic over the return type to keep things flexible and clean.

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
```diff
@@ -0,0 +1,52 @@
+use chrono::{DateTime, SubsecRound, Utc};
```
0xvasanth: All the types and Models are suppose to be under the Types module, can we move this there

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct Batch {
+    /// Unique identifier for the batch
+    pub id: Uuid,
+    /// Index of the batch
+    pub index: u64,
+    /// Number of blocks in the batch
+    pub size: u64,
```
0xvasanth: the attribute named as size is inappropriate here
since i thought it as is cumulative sum of all the block size to make sure it is under the the allowed since, event though i understand after reading the comment this is highly misintrpreted

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: This is the pattern we are following in update job as well. 
When I am calling this method in the code as well, I am assuming that the doc should already be present and I expect an error if it's not able to update it
If it's a hard requirement, can make the changes in the code. But I would prefer to keep it this way :)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: fixed this part in stateless compression part

orchestrator/crates/orchestrator/src/worker/mod.rs
prkpndy: yeah! removed it

orchestrator/crates/orchestrator/src/tests/database/mod.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
prkpndy: BatchUpdates only contains two fields
`pub struct BatchUpdates {
    pub end_block: u64,
    pub is_batch_ready: bool,
}`
We are not making any mutations in this inside `update_batch` function. I think it should be okay to pass it by reference? Or am I missing something?

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
prkpndy: I'll refactor this in stateless compression PR? - Also have added some more fields there
is `num_blocks` a good name?

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
prkpndy: Moved

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
0xvasanth: @prkpndy , we might have legacy code in update job ignore that for now we can correct that in future scope,
To answer to the @Mohiiit comment, it is not madatory to suppress the error in all the case with the warning, 

My recommendation would be like create two function 
- update_item_if_exists - > this function will not throw any error if nothing to been updated return `Optional<Item>`
- update_item - will update or raise error if the object is not present in the system

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+use starknet::providers::Provider;
+use starknet_core::types::MaybePendingStateUpdate::{PendingUpdate, Update};
+use std::cmp::{max, min};
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+// TODO: Remove this constant when `assign_batch_to_block` method is updated
+const MAX_BATCH_SIZE: u64 = 50;
+
+const STATE_UPDATE_DIR: &str = "state_update";
```
0xvasanth: move these variables into constant file

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+        let last_block_to_assign_batch = config
+            .service_config()
+            .max_block_to_process
+            .map_or(block_number_provider, |max_block| min(max_block, block_number_provider));
+
+        tracing::debug!(latest_block_number = %last_block_to_assign_batch, "Fetched latest block number from starknet");
+
+        // Getting the latest batch in DB
+        let latest_batch = config.database().get_latest_batch().await?;
+        let latest_block_in_db = latest_batch.map(|batch| batch.start_block).unwrap_or(0);
```
0xvasanth: If I am not wrong, end_block is suppose to be the lastest_block right

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+        let database = config.database();
+
+        // Get the storage client
+        let storage = config.storage();
+
+        // Get the state update for the block
+        let state_update = provider
+            .get_state_update(BlockId::Number(block_number))
+            .await
+            .map_err(|e| JobError::ProviderError(e.to_string()))?;
```
0xvasanth, line 75: Can we use the `#[from]` in JobError to handle the error better here

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    fn get_state_update_file_name(&self, batch_index: u64) -> String {
+        format!("{}/batch/{}.json", STATE_UPDATE_DIR, batch_index)
+    }
+
+    /// squash_state_updates merge all the StateUpdate into a single StateUpdate
+    fn squash_state_updates(&self, state_updates: Vec<StateUpdate>) -> Result<StateUpdate, JobError> {
+        if state_updates.is_empty() {
+            return Err(JobError::Other(OtherError(eyre!("Cannot merge empty state updates"))));
+        }
```
0xvasanth, line 164: Can we use anyhow here

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        // Take the last block hash and number from the last update as our "latest"
+        let last_update = state_updates.last().ok_or(JobError::Other(OtherError(eyre!("Invalid state updates"))))?;
```
0xvasanth, line 167: Use anyhow where ever you use new eyre code

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
0xvasanth: I think this been moved in second PR

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Will need to change the `OtherError` type here. Then I'll need to change `OtherError` usage in all the files where it's used. Might be too much out of the scope of this PR. Can create an issue for this. wdyt?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Sorted

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Again, this will require changes in a lot of other files o think. Can create a separate issue for this?

orchestrator/crates/orchestrator/src/types/batch.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub is_batch_ready: bool,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct Batch {
+    /// Unique identifier for the batch
+    #[cfg_attr(feature = "with_mongodb", serde(with = "uuid_1_as_binary"))]
+    pub id: Uuid,
+    /// Index of the batch
+    pub index: u64,
```
heemankv, line 19: Is index not already a unique key for the Batch ?

orchestrator/crates/orchestrator/src/types/batch.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Start and end block numbers of the batch (both inclusive)
+    pub start_block: u64,
+    pub end_block: u64,
+    /// Whether the batch is ready to be processed,
+    /// This will happen when adding a new block takes the size of the felt array beyond 6 * 4096
+    pub is_batch_ready: bool,
+    /// Path to the squashed state updates file,
+    /// This is done for optimization so we don't have to create a new squashed state update from scratch
+    pub squashed_state_updates_path: String,
+    /// timestamp when the job was created
```
heemankv: Change Job to Batch

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
heemankv:
```
let mut updates = update.to_document()?;
```
I believe this is allowing the current implementation to work, 
either we remove this and let updates be mutable by itself, or we keep it as is.

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +240,79 @@ async fn database_test_update_job() {
...
+    let services = TestConfigBuilder::new().configure_database(ConfigType::Actual).build().await;
+    let config = services.config;
+    let database_client = config.database();
+
+    // Create a new batch
+    let batch = build_batch(1, 100, 200);
+
+    database_client.create_batch(batch.clone()).await.unwrap();
+
+    // Waiting for sometime to ensure updated_at is different after the update
```
heemankv, line 284: I don't understand, what's the reason behind updating is_batch_ready with the same value

orchestrator/crates/orchestrator/src/types/batch.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[cfg_attr(feature = "with_mongodb", serde(with = "chrono_datetime_as_bson_datetime"))]
+    pub created_at: DateTime<Utc>,
+    /// timestamp when the job was last updated
+    #[cfg_attr(feature = "with_mongodb", serde(with = "chrono_datetime_as_bson_datetime"))]
+    pub updated_at: DateTime<Utc>,
```
heemankv, line 36: Can we use version here as well ? 
then we can enforce that if any other orchestrator instance updated the batch then the current update will not go through

orchestrator/crates/orchestrator/src/types/constant.rs
```diff
@@ -2,3 +2,6 @@ pub const BLOB_DATA_FILE_NAME: &str = "blob_data.txt";
 pub const SNOS_OUTPUT_FILE_NAME: &str = "snos_output.json";
 pub const PROGRAM_OUTPUT_FILE_NAME: &str = "program_output.txt";
 pub const CAIRO_PIE_FILE_NAME: &str = "cairo_pie.zip";
+pub const S3_STATE_UPDATE_DIR: &str = "state_update";
```
heemankv: let's not mention 'S3' here, rather use 'DB' if you really want to, else `STATE_UPDATE_DIR` is good enough for me!

orchestrator/crates/orchestrator/src/types/jobs/mod.rs
```diff
@@ -18,6 +18,7 @@ pub enum WorkerTriggerType {
     ProofRegistration,
     DataSubmission,
     UpdateState,
+    Batching,
```
heemankv, line 21: Check `impl FromStr for WorkerTriggerType {` below.
I believe we forgot to update the implementation with the `Batching` WorkerTriggerType.
This makes the code updating flow very susceptible to issues like these.
Can you try writing a macro that solves this ?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// 2. Fetch the last batch and check its `end_block`
+    /// 3. Assign batches to all the remaining blocks and store the squashed state update in storage
+    async fn run_worker(&self, config: Arc<Config>) -> color_eyre::Result<()> {
+        tracing::trace!(log_type = "starting", category = "BatchingWorker", "BatchingWorker started");
+
+        // Getting the latest block number from Starknet
+        let provider = config.madara_client();
+        let block_number_provider = provider.block_number().await?;
+
+        // Calculating the last block number to for which a batch needs to be assigned
```
heemankv: // Calculating the last block number to for which a batch needs to be assigned
 Can be changed to 
 // Calculating the latest block number that needs to be assigned to a batch.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+        let last_block_to_assign_batch = config
+            .service_config()
+            .max_block_to_process
+            .map_or(block_number_provider, |max_block| min(max_block, block_number_provider));
+
+        tracing::debug!(latest_block_number = %last_block_to_assign_batch, "Fetched latest block number from starknet");
```
heemankv: Nitpick :
> Fetched latest block number from starknet
Let's change this to 
> Calcualted latest block number to batch.

Coz the number is actually being minimised by the max_block.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: Nitpick : Let's use `map_or` | and not `unwrap_or`

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        // Getting the latest batch in DB
+        let latest_batch = config.database().get_latest_batch().await?;
+        let latest_block_in_db = latest_batch.map(|batch| batch.end_block).unwrap_or(0);
+
+        // Calculating the first block number to for which a batch needs to be assigned
```
heemankv, line 45: Same as above

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: I notice we have the same log in snos as well

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+                        // Update state update for the batch in storage
+                        storage
+                            .put_data(
+                                Bytes::from(serde_json::to_string(&new_state_update)?),
+                                &self.get_state_update_file_name(batch.index),
+                            )
+                            .await?;
+                        // Update batch status in the database
+                        database
+                            .update_batch(&batch, &BatchUpdates { end_block: block_number, is_batch_ready: false })
```
heemankv: ```is_batch_ready: false```
why are we hardcoding it to a false here,

I thought the logic here (without SF and SL compression) would be to check the size of the array against MAX_BATCH_SIZE.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+                    // No batch exists in the DB yet
+                    // Create a fresh batch
+
+                    let squashed_state_updates_path = self.get_state_update_file_name(batch_index + 1);
+                    // Put the state update in storage
+                    storage
+                        .put_data(Bytes::from(serde_json::to_string(&state_update)?), &squashed_state_updates_path)
+                        .await?;
+                    // Add the new batch info in the database
+                    database.create_batch(Batch::create(1, block_number, squashed_state_updates_path)).await?;
```
heemankv, line 143: Any reason we are starting batch from 1 and not 0 ?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: +1

orchestrator/crates/orchestrator/src/tests/database/mod.rs
heemankv: + let's utilise default() in `build_batch` @prkpndy

orchestrator/crates/orchestrator/src/worker/mod.rs
heemankv: Yeah, if this panics then the orch stops, we don't want that

heemankv (commented):
As discussed offline, 
Let's think about: 
- How running multiple orchestrators will/ will not break this worker 
- How can `version` (as they are used in jobs) can help us with the above problem

orchestrator/crates/orchestrator/src/types/constant.rs
0xvasanth: +1

orchestrator/crates/orchestrator/src/tests/database/mod.rs
prkpndy: I don't think default will work here as I need some customer values for testing

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: This is more of an optimisation, and will require changes in other methods as well. Can do this in a separate PR I think?

orchestrator/crates/orchestrator/src/types/batch.rs
prkpndy: Yes it is

orchestrator/crates/orchestrator/src/tests/database/mod.rs
prkpndy: Because we don't want to mark the batch as ready. Can make it different as well, don't think that will make any difference?

orchestrator/crates/orchestrator/src/types/constant.rs
prkpndy: made it `STORAGE_STATE_UPDATE_DIR` to keep it aligned with our convention - Added the `STORAGE` part to convey that the dir is in S3 or any other storage

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: yes, you are right. since the size is still less than `MAX_BATCH_SIZE`, we mark it as not ready. if adding this block makes the size greater than that, we mark the latest batch as ready and start a new batch (the logic in else block)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Not anything in particular. But now everything is coded keeping this in mind, so let's keep it this way

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/types/jobs/mod.rs
prkpndy: Fixed the FromStr implementation. We can create a macro in a separate PR (will need to change the code at many places for this)

orchestrator/crates/orchestrator/src/types/batch.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,110 @@ impl DatabaseClient for MongoDbClient {
...
+        let filter = doc! {
+            "id": batch.id,
+        };
+        let options = FindOneAndUpdateOptions::builder().upsert(false).return_document(ReturnDocument::After).build();
+
+        let mut updates = update.to_document()?;
+
+        // remove null values from the updates
+        let mut non_null_updates = Document::new();
+        updates.iter_mut().for_each(|(k, v)| {
```
Trantorian1: Why is this calling `iter_mut` if we are not mutating the elements `k` and `v` we are iterating over? Is the mutation taking place in `Document::insert`?

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +240,79 @@ async fn database_test_update_job() {
         panic!("Job not found in Database.")
     }
 }
+
+#[rstest]
```
Trantorian1, line 244: Why are we using `rstest` on a test with no fixtures?

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +240,79 @@ async fn database_test_update_job() {
...
+    database_client.create_batch(batch3.clone()).await.unwrap();
+
+    // Get latest batch should return batch3 since it has the highest index
+    let latest_batch = database_client.get_latest_batch().await.unwrap().unwrap();
+    assert_eq!(latest_batch, batch3);
+}
+
+#[rstest]
+#[tokio::test]
+async fn database_test_update_batch() {
```
Trantorian1: Same here.

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +240,79 @@ async fn database_test_update_job() {
...
+#[tokio::test]
+async fn database_test_get_latest_batch() {
+    let services = TestConfigBuilder::new().configure_database(ConfigType::Actual).build().await;
+    let config = services.config;
+    let database_client = config.database();
+
+    // Create multiple batches with different indices
+    let batch1 = build_batch(1, 100, 200);
+    let batch2 = build_batch(2, 201, 300);
+    let batch3 = build_batch(3, 301, 400);
```
Trantorian1: Batches could be a good thing to put into fixtures. You can re-use the same fixture for multiple arguments as follows:

```rust
#[rstest::fixture]
fn batch(
    #[default(0)] a: u32,
    #[default(0)] b: u32,
    #[default(0)] c: u32
) -> Batch {
    // batch initialization logic
}

#[rstest::rstest]
#[tokio::test]
async fn database_test_get_latest_batch(
    #[from(batch)]
    #[with(1, 100, 200)]
    batch1: Batch,
    #[from(batch)]
    #[with(2, 201, 300)]
    batch2: Batch,
    #[from(batch)]
    #[with(3, 301, 400)]
    batch3: Batch,
) {
    // test logic
}
```

Then again, this is mostly syntactic sugar :) But if you choose to go forwards with no fixtures, then please remove the `#[rstest]` annotation!

orchestrator/crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -239,3 +240,79 @@ async fn database_test_update_job() {
...
+    assert_eq!(updated_batch.size, updates.end_block - batch.start_block + 1);
+    assert_eq!(updated_batch.start_block, batch.start_block);
+    assert_eq!(updated_batch.end_block, updates.end_block);
+    assert_eq!(updated_batch.is_batch_ready, batch.is_batch_ready);
+    assert_eq!(updated_batch.squashed_state_updates_path, batch.squashed_state_updates_path);
+    assert_eq!(updated_batch.created_at, batch.created_at);
+    assert_ne!(updated_batch.updated_at, batch.updated_at);
+}
+
+#[rstest]
```
Trantorian1, line 305: Same here.

orchestrator/crates/orchestrator/src/tests/utils.rs
```diff
@@ -74,3 +75,17 @@ pub fn build_job_item(job_type: JobType, job_status: JobStatus, internal_id: u64
         updated_at: Utc::now().round_subsecs(0),
     }
 }
+
+pub fn build_batch(index: u64, start_block: u64, end_block: u64) -> Batch {
```
Trantorian1: This seems like it should be a fixture!

Trantorian1: (The same can be said for `build_job_item` above)

orchestrator/crates/orchestrator/src/types/batch.rs
Trantorian1: Bumping this up. If `index` is already a unique key then isn't `id: Uuid` redundant? I'm not familiar with the db schema of orchestrator, any more context would be much appreciated!

orchestrator/crates/orchestrator/src/types/jobs/mod.rs
Trantorian1: We shouldn't have to write a custom macro for this, [`strum`](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumString.html) already has something which does just that! Just add a `#[derive(strum_macros::EnumString)]`:

```rust
#[derive(..., strum_macros::EnumString)]
#[strum(serialize_all = "PascalCase")]
pub enum WorkerTriggerType {
    Snos,
    Proving,
    ProofRegistration,
    DataSubmission,
    UpdateState,
    Batching,
}

fn main() {
    // Same API
    let worker = WorkerTriggerType::from_str("Snos").unwrap();
}
```

Trantorian1: (In this case and given `strum` is already a project dep I do think this should be added!)

orchestrator/crates/orchestrator/src/worker/event_handler/service.rs
```diff
@@ -19,6 +19,7 @@ use crate::types::jobs::WorkerTriggerType;
 use crate::utils::metrics::ORCHESTRATOR_METRICS;
 #[double]
```
Trantorian1, line 20: What is this annotation? I've never seen it before so just curious :)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+
+pub struct BatchingTrigger;
+
+#[async_trait::async_trait]
+impl JobTrigger for BatchingTrigger {
+    /// 1. Fetch the latest completed block from Starknet chain
+    /// 2. Fetch the last batch and check its `end_block`
+    /// 3. Assign batches to all the remaining blocks and store the squashed state update in storage
+    async fn run_worker(&self, config: Arc<Config>) -> color_eyre::Result<()> {
+        tracing::trace!(log_type = "starting", category = "BatchingWorker", "BatchingWorker started");
```
Trantorian1: Why are `log_type` and `category` being set here? If you are trying to delimit this region of logs then you should probably be using `#[tracing::instrument]` instead.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Calculating the first block number to for which a batch needs to be assigned
+        let first_block_to_assign_batch = config
+            .service_config()
+            .min_block_to_process
+            .map_or(latest_block_in_db, |min_block| max(min_block, latest_block_in_db));
+
+        for block_num in first_block_to_assign_batch..last_block_to_assign_batch + 1 {
+            self.assign_batch_to_block(block_num, config.clone()).await?;
+        }
+        tracing::trace!(log_type = "completed", category = "BatchingWorker", "BatchingWorker completed.");
```
Trantorian1: Same here. Is logging set up in such a way that `log_type` and `category` have a special meaning?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+                // FIXME: Update this flow
+                // For now, adding a hardcoded limit on the number of blocks each batch can have
+                // The Correct implementation:
+                // 1. Squash all state updates to get a single state update
+                // 2. Perform stateful compression
+                // 3. Perform stateless compression
+                // 4. Create an array of felts
+                // 5. Check if its length is less than 6 * 4096
+                //    If yes, it can be added in the same batch
+                //    Else, start a new batch
```
Trantorian1, line 88: Could this be a potential DOS vector in the case where the orchestrator keeps getting supplied batches which generate arrays of size > 6 * 4096 and has to do the statefull and stateless compression just to keep throwing away the work? Is this even a possibility (ie: is this function called as the result of a publicly-facing endpoint)?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        // Create a new StateDiff to hold the merged state
+        let mut state_diff = StateDiff {
+            storage_diffs: Vec::new(),
+            deployed_contracts: Vec::new(),
+            declared_classes: Vec::new(),
+            deprecated_declared_classes: Vec::new(),
+            nonces: Vec::new(),
+            replaced_classes: Vec::new(),
+        };
```
Trantorian1, line 181: This can probably be replaced with a `StateDiff::default()` (maybe the trait isn't implemented, idk?)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+            self.assign_batch_to_block(block_num, config.clone()).await?;
+        }
+        tracing::trace!(log_type = "completed", category = "BatchingWorker", "BatchingWorker completed.");
+        Ok(())
+    }
+}
+
+impl BatchingTrigger {
+    /// assign_batch_to_block assigns a batch to the block
+    async fn assign_batch_to_block(&self, block_number: u64, config: Arc<Config>) -> Result<(), JobError> {
```
Trantorian1, line 61: Do we have any unit tests for this function? Might have missed them sorry

Trantorian1 (commented):
Cool changes! Do we have any unit test / e2e tests for the stateful compression here? I might have missed them but it doesn't seem like we are running `BatchingTrigger::run_worker` or `BatchingTrigger::assign_batch_to_block` inside any of the new tests, am I wrong?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
Mohiiit: this won't be a public function, the orch will be running in a private setup!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Yes, the trait isn't implemented

orchestrator/crates/orchestrator/src/worker/event_handler/service.rs
prkpndy: #[double] macro is a conditional macro - it determines at compile time whether to use the real implementation or the mock version of it

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: The logging functionality is coded in such a way that
log_type -> is the stage at which the operation is at
category -> is the operation which is being performed

Right now, we have not implemented logging properly in the orch code. A lot of improvement is needed in this honestly. Can pick this up when doing a complete cleanup of the code?
(BTW, this is the pattern being used in the rest of the code as well for orch. Will fix this up in a seperate PR?)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Commented above

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: We don't have any unit tests as of now. But I have added E2E tests (for the complete flow, which includes batching, stateful compression and stateless compression) in the stateful compression PR - https://github.com/madara-alliance/madara/pull/631

orchestrator/crates/orchestrator/src/tests/utils.rs
prkpndy: Sorted!

orchestrator/crates/orchestrator/src/tests/database/mod.rs
prkpndy: I have made a fixture. Thanks for the detailed explanation :)

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: Sorted!

orchestrator/crates/orchestrator/src/types/jobs/mod.rs
prkpndy: Sorted!

orchestrator/crates/orchestrator/src/types/batch.rs
prkpndy: Yeah, you're right! But I find having a UUID helpful a lot of times. For eg, it helps a lot while searching for a particular job in the logs. I would prefer keeping it 😅 (But if I am breaking some pattern, please let me know, I'll remove it)

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: Sorted

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: Fixed

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: Fixed

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,110 @@ impl DatabaseClient for MongoDbClient {
...
+            Some(updated_batch) => {
+                // Update done
+                let attributes = [KeyValue::new("db_operation_name", "update_batch")];
+                let duration = start.elapsed();
+                ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+                Ok(updated_batch)
+            }
+            None => {
+                // Not found
+                tracing::warn!(batch_id = %batch.id, category = "db_call", "Failed to update batch");
```
Mohiiit:
```suggestion
                tracing::error!(batch_id = %batch.id, category = "db_call", "Failed to update batch");
```

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +449,110 @@ impl DatabaseClient for MongoDbClient {
...
+        match self.get_batch_collection().insert_one(batch.clone(), InsertOneOptions::builder().build()).await {
+            Ok(_) => {
+                let duration = start.elapsed();
+                tracing::debug!(duration = %duration.as_millis(), "Batch created in MongoDB successfully");
+
+                let attributes = [KeyValue::new("db_operation_name", "create_batch")];
+                ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+                Ok(batch)
+            }
+            Err(err) => Err(DatabaseError::InsertFailed(format!(
```
Mohiiit: let's add tracing before throwing the error!

orchestrator/crates/orchestrator/src/tests/utils.rs
```diff
@@ -74,3 +76,22 @@ pub fn build_job_item(job_type: JobType, job_status: JobStatus, internal_id: u64
...
+    #[default(200)] end_block: u64,
+) -> Batch {
+    Batch {
+        id: Uuid::new_v4(),
+        index,
+        size: end_block - start_block + 1,
+        start_block,
+        end_block,
+        is_batch_ready: false,
+        squashed_state_updates_path: String::from("path/to/file.json"),
```
Mohiiit, line 93: nit: let's have a file in the constant just like `BLOB_DATA_FILE_NAME` and others and let's use that

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Calculating the first block number to for which a batch needs to be assigned
+        let first_block_to_assign_batch = config
+            .service_config()
+            .min_block_to_process
+            .map_or(latest_block_in_db, |min_block| max(min_block, latest_block_in_db));
+
+        for block_num in first_block_to_assign_batch..last_block_to_assign_batch + 1 {
+            self.assign_batch_to_block(block_num, config.clone()).await?;
+        }
+        tracing::trace!(log_type = "completed", category = "BatchingWorker", "BatchingWorker completed.");
```
Mohiiit:
```suggestion
        tracing::debug!(log_type = "completed", category = "BatchingWorker", "BatchingWorker completed.");
```

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+        let latest_batch = config.database().get_latest_batch().await?;
+        let latest_block_in_db = latest_batch.map_or(0, |batch| batch.end_block);
+
+        // Calculating the first block number to for which a batch needs to be assigned
+        let first_block_to_assign_batch = config
+            .service_config()
+            .min_block_to_process
+            .map_or(latest_block_in_db, |min_block| max(min_block, latest_block_in_db));
+
+        for block_num in first_block_to_assign_batch..last_block_to_assign_batch + 1 {
```
Mohiiit:
```suggestion
        for block_num in first_block_to_assign_batch..=last_block_to_assign_batch {
```

Mohiiit (approved):
gg! very few changes, mostly nit, looks good!

orchestrator/crates/orchestrator/src/tests/database/mod.rs
Trantorian1: No problem :D

orchestrator/crates/orchestrator/src/worker/event_handler/service.rs
Trantorian1: :raised_eyebrow:

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
Trantorian1: Oh yea totally, this is def out of scope for this PR haha. Was moreso curious as to how the logging worked in this context. Thanks for the explanation!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
Trantorian1: Noice :ok_hand: Can you confirm if these are passing with the stateful compression logic or not?

Trantorian1 (approved):
Lgtm!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Yes. They are passing :)

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
heemankv: +1 @itsparser, let's fix this

heemankv: +1 @itsparser, let's fix this

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Being resolved in, [Pr 653](https://github.com/madara-alliance/madara/pull/653)

orchestrator/crates/orchestrator/src/tests/utils.rs
prkpndy: Why do we need a constant for a dummy path in test case?

orchestrator/crates/orchestrator/src/types/batch.rs
0xvasanth: Based in the conversation with @prkpndy , here is the usage of the both the identifier 
index - to maintain the batch number 
id - UUID he is using this for traceability in the system with logs and other monitoring systems
and Mongo DB default _id 

And here is my question, due to this we have 3 identifier in the system which fells like redundant?

I want to have single identifier which seem not possible with @prkpndy requirement, better is make any one of the attribute to map to `_id` so we will have two identifier with unique index

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: +1

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: Let's make it info

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -0,0 +1,277 @@
...
+use std::sync::Arc;
+
+pub struct BatchingTrigger;
+
+#[async_trait::async_trait]
+impl JobTrigger for BatchingTrigger {
+    /// 1. Fetch the latest completed block from Starknet chain
+    /// 2. Fetch the last batch and check its `end_block`
+    /// 3. Assign batches to all the remaining blocks and store the squashed state update in storage
+    async fn run_worker(&self, config: Arc<Config>) -> color_eyre::Result<()> {
```
heemankv, line 26: let's change it to info!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
heemankv: Let's implement the trait

orchestrator/crates/orchestrator/src/worker/mod.rs
heemankv: Let's keep the panic here, since it's the parent of the event handler, (we don't wanna panic in the event handler)

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
prkpndy: @itsparser has done this in a separate PR 🎉

orchestrator/crates/orchestrator/src/worker/event_handler/jobs/models.rs
prkpndy: Have done it in this PR! Reviewed this on call as well

orchestrator/crates/orchestrator/src/types/batch.rs
prkpndy: Will add this in stateless compression PR :)

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
prkpndy: Done

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: I don't think we can implement Default for a type not defined in the current crate.  Will need to create a wrapper around StateDiff. This will require too much refactor + redundant code to be used only at one place. I think we can skip this? Let me know if I am missing something

heemankv (approved):

0xvasanth (approved):

PR #626, opened 2025-05-06
update: updated default values in the presets

Mohiiit:
we were using wrong values in the default presets, updated those:

1. number of events: 5000
2. max-state-diff size: 4000
3. message segment length: 3700

values are taken from: https://github.com/starkware-libs/sequencer/blob/9ebd907098473236d6470d38ec2b0f13b84d6f2a/config/sequencer/default_config.json#L22

### DIFF
--- a/configs/chain_config.example.yaml
+++ b/configs/chain_config.example.yaml
@@ -60,11 +60,11 @@ bouncer_config:
     # Maximum number of steps per block
     n_steps: 40000000
     # Maximum length of message segments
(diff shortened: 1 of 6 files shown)


Trantorian1 (approved):
Looks good! Note that some tests on the chain config are failing in the CI, you will need to update those to reflect the new defaults :+1:

Mohiiit:
my bad about that, it should be fixed after latest push!

heemankv (approved):

PR #633, opened 2025-05-08
ci(hive): added starknet hive to tests

Trantorian1:
## Pull Request type

- Testing

- No e2e RPC testing
- Release pipeline does not work
- Duplicate tests in CI

## What is the new behavior?

This pull requests introduces some improvements in the CI pipeline, namely:

- Added support for [`straknet-hive`](https://github.com/madara-alliance/starknet-hive).
- Updated and much faster release pipeline, with support for nightly builds and _much_ better artefact caching (building a release image takes ~35 min in CI, but only ~1:30 min when cached!).
- Updated Madara docker image with _much_ better layer caching.
- Removed duplicate tests in CI, now coverage tests are the default.

> [!NOTE]
> Starknet hive tests are currently **failing**. This is because we still have to fix some issues in Madara for them to pass.

## Runner Strategy

The plan is to implement **image building**, **hive tests** and **image publishing** as part of a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue). The required stubs have already been put in place for this to work, but atm the above steps will only run on `workflow-dispatch` so have to be triggered manually. The plan is to automate this once Madara passes all the Starknet Hive tests.

Once Merge queue has been enabled, the workflow would be as follows:

1. Open a pull request.
2. _light_ tests run for fast feedback.
3. Once a PR has received 2 reviews, an admin can mark it as 'merge when ready'.
4. This will put the PR in the merge queue and run _heavier_, more time intensive tests against it.
5. If all those test pass, the PR will be successfully merged into main and a new `nightly` release will be created. Otherwise, the PR reverts to step 3.

The idea behind this is to provide a tighter feedback loop to maintainers by reducing the amount of test we run upfront and keeping only those most essential. We still run the heavier stuff, just not immediately.

## Release Pipeline

> [!WARNING]
> **merge queue** has to be enabled before the release pipeline is put into play.

This PR introduces a proper working release pipeline for Madara. It has been written in such a way as to be easily extensible to the other targets in the repo. This pipeline is run **on each merge**, generating a new nighlty release each time. It introduces 4 new tags:

- `latest`: latest stable release.
- `v[0-9]+(\\.[0-9])*(\\.[0-9])*`: stable release version, based off the version in the root `Cargo.toml`.
- `nightly`: latest nightly release.
- `nightly-[a-f0-9]{8}`: nightly release wih a commit sha.

Here's a cool thing: creating a new stable release does not require a new docker build. Instead, it just downloads the latest nightly image, re-tags it, and pushes the changes to ghcr. This only takes about a minute or so.

## Other information

The current CI is _very_ slow. There remains a lot of low-hanging fruits that could make this _much_ faster:

- **Parallel builds**: currently we build Madara, Orchestrator, Snos and Bootstrapper in a single _serial_ job. This only makes sense because the repo is not set up to run `cargo build -p orchestrator` or similar yet. A lot of time can be gained by splitting each build into its own job, which would essentially allow for parallel compilation of each target.

- **Better caching**: I am not convinced that the current caching implementation works as intended. A lot of dependencies in the `setup-rust` step are downloaded over and over again and this does not seem to re-use results from previous runs. I think a simpler and faster setup is possible by using something like [`setup-rust-toolchain`](https://github.com/marketplace/actions/setup-rust-toolchain-for-github-ci).

- **CI based off file changes**: Orchestrator uses Madara but Madara does not use Orchestrator. It does not make sense to run Madara tests when we modify Orchestrator code, since there is not dependence in that direction (the reverse is not true!) . It would take a lot of time off Orchestrator PRs if we only ran Orchestrator-related CI.

No


.github/workflows/pull-request-main-ci.yml
```diff
@@ -46,14 +47,6 @@ jobs:
...
     secrets: inherit
 
-  # Run Madara tests
-  test-madara:
-    needs: [update-db-version, build-madara-cairo]
-    uses: ./.github/workflows/task-test-madara.yml
-    with:
-      madara-binary-hash: ${{ needs.build-madara-cairo.outputs.madara-binary-hash }}
-    secrets: inherit
-
```
Mohiiit, line 56: nice!

.github/workflows/task-build-madara.yml
```diff
@@ -66,8 +66,8 @@ jobs:
           python-version: ${{ env.BUILD_PYTHON_VERSION }}
           build-snos: true
 
-      - name: Build the artifacts
-        run: |
+      # - name: Build the artifacts
```
Mohiiit: this shouldn't be a comment?

.github/workflows/task-do-nothing-build-nightly.yml
Mohiiit, line 1: is this here for some future scope of work?

.github/workflows/task-do-nothing-publish-nightly.yml
Mohiiit, line 1: same!

.github/workflows/task-do-nothing-test-hive.yml
Mohiiit, line 1: same!

.github/workflows/task-build-madara.yml
Trantorian1: Yep, thanks for pointing that out!

.github/workflows/task-do-nothing-build-nightly.yml
Trantorian1: So the reason why this is here is related to branch status checks and the merge queue.

The way the merge queue works is by running some extra CI as part of a PR _right before_ it is merged into its target branch. Note that there is nothing which states that the CI must _pass_ (this is a bit of a pain in fact). What we want is the reverse: the merge queue should run some extra, time-intensive task and if they fail, abort the merge.

One way we can enforce this is using merge status checks. This is configured inside branch protection rules, and it allows us to specify some checks which need to pass before a branch can be merged (so for example `build-nightly / build-nightly`). This prevents the PR from being merged if certain tasks in the merge queue fail, but it produces another problem.

The issues here is that the Github CI runner does not differentiate merge status checks between the merge queue and `on-pull-request` CI runs. In fact, it does not provide any way to apply these constraints conditionally based off the type of event which fires off the CI (this would be nice!). What this means is that if we enforce that `build-nightly / build-nightly` should pass, then it _must_ pass _on all CI events_.

> A fun fact which really shows you how hopelessly broken the Github CI runner is, is that if you start a `on-pull-request` CI event that does _not_ contain a task which matches _exactly_ the one expected by the merge status checks, then the CI will just run _indefinitely_. For example, if we add `build-nightly / build-nightly` to the merge status check (`build-nightly` is different, and that won't pass! :D), the CI will expect this to pass before the PR can be merged. If `build-nightly / build-nightly` never ran, this will never pass, and the CI will just run indefinitely. Fun!

So any task which is part of the merge status checks must be present in _all_ CI runs. To do this, we need to introduce stubs like those inside of CI events which do not run those checks. Its not enough either to just define a task inside `pull-request-main-ci` as:

```yaml
build-nightly:                         
  runs-on: ubuntu-latest     
  steps:                     
    - name: build-nightly             
      run: exit 1
```

As this is identified as just `build-nightly` by the CI, whereas the _actual_ `task-build-nightly.yml` will identify as `build-nightly / build-nightly`, which does not match.

So really, this is just a hack around the Github CI to get things to work because the runner really isn't quite smart.

.github/workflows/task-publish-stable.yml
```diff
@@ -0,0 +1,151 @@
...
+              ]
+              | sort_by(split(".")
+              | map(tonumber))
+              | last // "zilch"
+            '
+          )
+          VERSION_NEW=$(grep -oP "(?<=^version = \").*(?=\"$)" Cargo.toml)
+
+          if [[ "$VERSION_LATEST" == "$VERSION_NEW" ]]; then
+            echo "ERROR - Release action cannot replace an existing version!"
```
jbcaron, line 75: for pre-releases like beta or RC versions, overwriting might be acceptable

.github/workflows/task-test-hive.yml
```diff
@@ -0,0 +1,73 @@
...
+            --paymaster-private-key "0x077e56c6dc32d40a67f6f7e6625c8dc5e570abe49c0a24e9202e4ae906abcc07" \
+            --udc-address "0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf" \
+            --account-class-hash "0xe2eb8f5672af4e6a4e8a8f1b44989685e668489b0a25437733756c5a34a1d6" \
+            --suite open-rpc
+
+      - name: Stop Madara
+        if: always()
+        run: |
+          echo ${{ steps.madara.outcome }}
+          echo ${{ steps.madara.outcome == 'success '}}
```
jbcaron: The ID `madara` is not defined.

.github/workflows/task-test-hive.yml
jbcaron, line 1: Do the logs from the Madara Docker container appear in case of a failure?

Dockerfile
```diff
@@ -1,40 +1,66 @@
...
+WORKDIR /app
+
+# Note that we do not install cargo chef and sccache through docker to avoid
+# having to compile them from source
+ENV SCCACHE_URL=https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz
+ENV SCCACHE_TAR=sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz
+ENV SCCACHE_BIN=/bin/sccache
+ENV SCCACHE_DIR=/sccache
+ENV SCCACHE=sccache-v0.10.0-x86_64-unknown-linux-musl/sccache
+ENV CHEF_URL=https://github.com/LukeMathWalker/cargo-chef/releases/download/v0.1.71/cargo-chef-x86_64-unknown-linux-gnu.tar.gz
```
jbcaron:
```suggestion
 ENV SCCACHE_VERSION=v0.10.0
ENV SCCACHE_URL=https://github.com/mozilla/sccache/releases/download/${SCCACHE_VERSION}/sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl.tar.gz
ENV SCCACHE_TAR=sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz
ENV SCCACHE_BIN=/bin/sccache
ENV SCCACHE_DIR=/sccache
ENV SCCACHE=sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl/sccache

ENV CHEF_VERSION=v0.1.71
ENV CHEF_URL=https://github.com/LukeMathWalker/cargo-chef/releases/download/${CHEF_VERSION}/cargo-chef-x86_64-unknown-linux-gnu.tar.gz
```

.github/workflows/task-do-nothing-build-nightly.yml
Mohiiit: thanks for the explanation!

Mohiiit (approved):

.github/workflows/task-test-hive.yml
Trantorian1: Sorry forgot to remove this, was just debug output.

.github/workflows/task-test-hive.yml
Trantorian1: Yep! By default the `docker/build-push-action` will build with `--progress plain` so each step is clearly visible.

.github/workflows/task-publish-stable.yml
Trantorian1: Idk, atm I would rather just have it explicitly forbidden to overwrite a release since this could cause issues down the line in pipeline which depend on those tags. We can always force-push manually using the `gh` or `docker` cli but I don't think we should allow this from within the CI. Or maybe we put this behind an `input`? idk.

Dockerfile
```diff
@@ -1,40 +1,68 @@
...
-# Install build dependencies
+# Step 0: setup tooling (rust)
+FROM rust:1.85 AS base-rust
+WORKDIR /app
+
+# Note that we do not install cargo chef and sccache through docker to avoid
+# having to compile them from source
+ENV SCCACHE_VERSION=v0.10.0
+ENV SCCACHE_URL=https://github.com/mozilla/sccache/releases/download/${SCCACHE_VERSION}/sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl.tar.gz
+ENV SCCACHE_TAR=sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz
```
jbcaron:
```suggestion
ENV SCCACHE_TAR=sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl.tar.gz
```

Dockerfile
```diff
@@ -1,40 +1,68 @@
...
+ENV SCCACHE_BIN=/bin/sccache
+ENV SCCACHE_DIR=/sccache
+ENV SCCACHE=sccache-v0.10.0-x86_64-unknown-linux-musl/sccache
```
jbcaron:
```suggestion
ENV SCCACHE=sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl/sccache
```

jbcaron (approved):

PR #643, opened 2025-05-12
refactor(ci): optimize CI run times

Trantorian1:
## Pull Request type

Current Madara CI is a bit of a mess, mostly in terms of CI runtimes. This PR aims to make the madara CI pipeline faster and more reliable.

### Build-related changes

- Unified the building process for all repository crates: now only `cargo build` is required at compilation time. Python, Forge and Cairo are no longer needed.
- Committed all non-rust artifacts to _github packages_. This includes cairo and solidity code. A build docker is provided which will re-generate all artifacts on demand: it can be run with `make artifacts`.
- Removed _all_ project submodules. This is no longer needed since all non-rust dependencies are statically compiled.
- Reworked workspace targets to make it easier to build `madara`, `orchestrator`, `bootstrapper` and `e2e` targets separately. Now `cargo build -p <target>` works for each of these targets. This is used in the CI for parallel builds.
- Use `mold` as the default linker for faster build times (this also shows improvements in the CI).

### CI-related changes

- `madara`, `orchestrator` and `bootstrapper` builds / check steps are now run in parallel.
- `madara`, `orchestrator` and `boostrapper` tests are now run in parallel.
- Remove unsuded CI artifacts (`bootsrapper` and `orchestrator` binaries were not being re-used between tests).
- Much faster setup actions: `rust`, `foundry` and `cairo` setup have been separated into their own action, making the setup faster for tasks which only need one of these deps.
- Better toolchain caching using [`setup-rust-toolchain`](https://github.com/marketplace/actions/setup-rust-toolchain-for-github-ci).
- Better dependency caching with manual caching. Previously dependency caching did not to work on coverage tests because `llvm-cov` stores build targets in a different directory. This has been fixed.
- Separate caching for build and test tasks to avoid cache overwrites.
- Added flaky tests prevention using `nextest` to automatically [retry certain failing tests](https://nexte.st/docs/features/retries/).
- [Better concurrency control](https://nexte.st/docs/configuration/test-groups/) of tests related to eth rpc queries to avoid rate limiting.
- Fixed merge queue being able to run on failed tests.

### Cairo artifacts pipeline

madara-artifacts

- Artifacts are auto-generated and published on prs labeled `artifact-update`.
- Artifacts are auto-downloaded, extracted and copied to source during the build process.
- Artifacts can be downloaded in parallel and will not clobber up existing docker containers in the process.
- If artifacts fail to download, the build process will default to compiling from source. This also lets users compile from source if they want to. This can be done by  calling `make artirfacts`
- Artifacts are downloaded sparingly and only if the `artifacts-version.yaml` is updated.
- Artifacts are versioned based on the pr number inside of `artifacts-version.yaml`.

### Other changes

Though not directly related to this PR, the ETH rpc endpoint we are using has been upgraded. This and the improvements above means that tests which query ethereum state should no longer be causing CI fails.

### Performance improvements

Overall, this has resulted in about a **1.5x** to **3x** improvement in CI speed. This is most notable for `madara`, with all tests now completing in about _10min_! CI times are bit longer for `bootstrapper` and `orchestrator`, taking about _20min_ and _25min_, but this is something which can be investigated in the future to find out if there is any way to optimize the tests there.

> [!TIP]
> As of now, CI runtimes are clearly dominated by [`e2e-orchestrator / e2e-orchestrator`](https://github.com/madara-alliance/madara/actions/runs/15261655907/job/42920606707?pr=643#logs). A low hanging fruit to make CI run even faster would be to optimize those tests to run faster.  **Caching dependencies is prohibitively expensive in this step and only adds time  overall** so this is not a solution.

<div align="center">
  
</div>

No


.github/actions/setup-rust/action.yml
```diff
@@ -0,0 +1,76 @@
...
+      with:
+        toolchain: ${{ inputs.rust-version }}
+        components: cargo, clippy, rustfmt
+        rustflags: ""
+
+    - name: Rust target
+      shell: bash
+      run: |
+        rustc --version --verbose
+        rustc --print cfg
```
jbcaron: Isn't that redundant?

.github/workflows/task-build-madara.yml
```diff
@@ -29,54 +18,30 @@ jobs:
...
-          python-version: ${{ env.BUILD_PYTHON_VERSION }}
-          build-snos: true
+          cache-key: madara
 
-      - name: Build the artifacts
+      - name: Debug
         run: |
-          make artifacts-linux
+          ls -la
+          cat .cargo/config.toml
```
jbcaron: remove it

bootstrapper/Cargo.toml
jbcaron, line 1: It would be better to use a repo on madara-alliance

jbcaron (approved):

.artifact-versions.yml
Mohiiit, line 1: this will get updated automatically? or it should be updated by the dev? 

edit: I got some idea after reading the comments and the code but I believe there could be gaps in my understanding still!

.github/CI_README.md
```diff
@@ -39,90 +33,37 @@ Located in `.github/actions/load-env/`, this action loads environment variables
...
 
-If you encounter issues:
+This repository makes use of the [Github Merge Queue] to separate test runs into two phases:
 
-1. Check if Act is using the correct Docker images
-2. Ensure all required secrets are provided
-3. Check for path-specific issues between different environments
-4. Look at act DEBUG logs `act -v`
+1. A first phase during which lighter and more essential tests are run to provide fast feedback.
+2. A second phase during which heavier tests and release pipelines are run before a merge into `main`.
```
Mohiiit, line 64: so merge-queue won't merge the PR until the second phase tests are passing?

.github/actions/setup-rust/action.yml
```diff
@@ -0,0 +1,70 @@
...
+      id: cache-cargo
+      with:
+        path: |
+          ~/.cargo/bin
+          ~/.cargo/registry
+          ~/.cargo/git
+        key: ${{ runner.os }}-cargo-registry-${{ inputs.cache-key }}-${{ hashFiles('**/Cargo.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-cargo-registry-${{ inputs.cache-key }}
+          ${{ runner.os }}-cargo-registry-
```
Mohiiit, line 39: is this being used to restore anything with this prefix?

.github/workflows/pull-request-merge.yml
```diff
@@ -25,15 +30,43 @@ jobs:
...
     permissions:
       contents: read
       packages: write
       attestations: write
       id-token: write
     secrets: inherit
+
+  # ========================================================================= #
+  #                                MERGE QUEUE                                #
+  # ========================================================================= #
```
Mohiiit, line 48: hmm interesting so we are not running tests as of now!

.github/workflows/task-build-artifacts.yml
```diff
@@ -0,0 +1,96 @@
...
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Load env
+        uses: ./.github/actions/load-env
+
+      - name: Compile artifacts
+        run: make artifacts
```
Mohiiit, line 52: isn't it make linux-artifacts?

edit: got it, it's make artifacts as of now!

.github/workflows/task-build-madara.yml
```diff
@@ -9,74 +11,42 @@ on:
...
-          install-ganache: true
-          ganache-version: ${{ env.BUILD_GANACHE_VERSION }}
-          install-solc: true
-          solc-version: ${{ env.BUILD_SOLC_VERSION }}
-          python-version: ${{ env.BUILD_PYTHON_VERSION }}
-          build-snos: true
+          cache-key: madara
 
-      - name: Build the artifacts
+      - name: Debug
```
Mohiiit: do we need this?

.github/workflows/task-build-madara.yml
```diff
@@ -91,123 +61,22 @@ jobs:
...
-          echo "orchestrator-hash=$HASH" >> $GITHUB_OUTPUT
-          echo "Hash of the orchestrator is: $HASH (with timestamp: $TIMESTAMP)"
-
-          # Path to the compiled Bootstrapper binary
-          BINARY_PATH=./target/release/madara-bootstrapper
-          HASH=$(echo "$TIMESTAMP" | cat - $BINARY_PATH | sha256sum | awk '{ print $1 }')
-          echo "bootstrapper-hash=$HASH" >> $GITHUB_OUTPUT
-          echo "Hash of the bootstrapper is: $HASH (with timestamp: $TIMESTAMP)"
+          # Path to the cairo artifacts
+          ARTIFACTS_PATH=./artifacts.tar.gz
```
Mohiiit, line 60: I am a bit confused about when are these artifacts arrive at this path?

.github/workflows/task-ci-version-file.yml
```diff
@@ -0,0 +1,70 @@
...
+    outputs:
+      update:
+        description: "Whether or not the version file was updated"
+        value: ${{ jobs.update-version-file.outputs.update }}
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Checking DB Version ('${{ inputs.label }}' label only)"
```
Mohiiit: it's checking db version or version file? 

edit: task-ci-version-file checks for both the version files right? the db version and the artifact version?

.github/workflows/task-test-js.yml
```diff
@@ -19,25 +21,28 @@ jobs:
...
         uses: actions/download-artifact@v4
         with:
-          name: cairo-artifacts-${{ inputs.cairo-artifacts-hash }}
-          path: cairo/target/
+          name: build-artifacts-${{ inputs.cairo-artifacts-hash }}
+          path: build-artifacts/
 
-      - name: Make Madara binary executable
-        run: chmod +x target/release/madara
+      - name: debug
```
Mohiiit: this can be removed right?

Makefile
```diff
@@ -177,224 +178,32 @@ restart: clean
 frestart: fclean
 	@$(MAKE) --silent start
 
-.PHONY: snos
-snos:
```
Mohiiit, line 181: shall we keep this tho? this is for creating the os_latest.json file which is snos essentially, wdyt? it's different than other artifacts and I don't see it's been created in the dockerfile too (might have missed)

edit: I missed that in the dockerfile, it's under cairo-lang got it, although it's using snos 0.13.2 which is fine for now!

bootstrapper/build.rs
Mohiiit, line 1: how does this function, can you share some docs on this?

edit: it's more or less about how passing different parent-level changes the output? idk

build-artifacts/build.docker
Mohiiit, line 1: super clean! thanks for this!

Mohiiit (commented):
thanks for the awesome work ser!

this is super clean!

I am a bit confused about the `get_or_compile_artifacts` function, how passing different parent_level is getting the right artifacts?

.artifact-versions.yml
Trantorian1: The way this works is:

1. When a pr is labelled with `artifacts-update`, this will automatically trigger `update-version-artifacts / update-version-file`.
2. This will read in the contents of `.artifacts-version.yml` and check to see if the version was already updated in this pr.
3. If the artifacts version has not yet been updated in this pr, the git bot will automatically push a new commit to the pr which updates the `.artifacts-version.yml`, incrementing the version by 1 and saving the pr number.
4. On a successful update, `update-version-artifacts / update-version-file` will emit as output `update-version-artifacts.outputs.update == true`, which is used by related tasks to build and publish the artifacts.

Essentially the flow is that this file is updated automatically each time you label a pr with `artifacts-update` and this will trigger a rebuild of the artifacts whenever this happens.

.github/CI_README.md
Trantorian1: The way things have been set up that should be the case :) This is a little bit more tricky, you can check out or previous discussion for more information.

https://github.com/madara-alliance/madara/pull/633#discussion_r2086165929

.github/actions/setup-rust/action.yml
Trantorian1: Tbh I didn't touch this part that much, this was the setup which was already being used previously and I didn't change the `key`  or `restore-keys`.

.github/workflows/pull-request-merge.yml
Trantorian1: Yep, that is on purpose! We don't want to duplicate the work we do before the merge queue by doing it again _inside_ the merge queue. Since we know that these tests must have passed for the merge queue to start, it is safe to replace them with stubs. This might seem a bit confusing and is just a limitation of the way in which the Github CI runner works. Check out our previous conversation for more information.

https://github.com/madara-alliance/madara/pull/633#discussion_r2086165929

.github/workflows/task-build-madara.yml
Trantorian1: Hippity hopity, removing this from my git history

.github/workflows/task-build-madara.yml
Trantorian1: `artifacts.tar.gz` are the artifacts which are downloaded from github packages as part of the Madara build script:

https://github.com/madara-alliance/madara/blob/8c107b0a151bbd56ff7d3ea9a4249cdd361ba80f/build-artifacts/src/lib.rs#L140

Trantorian1: Since we are building madara just before, these will be available at the root of the repository.

.github/workflows/task-ci-version-file.yml
Trantorian1: Yep that's it! This used to be a workflow which @jbcaron wrote which was used to update only the `.database-version.yml`, I just updated it to handle version files more generally but I forgot to remove this.

.github/workflows/task-test-js.yml
Trantorian1: Yep!

bootstrapper/build.rs
Trantorian1: Parent levels just refer to the number of directory levels above the current `CARGO_DIR` which are needed to reach the directory root. So for example:

```bash
.
|
|-bootsrapper/            # 1
|  |-build.rs             # parent-level = 1
|  |-Cargo.toml
|
|-madara/                 # 4
   |-crates/              # 3
       |-client/          # 2
          |-db/           # 1
             |-build.rs
             |-Cargo.toml # parent-level = 4

```

Trantorian1: Added this to the function docs

Trantorian1:
> I am a bit confused about the get_or_compile_artifacts function, how passing different parent_level is getting the right artifacts?

This isn't being used to _get_ the right artifacts, rather its used in build scripts so they know how many parent directories they need to climb to reach the root of the repo, which is where they need to download and extract the artifacts.

.github/CI_README.md
Mohiiit: ohh I see, I didn't realize it's part of the same problem

Mohiiit (approved):

PR #644, opened 2025-05-13
fix: madara to handle block 0 state

Mohiiit:
In order to make madara compatible with snos, we need to handle a magic number for the case when we are running snos for block 0 and there is no prev block number, in that case, as per the spec there is a magic number which can be used, but it doesn't fit with u64 range.

docs on it here: https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/starknet/solidity/StarknetState.sol#L19-L39

magic number is: `0x800000000000011000000000000000000000000000000000000000000000000`

so this PR updates madara code to handle this case!
## Pull Request type

- Bugfix

No


Trantorian1:
Hello @Mohiiit! Would it please be possible to add a test for this scenario to avoid potential regression in the future?

Mohiiit:
hey @Trantorian1, I think it would be covered in this PR: https://github.com/madara-alliance/madara/pull/647/files#diff-2bef00c1fcd007ca420ec8e5452b142af9b854b32d2ea1b3bfe95180c8f314deR53

and as you see, we are initialising the core-contract with -1 state, which internally returns 0x800000000000011000000000000000000000000000000000000000000000000 as mentioned here: https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/solidity/StarknetState.sol#L32 

EDIT: my bad, it return -1 in case of ethereum and magic number in case of starknet!

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -404,6 +407,53 @@ pub mod eth_client_getter_test {
             }
         )
     }
+
+    #[tokio::test]
+    async fn test_get_current_core_contract_state_with_initial_block_number() {
```
Trantorian1, line 419: Thanks a lot!

madara/crates/client/settlement_client/src/state_update.rs
```diff
@@ -29,19 +29,30 @@ pub struct StateUpdateWorker {
...
+                self.block_metrics.l1_block_number.record(block_num, &[]);
+                self.backend.write_last_confirmed_block(block_num).map_err(|e| {
+                    SettlementClientError::DatabaseError(format!("Failed to write last confirmed block: {}", e))
+                })?;
+                tracing::debug!("update_l1: wrote last confirmed block number: {}", block_num);
+            }
+            None => {
+                tracing::warn!("update_l1: no valid block number received from L1, skipping block number update");
+            }
+        }
```
Trantorian1: We are already checking this above, is there no way to merge this with the `match state_update.block_number` on line 32?

madara/crates/client/sync/src/sync.rs
```diff
@@ -146,23 +146,21 @@ impl<P: ForwardPipeline> SyncController<P> {
...
-        );
+        let current_head = self.current_l1_head.as_ref().and_then(|h| h.block_number);
+        let probe_block = self.probe.last_val().map(|v| v.block_number);
+
+        let target_block = match (current_head, probe_block) {
+            (Some(head), Some(probe)) => Some(cmp::max(head, probe)),
+            (Some(head), None) => Some(head),
+            (None, Some(probe)) => Some(probe),
+            (None, None) => None,
+        };
```
Trantorian1: Cf @cchudant does this seem correct to you? I'm not sure we should be using the current head as a sync target here, isn't that already checked by the probe?

madara/crates/client/sync/src/sync.rs
```diff
@@ -146,23 +146,21 @@ impl<P: ForwardPipeline> SyncController<P> {
...
 
         // Bound by stop_at_block_n
-        if let Some(stop_at) = self.config.stop_at_block_n {
-            if target_block >= Some(stop_at) {
-                target_block = Some(stop_at)
-            }
+        match (target_block, self.config.stop_at_block_n) {
+            (Some(target), Some(stop_at)) if target >= stop_at => Some(stop_at),
+            _ => target_block,
```
Trantorian1: Same here.

madara/crates/client/settlement_client/src/state_update.rs
Mohiiit: hmm, do you think we can skip the state-update log we have after the block_info? I can make this change then!

madara/crates/client/sync/src/sync.rs
Mohiiit: you mean we shouldn't compare the target and stop_at here? because it's already being compared in the probe?

madara/crates/client/sync/src/sync.rs
Mohiiit: what is the downside to this? I am unable to think of one 😅

cchudant: i think my return None was wrong,
but also your code could be simplified to
```rs
let target_block = cmp::max(current_head, probe_block);
```

because Some(_) > None

cchudant: nice catch! did you run into the issue or just foudn the bug by reading the code?

madara/crates/client/sync/src/sync.rs
cchudant: how about
```rs
fn target_height(&self) -> Option<u64> {
    fn merge_options<T>(a: Option<T>, b: Option<T>, f: impl FnOnce(T, T) -> T) -> Option<T> {
      match (a, b) {
        (Some(a), Some(b)) => Some(f(a, b)),
        (Some(a), _) => Some(a),
        (_, Some(b)) => Some(b),
        _ => None,
      }
    }

    let current_head = self.current_l1_head.as_ref().and_then(|h| h.block_number);
    let probe_block = self.probe.last_val().map(|v| v.block_number);

    let target_block = merge_options(current_head, probe_block, cmp::max);

    // Bound by stop_at_block_n
    let target_block = merge_options(target_block, self.config.stop_at_block_n, cmp::min);
    target_block
}
```

(i've checked the function does not already exist in the std unfortunately. would be great though!)

madara/crates/client/settlement_client/src/state_update.rs
```diff
@@ -13,7 +13,7 @@ use starknet_types_core::felt::Felt;
 
 #[derive(Debug, Clone, Deserialize, PartialEq)]
 pub struct StateUpdate {
-    pub block_number: u64,
+    pub block_number: Option<u64>,
```
cchudant, line 16: I think I don't get it, when is this null?

cchudant: oh i get it, so we get an initial StateUpdate for pre-genesis core contract
okay

Mohiiit: it is null in case we don't have any update on the core-contract, in case of ethereum it returns -1 and in case of starknet it returns magic number: 0x800000000000011000000000000000000000000000000000000000000000000

given both the numbers can't fit in u64, we return null, do you think there is a way to avoid this?

Mohiiit: this is before 0th block is settled

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -138,7 +138,7 @@ impl SettlementClientTrait for EthereumClient {
                     EthereumClientError::Contract(format!("Failed to get state block number: {e:#}")).into()
                 },
             )?;
-        let block_number: u64 = block_number._0.as_u64();
+        let block_number: Option<u64> = block_number._0.try_into().ok();
```
cchudant: i think it's better to be explicit here
```rs
let block_number: Option<u64> = if block_number._0 == I256::MINUS_ONE {
  None // initial contract state
} else { block_number._0.as_u64() };
```
just for documentation purposes

cchudant (approved):
oh okay so, initial core contract state is:
block_n = -1
block_hash = ZERO (matches genesis' previous_block_hash)
state_root = ZERO (all tries are empty, no nodes)

when updating core contract state for genesis block, the snos output must have
prev_output_number = 0x800000000000011000000000000000000000000000000000000000000000000
prev_block_hash = 0x0
prev_state_root = 0x0

and also, looking at the core contract code we may actually see gaps in the returned state updates block numbers (because of v0.14 multi-block state updates i would imagine).

might be worth documenting somewhere

looks good to me!

madara/crates/client/settlement_client/src/state_update.rs
Mohiiit: it's been resolved now!

madara/crates/client/settlement_client/src/eth/mod.rs
Mohiiit: resolved!

madara/crates/client/sync/src/sync.rs
Mohiiit: resolved!

madara/crates/client/settlement_client/src/starknet/mod.rs
```diff
@@ -138,7 +143,11 @@ impl SettlementClientTrait for StarknetClient {
     async fn get_current_core_contract_state(&self) -> Result<StateUpdate, SettlementClientError> {
         let state = self.get_state_call().await?; // Returns (StateRoot, BlockNumber, BlockHash).
         let global_root = state[0];
-        let block_number = u64::try_from(state[1]).ok();
+        let block_number = if state[1] == Felt::from_hex(INITIAL_STATE_BLOCK_NUMBER).unwrap() {
+            None
+        } else {
+            u64::try_from(state[1]).ok()
```
Trantorian1, line 149: Wouldn't it be an invalid state for this to return `None` here? If so, this is probably something which we want to `assert`.

Trantorian1 (approved):
lgtm!

PR #647, opened 2025-05-14
Bootstrapper bump

heemankv:
This PR bumps the bootstrapper to a working state.
- Adds delay in appropriate places (not the best solution, ideally we would want to do wait for transaction).
- Fixes core contract init with correct params. (Support for 0th block)
- Adds the latest stable version of [zaun](https://github.com/karnotxyz/zaun/releases/tag/v1.0.1).
- Changes `bin` name from `madara-bootstrapper` to `bootstrapper` to be consistent with `orchestrator`.
- Build fix:   
    ```
    .update_state(program_output.clone(), program_output, onchain_data_hash, onchain_data_size)
    ```
    - The first argument should be `snos_output`
    - This is fine for now, since this starknet settlement client is not being used and this code will be corrected by merging of L3 support PR on orchestrator.


Cargo.toml
```diff
@@ -349,7 +349,7 @@ stark_evm_adapter = "0.1.1"
 hex = "0.4"
 mockall_double = "0.3.1"
 testcontainers = "0.18.0"
-appchain-core-contract-client = { git = "https://github.com/notlesh/zaun", branch = "madara-monorepo" }
+appchain-core-contract-client = { git = "https://github.com/karnotxyz/zaun", tag = "v1.0.1" }
```
Trantorian1: If I'm correct this version includes my changes to `zaun`. I've been testing this locally and I am getting some errors with the madara CI bootstrapper tests. Are you having the same problems here?

bootstrapper/Dockerfile
```diff
@@ -89,7 +90,7 @@ WORKDIR /app
 COPY . .
 
 # Initialize and update submodules
-RUN git submodule update --init --recursive
+# RUN git submodule update --init --recursive
```
Trantorian1:
```suggestion
```

bootstrapper/Dockerfile
```diff
@@ -114,14 +115,14 @@ RUN cd lib/starkgate-contracts-old && \
     ganache-cli --version && \
     # Start ganache-cli in background with specific host and port
     nohup ganache-cli \
```
Trantorian1, line 115: Why `nohup` inside of Docker? I've never seen this before so just curious

bootstrapper/Dockerfile
```diff
@@ -159,25 +160,25 @@ RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && \
...
+    npm install -g --unsafe-perm ganache@7.9.0 && \
+    # Add environment variables
+    export PATH="/usr/local/lib/node_modules/.bin:$PATH" && \
+    # Add a small delay to let npm finish
+    sleep 5; \
+    fi; \
+    if [ $i -eq 5 ]; then \
+    echo "Failed to install ganache after 5 attempts" && \
+    exit 1; \
+    fi \
```
Trantorian1:
```suggestion
    if which ganache && ganache --version; then \
        echo "Ganache installation verified" && \
        break; \
    else \
        echo "Attempt $i: Ganache not found, trying again..." && \
        echo "Searching for ganache in PATH..." && \
        find / -name ganache 2>/dev/null && \
        # Clear npm cache and reinstall
        npm cache clean --force && \
        npm install -g --unsafe-perm ganache@7.9.0 && \
        # Add environment variables
        export PATH="/usr/local/lib/node_modules/.bin:$PATH" && \
        # Add a small delay to let npm finish
        sleep 5; \
    fi; \
    if [ $i -eq 5 ]; then \
        echo "Failed to install ganache after 5 attempts" && \
        exit 1; \
    fi \
```

bootstrapper/Dockerfile
```diff
@@ -195,17 +196,21 @@ RUN . "$HOME/.asdf/asdf.sh" && \
...
 
 # Copy only the compiled binary and artifacts
 COPY --from=builder /app/target/release/madara-bootstrapper /usr/local/bin/
 COPY --from=builder /app/artifacts /app/artifacts
+COPY --from=builder /app/src/contracts/ /app/src/contracts/
+
 
 # Set working directory
-WORKDIR /app
+#WORKDIR /app
```
Trantorian1:
```suggestion
```

bootstrapper/crates/bootstrapper/Cargo.toml
```diff
@@ -18,15 +22,15 @@ url = "2.4.1"
...
+ethereum-instance = { git = "https://github.com/karnotxyz/zaun", package = "ethereum-instance", tag = "v1.0.1" }
+starkgate-manager-client = { git = "https://github.com/karnotxyz/zaun", package = "starkgate-manager-client", tag = "v1.0.1" }
+starkgate-registry-client = { git = "https://github.com/karnotxyz/zaun", package = "starkgate-registry-client", tag = "v1.0.1" }
+starknet-core-contract-client = { git = "https://github.com/karnotxyz/zaun", package = "starknet-core-contract-client", tag = "v1.0.1" }
+starknet-erc20-client = { git = "https://github.com/karnotxyz/zaun", package = "starknet-erc20-client", tag = "v1.0.1" }
+starknet-eth-bridge-client = { git = "https://github.com/karnotxyz/zaun", package = "starknet-eth-bridge-client", tag = "v1.0.1" }
+starknet-proxy-client = { git = "https://github.com/karnotxyz/zaun", package = "starknet-proxy-client", tag = "v1.0.1" }
+starknet-token-bridge-client = { git = "https://github.com/karnotxyz/zaun", package = "starknet-token-bridge-client", tag = "v1.0.1" }
+zaun-utils = { git = "https://github.com/karnotxyz/zaun", package = "utils", tag = "v1.0.1" }
```
Trantorian1: Noice :partying_face:

bootstrapper/crates/bootstrapper/src/contract_clients/token_bridge.rs
```diff
@@ -57,6 +57,7 @@ impl BridgeDeployable for StarknetTokenBridge {
                 .await
                 .expect("Failed to deploy starkgate manager contract"),
         };
+        sleep(Duration::from_secs(20)).await;
```
Trantorian1: I know it was specified in the scope of this PR that this is just a temporary workaround. What are some solutions that we might consider down the line to wait on the actual bootstrapper? If this works rn then fine but this does have the risk of being quite flaky down the line so just something we need to keep in mind.

bootstrapper/crates/bootstrapper/src/contract_clients/token_bridge.rs
```diff
@@ -246,70 +253,92 @@ impl StarknetTokenBridge {
             .upgrade_to(Bytes::from(manager_calldata.clone()), self.manager.implementation_address(), false)
             .await
             .expect("Failed to initialize starkgate manager");
+        sleep(Duration::from_secs(20)).await;
```
Trantorian1: Gosh this is a lot of `sleep`s. Are we really sure we want to merge this as is? This could easily be waiting for several minutes just do to this :sweat_smile:

Trantorian1 (commented):
This is a lot of `sleep`s! I would be in favor of re-scoping this PR to add some generic transaction wait capabilities, it seems like that would be a good thing to have here. Let me know what you think. I'm really not sold on all the `sleep`s.

bootstrapper/crates/bootstrapper/src/setup_scripts/core_contract.rs
```diff
@@ -68,7 +75,7 @@ impl<'a> CoreContractStarknetL1<'a> {
         // )
         core_contract_client
             .upgrade_to_core_contract(
-                0u64.into(), // block number
+                I256::from_str("-1").unwrap(),
```
Trantorian1: Would it also be possible to add some context as to what this is doing? cf [this comment](https://github.com/madara-alliance/madara/pull/644#issuecomment-2885634489) by @Mohiiit

heemankv:
> This is a lot of `sleep`s! I would be in favor of re-scoping this PR to add some generic transaction wait capabilities, it seems like that would be a good thing to have here. Let me know what you think. I'm really not sold on all the `sleep`s.

Your views are correct, 
but the scope of this PR is to update the monorepo bootstrapper with all the changes that allow it to run at-least, meawhile the next few PRs would be working on re-scoping a lot of changes in the bootstrapper to be able to make it more resilient and prod ready!

bootstrapper/crates/bootstrapper/src/contract_clients/token_bridge.rs
heemankv: Yes, we are working on a re-work of the bootstrapper to make it more resilient, it must be followed in upcoming PRs.

bootstrapper/crates/bootstrapper/src/contract_clients/token_bridge.rs
heemankv: Yeah, the goal is to at-least make it work for now, will be working on removing all these soon!

bootstrapper/crates/bootstrapper/src/setup_scripts/core_contract.rs
heemankv: added the comment mentioning : https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/solidity/StarknetState.sol#L32

madara/crates/client/devnet/src/lib.rs
```diff
@@ -460,7 +460,7 @@ mod tests {
         Duration::from_secs(500000),
         true
     )]
-    #[case::should_work_across_block_boundary(true, true, None, Duration::from_millis(500), true)]
+    #[case::should_work_across_block_boundary(true, true, None, Duration::from_secs(1), true)]
```
heemankv, line 463: flaky test, need to take a deeper look | @Mohiiit

bootstrapper/src/main.rs
```diff
@@ -73,6 +73,37 @@ pub enum CoreContractMode {
...
+    pub dev: bool,
+    pub core_contract_mode: CoreContractMode,
+    pub l2_deployer_address: Option<String>,
+    pub core_contract_address: Option<String>,
+    pub core_contract_implementation_address: Option<String>,
+    pub udc_address: Option<String>,
+    pub l1_eth_bridge_address: Option<String>,
+    pub l2_eth_token_proxy_address: Option<String>,
+    pub l2_eth_bridge_proxy_address: Option<String>,
+}
```
Trantorian1: Why are we using a duplicate struct here? If this is trying to implement a type-safe builder pattern then consider using generics instead.

Trantorian1 (commented):
The `ConfigFile` and `IntermediateConfigFile` setup seems a bit jank, perhaps there is a better way to do this? It feels strange to be duplicating a struct just to validate that certain of its fields are present. Consider using a generics to enforce a type-safe builder pattern instead.

Trantorian1 (commented):
.

bootstrapper/src/main.rs
Mohiiit: I have added something, not sure if you were referring the same thing tho!

bootstrapper/src/main.rs
```diff
@@ -135,14 +102,14 @@ pub struct ConfigFile {
     pub l2_eth_bridge_proxy_address: Option<String>,
 }
 
-impl Default for IntermediateConfigFile {
+impl Default for ConfigBuilder {
     fn default() -> Self {
         Self {
             eth_rpc: Some("http://127.0.0.1:8545".to_string()),
             eth_priv_key: Some("0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80".to_string()),
+            rollup_priv_key: Some("0xabcd".to_string()),
```
Trantorian1, line 110: Shouldn't this be `None` by default?

bootstrapper/src/main.rs
Trantorian1, line 1: Thanks for the changes!

bootstrapper/src/tests/mod.rs
```diff
@@ -150,8 +150,7 @@ fn kill_process_on_port(port: u16) {
 }
 
 fn get_test_config_file() -> ConfigFile {
-    let intermediate_config = IntermediateConfigFile::default();
-    ConfigFile::try_from(intermediate_config).expect("Failed to convert intermediate config to final config")
+    ConfigBuilder::default().build().expect("Failed to convert config builder to final config")
```
Trantorian1, line 153: Should we perhaps not also be calling `merge_with_env` here as well?

Trantorian1 (approved):

0xvasanth (approved):

bootstrapper/src/tests/mod.rs
Mohiiit: I have put default values for tests to pass, so no need to merge it with env

bootstrapper/src/main.rs
Mohiiit: this is not an issue as a default value, we create an account using this key which performes the transactions then, so this is good to go!

Issue #613, opened 2025-04-25
Orchestrator Prod Readiness : Remove hardcoding


heemankv:
Make timeouts and max retries .env driven

Noting down all the hardcoded values here : 

Compiled OS path 
    - It's hardcoded to : `../../../../build/os_latest.json` (but it compiles internally)

- DA job 
    - max_process_attempts
    - max_verification_attempts
    - verification_polling_delay_seconds

- Proving Job
    - max_process_attempts
    - max_verification_attempts
    - verification_polling_delay_seconds
    - max_paralle_jobs_to_create

- State Update job 
    - max_process_attempts
    - max_verification_attempts
    - verification_polling_delay_seconds

- SNOS Job
    - max_process_attempts
    - max_verification_attempts
    - verification_polling_delay_seconds

heemankv:
Status update from codebase review on 2026-05-05.

Classification: Still relevant.

Current state:
Still relevant. Many retry counts, queue constants, and polling delays are still fixed in code.

Keeping this issue open based on the current tree.

PR #651, opened 2025-05-16
Orchestrator : SNOS Job Trigger Enhancements

### LINKED ISSUE #591: Orchestrator Prod Readiness  : Optimise SNOS job picking logic
Currently we have observed when multiple jobs are to be picked, the orchestrators pick SNOS in a `random` order.
Given SNOS is a blocking job, we want to optimise it so that the first few (any pre-decided number) blocks (in respect to the latest block of chain) are processed first.

It should be a practice in orchestrator to not delete jobs

Create jobs in gridy fashion and process them conservatively  
OR 
Create jobs in conservative fashion and process them accordingly

- can put a limit of number of created jobs

Scenario : 
- Madara has created many blocks 
- Orchestrators is lagging behind in doing snos jobs by a big number (say 500)

Problems Faced in SNOS : 
- Orchestrator picked up SNOS at random without any order. (Why this happens as such ? Is our queue not FCFS)
- Orchestrator doesn't validate if any previous jobs were completed or not, worker's logic should be to create jobs for blocks in increasing order + with a picking logic mentioned below.

QUEUE : 
```
For standard queues:

They provide "at-least-once" delivery, but don't guarantee the exact order of messages
Messages can occasionally be delivered in a different order than they were sent
There's no strict FCFS behavior due to the distributed nature of the service
```

Picking Logic : Next 20 ? 50 ? 100 ? or Pick any

SNOS worker : 
- Need to optimise for batched picking of jobs 
- uses picking Logic

ProofCreation worker : 
- Validate against MAX_PARALLEL limit.
- uses picking Logic

DataSubmission worker : 
- uses picking Logic

StateTransition worker : 
- Pick based on last block.

How does this job creation logic support the job picking logic

heemankv:
# SNOS Job Trigger Enhancements

## Pull Request type

- Feature
- Testing

Resolves: #591 

The current implementation sequentially processes blocks from the last completed block to the latest block from the sequencer. It doesn't handle blocks that might have been missed in previous runs and doesn't have a mechanism to limit the number of jobs created at once.

## What is the new behavior?

- Added detection and prioritization of missing blocks within already processed ranges
- Implemented configurable concurrent job limits through `max_concurrent_created_snos_jobs` setting
- Enhanced block processing logic to prioritize missing blocks before creating jobs for new blocks
- Improved logging with detailed statistics on block processing decisions
- Added comprehensive test coverage for both limited and unlimited job creation scenarios

No

## Other information

This enhancement ensures the completeness of blockchain processing by detecting and prioritizing any blocks that were missed in previous runs. The new configurable job limits prevent system overload during high-throughput periods while maintaining optimal processing efficiency.


.env.test
```diff
@@ -139,3 +139,14 @@ L2_MULTISIG_ADDRESS=""
 VERIFIER_ADDRESS=""
 
 ETH_FORK_URL=https://api.zan.top/eth-mainnet
+
+
+MADARA_ORCHESTRATOR_MAX_BLOCK_NO_TO_PROCESS=99999
+MADARA_ORCHESTRATOR_MIN_BLOCK_NO_TO_PROCESS=0
+MADARA_ORCHESTRATOR_MAX_CONCURRENT_CREATED_SNOS_JOBS=3
+
+MADARA_ORCHESTRATOR_MAX_CONCURRENT_PROOF_REGISTRATION_JOBS=10
```
Mohiiit: can you add a comment about what's the difference between concurrent and parallel jobs here?

orchestrator/crates/orchestrator/src/tests/workers/snos/mod.rs
Mohiiit, line 1: I think this could be simplified using the fixture

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
 
 #[async_trait]
 impl JobTrigger for SnosJobTrigger {
-    /// 1. Fetch the latest completed block from the Starknet chain
-    /// 2. Fetch the last block that had a SNOS job run.
-    /// 3. Create SNOS run jobs for all the remaining blocks
```
Mohiiit: let's update this comment!

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +442,29 @@ impl DatabaseClient for MongoDbClient {
         ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
         Ok(jobs)
     }
+
+    #[tracing::instrument(skip(self), fields(function_type = "db_call"), ret, err)]
+    async fn get_jobs_by_type_and_status(
+        &self,
+        job_type: JobType,
+        job_status: JobStatus,
```
Mohiiit: do you think it would be nice to have vec[JobStatus] here?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
 
-        tracing::debug!(latest_block_number = %latest_block_number, "Fetched latest block number from starknet");
+        // Get max block number to process up to
+        let max_block_to_process = match config.service_config().max_block_to_process {
+            Some(max_block) => min(max_block, latest_block_from_sequencer),
+            None => latest_block_from_sequencer,
+        };
 
-        let latest_job_in_db = config.database().get_latest_job_by_type(JobType::SnosRun).await?;
+        tracing::debug!(max_block_to_process = %max_block_to_process, "Fetched latest block number from starknet");
```
Mohiiit: the comment is wrong here!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            .collect();
+
+        // Calculate total pending and created jobs
+        let total_pending_and_created: u64 = (pending_jobs.len() + created_jobs.len()) as u64;
+
+        // Check if we have a job limit from config
+        let available_job_slots = match config.service_config().max_concurrent_created_snos_jobs {
+            // If limit is set, check if we have room for more jobs
+            Some(max_jobs) => {
+                if total_pending_and_created >= max_jobs as u64 {
```
Mohiiit: shouldn't we include the lockedForProcessing jobs as well?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+        let mut jobs_to_create = Vec::new();
+
+        // 1. First identify missing blocks in already "completed" range
+        // Ensure we don't try to process blocks below min_block_to_process
+        let range_start = if min_block_to_process > 0 { min_block_to_process } else { 0 };
+        let missing_blocks: Vec<u64> = (range_start..=last_completed_block)
+            .filter(|block| !processed_blocks.contains(block))
+            // FIX: Also filter out blocks that are already pending or created
+            .filter(|block| !pending_and_created_blocks.contains(block))
+            .collect();
```
Mohiiit: we are missing the lockedForProcessing here too

also this is a very heavy operation fs, one way I can think of to make it cheaper is to have range_start compared with the latest state of the core-contract, given a block has been settled, we don't really need to run anything on that again!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            })
+            .max();
+
+        let last_completed_block = match last_completed_block {
+            Some(block) => block,
+            None => {
+                tracing::info!("No previously completed SNOS jobs found, starting from min_block_to_process");
+                // We use min_block_to_process - 1 so that the first block to be processed will be min_block_to_process
+                // This ensures we don't incorrectly assume block 0 is processed when no blocks are processed
+                min_block_to_process.saturating_sub(1)
```
Mohiiit: this will return 0 in case of min_block_to_process is zero

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            }
+        };
+
+        tracing::debug!(last_completed_block = %last_completed_block, "Found last completed SNOS block");
```
Mohiiit: this would result in wrong log in case min_block_to_process is 0

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+                Err(e) => {
+                    tracing::warn!(
+                        job_id = %job.id,
+                        internal_id = %job.internal_id,
+                        error = %e,
+                        "Failed to parse job internal ID while building pending/created blocks set"
+                    );
+                    None
+                }
+            })
```
Mohiiit: I think we should make it an util function as of now, it's being repeated quite a bit!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            }
+        };
 
-        for block_num in block_start..latest_block_number + 1 {
+        // Get all jobs we need to create, starting with missing blocks
+        let mut jobs_to_create = Vec::new();
+
+        // 1. First identify missing blocks in already "completed" range
+        // Ensure we don't try to process blocks below min_block_to_process
+        let range_start = if min_block_to_process > 0 { min_block_to_process } else { 0 };
```
Mohiiit: this check doesn't make sense, it would be zero or min_block_to_process anyway!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            // If no limit is set, we can create jobs up to the max block
+            None => {
+                tracing::debug!("No maximum concurrent SNOS jobs limit specified in config");
+                max_block_to_process - last_completed_block
+            }
+        };
 
-        for block_num in block_start..latest_block_number + 1 {
+        // Get all jobs we need to create, starting with missing blocks
+        let mut jobs_to_create = Vec::new();
```
Mohiiit: we know the size of this vector already right?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+            missing_blocks.iter().take(available_job_slots as usize).copied().collect::<Vec<_>>();
+
+        jobs_to_create.extend(missing_blocks_to_add.iter());
+
+        // Calculate remaining available slots
+        let remaining_slots = if missing_blocks_to_add.len() as u64 <= available_job_slots {
+            available_job_slots - missing_blocks_to_add.len() as u64
+        } else {
+            0
+        };
```
Mohiiit:
```suggestion
       let missing_blocks_count = min(missing_blocks.len(), available_job_slots as usize);
       jobs_to_create.extend(missing_blocks.iter().take(missing_blocks_count).copied());

       let remaining_slots = available_job_slots - missing_blocks_count as u64;
```

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
Mohiiit: even better if we can make it: BTreeSet<u64> given we are sorting it finally!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,39 +19,231 @@ pub struct SnosJobTrigger;
...
+
+            // Collect all candidate blocks from block_start up to max_block_to_process
+            // that aren't already in pending or created status
+            let candidate_blocks: Vec<u64> = (block_start..=max_block_to_process)
+                .filter(|block| !pending_and_created_blocks.contains(block))
+                .collect();
+
+            // Take only as many blocks as we have slots for
+            let new_blocks_count = if config.service_config().max_concurrent_created_snos_jobs.is_none() {
+                candidate_blocks.len() // No limit, take all candidate blocks
```
Mohiiit: this looks very heavy, we would have a limit always tho? that's whole point right? we won't be putting none post this? or we can?

.env.test
heemankv: These are removed, not needed.

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Valid, I have implemented it now and optimised further by using it for pending and created jobs

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: There are two optimisation happening here : 
- Ensuring that all missing blocks are created and sent to be processed at priority.
- Ensuring that only a certain amount of jobs are being created at a time based on the pre-existing jobs in the db.

Having `max_concurrent_created_snos_jobs` as `None` allows the user to not have the second optimisation in place. which is bound for scenarios when we don't care about the order of processing as such.

Although we could argue for a `default value` for this variable.
Upon further inspection I notice that we have this trend in all the `ServiceParams` 
```
#[derive(Debug, Clone)]
pub struct ServiceParams {
    pub max_block_to_process: Option<u64>,
    pub min_block_to_process: Option<u64>,
    pub max_concurrent_created_snos_jobs: Option<usize>,
    pub max_concurrent_snos_jobs: Option<usize>,
    pub max_concurrent_proving_jobs: Option<usize>,
}
```

As of now, I would like to proceed ahead with the `None` option and fix this for all the `ServiceParams` as a part of the `file based config` issue. (#613)

WDYT @Mohiiit  ???

EDIT: After offline discussion we have decided to change it to u64 in this [commit](https://github.com/madara-alliance/madara/pull/651/commits/2366f27d40d03d3e7b44cf9a40acb813cb3fa546), and we'll later change the other params in a separate PR.

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +442,35 @@ impl DatabaseClient for MongoDbClient {
...
+        job_type: JobType,
+        job_status: Vec<JobStatus>,
+    ) -> Result<Vec<JobItem>, DatabaseError> {
+        let start = Instant::now();
+
+        // For MongoDB to correctly filter by multiple statuses, we need an $in query
+        let filter = doc! {
+            "job_type": bson::to_bson(&job_type)?,
+            "status": {
+                // TODO: Check that the conversion leads to valid output!
```
Trantorian1: This seems like a good thing to test :sweat_smile: Is this still the case?

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -556,9 +556,15 @@ pub(crate) fn get_env_params() -> EnvParams {
     let max_concurrent_proving_jobs: Option<usize> =
         env.and_then(|s| if s.is_empty() { None } else { Some(s.parse::<usize>().unwrap()) });
 
+    let env = get_env_var_optional("MADARA_ORCHESTRATOR_MAX_CONCURRENT_CREATED_SNOS_JOBS")
```
Trantorian1: Is there anything which `get_env_var_optional` does which is different from `std::env::var`?

orchestrator/crates/orchestrator/src/tests/config.rs
```diff
@@ -556,9 +556,15 @@ pub(crate) fn get_env_params() -> EnvParams {
...
+        .expect("Couldn't get max concurrent proving jobs");
+    let max_concurrent_created_snos_jobs: Option<usize> =
+        env.and_then(|s| if s.is_empty() { None } else { Some(s.parse::<usize>().unwrap()) });
```
Trantorian1:
```suggestion
        env.and_then(|s| s.is_empty().then(|| s.parse::<usize>().unwrap()));
```

orchestrator/crates/orchestrator/src/tests/workers/snos/mod.rs
```diff
@@ -108,14 +134,16 @@ async fn test_snos_worker(#[case] db_val: bool) -> Result<(), Box<dyn Error>> {
...
 
-    // mocking block call
+    // Mock block_number RPC call
     let rpc_block_call_mock = server.mock(|when, then| {
         when.path("/").body_includes("starknet_blockNumber");
         then.status(200).body(serde_json::to_vec(&response).unwrap());
     });
 
+    // Run the worker
     crate::worker::event_handler::triggers::snos::SnosJobTrigger.run_worker(services.config).await?;
```
Trantorian1: I am a bit confused by this test. Are we running any assertions against the state of the run? ie: making sure that the correct jobs have been processed.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,53 +19,120 @@ pub struct SnosJobTrigger;
...
+        // TODO: This process to find last_completed_block is operation heavy, i.e for big list it will take time,
+        // maybe we can ask mongodb to return in descending order
+        let completed_jobs = db
+            .get_jobs_by_type_and_status(JobType::SnosRun, vec![JobStatus::Completed])
+            .await
+            .wrap_err("Failed to fetch completed SNOS jobs")?;
+
+        // Create set of completed block numbers for efficient lookup
+        let processed_blocks: HashSet<u64> =
+            completed_jobs.iter().filter_map(|job| parse_block_number(&job.internal_id)).collect();
```
Trantorian1: It is possible for a completed job not to have a block number? Wondering since `filter_map` is being used here.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,53 +19,120 @@ pub struct SnosJobTrigger;
...
+
+        // Get all jobs with relevant statuses
+        let db = config.database();
+
+        // TODO: This process to find last_completed_block is operation heavy, i.e for big list it will take time,
+        // maybe we can ask mongodb to return in descending order
+        let completed_jobs = db
+            .get_jobs_by_type_and_status(JobType::SnosRun, vec![JobStatus::Completed])
+            .await
+            .wrap_err("Failed to fetch completed SNOS jobs")?;
```
Trantorian1: How big can this get? Is there a bound to the number of values we are retrieving from DB or will this just keep growing as SNOS completes more and more jobs?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,53 +19,120 @@ pub struct SnosJobTrigger;
...
+        tracing::debug!(last_completed_block = %last_completed_block, "Found last completed SNOS block");
+
+        // Fetch jobs in a single query with combined statuses
+        let pending_statuses = vec![JobStatus::PendingRetry, JobStatus::Created];
+
+        // Get pending/created jobs
+        let pending_jobs = db
+            .get_jobs_by_type_and_status(JobType::SnosRun, pending_statuses)
+            .await
+            .wrap_err("Failed to fetch pending/created SNOS jobs")?;
```
Trantorian1: Same here, is there a bound on the number of results we are collecting?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,53 +19,120 @@ pub struct SnosJobTrigger;
...
+                candidate_blocks.len()
+            } else {
+                min(remaining_slots as usize, candidate_blocks.len())
             };
 
+            blocks_to_process.extend(candidate_blocks.into_iter().take(new_blocks_count));
+        }
+
+        // Sort blocks to ensure we process in ascending order
+        blocks_to_process.sort();
```
Trantorian1: Maybe we should mention here that this is necessary because blocks can end up out-of-order if we miss a previous block and then retry it?

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Oh, I believe this got created when we copied the code from above. have removed this.
It was a TODO as a part of orchestrator/restructure PR, which has been merged.

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: It does validate if the field is empty and passes `None` accordingly, if this is what you are asking.
` Ok(s) if s.is_empty() => Ok(None),`


Note : 
Have updated the code to use `get_env_var_default`, since we are no longer keeping `max_concurrent_created_snos_jobs` as Option.

orchestrator/crates/orchestrator/src/tests/config.rs
heemankv: Note :
Have updated the code to use `get_env_var_default`, since we are no longer keeping `max_concurrent_created_snos_jobs` as Option, leading to this being outdated.

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -442,4 +441,34 @@ impl DatabaseClient for MongoDbClient {
...
+        job_type: JobType,
+        job_status: Vec<JobStatus>,
+    ) -> Result<Vec<JobItem>, DatabaseError> {
+        let start = Instant::now();
+
+        // For MongoDB to correctly filter by multiple statuses, we need an $in query
+        let filter = doc! {
+            "job_type": bson::to_bson(&job_type)?,
+            "status": {
+                "$in": job_status.iter().map(|job_status| bson::to_bson(job_status).unwrap_or(Bson::Null)).collect::<Vec<Bson>>()
```
0xvasanth: since Null might cause issue in Mongo And expect alway a list as input
kindly make sure Null is not cause any issue

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
0xvasanth: +1, We can move this Identifying / dectecting failed block mechanisam to mongodb pipeline 
Ref: https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/
with a grain of salt to not knock out mongo in the process

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -18,52 +19,111 @@ pub struct SnosJobTrigger;
...
+        blocks_to_process.extend(missing_blocks_to_add.iter().copied());
+
+        // Calculate remaining slots
+        let remaining_slots = available_job_slots.saturating_sub(blocks_to_process.len() as u64);
+
+        // Add new blocks if slots available
+        if remaining_slots > 0 {
+            let new_blocks_count = min(remaining_slots as usize, candidate_blocks.len());
+            blocks_to_process.extend(candidate_blocks.into_iter().take(new_blocks_count));
+        }
```
0xvasanth: Optimise this code.😍

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Simplified!

heemankv (commented):
Done with self review.

e2e-tests/tests.rs
```diff
@@ -102,7 +102,8 @@ impl Setup {
             "MADARA_ORCHESTRATOR_L1_CORE_CONTRACT_ADDRESS".to_string(),
             starknet_core_contract_address.to_string(),
         );
-        env_vec.insert("MADARA_ORCHESTRATOR_MAX_BLOCK_NO_TO_PROCESS".to_string(), l2_block_number);
+        env_vec.insert("MADARA_ORCHESTRATOR_MAX_BLOCK_NO_TO_PROCESS".to_string(), l2_block_number.clone());
+        env_vec.insert("MADARA_ORCHESTRATOR_MIN_BLOCK_NO_TO_PROCESS".to_string(), "0".to_string());
```
0xvasanth: It is expected to run only one block during test case in e2e right?

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
```diff
@@ -47,19 +47,32 @@ pub trait DatabaseClient: Send + Sync {
...
         limit: Option<i64>,
     ) -> Result<Vec<JobItem>, DatabaseError>;
+
+    /// get_missing_jobs_by_type_and_caps - Get all the missed jobs by type and block number limits
+    async fn get_missing_block_numbers_by_type_and_caps(
+        &self,
+        job_type: JobType,
+        lower_cap: u64,
+        upper_cap: u64,
+        limit: Option<i64>,
```
0xvasanth: can we considering change this to u64

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+                    "complete_range": {
+                        "$range": [lower_limit, upper_limti]
+                    }
+                }
+            },
+            doc! {
+                "$project": {
+                    "missing_blocks": {
+                        "$setDifference": [
+                            "$complete_range",
```
0xvasanth: can you confirm this with some test case so we can we confident with this pipeline

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+        // Constructing the aggregation pipeline
+        let mut pipeline = vec![
+            doc! {
+                "$facet": {
+                    "existing_data": [
+                        doc! {
+                            "$match": {
+                                "job_type": job_type_bson,
+                                "metadata.specific.block_number": {
+                                    "$gte": lower_limit,
```
0xvasanth:
```suggestion
// 🔹 Stage 1: $facet
// Filter documents by job_type and block_number range.
// Collect all unique block numbers into an existing_blocks array.
// 🔹 Stage 2: $project
// Extract the first (and only) existing_blocks array from the facet output.
// If none found, default to an empty array.
// 🔹 Stage 3: $addFields
// Generate a list of all expected block numbers in the range using $range.
// 🔹 Stage 4: $project
// Find missing blocks by subtracting existing ones from the full range.


```

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+
+        // Constructing the aggregation pipeline
+        let mut pipeline = vec![
+            doc! {
+                "$facet": {
+                    "existing_data": [
+                        doc! {
+                            "$match": {
+                                "job_type": job_type_bson,
+                                "metadata.specific.block_number": {
```
0xvasanth: I belive this is miss match in conflict resolution

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+        let job_type_bson = mongodb::bson::to_bson(&job_type)?;
+
+        // NOTE: This implementation is limited by mongodb's capbility to not support u64.
+        // i.e it will fail if upper_limit / lower_limit exceeds u32::MAX.
+
+        let lower_limit = u32::try_from(lower_cap).map_err(|e| {
+            tracing::error!(error = %e, category = "db_call", "Deserialization error");
+            DatabaseError::FailedToSerializeDocument(format!("Failed to deserialize: {}", e))
+        })?;
+        let upper_limti = u32::try_from(upper_cap).map_err(|e| {
```
0xvasanth:
```suggestion
        let upper_limit = u32::try_from(upper_cap).map_err(|e| {
```

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+            doc! {
+                "$facet": {
+                    "existing_data": [
+                        doc! {
+                            "$match": {
+                                "job_type": job_type_bson,
+                                "metadata.specific.block_number": {
+                                    "$gte": lower_limit,
+                                    "$lte": upper_limti
+                                }
```
0xvasanth: This pipeline might not be working since a sigle item of snos might have multiple block in it [I.e: 0.14.0]
ref @Mohiiit

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,29 +587,220 @@ impl DatabaseClient for MongoDbClient {
...
+            job_type = ?job_type,
+            lower_cap = lower_cap,
+            upper_cap = upper_cap,
+            category = "db_call",
+            "Fetching missing jobs by type and caps"
+        );
+
+        let collection: Collection<JobItem> = self.get_job_collection();
+
+        // Execute pipeline and extract block numbers
```
0xvasanth: If possible can move this with top stage of pipeline with the range so it will reduce the computation

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -371,32 +430,27 @@ impl DatabaseClient for MongoDbClient {
...
+                Ok(None)
+            }
+            1 => {
+                ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+                Ok(results.into_iter().next())
+            }
+            n => {
+                tracing::error!("Expected at most 1 result, got {}", n);
+                Err(DatabaseError::FailedToSerializeDocument(format!("Expected at most 1 result, got {}", n)))
             }
```
0xvasanth: If possible can we have utils for this block code since this been used in multiple place

0xvasanth: And also can you explain the reason behind "missing metric in error"?

orchestrator/crates/orchestrator/src/tests/workers/utils/mod.rs
```diff
@@ -37,7 +37,7 @@ pub fn get_job_item_mock_by_id(id: String, uuid: Uuid) -> JobItem {
         id: uuid,
         internal_id: id.clone(),
         job_type: JobType::SnosRun,
-        status: JobStatus::Created,
+        status: JobStatus::Completed,
```
0xvasanth: why did we change this

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,84 +1,229 @@
...
-                    program_output_path: Some(format!("{}/{}", block_num, PROGRAM_OUTPUT_FILE_NAME)),
-                    ..Default::default() // Ensure all other fields are set to default
-                }),
+        // Possible Values : 0..u64::MAX or Failed (network error or no block in Madara)
+        let latest_created_block_from_sequencer =
+            provider.block_number().await.wrap_err("Failed to fetch latest block number from sequencer")?;
+
+        // Get processing boundaries from config
+        let service_config = config.service_config();
+        // Will always be in range 0..u64::MAX
```
0xvasanth: can we metion here that it can be none too

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,84 +1,229 @@
...
+                        panic! {"This case should never have happened!"}
+                    }
+                },
+            };
+
+        let lower_limit = match latest_su_completed_block_number {
+            None => min_block_to_process_bound,
+            Some(value) => max(value, min_block_to_process_bound),
+        };
+
```
0xvasanth: If possible can we merge this multiple match to single match?

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,84 +1,229 @@
...
+            assert!(candidate_blocks.len() <= available_slots as usize);
+
+            // create_jobs_snos
+            tracing::info!(
+                "Creating SNOS jobs for {:?} blocks, with {} left slots",
+                &candidate_blocks,
+                available_slots
+            );
+            create_jobs_snos(config, candidate_blocks).await?;
```
0xvasanth: If i am not wrong, intend of reruning the missing, missing block might not work as expected here

e2e-tests/tests.rs
heemankv: Yes, the e2e test only runs for 1 specific block, the e2e still works here because we do not test the `WorkerTrigger` in the e2e test.

We can keep these as or change it to match the e2e case, but it wouldn't make a difference IMO.
wdyt ?

orchestrator/crates/orchestrator/src/core/client/database/mod.rs
heemankv: I believe you are talking about `limit`;

Mongodb internally accepts `Option<i64>`  hence we are following it's declaration
[Ref](https://docs.rs/mongodb/latest/mongodb/options/struct.FindOptions.html#structfield.limit)

lemme know your thoughts

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Surely, pushing some tests for this!

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: I don't understand this quite well, are you suggesting to add docs ?

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: I don't quite understand this comment, more context will be appreciated !

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Thanks!

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Oh, then let's discuss a way to handle that scenario as well!

However, I believe we can defer this one to be reworked after our release for `0.13.2`
wdyt ?

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: Our-pipeline currently goes by the process of elimination 
`Total Needed - Total Active = Total Missing`
Adding a limit to the top will disrupt this pipeline.

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
heemankv: > If possible can we have utils for this block code since this been used in multiple place

Done

> And also can you explain the reason behind "missing metric in error"?

Our Grafana dashboard currently uses the tracing logs from each function to let know for errored calls, these duration metrics are to let know of the successful calls only

orchestrator/crates/orchestrator/src/tests/workers/utils/mod.rs
heemankv: Oh, this change slid through my self review, reverting
Thanks!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Done!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Given the complexity of the coded algorithm, I have tried to keep the code as readable as possible, implementing partitioning (Part 1,2,3,4a,4b,5).
This is done intentionally as a part of that practice!

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Can you elaborate on this ?

orchestrator/cairo-lang
Mohiiit, line 1: this can be removed now right?

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -485,65 +511,45 @@ impl DatabaseClient for MongoDbClient {
...
+    //     };
+    //     let find_options = FindOneOptions::builder().sort(doc! { "internal_id": -1 }).build();
+
+    //     tracing::debug!(job_type = ?job_type, job_status = ?job_status, category = "db_call", "Fetched latest job by type and status");
+    //     let attributes = [KeyValue::new("db_operation_name", "get_latest_job_by_type_and_status")];
+    //     let duration = start.elapsed();
+    //     ORCHESTRATOR_METRICS.db_calls_response_time.record(duration.as_secs_f64(), &attributes);
+    //     Ok(self.get_job_collection().find_one(filter, find_options).await?)
+    // }
+
```
Mohiiit:
```suggestion
```

orchestrator/src/tests/workers/snos/mod.rs
```diff
@@ -1,105 +1,270 @@
...
-        start_job_index = 2;
-        block = 6;
-    }
+    // Mock sequencer response
+    let sequencer_response = json!({
+        "id": 1,
+        "jsonrpc": "2.0",
+        "result": latest_sequencer_block
+    });
+    let _rpc_mock = server.mock(|when, then| {
```
Mohiiit: we should assert is this call was made or not? if this is not being used, let's remove it?

Mohiiit (approved):

orchestrator/src/tests/workers/snos/mod.rs
heemankv: `rpc_mock` is not being used, but we are indeed mocking the server, and describing the server at the end like :
```rust
    // Setup provider
    let provider = JsonRpcClient::new(HttpTransport::new(
        Url::parse(&format!("http://localhost:{}", server.port())).expect("Failed to parse URL"),
    ));

    // Build test configuration
    let services = TestConfigBuilder::new()
        .configure_starknet_client(provider.into())
        .configure_database(db.into())
        .configure_queue_client(queue.into())
        .configure_da_client(da_client.into())
        .build()
        .await;

```

hence this can't be removed, 
One can argue that we can just not use `let _rpc_mock` and would be correct, 
to resolve this confusion, I have now simplified things to :
```rust
    // Mock sequencer response
    let sequencer_response = json!({
        "id": 1,
        "jsonrpc": "2.0",
        "result": latest_sequencer_block
    });
    server.mock(|when, then| {
        when.path("/").body_includes("starknet_blockNumber");
        then.status(200).body(serde_json::to_vec(&sequencer_response).unwrap());
    });
```

orchestrator/cairo-lang
Trantorian1: Seconding this, this should definitely be removed!

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,57 +556,404 @@ impl DatabaseClient for MongoDbClient {
...
+        lower_cap: u64,
+        upper_cap: u64,
+        limit: Option<i64>,
+    ) -> Result<Vec<u64>, DatabaseError> {
         let start = Instant::now();
+        // Converting params to Bson
+        let job_type_bson = mongodb::bson::to_bson(&job_type)?;
+
+        // NOTE: This implementation is limited by mongodb's capbility to not support u64.
+        // i.e it will fail if upper_limit / lower_limit exceeds u32::MAX.
```
Trantorian1, line 618: Why are we passing those as `u64` then? These should be a `u32` instead. If we need to operate on u64 values inside of this function then we should so by casting `as u64` instead

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -570,57 +556,404 @@ impl DatabaseClient for MongoDbClient {
...
+        updates.iter().for_each(|(k, v)| {
+            if v != &Bson::Null {
+                non_null_updates.insert(k, v);
+            }
+        });
+
+        // throw an error if there's no field to be updated
+        if non_null_updates.is_empty() {
+            return Err(DatabaseError::NoUpdateFound("No field to be updated, likely a false call".to_string()));
+        }
```
Trantorian1: In what conditions can this be a false call? Is this just here as a defensive measure or is this something that has happened?

orchestrator/src/tests/jobs/mod.rs
```diff
@@ -112,7 +112,10 @@ async fn create_job_job_exists_in_db_works() {
         .is_ok());
 
     // There should be only 1 job in the db
-    let jobs_in_db = database_client.get_jobs_by_statuses(vec![JobStatus::Created], None).await.unwrap();
+    let jobs_in_db = database_client
+        .get_jobs_by_types_and_statuses(vec![JobType::ProofCreation], vec![JobStatus::Created], None)
+        .await
+        .unwrap();
```
Trantorian1, line 118: Same as before: do we have any cap on the number of jobs returned by this or can this just grow indefinitely?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,176 +1,497 @@
...
+/// - `middle_limit`: The highest block number that has completed SNOS processing (if any)
+/// - `upper_limit`: The maximum block number to process (from sequencer or config limit)
+#[derive(Debug)]
+struct ProcessingBounds {
+    /// Minimum block number to process, derived from completed state updates
+    lower_limit: u64,
+    /// Latest block number that has completed SNOS processing (None if no jobs completed)
+    middle_limit: Option<u64>,
+    /// Maximum block number to process, limited by sequencer or configuration
+    upper_limit: u64,
```
Trantorian1: > [!NOTE]
> _**Bikeshedding**_ :motorcycle: 

These names feel a bit confusing, maybe we should go for something clearer? Like
```suggestion
    /// Minimum block number to process, derived from completed state updates
    block_n_min: u64,
    /// Latest block number that has completed SNOS processing (None if no jobs completed)
    block_n_completed: Option<u64>,
    /// Maximum block number to process, limited by sequencer or configuration
    block_n_max: u64,
```

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,176 +1,497 @@
...
+    /// * `Result<JobSchedulingContext>` - Initialized context or error
+    async fn initialize_scheduling_context(
+        &self,
+        config: &Arc<Config>,
+        bounds: ProcessingBounds,
+    ) -> Result<JobSchedulingContext> {
+        let service_config = config.service_config();
 
-        // TODO: This process to find last_completed_block is operation heavy, i.e for big list it will take time,
-        // maybe we can ask mongodb to return in descending order
```
Trantorian1: This! Are there any other parts in the codebase where we iterate over all `JobStatus::Completed` jobs?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,176 +1,497 @@
...
-        // Find the highest block that was already processed
-        let last_completed_block =
-            processed_blocks.iter().max().copied().unwrap_or(min_block_to_process.saturating_sub(1));
+    /// Schedules jobs for processing based on the current context and processing strategy.
+    ///
+    /// This method implements a two-phase scheduling strategy:
+    ///
+    /// # Phase 1: Initial Jobs (when middle_limit is None)
+    /// If no SNOS jobs have completed yet:
+    /// - Process all missing blocks in range [lower_limit, upper_limit]
```
Trantorian1: Are you meaning to add a hyperlink for `[lower_limit, upper_limit]`?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,176 +1,497 @@
...
+    async fn schedule_jobs_for_range(
+        &self,
+        config: &Arc<Config>,
+        context: &mut JobSchedulingContext,
+        start: u64,
+        end: u64,
+        range_name: &str,
+    ) -> Result<()> {
+        // Skip if range is invalid or empty
+        if start > end || (start == end && end == 0) {
```
Trantorian1: > [!WARNING]
> _**Invalid condition**_

Wouldn't `start == end` also qualify as an empty range? Right now this can be simplified to `start > end || end == 0`, since if `start != end` and `end == 0` then `start > end` since `start >= 0`.

```suggestion
        if start > end || start == end || end == 0 {
```

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -1,176 +1,497 @@
...
+    ///
+    /// # Returns
+    /// * `Result<u64>` - Count of pending jobs or database error
+    async fn count_pending_snos_jobs(&self, config: &Arc<Config>) -> Result<u64> {
+        let db = config.database();
+        let pending_statuses = vec![JobStatus::PendingRetry, JobStatus::Created];
 
-        Ok(())
+        let pending_jobs = db
+            .get_jobs_by_types_and_statuses(vec![JobType::SnosRun], pending_statuses, None)
```
Trantorian1: > [!NOTE]
> _**Perf**_

Is there a way to have mongo count the number of jobs directly instead of querying the whole job data? I assume this shouldn't be much of a problem though since the number of pending jobs should be bounded quite low by the max concurrency.

Trantorian1 (changes requested):
There appears to be some merge artifacts from main (submodules): these need to be removed! Please check if any files which were removed in #643 have not been accidentally re-added into the repo.

orchestrator/crates/orchestrator/src/tests/workers/snos/mod.rs
heemankv: So we are actually not creating jobs here, rather `expecting` various function calls in a manner that will lead to creating those jobs, the point of tests here is to validate the logic and more than the functionality. 
So for eg. we expect that `create_job` will be called with our specified block number.
```rust
job_handler
            .expect_create_job()
            .with(eq(block_num_str.clone()), mockall::predicate::always())
```

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Nope, not for the current functionality of SNOS job.

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Done!

heemankv: > How big can this get? Is there a bound to the number of values we are retrieving from DB or will this just keep growing as SNOS completes more and more jobs?

Potentially a lot, we have moved this to a mongodb pipeline now, it's much scalable now (given we don't move away from mongodb anytime soon 😁)

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Same as above

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Not needed with the new pipeline fn

orchestrator/cairo-lang
heemankv: Valid, removed!

orchestrator/src/core/client/database/mongodb.rs
heemankv: I am unable to understand this comment properly, 

Side note : wdyt about changing the values from top to bottom to `u32` , and define in the Clap args that these values can't surpass `u32::MAX` for now ?

heemankv: I am unable to understand this comment properly, 

Side note : wdyt about changing the values from top to bottom to `u32` , and define in the Clap args that these values can't surpass `u32::MAX` ?

orchestrator/src/core/client/database/mongodb.rs
heemankv: Unsure, I believe this is `defensive measure`. I don't see any other scenario. 
@Mohiiit do you ?

Mohiiit: it's a defensive measure only!

heemankv: Here's an example usage of `JobItemUpdates` struct, I believe we can utilise this condition `if non_null_updates.is_empty()` inside the `build` fn,
The build fn will throw an error if there are no updates, I believe this is much better than what we currently have, 
wdyt @Mohiiit @Trantorian1  
```rust
 let mut job = config
            .database()
            .update_job(
                &job,
                JobItemUpdates::new()
                    .update_status(JobStatus::LockedForProcessing)
                    .update_metadata(job.metadata.clone())
                    .build(),
            )
            .await
            .map_err(|e| {
                tracing::error!(job_id = ?id, error = ?e, "Failed to update job status");
                e
            })?;
```

heemankv: In the `build` fn we'll check if all the values are `None` then panic 
```rust
#[derive(Serialize, Debug)]
pub struct JobItemUpdates {
    pub status: Option<JobStatus>,
    pub external_id: Option<ExternalId>,
    pub metadata: Option<JobMetadata>,
}
```

orchestrator/src/tests/jobs/mod.rs
heemankv: Interesting, I believe we can use the limit by `max_slots` here : 
```rust
let max_slots = service_config.max_concurrent_created_snos_jobs;
let pending_jobs_count = self.count_pending_snos_jobs(config).await?;
let available_slots = max_slots.saturating_sub(pending_jobs_count);
```
given we only need the max value of pending_jobs_count as far as `max_slots` 
this would simplify the call here, wdyt ?

orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: HAHA, seems fine, have added!

orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: I believe all the fns that are fetching `jobs` would have big iterations internally. 

```rust
pub trait DatabaseClient: Send + Sync {
    /// switch_database - switch to a different database
    async fn switch_database(&mut self, database_name: &str) -> Result<(), DatabaseError>;

    /// disconnect - Disconnect from the database
    async fn disconnect(&self) -> Result<(), DatabaseError>;

    /// ENHANCEMENT: following method are supposed to be generic, but we need to figure out how to serialize them
    /// create_job - Create a new job in the database
    async fn create_job(&self, job: JobItem) -> Result<JobItem, DatabaseError>;
    /// get_job_by_id - Get a job by its ID
    async fn get_job_by_id(&self, id: uuid::Uuid) -> Result<Option<JobItem>, DatabaseError>;
    /// get_job_by_internal_id_and_type - Get a job by its internal ID and type
    async fn get_job_by_internal_id_and_type(
        &self,
        internal_id: &str,
        job_type: &JobType,
    ) -> Result<Option<JobItem>, DatabaseError>;
    /// update_job - Update a job in the database
    async fn update_job(&self, current_job: &JobItem, update: JobItemUpdates) -> Result<JobItem, DatabaseError>;
    /// get_latest_job_by_type - Get the latest job of a specific type
    async fn get_latest_job_by_type(&self, job_type: JobType) -> Result<Option<JobItem>, DatabaseError>;
    /// get_jobs_without_successor - Get jobs without a successor
    async fn get_jobs_without_successor(
        &self,
        job_a_type: JobType,
        job_a_status: JobStatus,
        job_b_type: JobType,
    ) -> Result<Vec<JobItem>, DatabaseError>;

    /// get_latest_job_by_type_and_status - Get the latest job of a specific type and status
    async fn get_latest_job_by_type_and_status(
        &self,
        job_type: JobType,
        job_status: JobStatus,
    ) -> Result<Option<JobItem>, DatabaseError>;

    /// get_jobs_after_internal_id_by_job_type - Get jobs after a specific internal id by job type
    async fn get_jobs_after_internal_id_by_job_type(
        &self,
        job_type: JobType,
        job_status: JobStatus,
        internal_id: String,
    ) -> Result<Vec<JobItem>, DatabaseError>;

    /// get_jobs_by_statuses -  Get all the jobs by types and status
    async fn get_jobs_by_types_and_statuses(
        &self,
        job_type: Vec<JobType>,
        status: Vec<JobStatus>,
        limit: Option<i64>,
    ) -> Result<Vec<JobItem>, DatabaseError>;

    /// get_missing_jobs_by_type_and_caps - Get all the missed jobs by type and block number limits
    async fn get_missing_block_numbers_by_type_and_caps(
        &self,
        job_type: JobType,
        lower_cap: u64,
        upper_cap: u64,
        limit: Option<i64>,
    ) -> Result<Vec<u64>, DatabaseError>;

    /// get_latest_batch - Get the latest batch from DB. Returns `None` if the DB is empty
    async fn get_latest_batch(&self) -> Result<Option<Batch>, DatabaseError>;
    /// update_batch - Update the bath
    async fn update_batch(&self, batch: &Batch, update: BatchUpdates) -> Result<Batch, DatabaseError>;
    /// create_batch - Create a new batch
    async fn create_batch(&self, batch: Batch) -> Result<Batch, DatabaseError>;
}
```

orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Nope, 
using this Convention :  
The endpoint adjoining the square bracket is known as closed, whereas the endpoint adjoining the parenthesis is known as open. [[here]](https://en.wikipedia.org/wiki/Bracket#:~:text=The%20endpoint%20adjoining%20the%20square%20bracket%20is%20known%20as%20closed%2C%20whereas%20the%20endpoint%20adjoining%20the%20parenthesis%20is%20known%20as%20open.%5B43%5D)

Do you mean something else ?

orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: Great catch !
Added.

orchestrator/src/worker/event_handler/triggers/snos.rs
heemankv: I believe my reply [here](https://github.com/madara-alliance/madara/pull/651#discussion_r2136049547) matches your suggestion

orchestrator/crates/orchestrator/src/worker/event_handler/triggers/snos.rs
Trantorian1: Neat, thanks for the changes!

orchestrator/src/core/client/database/mongodb.rs
Trantorian1: > I am unable to understand this comment properly,

It is a bit strange to be passing `lower_cap` and `upper_cap` as `u64` in the function only to check if they are in the range of `u32` and return an error otherwise. Why are these not passed as `u64` directly?

Trantorian1: > Side note : wdyt about changing the values from top to bottom to u32 , and define in the Clap args that these values can't surpass u32::MAX

If that makes it so we can remove the `u32` bound check then all for it!

orchestrator/src/tests/jobs/mod.rs
Trantorian1: That would be good if it could avoid a potentially unbounded database query :sweat_smile:

orchestrator/src/worker/event_handler/triggers/snos.rs
Trantorian1: All good, I was just confused because this connotes a link in Rustdoc and so I was wondering if the url was just missing :)

orchestrator/src/worker/event_handler/triggers/snos.rs
Trantorian1: _Actually this can be simplified down to `start > end || start == end` but at this point I'm nitpicking :sweat_smile:_

Trantorian1 (commented):
Looks good! 

Some of my previous comments relating to unbounded database queries have not been resolved, it would be good if those could be addressed :sweat_smile:

orchestrator/src/core/client/database/mongodb.rs
heemankv: > Why are these not passed as u64 directly? 

Because, mongodb doesn't support u64.

heemankv: >If that makes it so we can remove the u32 bound check then all for it!

lemme check the code changes for this

orchestrator/src/tests/jobs/mod.rs
heemankv: yup, adding it

orchestrator/src/core/client/database/mongodb.rs
Trantorian1: This sounds like a much more streamlined and less error-prone approach, looks good! :)

Trantorian1 (approved):

0xvasanth (approved):

PR #648, opened 2025-05-15
style: format code and improve readability in tests and initialization scripts

0xvasanth:
## Pull Request type

style: format code and improve readability in tests and initialization scripts

Please add the labels corresponding to the type of changes your PR introduces:

- Code style update (formatting, renaming)

PR #653, opened 2025-05-17
feat(database): enhance MongoDB client with Pre Setup PR for L3

0xvasanth:
- Added new setup targets `setup-l2` and `setup-l3` in the Makefile for orchestrator layer configurations.
- Updated `SetupCmd` struct to include a `layer` argument, allowing specification of L2 or L3 during setup.
- Refactored resource setup methods to accept the layer parameter, ensuring compatibility with both L2 and L3 configurations.
- Added `UpdateResult` and `DeleteResult` structs for better handling of update and delete operations.
- Implemented `find_one`, `update_one`, and `delete_one` methods in the MongoDbClient for improved database interactions.
- Introduced a flexible `find` method to support aggregation queries with sorting, limiting, and projection options.
- Added unit tests for the new methods to ensure functionality and correctness.

## Pull Request type

- Bugfix
- Feature

Resolves: #NA


Makefile
```diff
@@ -410,6 +411,11 @@ check:
 
 .PHONY: fmt
 fmt:
+	@echo -e "$(DIM)Running code formatters...$(RESET)"
+	@echo -e "$(INFO)Running taplo formatter...$(RESET)"
+	@npm install
+	@npx prettier --write .
+	@echo -e "$(PASS)Code formatting complete!$(RESET)"
```
Trantorian1: Seems like this should be merge after #648 then?

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -55,6 +67,116 @@ impl MongoDbClient {
...
+            });
+        }
+        if let Some(projection) = projection {
+            pipeline.push(doc! {
+                "$project": projection
+            });
+        }
+
+        let mut cursor = collection.aggregate(pipeline, None).await?;
+        let mut vec_items: Vec<T> = Vec::new();
```
Trantorian1: Is there a way to know the size of this vector ahead-of-time and pre-allocate it using `Vector::with_capacity`? I don't know if `cursor` is just an iterator (or if it has a [size hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint)) but we should try and avoid re-allocating this `Vec` if possible.

orchestrator/crates/orchestrator/src/setup/queue.rs
```diff
@@ -38,6 +38,16 @@ lazy_static! {
             visibility_timeout: 300,
             dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure })
         },
+        QueueConfig {
```
Mohiiit, line 52: this will make the orchestrator create these queues right? but ideally we want it to be conditional right? I mean we don't really need this in case of L2, is this being handled later on?

Makefile
0xvasanth: yes sir done

orchestrator/crates/orchestrator/src/core/client/database/mongodb.rs
0xvasanth: However, in this case, we can't know exactly how many items we'll get from the database ahead of time,
but have Rewritten the code bit better to solve it 

```
        let cursor = collection.aggregate(pipeline, None).await?;
        let vec_items: Vec<T> = cursor
            .map_err(|e| {
                tracing::error!(error = %e, category = "db_call", "Error retrieving document");
                DatabaseError::FailedToSerializeDocument(format!(
                    "Failed to retrieve document: {}",
                    e
                ))
            })
            .and_then(|doc| {
                futures::future::ready(mongodb::bson::from_document::<T>(doc).map_err(|e| {
                    tracing::error!(error = %e, category = "db_call", "Deserialization error");
                    DatabaseError::FailedToSerializeDocument(format!(
                        "Failed to deserialize document: {}",
                        e
                    ))
                }))
            })
            .try_collect()
            .await?;
            ```

File: orchestrator/crates/orchestrator/src/cli/mod.rs [hunk 4]
```diff
@@ -141,6 +141,8 @@ pub struct RunCmd {
 
     #[arg(env = "MADARA_ORCHESTRATOR_MADARA_RPC_URL", long, required = true)]
     pub madara_rpc_url: Url,
+    #[arg(env = "MADARA_ORCHESTRATOR_LAYER", long, default_value = Some("L2"), value_enum)]
+    pub layer: Option<Layer>,
```
Mohiiit: would there be any case when layer is none?

orchestrator/crates/orchestrator/src/setup/aws/event_bus.rs
```diff
@@ -53,6 +55,9 @@ impl Resource for EventBridgeClient {
         sleep(Duration::from_secs(15)).await;
 
         for trigger in WORKER_TRIGGERS.iter() {
+            if *trigger == WorkerTriggerType::ProofRegistration && layer != Layer::L3 {
```
Mohiiit, line 60: a comment here would be nice!

Makefile
```diff
@@ -456,3 +461,13 @@ pre-push:
 .PHONY: git-hook
 git-hook:
 	@git config core.hooksPath .githooks
+
+.PHONY: setup-l2
+setup-l2:
+	@echo -e "$(DIM)Setting up orchestrator with L2 layer...$(RESET)"
+	@cargo run --package orchestrator-core -- setup  --layer l2 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type schedule
```
Mohiiit, line 468: this won't work with local-stack tho, in case of local-stack we need --event-bridge-type rule, shall we change this to prod-setup-l2? idk, wdyt?

Mohiiit (commented):
gg! this is nice!

Makefile
0xvasanth: Yes, I have created this as starting point, i think we can do this enhancement in follow up PR

orchestrator/crates/orchestrator/src/cli/mod.rs
0xvasanth: yes your are right

Mohiiit (approved):

Trantorian1 (approved):
Looks good! Please note that some CI is not passing in the e2e orchestrator tests  (madara failure is normal, I have restarted the CI so we can use the new RPC url)

PR #657, opened 2025-05-20
Chain Config : L1DataAvailabilityMode added

heemankv:
This PR allows to configure the `L1DataAvailabilityMode` by providing it inside the chain config.

Currently the code as is returning `L1DataAvailabilityMode::Blob` by default, which works for `L2s` that want to settle using `Ethereum Blobs`. 

## What is the new behavior?
This PR allows the user to choose between `L1DataAvailabilityMode::Blob` and `L1DataAvailabilityMode::Calldata` based on which ever they want to use. 
For eg. 
The most common setting right now would follow : 
L2s : `L1DataAvailabilityMode::Blob` for settlement on `Ethereum`
L3s: `L1DataAvailabilityMode::Calldata` for settlement on `Starknet`

Note: 
This PR moves `L1DataAvailabilityMode` from `mp_block` to `mp_chain_config` due to a cyclic dependency that gets created when we try add `mp_block` as a dependency for `mp_chain_config` :
- `mp_chain_config` requires `L1DataAvailabilityMode`.
-  but `mp_block` is already using `StarknetVersion` from `mp_chain_config`.
Hence we took the decision to move `L1DataAvailabilityMode`  as a part of `mp_chain_config`.

Please use this in your chain config file : 
```
l1_da_mode : "Calldata" or "Blob" 
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9716,6 +9716,7 @@ dependencies = [
  "anyhow",
  "blockifier",
  "lazy_static",
+ "mp-rpc",
  "mp-utils",
  "primitive-types",
  "rstest 0.18.2",
--- a/configs/chain_config.example.yaml
+++ b/configs/chain_config.example.yaml
@@ -3,6 +3,10 @@ chain_name: "Starknet Mainnet"
 
 chain_id: "SN_MAIN"
 
+# The DA mode used by L1.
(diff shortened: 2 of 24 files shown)


Mohiiit (approved):

Trantorian1 (approved):
lgtm

PR #660, opened 2025-05-22
Orchestrator: Better resource handling & ARN based management

heemankv:
# Implement ARN/Name handling for AWS resources

## Overview
This PR streamlines AWS resource configuration by implementing flexible ARN (Amazon Resource Name) and name handling for various AWS resources. The changes remove the need to pass AWS configuration through CLI arguments, allowing the AWS SDK to automatically resolve configuration from environment variables.

### Environment Variable Consolidation
**Removed environment variables:**
- `MADARA_ORCHESTRATOR_EVENT_BRIDGE_TYPE`
- `MADARA_ORCHESTRATOR_EVENT_BRIDGE_TRIGGER_RULE_NAME`
- `MADARA_ORCHESTRATOR_EVENT_BRIDGE_TRIGGER_ROLE_NAME`
- `MADARA_ORCHESTRATOR_EVENT_BRIDGE_TRIGGER_POLICY_NAME`
- `MADARA_ORCHESTRATOR_AWS_SNS_TOPIC_NAME`
- `MADARA_ORCHESTRATOR_AWS_S3_BUCKET_NAME`
- `MADARA_ORCHESTRATOR_SQS_PREFIX`
- `MADARA_ORCHESTRATOR_SQS_SUFFIX`
- `MADARA_ORCHESTRATOR_EVENT_BRIDGE_TARGET_QUEUE_NAME`

**Added environment variables:**
- `MADARA_ORCHESTRATOR_AWS_SNS_TOPIC_IDENTIFIER`
- `MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER`
- `MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER`

### Enhanced Resource Identification
The new environment variables support flexible resource identification:

**ARN Support:** When a full ARN is provided, it will be used directly without modification.

**Name-based Configuration:** When only a resource name is provided:
- If `AWS_PREFIX` environment variable is set, it will be automatically prepended to the resource name
- If `AWS_PREFIX` is not set, the resource name will be used as-is

### Note
QUEUE IDENTIFIER currently uses format `resource_{}_name`, it requires the `{}` in between currently since we setup multiple queues within the same setup and we replace the `{}` with the queue's type.

## Example Usage
```bash
Example Usage
Using ARNs (used directly)
bashMADARA_ORCHESTRATOR_AWS_SNS_TOPIC_IDENTIFIER=arn:aws:sns:{region}:{account_id}:{resource_name}
MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER=arn:aws:sqs:{region}:{account_id}:{resource_{}_name}
MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER=arn:aws:s3:::{resource-name}

Using names without prefix
bashMADARA_ORCHESTRATOR_AWS_SNS_TOPIC_IDENTIFIER=resource_name
MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER=resource_{}_name
MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER=resource-name

Using names with prefix
bashMADARA_ORCHESTRATOR_AWS_PREFIX=prefix
MADARA_ORCHESTRATOR_AWS_SNS_TOPIC_IDENTIFIER=resource_name
MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER=resource_{}_name
MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER=resource-name
# Results in: "prefix_resource_name", "prefix_resource_{}_name", "prefix-resource-name"
```


orchestrator/src/setup/aws/event_bus.rs
```diff
@@ -111,6 +127,217 @@ impl Resource for EventBridgeClient {
...
+
+        let role_name = format!("{}-{}", trigger_role_name, short_id);
+
+        let role_arn = self.create_iam_role(&role_name).await?;
+
+        let policy_name = format!("{}-{}", trigger_policy_name, short_id);
+
+        self.create_and_attach_sqs_policy(&policy_name, &role_name, &queue_arn).await?;
+
+        Ok(TriggerArns { queue_arn, role_arn })
```
0xvasanth, line 245: why are we adding the random number here

orchestrator/src/cli/alert/aws_sns.rs
```diff
@@ -8,8 +8,8 @@ pub struct AWSSNSCliArgs {
     #[arg(long)]
     pub aws_sns: bool,
 
-    /// The ARN of the SNS topic.
-    #[arg(env = "MADARA_ORCHESTRATOR_AWS_SNS_TOPIC_NAME", long, default_value = Some("arn")
-    )]
-    pub alert_topic_name: Option<String>,
+    /// The ARN / Name of the SNS topic.
+    /// ARN: arn:aws:sns:region:accountID:name
```
0xvasanth, line 12:
```suggestion
    /// ARN: arn:aws:sns:region:accountID:name
    /// Name: name
```

orchestrator/src/cli/alert/aws_sns.rs
```diff
@@ -8,8 +8,8 @@ pub struct AWSSNSCliArgs {
     #[arg(long)]
     pub aws_sns: bool,
 
-    /// The ARN of the SNS topic.
-    #[arg(env = "MADARA_ORCHESTRATOR_AWS_SNS_TOPIC_NAME", long, default_value = Some("arn")
-    )]
-    pub alert_topic_name: Option<String>,
+    /// The ARN / Name of the SNS topic.
```
0xvasanth:
```suggestion
    /// The ARN / Name of the SNS topic. it can have either name or ARN string
```

orchestrator/src/cli/cron/event_bridge.rs
```diff
@@ -18,28 +18,12 @@ pub struct AWSEventBridgeCliArgs {
...
-    )]
-    pub trigger_role_name: Option<String>,
-
-    /// The name of the queue for the event bridge
-    #[arg(env = "MADARA_ORCHESTRATOR_EVENT_BRIDGE_TRIGGER_POLICY_NAME", long, default_value = Some("madara-orchestrator-worker-trigger-policy"), help = "The name of the Trigger Policy to assign to the event bridge"
+    #[arg(
+        env = "MADARA_ORCHESTRATOR_EVENT_BRIDGE_INTERVAL_SECONDS",
+        long,
+        default_value = "60",
+        help = "The interval in seconds for the event bridge trigger rule. Defaults to 60 seconds."
```
0xvasanth, line 26: I am not sure why we have removed these access from here,
since it is possible for user / developer / enduser who want to use the existing Rule, policy ARN or to create new one

orchestrator/src/cli/provider/aws.rs
```diff
@@ -3,17 +3,12 @@ use serde::Serialize;
...
-    #[arg(env = "AWS_REGION", long)]
-    pub aws_region: String,
-
-    /// The region.
-    #[arg(env = "MADARA_ORCHESTRATOR_AWS_PREFIX", long, default_value = "madara-orchestrator")]
-    pub aws_prefix: String,
+    /// The prefix value.
+    /// And added to the start of each resource name if available
+    #[arg(env = "MADARA_ORCHESTRATOR_AWS_PREFIX", long, default_value = None)]
+    pub aws_prefix: Option<String>,
```
0xvasanth, line 13:
```suggestion
    #[arg(env = "MADARA_ORCHESTRATOR_AWS_PREFIX", long, default_value = "mo")]
    pub aws_prefix: String,
```
since at anypoint we can have none if other parameters are given as name

orchestrator/src/cli/queue/aws_sqs.rs
```diff
@@ -8,11 +8,10 @@ pub struct AWSSQSCliArgs {
...
-
-    /// The suffix of the queue.
-    #[arg(env = "MADARA_ORCHESTRATOR_SQS_SUFFIX", long, default_value = Some("queue"))]
-    pub sqs_suffix: Option<String>,
+    /// The ARN / Name of the queue.
+    /// ARN: arn:aws:sqs:region:accountID:name
+    /// {} will be replaced by Queue Type, when independent setup for each queue.
+    /// i.e for WorkerTrigger queue : mo_worker_trigger_queue
+    #[arg(env = "MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER", long, default_value = Some("mo_{}_queue"))]
+    pub queue_identifier: Option<String>,
```
0xvasanth, line 16: if user have ARN as : arn:aws:sqs:region:accountID:mo_wt_queue
and want to use this ARN for worker trigger queue, how will this work

orchestrator/src/cli/storage/aws_s3.rs
```diff
@@ -8,11 +8,10 @@ pub struct AWSS3CliArgs {
...
-
-    /// The S3 Bucket Location Constraint.
-    #[arg(env = "MADARA_ORCHESTRATOR_AWS_BUCKET_LOCATION_CONSTRAINT", long)]
-    pub bucket_location_constraint: Option<String>,
+    /// The ARN / Name of the S3 bucket.
+    /// ARN: arn:aws:s3:::name
+    /// We don't need to provide the region and accountID in s3
+    /// because s3 is unique globally.
+    #[arg(env = "MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER", long, default_value = Some("mo-bucket"))]
+    pub bucket_identifier: Option<String>,
```
0xvasanth, line 16:
```suggestion
    #[arg(env = "MADARA_ORCHESTRATOR_AWS_S3_BUCKET_IDENTIFIER", long, default_value = "bucket")]
    pub bucket_identifier: String,
```
I dont find a reason to have this field as option, can you help me understand this

orchestrator/src/core/client/alert/error.rs
```diff
@@ -8,6 +8,9 @@ pub enum AlertError {
     #[error("Topic ARN is empty")]
     TopicARNEmpty,
 
+    #[error("Topic ARN is invalid")]
+    TopicARNInvalid,
```
0xvasanth:
```suggestion
    #[error("Topic ARN is invalid, ARN : {0}")]
    TopicARNInvalid(String),
```

since it is invalid ARN can you parse the invalid string here as part of the error message and argument

orchestrator/src/core/client/alert/sns.rs
```diff
@@ -1,15 +1,73 @@
...
+                        tracing::debug!("ARN not parsable");
+                        AlertError::TopicARNInvalid
+                    })?;
+                    return Ok(arn);
+                }
+            }
+        }
+
+        Err(AlertError::TopicNotFound(topic_name.to_string()))
+    }
```
0xvasanth: This code have lot of IO overhead which seems needless

orchestrator/src/types/params/mod.rs
```diff
@@ -9,37 +9,135 @@ pub mod snos;
 
 use crate::cli::cron::event_bridge::EventBridgeType;
 use crate::cli::{RunCmd, SetupCmd};
+use crate::core::client::queue::sqs::InnerSQS;
+use crate::types::queue::QueueType;
 use crate::OrchestratorError;
 pub use otel::OTELConfig;
-use tracing::info;
+
+pub type ResourceName = String;
```
0xvasanth: this variable is not been used anywhere can we remove this

orchestrator/src/cli/queue/aws_sqs.rs
0xvasanth: arn:aws:sqs:region:accountID:mo_wt_queue
arn:aws:sqs:region:accountID:mo_snos_processing_queue

.tool-versions
Trantorian1, line 1: Is this supposed to be here or is this an artifact from some tests which were run locally? We should make sure this does not interfere with cairo compilation when running `make artifacts`!

e2e-tests/tests.rs
```diff
@@ -136,9 +137,22 @@ async fn test_orchestrator_workflow(#[case] l2_block_number: String) {
...
+                prefix,
+                get_env_var_or_panic("MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER",)
+            )),
+        },
+        None => QueueArgs {
+            queue_template_identifier: orchestrator::types::params::AWSResourceIdentifier::Name(get_env_var_or_panic(
+                "MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER",
+            )),
+        },
     };
```
Trantorian1: > [!NOTE]
> _**Nitpick**_

As far as refactoring goes, something like this would probably be cleaner:
```suggestion
    // E2E tests assume that we are not passing ARNS
    let aws_prefix = get_env_var_optional_or_panic("MADARA_ORCHESTRATOR_AWS_PREFIX");
    let aws_identifier =  get_env_var_or_panic("MADARA_ORCHESTRATOR_AWS_SQS_QUEUE_IDENTIFIER");

    let queue_params = match aws_prefix {
        Some(prefix) => QueueArgs {
            queue_template_identifier: orchestrator::types::params::AWSResourceIdentifier::Name(format!(
                "{}_{}",
                prefix,
                aws_identifier,
            )),
        },
        None => QueueArgs {
            queue_template_identifier: orchestrator::types::params::AWSResourceIdentifier::Name(aws_identifier),
        },
    };
```

lib/argent-contracts-starknet
Trantorian1, line 1: > [!CAUTION]
> **_Potential merge artifacts_**

All submodules have been removed in the latest CI update #643 as they are no longer needed to build artifacts (compilation has been dockerised). Is adding them back here intentional or was this not caught during merge?

You might have to check the root `.gitmodules` to make sure it has been cleared.

lib/braavos-account-cairo
Trantorian1, line 1: Ditto.

lib/starkgate-contracts-82e651f
Trantorian1, line 1: Ditto.

lib/starkgate-contracts-latest
Trantorian1, line 1: Ditto.

lib/starkgate-contracts-old
Trantorian1, line 1: Ditto.

orchestrator/cairo-lang
Trantorian1, line 1: Ditto.

orchestrator/crates/utils/src/env_utils.rs
```diff
@@ -22,6 +22,6 @@ pub fn get_env_var_optional(key: &str) -> Result<Option<String>, VarError> {
     }
 }
 
-pub fn get_env_car_optional_or_panic(key: &str) -> Option<String> {
+pub fn get_env_var_optional_or_panic(key: &str) -> Option<String> {
```
Trantorian1, line 25: image

orchestrator/src/core/client/alert/error.rs
Trantorian1: Seconding this!

orchestrator/src/core/client/alert/sns.rs
```diff
@@ -1,15 +1,73 @@
...
 use aws_config::SdkConfig;
 use aws_sdk_sns::Client;
-
-use super::AlertError;
-use crate::{core::client::alert::AlertClient, types::params::AlertArgs};
 use std::sync::{Arc, OnceLock};
 
+/// AWSS3 is a struct that represents an AWS S3 client.
+#[derive(Clone, Debug)]
+pub(crate) struct InnerAWSSNS(pub(crate) Arc<Client>);
```
Trantorian1: If we already have a `pub` getter for the inner client then why is `Arc<Client>` marked as `pub(crate)`?

orchestrator/src/core/client/alert/sns.rs
```diff
@@ -1,15 +1,73 @@
...
+
+    pub fn is_valid_topic_name(&self, name: &str) -> bool {
+        // AWS SNS topic name requirements:
+        // - Can include numbers, letters, hyphens, and underscores
+        // - Length between 1 and 256
+        if name.is_empty() || name.len() > 256 {
+            return false;
+        }
+
+        name.chars().all(|c| c.is_alphanumeric() || c == '-' || c == '_')
```
Trantorian1, line 64: Neat!

orchestrator/src/core/client/alert/sns.rs
Trantorian1: This is not doing anything different from what `get_topic_arn` did previously so this should be ok?

orchestrator/src/core/client/event_bus/event_bridge.rs
```diff
@@ -1,245 +1,33 @@
...
-    queue_client: Arc<aws_sdk_sqs::Client>,
-    iam_client: Arc<aws_sdk_iam::Client>,
-    pub event_bridge_type: Option<EventBridgeType>,
-    pub target_queue_name: Option<String>,
-    pub cron_time: Option<Duration>,
-    pub trigger_rule_name: Option<String>,
-    pub trigger_role_name: Option<String>,
-    pub trigger_policy_name: Option<String>,
+    pub(crate) queue_client: Arc<aws_sdk_sqs::Client>,
+    pub(crate) iam_client: Arc<aws_sdk_iam::Client>,
```
Trantorian1, line 11: Thanks for the change! Looks much easier to work with :D

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -13,55 +16,23 @@ use std::sync::Arc;
...
         let client = Client::from_conf(sqs_config_builder.build());
-        Self {
-            client: Arc::new(client),
-            prefix: args.map(|a| a.prefix.clone()),
-            suffix: args.map(|a| a.suffix.clone()),
-        }
+        Self(Arc::new(client))
     }
 
     pub fn client(&self) -> Arc<Client> {
```
Trantorian1: Could it perhaps not be enough to just send a `&'_ Client` instead of cloning the whole `Arc`?

orchestrator/src/setup/aws/event_bus.rs
```diff
@@ -111,6 +127,217 @@ impl Resource for EventBridgeClient {
     /// # Returns
     /// * `OrchestratorResult<bool>` - A result indicating if the event bridge rule is ready to use
     async fn is_ready_to_use(&self, _layer: &Layer, args: &Self::SetupArgs) -> OrchestratorResult<bool> {
-        Ok(self.eb_client.describe_rule().name(&args.trigger_rule_name).send().await.is_ok())
+        let mut flag = false;
+        for trigger_type in WORKER_TRIGGERS.iter() {
+            let trigger_name = Self::get_trigger_name_from_trigger_type(&args.trigger_rule_template_name, trigger_type);
+            flag = flag && self.eb_client.describe_rule().name(trigger_name).send().await.is_ok()
+        }
+        Ok(flag)
```
Trantorian1: > [!WARNING]
> **_Circular condition_**

Wont' this always be false since `flag && self.eb_client... == false` if `flag == false` initially?

orchestrator/src/setup/aws/s3.rs
Trantorian1, line 1: > [!TIP]
> _**Refactor**_

It would be nice common values like `"us-east-1"` could be hoisted to their own `const`. Maybe we could even regroup such related constants into their own module.

orchestrator/src/setup/aws/sns.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::core::cloud::CloudProvider;
+use crate::core::traits::resource::Resource;
+use crate::types::params::AWSResourceIdentifier;
+use crate::types::params::AlertArgs;
+use crate::{OrchestratorError, OrchestratorResult};
+use anyhow::Context;
+use async_trait::async_trait;
+use std::sync::Arc;
+
+#[async_trait]
```
Trantorian1, line 12: > [!TIP]
> _**Unnecessary `async-trait`**_

Have you tried implementing this without `async-trait`? I only recently found out ( :sweat_smile:  ) that `async fn`s in traits have been [supported since Rust 1.78](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0/#async-fn-and-return-position-impl-trait-in-traits) and [`async-traint` is only needed in the case of traits which return a `dyn Train`](https://docs.rs/async-trait/latest/async_trait/).

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -36,32 +39,83 @@ impl Resource for SQS {
...
+                    // If ARN is provided, we just check if it exists
+                    let queue_name = InnerSQS::get_queue_name_from_type(&arn.resource, &queue.name);
+                    tracing::info!("Queue Arn provided, skipping setup for {}", &queue_name);
+                    continue;
+                }
+
+                AWSResourceIdentifier::Name(name) => {
+                    let queue_name = InnerSQS::get_queue_name_from_type(name, &queue.name);
+
+                    println!("printing queue that will be created: {} ", queue_name);
```
Trantorian1: > [!Note]
> _**Nitpick**_

Leftover `println!` statement?

orchestrator/src/types/jobs/mod.rs
```diff
@@ -32,6 +32,12 @@ pub enum WorkerTriggerTypeError {
     UnknownType(String),
 }
 
+impl WorkerTriggerType {
+    pub fn to_lowercase(&self) -> String {
+        self.to_string().to_lowercase()
+    }
+}
+
```
Trantorian1: Is this really necessary? :sweat_smile:

orchestrator/src/types/params/mod.rs
```diff
@@ -9,37 +9,135 @@ pub mod snos;
...
+
+#[derive(Debug, Clone)]
+pub struct ARN {
+    pub partition: String,  // Usually "aws" (e.g., "aws-us-gov", "aws-cn")
+    pub service: String,    // AWS service (e.g., "s3", "sns", "sqs")
+    pub region: String,     // AWS region (e.g., "us-east-1", can be empty for global services)
+    pub account_id: String, // AWS account ID (12-digit number, can be empty for some resources)
+    pub resource: String,   // Resource identifier (e.g., "topic-name", "bucket-name", "queue-name")
```
Trantorian1: > [!Tip]
> _**Docs**_

Consider moving those comments to doc comments.

orchestrator/src/types/params/mod.rs
```diff
@@ -9,37 +9,135 @@ pub mod snos;
...
+}
+impl ARN {
+    /// Parse an ARN string into its components
+    /// Format: arn:partition:service:region:account-id:resource
+    pub fn parse(arn_str: &str) -> Result<Self, &'static str> {
```
Trantorian1, line 33: > [!WARNING]
> _**Testing**_

I would suggest adding some simple unit tests to make sure the parsing works and will not break silently in the future.

package-lock.json
Trantorian1, line 1: Is this a leftover artifact from local testing?

Trantorian1 (changes requested):
Nice changes!

I really like the the `Inner*` structs and the separation of concerns they bring: this part of the code feels much better structured now.

I have left a few comments, mostly nitpick and suggestions for improvements here and there. It would be nice if these are addressed but is not what I would consider paramount. In order of increasing (subjective) priority:

- **note**
- **tip**
- **warning**
- **caution**

If you could pay special attention to comments marked as **caution**, these are related to potential issues which I feel _should_ be fixed before this PR is merged.

Please let me know what you think of this and if you have any questions feel free to reach out on Slack.

orchestrator/src/setup/aws/event_bus.rs
heemankv: These are used to create new identifiers for everytime the setup runs, 
allowing for policy and role name to be unique each time.
Here we have a simplified version of the len 16 unique UUID we were using earlier

orchestrator/src/cli/cron/event_bridge.rs
heemankv: These are non-orchestrator facing resources, 
i.e these are directed to cron and queues.
> it is possible for user / developer / enduser who want to use the existing Rule, policy ARN  

For this they must already configure the rule to the queue.

> create new one

They can create new ones with this setup

orchestrator/src/cli/provider/aws.rs
heemankv: we are following this structure because of clap,
I am currently unsure as to how it will be done,
but we can take up refactoring clap in next few PRs.
I believe this was discussed offline.

orchestrator/src/cli/queue/aws_sqs.rs
heemankv: We don't currently support customising names on dependent resources for orchestrator.
We follow a convention. There can obviously be an argument against it not being possible, but I believe we should keep things simple and ask if we really need it or not!

Lemme know if you feel the need for a scenario to have this versatility.

orchestrator/src/cli/storage/aws_s3.rs
heemankv: Same as above: 
we are following this structure because of clap,
I am currently unsure as to how it will be done,
but we can take up refactoring clap in next few PRs.
I believe this was discussed offline.

orchestrator/src/core/client/alert/error.rs
heemankv: Valid, added.

orchestrator/src/core/client/alert/sns.rs
heemankv: Yeah, unlike SQS there's no other fn is aws that allows for directly fetching SNS.

orchestrator/src/types/params/mod.rs
heemankv: Valid, removed

.tool-versions
heemankv: Valid, removing this.
Great catch!

e2e-tests/tests.rs
heemankv: Oh, alright, seems valid!

lib/argent-contracts-starknet
heemankv: I believe the merge with main after PR #643 didn't catch this.
Have removed this now
This is a great catch!

orchestrator/crates/utils/src/env_utils.rs
heemankv: Yes! the greatest fix of them all !

orchestrator/src/core/client/alert/sns.rs
heemankv: This is a great catch, 
Your point is valid that this `pub(crate)` on `Arc<Client>` can be removed.
I revisited other parts as well and have further reduced the use of `pub` identifier from wherever it was not necessary.
Thanks!

heemankv: [Ref](https://github.com/madara-alliance/madara/commit/64eb3f638ad4040f4e52bcf7a1b6d678296a6b45)

orchestrator/src/setup/aws/s3.rs
heemankv: I believe we are soon to try something like you suggest, in one of the following PRs, essentially regrouping all the `hardcoded` values into 
either `config.json` 
or some `constants.rs`

If it's not a pressing issue, we'd prefer to handle it then!

heemankv: [Ref](https://github.com/madara-alliance/madara/issues/613)

orchestrator/src/setup/aws/sqs.rs
heemankv: Removed, thanks

orchestrator/src/types/jobs/mod.rs
heemankv: No, have removed!

orchestrator/src/types/params/mod.rs
heemankv: Valid, moved!

orchestrator/src/types/params/mod.rs
heemankv: Added!

package-lock.json
heemankv: Yeah, every-time we run the `git-hook` on `git push` , the `package-lock.json` is created.
wdyt about adding this to `.gitignore` ?

orchestrator/src/setup/aws/event_bus.rs
Mohiiit: +1

orchestrator/src/setup/aws/event_bus.rs
```diff
@@ -111,6 +127,217 @@ impl Resource for EventBridgeClient {
...
+        &self,
+        trigger_type: &WorkerTriggerType,
+        trigger_arns: &TriggerArns,
+        trigger_rule_template_name: String,
+        event_bridge_type: EventBridgeType,
+        cron_time: Duration,
+    ) -> color_eyre::Result<()> {
+        let message = trigger_type.clone().to_string();
+        let trigger_name = Self::get_trigger_name_from_trigger_type(&trigger_rule_template_name, trigger_type);
+        tracing::info!("Creating Event Bridge Rule trigger: {} ", trigger_name);
```
Mohiiit: this info doesn't belong here, it should be after match ::Rule

orchestrator/src/setup/aws/s3.rs
```diff
@@ -28,42 +29,64 @@ impl Resource for AWSS3 {
...
         }
     }
     /// Set up a new S3 bucket
     async fn setup(&self, _layer: &Layer, args: Self::SetupArgs) -> OrchestratorResult<Self::SetupResult> {
         // Check if the bucket already exists
         // If it does, return the existing bucket name and location
-        if self.check_if_exists(args.bucket_name.clone()).await? {
-            warn!(" ⏭️  S3 bucket '{}' already exists", args.bucket_name);
+        if self.check_if_exists(&args.bucket_identifier).await? {
+            warn!(" ⏭️  S3 bucket already exists, skipping creation");
```
Mohiiit: I think the warning before was more informative

orchestrator/src/setup/aws/s3.rs
```diff
@@ -28,42 +29,64 @@ impl Resource for AWSS3 {
...
-        // Get the current region from the client config
-        let region = self.client.config().region().map(|r| r.to_string()).unwrap_or_else(|| "us-east-1".to_string());
-        info!("Creating bucket in region: {}", region);
+        // s3 can have empty region in it's arn : e.g: arn:aws:s3:::mo-bucket
+        // in such scenarios we would want to default to provided region
 
-        let mut bucket_builder = self.client.create_bucket().bucket(&args.bucket_name);
+        match &args.bucket_identifier {
+            AWSResourceIdentifier::ARN(arn) => {
+                // If ARN is provided in setup, we only check if it exists
```
Mohiiit: we are not checking here tho?

Mohiiit: PS: I get it, you check it before only with the bucket_identifier, the comment here is a bit misleading

orchestrator/src/setup/aws/sns.rs
```diff
@@ -0,0 +1,80 @@
...
+            AWSResourceIdentifier::Name(name) => Ok(self.fetch_topic_arn_by_name_via_client(name).await.is_ok()),
+        }
+    }
+
+    async fn is_ready_to_use(&self, _layer: &Layer, args: &Self::SetupArgs) -> OrchestratorResult<bool> {
+        Ok(self.check_if_exists(&args.alert_identifier).await?)
+    }
+}
+
+impl InnerAWSSNS {}
```
Mohiiit: is this required?

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -77,23 +130,58 @@ impl Resource for SQS {
...
+                    .is_ok())
+            }
+            AWSResourceIdentifier::Name(name) => {
+                let queue_name = InnerSQS::get_queue_name_from_type(&name, &check_args.1);
+                Ok(self.client().get_queue_url().queue_name(queue_name).send().await.is_ok())
+            }
+        }
     }
 
     async fn is_ready_to_use(&self, layer: &Layer, args: &Self::SetupArgs) -> OrchestratorResult<bool> {
```
Mohiiit, line 153: let's create an issue regarding de-duplication of the code here in `check_if_exists` and `is_ready_to_use`

Mohiiit (approved):
mostly nit, lfg

PS: I think the code where we have TryFrom from RunCmd and SetupCmd to the required clients/configs, I think that can simplified and test can be added regarding the same as well

orchestrator/src/setup/aws/event_bus.rs
heemankv: Valid, shifted, also added a log for schedule

orchestrator/src/setup/aws/s3.rs
heemankv: Alright, shifted to : 
```rust 
 warn!(" ⏭️  S3 bucket {} already exists , skipping creation", &args.bucket_identifier);
 ```

orchestrator/src/setup/aws/s3.rs
heemankv: Valid, changed to : 
// If ARN is provided in setup, we have already checked if it exists

orchestrator/src/setup/aws/sns.rs
heemankv: Great catch !
Removed

orchestrator/src/setup/aws/sqs.rs
heemankv: https://github.com/madara-alliance/madara/issues/680

orchestrator/src/setup/aws/event_bus.rs
heemankv: Valid, marked as true now, tested locally it's working now

orchestrator/src/core/client/queue/sqs.rs
heemankv: Interesting, I have reworked to code a bit to not require the Arc at all, [check this out](https://github.com/madara-alliance/madara/pull/660/commits/91bf96a18a5ef74bd25a62b53b3860b65a45fb60)

This is a really great catch!

orchestrator/src/setup/aws/sns.rs
heemankv: Tracking [here](https://github.com/madara-alliance/madara/issues/681)

orchestrator/src/setup/aws/s3.rs
Trantorian1: Oh yea definitely nothing pressing, just me nitpicking

package-lock.json
Trantorian1: Sounds like a good idea, let's do that!

orchestrator/src/types/params/mod.rs
Trantorian1, line 1: Thanks for the added tests!

Trantorian1 (approved):
lgtm!

0xvasanth (approved):

PR #666, opened 2025-05-27
refactor: Use type imports for Starknet types

tongjicoder:
## Pull Request type

Resolves: #NA

## What is the new behavior?

This PR updates the import statements for `CompiledSierra` and `CompiledSierraCasm` from `starknet` to use `import type`. This change explicitly marks these imports as type-only, improving code clarity and enabling better tree-shaking by bundlers. 

-
-
-


Trantorian1:
We thank you for your work. However, we are currently not accepting any outside contributions. If you want to provide a fix, please open an issue first.

PR #667, opened 2025-05-27
refactor: streamline queue handling

### LINKED ISSUE #627: feat(restructure) - Message-Level Concurrency Control to Event Handler Layer

- [x] I have searched the existing issues

### Description of bug

Current Situation
We’re currently using a semaphore within specific flows (e.g., snosprocessing) to limit the number of concurrent jobs. However, this control is implemented deep inside the business logic (e.g., within SnosHandler::process). As a result, the system may pull a message from the queue, attempt to acquire a permit, and—if unsuccessful—wait briefly and eventually NACK the message. This is both inefficient and unnecessary queue churn.

Proposal
Instead, we should move the concurrency control to the event_handler layer. This would allow us to check for availability before fetching a message from the queue, effectively avoiding unnecessary processing and better aligning concurrency control with queue management responsibilities.

Reference
Discussion: https://github.com/madara-alliance/madara/pull/619#discussion_r2066502345

### Steps to reproduce

_No response_

0xvasanth:
## What is the new behavior?

- Eliminated the ProcessingLocks struct and its usage across the Config and job handler implementations.
- Updated references in the `sqs.rs` and `event_worker.rs` files to use the new `QUEUES` structure from `q_control`.
- Adjusted the `EventWorker` to utilize `QueueControlConfig` for managing queue settings.
- Added `q_control` module for improved structure.
- Refactored `QueueControlConfig` methods to reduce verbosity in struct initialization.
- Enhanced `EventWorker` to maintain clarity while processing messages and handling errors.
- Updated job handler methods to remove references to job processing locks, simplifying job processing logic.
- Adjusted tests to reflect the removal of processing locks, ensuring continued functionality without them.

## Pull Request type

- Code style update (streamline queue handling)
- Refactor (streamline queue handling)

- This introduce new Concurrency control system in Queue / EventWorker


orchestrator/src/setup/aws/sqs.rs
```diff
@@ -52,7 +56,8 @@ impl Resource for SQS {
             attributes.insert(QueueAttributeName::VisibilityTimeout, queue.visibility_timeout.to_string());
 
             if let Some(dlq_config) = &queue.dlq_config {
-                let dlq_url = self.get_queue_url_from_client(&queue_name).await?;
+                let dql_queue_name = queue_fn(&dlq_config.dlq_name);
```
heemankv: Does this mean the setup was wrong before !?

heemankv: Made similar changes in ARN pr

heemankv: It's interesting that we never picked it up!

orchestrator/src/types/mod.rs
```diff
@@ -3,5 +3,6 @@ pub mod constant;
 pub mod error;
 pub mod jobs;
 pub mod params;
+pub mod q_control;
```
heemankv: Let's rename this to `queue_control`.
wdyt ?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -36,7 +40,9 @@ impl EventWorker {
     /// * `EventWorker` - A new EventWorker instance
     pub fn new(queue_type: QueueType, config: Arc<Config>) -> Self {
         info!("Kicking in the Worker to Monitor the Queue {:?}", queue_type);
-        Self { queue_type, config }
+        let queue_control =
+            QUEUES.get(&queue_type).map(|q| q.queue_control.clone().unwrap_or_default()).unwrap_or_default();
```
heemankv: Should we not throw an error or panic if the QUEUES.get is returning a `None` ?
This should possibly reduce the outer `unwrap_or_default` and have better error handling
wdyt? 

Also the reason we have the queue_control as Option because we don't want any control over listening to QueueType::JobHandleFailure.
but the code below : 
```
let max_concurrent_tasks = self.queue_control.max_message_count.unwrap_or(10) as usize;
```
will also put a max limit over it, which seems to differ from the config that we set in QUEUE. 
wdyt?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -83,65 +89,95 @@ impl EventWorker {
...
+    /// * `Result<(), EventSystemError>` - A result indicating whether the operation was successful or not
+    /// # Errors
+    /// * Returns an EventSystemError if the message cannot be handled
+    async fn handle_message(&self, message: ParsedMessage) -> EventSystemResult<()> {
+        match self.queue_type {
+            QueueType::WorkerTrigger => self.handle_worker_trigger(message).await,
+            QueueType::JobHandleFailure => self.handle_job_failure(message).await,
             _ => {
                 let job_state: JobState = JobState::try_from(self.queue_type.clone())?;
                 info!("Received message: {:?}, state: {:?}", message, job_state);
```
heemankv: Do you think we can move this statement inside the `handle_job_queue` fn?
Do you have a specific reason to keep it here ?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +284,33 @@ impl EventWorker {
     /// * It will also sleep for a longer duration if an error occurs to prevent a tight loop
     /// * It will log errors and messages for debugging purposes
     pub async fn run(&self) -> EventSystemResult<()> {
+        let mut tasks = JoinSet::new();
+        let max_concurrent_tasks = self.queue_control.max_message_count.unwrap_or(10) as usize;
```
heemankv: `.unwrap_or(10)`

let's remove this hardcoded value and allow the orchestrator to receive all the messages without limitation if not specified!
else this will limit the high throughput queue like DataSubmission

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +284,33 @@ impl EventWorker {
     /// * It will also sleep for a longer duration if an error occurs to prevent a tight loop
     /// * It will log errors and messages for debugging purposes
     pub async fn run(&self) -> EventSystemResult<()> {
```
heemankv, line 287: let's go to where this fn is being called : 
```
 match self.create_event_handler(q).await {
            Ok(handler) => match handler.run().await {
                Ok(_) => info!("Worker for queue type {:?} started successfully", q),
                Err(e) => {
                    let _ = eyre!("🚨Failed to start worker: {:?}", e);
                    tracing::error!("🚨Failed to start worker: {:?}", e)
                }
            },
            Err(e) => tracing::error!("🚨Failed to create handler: {:?}", e),
        };
```

given the run fn is a loop, it will keep looping indefinitely under normal operation and only end when failed.
we should update the `info!("Worker for queue type {:?} started successfully", q),` to say something more precise, it confused for a sec to see `started` there. 
wdyt?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +284,33 @@ impl EventWorker {
...
+                    error!("Task failed: {:?}", e);
+                }
+            }
+
             match self.get_message().await {
                 Ok(Some(message)) => match self.parse_message(&message) {
                     Ok(parsed_message) => {
-                        self.process_message(message, parsed_message).await?;
+                        let worker = self.clone();
+                        let parsed_message = parsed_message.clone();
```
heemankv: I believe we can work without using `.clone()` for `parsed_message`

heemankv: I believe we can further reduce the clones in this manner.
Screenshot 2025-06-03 at 12 51 16 AM

heemankv (changes requested):

orchestrator/src/setup/aws/sqs.rs
0xvasanth: yes, its an existing bug, fixed now

orchestrator/src/types/mod.rs
0xvasanth: i don't think this as much value

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Yes, your first point is absolutely correct. Regarding the second point, it’s generally recommended to set a higher ceiling value. We can always adjust it if needed, but without a defined ceiling, we risk planning container resource sizing

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: This code have some other change adding to this in shutdown Code, will take it then

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: same as about comment, we can increase the default but upper ceiling is must

orchestrator/src/setup/aws/sqs.rs
0xvasanth: dql_queue_name

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+                queue_control: Some(QueueControlConfig::new(None, 300, 30)),
+                supported_layers: vec![Layer::L2, Layer::L3],
+            },
+        );
+        map.insert(
+            QueueType::ProofRegistrationJobProcessing,
+            QueueConfig {
+                visibility_timeout: 300,
+                dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
+                queue_control: Some(QueueControlConfig::new(None, 300, 30)),
```
0xvasanth: switch this to default

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+pub struct DlqConfig {
+    pub max_receive_count: u32,
+    #[allow(dead_code)]
+    pub dlq_name: QueueType,
+}
+
+#[derive(Clone)]
+pub struct QueueControlConfig {
+    // Max message count is the maximum number of messages to receive from the queue.
+    pub max_message_count: Option<u32>,
```
0xvasanth: can we remove this with default to 10

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -34,9 +38,13 @@ impl EventWorker {
     /// * `config` - The configuration for the EventWorker
     /// # Returns
     /// * `EventWorker` - A new EventWorker instance
```
0xvasanth: update this docs with result

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -83,65 +91,95 @@ impl EventWorker {
...
+    /// * `message` - The message to be handled
+    /// * `job_state` - The state of the job
+    /// # Returns
+    /// * `Result<(), EventSystemError>` - A result indicating whether the operation was successful or not
+    /// # Errors
+    /// * Returns an EventSystemError if the message cannot be handled
+    async fn handle_job_queue(&self, message: &ParsedMessage, job_state: JobState) -> EventSystemResult<()> {
+        info!("Received message: {:?}, state: {:?}", message, job_state);
+        if let ParsedMessage::JobQueue(queue_message) = message {
+            let span = info_span!("job_queue", q = %self.queue_type, worker_id = %queue_message.id);
```
0xvasanth: worker_id - should been id

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +286,33 @@ impl EventWorker {
...
+                    error!("Task failed: {:?}", e);
+                }
+            }
+
             match self.get_message().await {
                 Ok(Some(message)) => match self.parse_message(&message) {
                     Ok(parsed_message) => {
-                        self.process_message(message, parsed_message).await?;
+                        let worker = self.clone();
+                        let parsed_message = parsed_message.clone();
```
0xvasanth: will remove this

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: update the message

Mohiiit (approved):

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -32,11 +35,12 @@ impl Resource for SQS {
     /// TODO: The dead letter queues will have a visibility timeout of 300 seconds and a max receive count of 5.
     /// If the dead letter queue is not configured, the dead letter queue will not be created.
     async fn setup(&self, layer: &Layer, args: Self::SetupArgs) -> OrchestratorResult<Self::SetupResult> {
-        for queue in QUEUES.iter() {
+        let queue_fn = |q_type| format!("{}_{}_{}", args.prefix, q_type, args.suffix);
```
heemankv: I believe this `queue_fn` will now not be needed, 
The ARN merge changes the logic here to not need `args.prefix, q_type, args.suffix`,
Ideally, we'd use something like : 
```rust
let queue_name = InnerSQS::get_queue_name_from_type(name, &queue.name);
```

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -52,7 +56,8 @@ impl Resource for SQS {
             attributes.insert(QueueAttributeName::VisibilityTimeout, queue.visibility_timeout.to_string());
 
             if let Some(dlq_config) = &queue.dlq_config {
-                let dlq_url = self.get_queue_url_from_client(&queue_name).await?;
+                let dlq_queue_name = queue_fn(&dlq_config.dlq_name);
```
heemankv: Same as above, 
```rust
let queue_name = InnerSQS::get_queue_name_from_type(name, &queue.name);
```

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -83,11 +88,11 @@ impl Resource for SQS {
 
     async fn is_ready_to_use(&self, layer: &Layer, args: &Self::SetupArgs) -> OrchestratorResult<bool> {
         let client = self.client().clone();
-        for queue in QUEUES.iter() {
+        for (query_type, queue) in QUEUES.iter() {
             if !queue.supported_layers.contains(layer) {
                 continue;
             }
-            let queue_name = format!("{}_{}_{}", args.prefix, queue.name, args.suffix);
+            let queue_name = format!("{}_{}_{}", args.prefix, query_type, args.suffix);
```
heemankv: Same as above 
```rust
let queue_name = InnerSQS::get_queue_name_from_type(name, &queue.name);
```

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -188,30 +217,41 @@ impl EventWorker {
...
             // self.config.alerts().send_message(error_context).await?;
 
-            // Return the specific error
             return Err(consumption_error.into());
         }
 
-        // Only acknowledge if processing was successful
         message.ack().await.map_err(|e| ConsumptionError::FailedToAcknowledgeMessage(e.0.to_string()))?;
-
         Ok(())
```
heemankv, line 237: Any reason we have removed the specific comments here ?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -155,9 +157,12 @@ impl SnosJobTrigger {
         config: &Arc<Config>,
         bounds: ProcessingBounds,
     ) -> Result<JobSchedulingContext> {
-        let service_config = config.service_config();
+        let max_slots = QUEUES
+            .get(&QueueType::SnosJobProcessing)
+            .ok_or_else(|| color_eyre::eyre::eyre!("SnosJobProcessing queue config not found"))?
+            .queue_control
+            .max_message_count as u64;
```
heemankv: Let's discuss this offline once! @itsparser

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+                queue_control: QueueControlConfig::default_with_message_count(50),
+                supported_layers: vec![Layer::L2, Layer::L3],
+            },
+        );
+        map.insert(
+            QueueType::SnosJobProcessing,
+            QueueConfig {
+                visibility_timeout: 300,
+                dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
+                queue_control: QueueControlConfig::default_with_message_count(200),
```
heemankv: I believe we should have capabilities to set this from the env as well.
Let's discuss this offline as well @itsparser

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -35,29 +35,30 @@ impl Resource for InnerSQS {
     /// TODO: The dead letter queues will have a visibility timeout of 300 seconds and a max receive count of 5.
     /// If the dead letter queue is not configured, the dead letter queue will not be created.
     async fn setup(&self, layer: &Layer, args: Self::SetupArgs) -> OrchestratorResult<Self::SetupResult> {
-        for queue in QUEUES.iter() {
+        // let queue_fn = |q_type| format!("{}_{}_{}", args.prefix, q_type, args.suffix);
```
Trantorian1: > [!NOTE]
> _**Style**_

Leftover code from testing?

```suggestion
```

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(Clone)]
+pub struct QueueConfig {
+    pub visibility_timeout: u32,
+    pub queue_control: QueueControlConfig,
+    pub dlq_config: Option<DlqConfig>,
+    pub supported_layers: Vec<Layer>,
+}
+
+lazy_static! {
```
Trantorian1: > [!NOTE]
> _nit:_  _**dependency**_

You can replace `lazy_static` usage with [`std::sync::LazyLock`](https://doc.rust-lang.org/stable/std/sync/struct.LazyLock.html#examples) or [`std::cell::LazyCell`](https://doc.rust-lang.org/stable/std/cell/struct.LazyCell.html#examples).

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+    // Max message count is the maximum number of messages to receive from the queue.
+    pub max_message_count: usize,
+    // Retry count is the number of times to retry the job if it fails.
+    pub retry_count: usize,
+    // Retry delay is the delay between retries in seconds.
+    pub retry_delay: u32,
+}
+impl QueueControlConfig {
+    pub fn default_with_message_count(max_message_count: usize) -> Self {
+        Self { max_message_count, retry_count: 3, retry_delay: 0 }
```
Trantorian1: > [!NOTE]
> _nit:_ _**style**_

```suggestion
        Self { max_message_count, ..Default::default() }
```

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub visibility_timeout: u32,
+    pub queue_control: QueueControlConfig,
+    pub dlq_config: Option<DlqConfig>,
+    pub supported_layers: Vec<Layer>,
+}
+
+lazy_static! {
+    pub static ref QUEUES: HashMap<QueueType, QueueConfig> = {
+        let mut map = HashMap::new();
+        map.insert(
```
Trantorian1: > [!TIP]
> _nit:_ _**style**_

Just in case you find this kind of map initialization a pain, you have the [`maplit`](https://docs.rs/maplit/1.0.2/maplit/index.html) crate which makes this sort of stuff easier :)

orchestrator/src/types/queue_control.rs
Trantorian1: Seconding this.

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+                queue_control: QueueControlConfig::new(10, 3, 30),
+                supported_layers: vec![Layer::L2, Layer::L3],
+            },
+        );
+        map.insert(
+            QueueType::ProvingJobVerification,
+            QueueConfig {
+                visibility_timeout: 300,
+                dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
+                queue_control: QueueControlConfig::new(10, 300, 30),
```
Trantorian1: > [!WARNING]
> _**Configuration**_

This seems like a very high retry count, is this normal? If we need to retry a proving task up to 300 times I would consider this an bug no? We might want to increase the `max_message_count` and `retry_delay` instead to avoid congestion.

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+                queue_control: QueueControlConfig::new(10, 3, 30),
+                supported_layers: vec![Layer::L3],
+            },
+        );
+        map.insert(
+            QueueType::ProofRegistrationJobVerification,
+            QueueConfig {
+                visibility_timeout: 300,
+                dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
+                queue_control: QueueControlConfig::new(10, 300, 30),
```
Trantorian1: > [!WARNING]
> _**Configuration**_

Same here.

orchestrator/src/types/queue_control.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#[derive(Clone)]
+pub struct QueueControlConfig {
+    // Max message count is the maximum number of messages to receive from the queue.
+    pub max_message_count: usize,
+    // Retry count is the number of times to retry the job if it fails.
+    pub retry_count: usize,
+    // Retry delay is the delay between retries in seconds.
+    pub retry_delay: u32,
```
Trantorian1: > [!TIP]
> _**Type**_

Why is this not an `std::time::Duration` instead?

orchestrator/src/error/consumer.rs
```diff
@@ -22,4 +22,7 @@ pub enum ConsumptionError {
 
     #[error("Failed to parse message from queue, error {0}")]
     FailedToAcknowledgeMessage(String),
+
+    #[error("Queue Type not found: {0}")]
+    QueueNotFound(String),
```
Trantorian1, line 27: > [!TIP]
> **Type**

Should we not keep the `crate::types::queue::QueueType` here instead of `String`?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -83,65 +84,93 @@ impl EventWorker {
...
-            QueueType::WorkerTrigger => {
-                info!("Received message to Handler: {:?}", message);
-                if let ParsedMessage::WorkerTrigger(worker_mes) = message {
-                    let span = info_span!("worker_trigger", q = %self.queue_type, worker_id = %worker_mes.worker);
-                    let _guard = span.enter();
-                    let worker_handler =
-                        JobHandlerService::get_worker_handler_from_worker_trigger_type(worker_mes.worker.clone());
-                    worker_handler
-                        .run_worker_if_enabled(self.config.clone())
+    async fn handle_worker_trigger(&self, message: &ParsedMessage) -> EventSystemResult<()> {
```
Trantorian1: > [!TIP]
> _**Type**_

Why not just enforce message be of `Box<crate::worker::parser::worker_trigger_message::WorkerTriggerMessage>` type? That way we can remove the error handling and make sure this function will only be called with valid data.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -83,65 +84,93 @@ impl EventWorker {
...
+    /// handle_job_failure - Handle the message received from the queue for JobHandleFailure type
+    /// This function processes the message based on its type
+    /// It returns a Result<(), EventSystemError> indicating whether the operation was successful or not
+    /// # Arguments
+    /// * `message` - The message to be handled
+    /// # Returns
+    /// * `Result<(), EventSystemError>` - A result indicating whether the operation was successful or not
+    /// # Errors
+    /// * Returns an EventSystemError if the message cannot be handled
+    async fn handle_job_failure(&self, message: &ParsedMessage) -> EventSystemResult<()> {
```
Trantorian1: > [!TIP]
> _**Type**_

Same here, why are we not requiring a `Box<crate::worker::parser::worker_trigger_message::JobQueueMessage>` as the type for `message`?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -83,65 +84,93 @@ impl EventWorker {
...
+    /// # Arguments
+    /// * `message` - The message to be handled
+    /// * `job_state` - The state of the job
+    /// # Returns
+    /// * `Result<(), EventSystemError>` - A result indicating whether the operation was successful or not
+    /// # Errors
+    /// * Returns an EventSystemError if the message cannot be handled
+    async fn handle_job_queue(&self, message: &ParsedMessage, job_state: JobState) -> EventSystemResult<()> {
+        info!("Received message: {:?}, state: {:?}", message, job_state);
+        if let ParsedMessage::JobQueue(queue_message) = message {
```
Trantorian1: > [!TIP]
> _**Type**_

Ditto.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +277,32 @@ impl EventWorker {
     /// * It will also sleep for a longer duration if an error occurs to prevent a tight loop
     /// * It will log errors and messages for debugging purposes
     pub async fn run(&self) -> EventSystemResult<()> {
+        let mut tasks = JoinSet::new();
+        let max_concurrent_tasks = self.queue_control.max_message_count;
+        info!("Starting worker with thread pool size: {}", max_concurrent_tasks);
+
         loop {
+            while tasks.len() >= max_concurrent_tasks {
```
Trantorian1: > [!CAUTION]
> _**Invariant**_

How is it that we can have more tasks in the `JoinSet` than the max concurrency limit? Then what is the point of that limit?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +277,32 @@ impl EventWorker {
...
+                    error!("Task failed: {:?}", e);
+                }
+            }
+
             match self.get_message().await {
                 Ok(Some(message)) => match self.parse_message(&message) {
                     Ok(parsed_message) => {
-                        self.process_message(message, parsed_message).await?;
+                        let worker = self.clone();
+                        tasks.spawn(async move { worker.process_message(message, parsed_message).await });
```
Trantorian1: > [!CAUTION]
> _**Invariant**_

Shouldn't we be adding a check here to make sure that we are not adding tasks to the `JoinSet` beyond the max concurrency limit?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -237,23 +277,32 @@ impl EventWorker {
...
                     }
                     Err(e) => {
                         error!("Failed to parse message: {:?}", e);
                     }
                 },
                 Ok(None) => {
-                    // Sleep to prevent tight loop and allow memory cleanup
                     sleep(Duration::from_secs(1)).await;
```
Trantorian1: > [!CAUTION]
> _**Sleep in async code**_

Why do we have a `sleep` inside of `async`  gode with the goal of "preventing a tight loop". This to me indicates that:

1. We do not have proper async waiting in place and that `get_message` is not properly signaling when the next message is available. In this case the code is basically `sync` with an `async` wrapper around.

2. `get_message` implements proper `async` waiting and will only yield when the next message is available. In this case there is no need to `sleep` since the runtime will automatically wake up this task when progress can be made (ie: there is risk of busy-waiting / "tight loops").

Am I correct in assuming that `get_message` does not correctly wait for the next message to be made available? In which case why is it `async`?

Trantorian1 (changes requested):
> [!CAUTION]
> The `async` code in `orchestrator/src/worker/controller/event_worker.rs` has some issues and needs to be reworked.

As it stands now I don't understand why we are calling `sleep` to wait on task advancement and I don't see how max concurrency limits are implemented. This feels rather strange.

Trantorian1 (changes requested):
Missed some files in my last review, mb.

orchestrator/src/worker/controller/event_worker.rs
Trantorian1: As a side note here, what was the previous comment referring to in terms of "allow memory cleanup"?

orchestrator/src/error/consumer.rs
0xvasanth: Than might force us to add all the queue to queue_type, it need not to been in future, i think it is better to been in string, since we might remove the QueueType in future

orchestrator/src/setup/aws/sqs.rs
0xvasanth: removed

orchestrator/src/types/queue_control.rs
0xvasanth: interesting, will follow with LazyLock

orchestrator/src/types/queue_control.rs
0xvasanth: The maplit crate offers the same functionality, so adding it would only introduce an unnecessary dependency. I’ll keep the current implementation instead.

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Have moved the type conversion to parent function, but the error handling can't be unavoidable  in the parent function

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Have moved the type conversion to parent function, but the error handling can't be unavoidable in the parent function

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: updated the code to async with non blocking IO

```mermaid
sequenceDiagram
    participant RT as Tokio Runtime
    participant T1 as Task 1: get_message()
    participant T2 as Task 2: cleanup_tasks()
    participant T3 as Task 3: process_message()
    participant SQS as AWS SQS
    participant DB as Database
    
    Note over RT: ✅ Single OS Thread - No Blocking
    
    RT->>T1: Schedule Task 1
    activate T1
    T1->>SQS: consumer.receive() [async]
    Note over T1,SQS: IO initiated, but thread not blocked
    T1-->>RT: yield (await point)
    deactivate T1
    
    RT->>T2: Schedule Task 2 
    activate T2
    T2->>T2: try_join_next() [non-blocking]
    T2-->>RT: yield (no tasks complete)
    deactivate T2
    
    RT->>T3: Schedule Task 3
    activate T3  
    T3->>DB: database.query() [async]
    Note over T3,DB: IO initiated, but thread not blocked
    T3-->>RT: yield (await point)
    deactivate T3
    
    Note over SQS: Message arrives after 2s
    SQS-->>T1: Message ready
    RT->>T1: Resume Task 1
    activate T1
    T1->>T1: Got message!
    T1-->>RT: return message
    deactivate T1
    
    Note over DB: Query completes
    DB-->>T3: Result ready  
    RT->>T3: Resume Task 3
    activate T3
    T3->>T3: Process result
    T3-->>RT: task complete
    deactivate T3
```

orchestrator/src/types/queue_control.rs
0xvasanth: Since this delay is not been used in the code now and need some clarity on the this with @heemankv case will take this to next PR , and focus only on the Q- Control

orchestrator/src/types/queue_control.rs
0xvasanth: same as above

orchestrator/src/types/queue_control.rs
0xvasanth: Since this delay is not been used in the code now and need some clarity on the this with @heemankv case will take this to next PR , and focus only on the Q- Control

orchestrator/src/types/queue_control.rs
0xvasanth: Since this delay is not been used in the code now and need some clarity on the this with @heemankv case will take this to next PR , and focus only on the Q- Control

orchestrator/src/types/queue_control.rs
Trantorian1: Oh yea def, was just mentioning this in case you weren't aware of it :)

orchestrator/src/types/queue_control.rs
Trantorian1: I still think this should be set to a lower limit similar to the other queues so we don't en up with a very high retry count in the future when this delay _becomes_ used.

orchestrator/src/types/queue_control.rs
```diff
@@ -43,117 +44,115 @@ pub struct QueueConfig {
...
-                visibility_timeout: 300,
-                dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
-                queue_control: QueueControlConfig::new(10, 10, 60),
-                supported_layers: vec![Layer::L2, Layer::L3],
-            },
-        );
-        map
-    };
-}
+pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(|| {
```
Trantorian1, line 42: > [!WARNING]
> _**Locking**_

If these resources are being used in a `sync` context then using a `std::cell::LazyCell` would be enough. Might be some issues in terms of `Send` bounds. Is this the case?

Trantorian1: Also note that relying on thread-blocking std locks in an async context can pose some issues. Again, I don't know if the use of `std::cell::LazyCell` was posing some issues so maybe this is the correct way to go.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -39,23 +41,29 @@ impl EventWorker {
...
-            Err(QueueError::NoData) => Ok(None),
-            Err(e) => Err(ConsumptionError::FailedToConsumeFromQueue { error_msg: e.to_string() })?,
+        loop {
+            debug!("Long polling for message from queue {:?}", self.queue_type);
+            match consumer.receive().await {
+                Ok(delivery) => return Ok(delivery),
+                Err(QueueError::NoData) => {
+                    tokio::time::sleep(Duration::from_millis(100)).await;
+                    continue;
+                }
```
Trantorian1, line 60: Should we perhaps implement a max retry here?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -279,31 +287,35 @@ impl EventWorker {
...
                 }
-                Err(e) => {
-                    error!("Error receiving message: {:?}", e);
-                    sleep(Duration::from_secs(5)).await;
+            } else {
+                if let Some(result) = tasks.join_next().await {
+                    if let Err(e) = result {
+                        error!("Task failed: {:?}", e);
+                    }
                 }
```
Trantorian1: There remains an issue with this code. I will try and do my best to outline it here. First of all, here is what I understand as being the objective of this function:

1. We implement a queue-based approach where tasks are continuously polled to make progress in execution.
2. The number of tasks executing is bounded to  `max_concurrent_tasks` to avoid congestion.

Here are some properties which we would like this function to have:

1. Task selection should be _fast_, meaning that as soon as we have less than `max_concurrent_tasks` we should be polling a new task (waiting for the next message).
2. Task selection should be bounded, so that we cannot add more tasks than `max_concurrent_tasks`.

## Emptying the `JoinSet`

```rust
while let Some(result) = tasks.try_join_next() {
    if let Err(e) = result {
        error!("Task failed: {:?}", e);
    }
}
```

This portion of the code isn't really doing anything. To understand why this is the case, let's check out the [tokio docs for `try_join_next`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.try_join_next):

> Tries to join one of the tasks in the set that has completed and return its output.
> Returns `None` **if there are no completed tasks**, or if the set is empty.

The key part here being **if there are not completed tasks**. So this is not _driving_ any task to completion, just collecting the results and checking for failures. This code isn't necessary since we are already waiting for tasks to complete later on in the function:

```rust
if let Some(result) = tasks.join_next().await {
    if let Err(e) = result {
        error!("Task failed: {:?}", e);
    }
}
```

This is also a problem if we consider that **Task selection should be _fast_**, since this is taking up time which could be spent spawning a message processing task instead. This is not much time, but this code isn't really doing anything so having it doesn't make sense.

Trantorian1: ## Limiting concurrency

As it is, it would be enough to just have:

```rust
loop {
    if tasks.len() < max_concurrent_tasks {
        // Get message - this now blocks until a message is available
        match self.get_message().await {
            Ok(message) => match self.parse_message(&message) {
                Ok(parsed_message) => {
                    let worker = self.clone();
                    tasks.spawn(async move { 
                        worker.process_message(message, parsed_message).await 
                    });
                }
                Err(e) => {
                    error!("Failed to parse message: {:?}", e);
                }
            },
            Err(e) => {
                error!("Error receiving message: {:?}", e);
            }
        }
    } else {
        if let Some(result) = tasks.join_next().await {
            if let Err(e) = result {
                error!("Task failed: {:?}", e);
            }
        }
    }
}
```

The reason why this works is that:

1. If we have less tasks in the set than `max_concurrent_tasks`, we wait for the next message and spawn a processing task with `tokio::task::JoinSet::spawn`. This will start making progress immediately, as per the [tokio docs for `spawn`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.spawn):

> The provided future will start running in the background immediately when this method is called, even if you don’t await anything on this `JoinSet`.

**This way we keep rapidly adding new message processing tasks to the join set as soon as we can. This satisfies property 1.**

2. If we have reached the concurrency limit (ie, `tasks.len() == max_concurrent_tasks`), then we will wait for any of the tasks in the set to finish with `tasks.join_next().await`. This `await` point will either complete immediately if a task has already finished or it will wait for the next task to finish if that is not the case. Either way, we can be sure that once this has completed that we will have _at least_ one more space available in `task`.

> The reason why we will have _at least_ on more space available in `task` is because `tokio::task::JoinSet::join_next()` does not actually drive any of its futures to completion, instead it just waits for them to complete. So while `join_next` will finish once _a_ future in the set has completed, it is possible that multiple futures complete at once, in which case we could have up to `max_concurrent_tasks` empty space in the set.

**This way when the join set is full, we wait for some of its tasks to complete so we can add new ones. This satisfies property 2.**

Trantorian1: All in all, this helps us maintain a _tight execution loop_, which is something we actually want as long as every `async` function is only making progress once it is ready, since the async runtime will handle waiting until each future can make progress.

I would highly suggest giving [this section of the docs](https://doc.rust-lang.org/stable/std/future/trait.Future.html#tymethod.poll) a read to get more context on how Rust's async futures are implemented under the hood.

Hopefully this can help clear up any misunderstandings in regards to this part of the code. Please let me know if you have any questions and I will do my best to follow up on those. Please do note that this kind of review does take a long time so it might be a few days before I come back to this.

Trantorian1 (changes requested):
One last async issue, otherwise seems good :+1:

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Thank you for the suggestion! However, I believe the current implementation is appropriate for this use case. The `QueueError::NoData` error simply indicates that the queue is currently empty — a normal operational state, not a system failure.

### What would happen with max retries:
> If we added a retry limit for `NoData`, the worker could shut down after several empty polls, even if the system is functioning correctly.  This could result in messages being missed if they arrive after the worker exits.

Here’s a helpful reference: [Link](https://docs.rs/omniqueue/0.2.0/src/omniqueue/backends/sqs.rs.html#260)

Let me know if I misunderstood your comment — happy to revisit it!

orchestrator/src/types/queue_control.rs
0xvasanth: Thanks for the input! We had an internal discussion (myself and @heemankv ) about this. While all current configurations are tied to environment variables, adding more at this point could make things harder to manage.

To address this, we're planning to introduce a config-based system that will better handle values like retry intervals and delays. Once that’s in place, this logic will be moved into that system in a future update.

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: this while might be need for the shutdown code, that is been move to new PR this is irrelevant now

orchestrator/src/types/queue_control.rs
heemankv: Being tracked here : #706

orchestrator/src/types/queue_control.rs
0xvasanth: LazyCell is not thread-safe (Sync), So can't be used in static variables.

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: since, it straight-forward seems this useless here

heemankv (approved):

orchestrator/src/types/queue_control.rs
```diff
@@ -15,11 +15,6 @@ pub struct DlqConfig {
 pub struct QueueControlConfig {
     // Max message count is the maximum number of messages to receive from the queue.
     pub max_message_count: usize,
-    // TODO: Add these fields back when we have a retry mechanism from the queue
-    // // Retry count is the number of times to retry the job if it fails.
-    // pub retry_count: usize,
-    // // Retry delay is the delay between retries in seconds.
-    // pub retry_delay: u32,
```
Trantorian1: Why was this removed?

orchestrator/src/worker/controller/event_worker.rs
Trantorian1, line 1: Previous comment was not addressed :/

Trantorian1 (changes requested):

orchestrator/src/worker/controller/event_worker.rs
Trantorian1: https://github.com/madara-alliance/madara/pull/667#discussion_r2153952897

orchestrator/src/types/queue_control.rs
0xvasanth: We’re currently having an active discussion around how to handle config vs secret management, so we’ve decided to scope this PR strictly to the QController logic for now.

We’ll revisit the retry mechanism separately once there’s clarity on the broader config handling approach.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -153,31 +158,50 @@ impl ProverClient for SharpProverService {
...
+
+    /// TODO: We need to implement this function for the prover client while adding the testcase
+    /// or while using the sharp prover client.
+    async fn submit_l2_query(
+        &self,
+        _task_id: &str,
+        _fact: &str,
+        _n_steps: Option<usize>,
+    ) -> Result<String, ProverClientError> {
+        todo!()
```
Trantorian1, line 176: No risk of this being called in prod?

orchestrator/src/utils/instrument.rs
```diff
@@ -46,10 +45,13 @@ impl OrchestratorInstrumentation {
...
 
-                tracing_subscriber
+                let subscriber = tracing_subscriber
                     .with(OpenTelemetryLayer::new(tracer))
-                    .with(OpenTelemetryTracingBridge::new(&logger))
-                    .init();
+                    .with(OpenTelemetryTracingBridge::new(&logger));
+
+                // Force overwrite any existing global subscriber
+                let _ = tracing::subscriber::set_global_default(subscriber);
```
Trantorian1, line 53: This could pose some issues with initializing logging in a test context. Are we sure this function is never called in a test path?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -290,11 +290,6 @@ impl EventWorker {
         info!("Starting worker with concurrency limit: {}", max_concurrent_tasks);
 
         loop {
-            while let Some(result) = tasks.try_join_next() {
-                if let Err(e) = result {
-                    error!("Task failed: {:?}", e);
-                }
-            }
```
Trantorian1: Thanks for the change!

Trantorian1 (approved):

orchestrator/src/utils/instrument.rs
0xvasanth: I’m sure this shouldn’t cause any issues in test contexts, as we’re not using OTEL (OpenTelemetry) in our test cases.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
0xvasanth: This shouldn’t cause any issues — we’re not using the Sharp service in production at the moment.
Our current focus is entirely on the Atlantic service, so this path remains unused for now.

PR #675, opened 2025-06-03
Update workflow to create builds for all components on merge

prkpndy:
Update workflow to create builds for all components on merge and fix Dockerfiles for all components

## Pull Request type

- Feature

Currently, we have dockerfiles scattered around in the code - some in root while others in their respective directories. Also, we don't create images for orchestrator and bootstrapper during merge with main.

Resolves: #NA

## What is the new behavior?

Moved all the dockerfiles to the root. Also, updated the merge to main workflow to create nightly builds for all 3 components and publish them on ghcr.

NO


bootstrapper.dockerfile
```diff
@@ -0,0 +1,125 @@
...
+    asdf global scarb 2.8.4
+
+# Set working directory
+WORKDIR /app
+
+# Copy the entire project
+COPY . .
+
+RUN apt-get update && apt-get install -y \
+    wget
```
Mohiiit: we can move this to top where we are installing all the dependencies?

bootstrapper.dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+RUN curl https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.6.12+commit.27d51765 -o /usr/local/bin/solc-0.6.12
+RUN echo 'f6cb519b01dabc61cab4c184a3db11aa591d18151e362fcae850e42cffdfb09a /usr/local/bin/solc-0.6.12' | sha256sum --check
+RUN chmod +x /usr/local/bin/solc-0.6.12
+RUN npm install -g --unsafe-perm ganache-cli@6.12.2
+
+# Setting it to avoid building artifacts again inside docker
+ENV RUST_BUILD_DOCKER=true
```
Mohiiit: interesting, if we are not building the artifacts in the docker, then there are a few things that we don't really need right? like scarb, ganache, solc etc etc

I think we can optimize this a lot with this

madara.dockerfile
```diff
@@ -15,42 +15,38 @@ ENV CHEF_VERSION=v0.1.71
 ENV CHEF_URL=https://github.com/LukeMathWalker/cargo-chef/releases/download/${CHEF_VERSION}/cargo-chef-x86_64-unknown-linux-gnu.tar.gz
 ENV CHEF_TAR=cargo-chef-x86_64-unknown-linux-gnu.tar.gz
 
-ENV VERSION_MOLD="2.39.1"
-ENV MOLD_URL=https://github.com/rui314/mold/releases/download/v$VERSION_MOLD/mold-$VERSION_MOLD-x86_64-linux.tar.gz
-
 ENV RUSTC_WRAPPER=/bin/sccache
 
-ENV WGET="-O- --timeout=10 --waitretry=3 --retry-connrefused --progress=dot:mega"
```
Mohiiit: why are we removing this? looks like a good thing to have in case wget fails the first time!

orchestrator.dockerfile
```diff
@@ -0,0 +1,94 @@
...
+RUN python -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+
+# Install Python dependencies in the virtual environment
+RUN pip install --upgrade pip
+RUN pip install cairo-lang==0.13.2 "sympy<1.13.0"
+
+# Continue with the build
+RUN mkdir -p orchestrator/build
+RUN cairo-compile orchestrator/cairo-lang/src/starkware/starknet/core/os/os.cairo --output orchestrator/build/os_latest.json --cairo_path orchestrator/cairo-lang/src
```
Mohiiit: I think we can remove this as well!

.github/workflows/pull-request-merge.yml
```diff
@@ -11,31 +11,74 @@ jobs:
   #                              NIGHTLY RELEASE                              #
   # ========================================================================= #
 
-  build-nightly:
+  build-nightly-madara:
     uses: ./.github/workflows/task-build-nightly.yml
     with:
       image-name: madara
-      image-file: ./madara/Dockerfile
+      image-file: ./madara.dockerfile
```
Trantorian1: I don't like the idea of having all the Dockerfiles at the root. It seems to me this PR might have been started before the CI update, which moved each dockerfile to their respective project directories. Imo it is easier to track and more compartmentalized if we keep each dockerfile there (so `madara/Dockerfile`, `bootstrapper/Dockerfile`...).

.github/workflows/pull-request-merge.yml
```diff
@@ -11,31 +11,74 @@ jobs:
...
       attestations: write
       id-token: write
     secrets: inherit
 
-  # test-hive:
-  #   needs: [build-nightly]
-  #   uses: ./.github/workflows/task-test-hive.yml
-  #   with:
-  #     nightly-sha: ${{ needs.build-nightly.outputs.nightly-sha }}
-  #   secrets: inherit
```
Trantorian1: Please keep this! This action works but the hive tests have to be updated before we can run it as part of the Madara CI

Trantorian1: Left this so it is easy to re-activate them in the future.

.github/workflows/task-build-nightly.yml
```diff
@@ -42,6 +46,12 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Download Cairo artiracts
+        uses: actions/download-artifact@v4
+        with:
+          name: build-artifacts-${{ inputs.cairo-artifacts-hash }}
+          path: build-artifacts/
+
```
Trantorian1: Please remove this! It is unsound to rely on the build artifacts gh artifacts since their retention policy is set to 1 day! Instead, you should call `cargo check` which will cause the madara build script to be triggered and the artifacts to be downloaded. Actually this already present in this file:

```yaml
      - name: Download artifacts
        run: cargo check
```

bootstrapper.dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Karnot Bootstrapper
+# ==============================================
+FROM ubuntu:22.04 AS builder
+
+# Install basic dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    build-essential \
+    clang \
+    mold \
```
Trantorian1: Mold is no longer used a linker since #647!

bootstrapper.dockerfile
Trantorian1: > interesting, if we are not building the artifacts in the docker, then there are a few things that we don't really need right? like scarb, ganache, solc etc etc

Seconding this

bootstrapper.dockerfile
```diff
@@ -0,0 +1,61 @@
...
+WORKDIR /app
+
+# Copy the entire project
+COPY . .
+
+# Setting it to avoid building artifacts again inside docker
+ENV RUST_BUILD_DOCKER=true
+
+# Build the Rust project with specific binary name
+RUN cargo build --release --workspace --bin bootstrapper
```
Trantorian1: Please use `cargo chef` along with `sccache` and bind mounts to speed up docker builds. You can follow the current [madara Dockerfile](https://github.com/madara-alliance/madara/blob/main/madara/Dockerfile) as an example.

bootstrapper.dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+
+# Set working directory
+WORKDIR /app
+
+# Environment variables
+ENV RUST_LOG=info
+
+# Run the binary
+ENTRYPOINT ["/usr/local/bin/bootstrapper"]
```
Trantorian1: Please add `tini` to handle interrupt forwarding and clean exits!

orchestrator.dockerfile
Trantorian1, line 1: Same as before, it would be great if this could take advantage of `cargo-chef`, `sccache` and bind mounts to speed up building!

orchestrator.dockerfile
```diff
@@ -0,0 +1,82 @@
...
+COPY --from=builder /usr/src/madara/orchestrator/crates/da-clients/ethereum/trusted_setup.txt /tmp/trusted_setup.txt
+COPY --from=builder /tmp/kzg_dirs.txt /tmp/kzg_dirs.txt
+
+# Recreate the dynamic dirs and copy the trusted setup into them
+RUN while read dir; do \
+    mkdir -p "$dir" && \
+    cp /tmp/trusted_setup.txt "$dir/trusted_setup.txt"; \
+    done < /tmp/kzg_dirs.txt
+
+ENTRYPOINT ["/bin/bash"]
```
Trantorian1: Please add `tini`!

orchestrator.dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN cd orchestrator && npm install
+
+# Dump information where kzg files must be copied
+RUN mkdir -p /tmp && \
+    (find /usr/local/cargo -type d -path "*/crates/starknet-os/kzg" > /tmp/kzg_dirs.txt 2>/dev/null || \
+     find $HOME/.cargo -type d -path "*/crates/starknet-os/kzg" >> /tmp/kzg_dirs.txt 2>/dev/null || \
+     touch /tmp/kzg_dirs.txt)
+
+
+FROM debian:bookworm
```
Trantorian1: This should be `bookworm-slim` if possible.

Trantorian1 (dismissed):

bootstrapper.dockerfile
Mohiiit: I have resolved this

.github/workflows/pull-request-merge.yml
Mohiiit: I think the relative paths for the artifacts becomes an issue, but I am not sure, I will check

.github/workflows/task-build-nightly.yml
Mohiiit: cargo check was added in one of the recent PRs I think, will take care of this!

.dockerignore
```diff
@@ -1,8 +1,21 @@
...
+
+# Ignore editor and IDE files
+.vscode/
+.idea/
+*.swp
+*.swo
+
+# Ignore scripts and target directory
 scripts
 target/
```
byteZorvin, line 14: I guess we should have this `**/target/**`. As there could be target folder inside `madara` folder too. When I was trying to build build it was trying to copy `madara/target` folder which was 34GB for me and my system was going out of resources due it

byteZorvin (changes requested):

.dockerignore
byteZorvin: It would be good to add `node_modules/*` too, although not entirely sure on this one

.github/workflows/pull-request-merge.yml
Mohiiit: @prkpndy let's keep this for now as mentioned by @Trantorian1

.github/workflows/task-build-nightly.yml
Mohiiit: @prkpndy instead of downloading the artifacts, if we use cargo check, it will handle it automatically

.github/workflows/pull-request-merge.yml
prkpndy: done

.github/workflows/task-build-nightly.yml
prkpndy: done

.dockerignore
prkpndy: done

.github/workflows/task-build-nightly.yml
```diff
@@ -18,6 +18,10 @@ on:
         required: false
         default: ghcr.io
         type: string
+      cairo-artifacts-hash:
+        description: "Hash used to retrieve the cairo artifacts"
+        required: true
+        type: string
```
Mohiiit: we can remove this now I think?

Mohiiit (approved):
LGTM

.github/workflows/task-build-nightly.yml
prkpndy: done

orchestrator/Dockerfile
```diff
@@ -1,92 +1,116 @@
...
+ENV TINI_VERSION=v0.19.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /bin/tini
+RUN chmod +x /bin/tini
 
-# Set the entrypoint to the startup script
-ENTRYPOINT ["./start.sh"]
\ No newline at end of file
+# Set the entrypoint
+ENTRYPOINT ["tini", "--", "orchestrator"]
+CMD ["--help"]
```
0xvasanth, line 116: Since after the entry point this command does not make anysence here might be mistake

orchestrator/Dockerfile
```diff
@@ -1,92 +1,116 @@
...
-COPY ./crates/da-clients/ethereum/trusted_setup.txt /usr/src/madara-orchestrator/crates/settlement-clients/ethereum/src/trusted_setup.txt
+EXPOSE 3000
 
-# Create a startup script
-RUN echo '#!/bin/bash\n\
-    npm run migrate up\n\
-    ./orchestrator' > start.sh && chmod +x start.sh
+ENV TINI_VERSION=v0.19.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /bin/tini
+RUN chmod +x /bin/tini
```
0xvasanth, line 112: Is there any specific reason behind the usage of this Tini ?

orchestrator/Dockerfile
```diff
@@ -1,92 +1,116 @@
...
+WORKDIR /app
+
+# Note that we do not install cargo chef and sccache through docker to avoid
+# having to compile them from source
+ENV SCCACHE_VERSION=v0.10.0
+ENV SCCACHE_URL=https://github.com/mozilla/sccache/releases/download/${SCCACHE_VERSION}/sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl.tar.gz
+ENV SCCACHE_TAR=sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl.tar.gz
+ENV SCCACHE_BIN=/bin/sccache
+ENV SCCACHE_DIR=/sccache
+ENV SCCACHE=sccache-${SCCACHE_VERSION}-x86_64-unknown-linux-musl/sccache
```
0xvasanth, line 12: why sccache, does cargo-chef is enough to do the caching?

orchestrator/Dockerfile
prkpndy: EntryPoint defines the executable that will be run when we don `docker run <image-name>`
Cmd is the default command that will be used while running the docker image
So simply doing `docker run orchestrator` will run `orchestrator --help`
Doing `docker run orchestrator run [other args]` will do `run` the orchestrator
This is a standard practice I believe. Or am I missing something?

orchestrator/Dockerfile
prkpndy: This was used in madara by Kasar. So I used the same version here as well

orchestrator/Dockerfile
prkpndy: sscache is used to do fine-grain dependency caching.
For eg, if we change even a single dependency, the cargo chef recipe.json will change and it'll force re-compilation of everything. 
sscache will prevent this

byteZorvin (approved):

orchestrator/Dockerfile
0xvasanth: cargo-chef does the same job right ?

orchestrator/Dockerfile
0xvasanth: I dont think this is standard practice, since mostly the entry-point will be blocked CMD does not make any reason to be there

PR #676, opened 2025-06-03
refactor: enhance worker initialization and shutdown handling

0xvasanth:
## Pull Request type

- Refactoring (shutdown handling)

## What is the new behavior?

- Updated `initialize_worker` to return a `WorkerController` for better management of worker lifecycle.
- Implemented graceful shutdown for workers in the orchestrator's main run loop.
- Added `shutdown` method to `EventWorker` and `WorkerController` for coordinated shutdown of all workers.
- Improved error handling in event processing and added a new `MutexPoisonError` variant to `EventSystemError` for better diagnostics.


orchestrator/src/setup/aws/sqs.rs
```diff
@@ -32,11 +35,12 @@ impl Resource for SQS {
     /// TODO: The dead letter queues will have a visibility timeout of 300 seconds and a max receive count of 5.
     /// If the dead letter queue is not configured, the dead letter queue will not be created.
     async fn setup(&self, layer: &Layer, args: Self::SetupArgs) -> OrchestratorResult<Self::SetupResult> {
-        for queue in QUEUES.iter() {
+        let queue_fn = |q_type| format!("{}_{}_{}", args.prefix, q_type, args.suffix);
```
prkpndy: Can give this a better name I think? something like `get_queue_name`?

orchestrator/src/tests/server/mod.rs
```diff
@@ -41,13 +41,7 @@ async fn test_health_endpoint() {
...
     let services = TestConfigBuilder::new().configure_queue_client(ConfigType::Actual).build().await;
 
-    let result = tokio::select! {
-        res = initialize_worker(services.config) => res,
-        _ = tokio::time::sleep(std::time::Duration::from_secs(10)) => {
-            println!("Timed out waiting for consumers to initialize");
-            Ok(())
-        }
-    };
+    let result = initialize_worker(services.config).await;
```
prkpndy: Is this test passing now? If yes, we can update the comment above this test

orchestrator/src/core/config.rs
```diff
@@ -84,21 +81,9 @@ impl Config {
...
-        Self {
-            params,
-            madara_client,
-            database,
-            storage,
-            alerts,
-            queue,
-            prover_client,
-            da_client,
-            processing_locks,
```
prkpndy: Why are we removing processing locks from the code ?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -221,12 +278,12 @@ impl EventWorker {
                 }
                 let _ = eyre!("Failed to process message: {:?}", e);
```
prkpndy: Can we remove this or is it doing something ?

prkpndy (commented):
LFG 🚀

Mohiiit:
yo, I was trying to make it easy for me to review! given this PR is dependent on the #667

orchestrator/src/core/config.rs
heemankv: Better to review : https://github.com/madara-alliance/madara/pull/667 independently first

Mohiiit (approved):

orchestrator/src/main.rs
```diff
@@ -71,14 +72,48 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                worker_controller.shutdown().await?;
+
+                // Analytics Shutdown
+                instrumentation.shutdown()?;
+
+                info!("All components shutdown successfully");
+                Ok(())
+            },
+            300, // 300 seconds timeout - matches Docker's default graceful shutdown period
+                 // Note: This Timeout might be to much for this case might need to revisit after consuting with mohit
```
0xvasanth: @Mohiiit can you confirm this, I am just adding a max timeout to shutdown the Orchestrator is unable to shutdown in time

eenthiran: Refernce: have confimed this we are going with 300 for now and observe for time being

0xvasanth: NOTE: we have discussed this and we are going with 300 for now and observe for time being

cchudant (approved):
the Notify can be replaced with a CancellationToken, I think it would be cleaner
other than that, I dont have enough context to do a full review but I think it looks fine

orchestrator/src/setup/aws/sqs.rs
0xvasanth: cleaned this code

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -32,13 +35,35 @@ impl EventWorker {
...
         let queue_config = QUEUES.get(&queue_type).ok_or(ConsumptionError::QueueNotFound(queue_type.to_string()))?;
         let queue_control = queue_config.queue_control.clone();
-        Ok(Self { queue_type, config, queue_control })
+        Ok(Self { queue_type, config, queue_control, cancellation_token })
+    }
+
+    /// Create a child token for this worker's specific operations
+    pub fn child_token(&self) -> CancellationToken {
+        self.cancellation_token.child_token()
+    }
```
Mohiiit: child token won't be used in the event worker itself right? I get it that we are using the controller's child_token for each event_worker but will there be any need after event_worker?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -287,27 +312,81 @@ impl EventWorker {
...
+                    } else {
+                        debug!("Task completed, active tasks: {}", tasks.len());
+                    }
+                }
+
+                // Handle shutdown signal
+                _ = self.cancellation_token.cancelled() => {
+                    info!("Shutdown signal received, breaking from main loop");
+                    break;
+                }
```
Mohiiit, line 335:
```suggestion
                _ = Self::is_shutdown_requested() => {
                    info!("Shutdown signal received, breaking from main loop");
                    break;
                }
```

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -287,27 +312,81 @@ impl EventWorker {
...
+            }
+
+            tokio::select! {
+                biased;
+
+                // Immediate cleanup when tasks complete
+                Some(result) = tasks.join_next(), if !tasks.is_empty() => {
+                    Self::handle_task_result(result);
+
+                    if tasks.len() >= max_concurrent_tasks - 1 {
```
Mohiiit: should this be like:
```suggestion
                    if tasks.len() >= max_concurrent_tasks  {
```

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -287,27 +312,81 @@ impl EventWorker {
...
+                                tasks.spawn(async move {
+                                    worker.process_message(message, parsed_message).await
+                                });
+                                debug!("Spawned task, active: {}", tasks.len());
+                            }
                         }
                         Err(e) => {
-                            error!("Failed to parse message: {:?}", e);
+                            error!("Error receiving message: {:?}", e);
+                            sleep(Duration::from_secs(1)).await;
```
Mohiiit, line 351: let's move this duration to const above, also just curious, what are the cases where we can encounter this? like this condition

orchestrator/src/worker/controller/worker_controller.rs
```diff
@@ -19,15 +24,27 @@ impl WorkerController {
...
+    }
+
+    /// workers - Get the list of workers
+    /// This function returns the list of workers
+    /// # Returns
+    /// * `Vec<Arc<EventWorker>>` - A list of workers
+    /// # Errors
+    /// * `EventSystemError` - If there is an error during the operation
+    pub fn workers(&self) -> EventSystemResult<Vec<Arc<EventWorker>>> {
+        let workers = self.workers.lock().map_err(|e| EventSystemError::MutexPoisonError(e.to_string()))?;
```
Mohiiit, line 41: this lock get dropped after this function call ends?

orchestrator/src/main.rs
```diff
@@ -71,14 +72,47 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                // Graceful shutdown for workers
+                worker_controller.shutdown().await?;
+
+                // Analytics Shutdown
+                instrumentation.shutdown()?;
+
+                info!("All components shutdown successfully");
+                Ok(())
+            },
+            300, // 300 seconds timeout - matches Docker's default graceful shutdown period
```
Mohiiit: let's move this to const as well

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Actually it is in correct parameter if we remove the -1 then we need to remove the = too make this Simplifier have done it

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: @Mohiiit You were right, The `child_token()` method in EventWorker was never called.

orchestrator/src/worker/controller/worker_controller.rs
0xvasanth: yes, but the object will be accessible

PR #668, opened 2025-05-27
feat: l2 gas support

jbcaron:
## Pull Request type

Please add the labels corresponding to the type of changes your PR introduces:

- Feature

## What is the current behavior?

- adds L2 gas handling for gateway and primitives
- refactors the L1 gas provider
- adds L2 gas calculation for block production

Resolves: #NA

## What is the new behavior?

-

## Does this introduce a breaking change?

Yes

PR #678, opened 2025-06-04
l1 to l2 messaging refactor: fix a number of bug & tidy up

cchudant:
The main aim of this PR is:
- Remove l1 handle transactions from the mempool & remove declare v0 transactions from the mempool too
  - this is a prerequisite needed for a future PR for v0.14 mempool support.
- Fix a number of bugs around l1 to l2 message handling and strenghten correctness around these parts: cases where you could actually softlock your chain.

# In details

This PR fixes a number of bugs around the L1 messaging worker, specifically:
- We now support restarting the sequencer on a clean database, without having run the nonce worker before. Previously, this setup was very likely to repeat L1 to L2 messages.
- L1 to L2 message deduplication was possible in some specific circonstances. This was really really bad, as producing an invalid block that consumes an L1 to L2 message that actually doesn't exist anymore on the core contract will lead to the core contract declining the state update, thus softlocking the chain
  - This is not possible anymore: block production now has some extra logic to offer a much stronger guarantee that this can never be the case.
- Fix the possibility of consuming messages that were canceled on l1. This should never be possible now, as the messages always re-checked for cancellation & validity just before creating a batch in block production, instead of checking only when the message is added to db.
- When starting the messaging worker for the first time, it will now look in the past for messages that was potentially missed. This should be much better UX, and allow for starting a sequencer on an already existing state without loss of l1 to l2 messages.
- Fixed the Starknet messaging worker. It did not follow the settlement chain properly, and polling was implemented at the wrong place.

Furthermore, this PR has a bit of refactor around how batches are created:
- L1HandlerTransactions are not in the mempool anymore.
- Instead, the batcher task now picks transactions from 3 different places at once when making a batch:
  - Transactions from mempool
  - L1 to L2 messages from the settlement layer client, through a new trait.
  - And, a newly created "Bypass" channel.

### Motivation

When bootstrapping a new chain from scratch, the state is completely empty. You thus need 2 things to make the chain actually usable:
- **A way to bypass fees**: the ERC20 contract for the fee token is not deployed yet. This means that any normal transaction that has any cost will fail, because the fee `transfer` invocation will error.
- **A way to declare the first class**: you need a deployed account to declare a class (so that `__validate_declare__` can be called). That's a chicken and egg problem: you can't have a deployed account without having at least a class declared.

To solve the first issue, we used to set the gas prices to 0 in the header of the genesis block. This is a weird behavior and we found that it is actually not necessary at all! As it turns out, SNOS skip every fee check when the max gas of the transaction = 0. Blockifier also has this behavior, but will fail validation when using StatefulValidator (the validation for mempool entry).

To solve the second issue, starknet used to have a Declare v0 transaction type. In v0.14 chains, this declare v0 transaction does not work with SNOS anymore; instead, you're now supposed to create a declare v3 transaction where the sender_address has the magic value of `'BOOTSTRAP'`.

However, this poses a new issue: when adding a transaction to the mempool, we validate it. This execution environment rejects any transaction that use these two mechanisms. This is a good thing: we don't want users to be able to use these two capabilities!
But this creates the need for a new way to add transactions to the chain: the bypass channel.

### Implementation

The bypass channel is a channel that shouldn't be accessible to users. You can send transactions through it using the admin RPC, and they will be included in a first-come-first-serve fashion, with a priority over the other channels. Meaning, even if the mempool is full, sending a transaction through the bypass channel will ensure that this transaction is included directly with priority over any other.
This means that the bypass channel might also be useful even when not doing boostrapping :) it's just a general purpose "force transaction insertion, bypassing the mempool" set of endpoints. You will always be able to send transactions with max_gas=0, bypassing all fee checks through this channel - that might prove useful i guess? for testing, disaster recovery, or anything like that.

Also: Declare v0 transactions are still supported, if you want to bootstrap pre v0.14 chains. They go through this bypass channel now instead of mempool.

Also: the `close_block` function is now available on the admin rpc 🥳 When called, this forces the closing of the current block. This will allow us to implement much less flaky tests! in your e2e tests, you can put a very high block time and close the blocks by hand to allow for repeatable state :)
The functionality was already there, I just didn't hook that up to admin rpc until now.


build-artifacts/build.docker
Trantorian1, line 1: > [!CAUTION]
> _**Build error**_

This is failing to build when calling `make artifacts` on my end.

```
 > [orchestrator-tests 5/5] RUN ./scarb/bin/scarb build:                                                          
1.033 /bin/sh: 1: ./scarb/bin/scarb: Exec format error                                                            
------                                                                                                            
build.docker:203                                                                                                  
--------------------                                                                                              
 201 |                                                                                                            
 202 |     # Compile contracts                                                                                    
 203 | >>> RUN ./scarb/bin/scarb build                                                                            
 204 |                                                                                                            
 205 |     # =========================================================================== #                        
--------------------                                                                                              
ERROR: failed to solve: process "/bin/sh -c ./scarb/bin/scarb build" did not complete successfully: exit code: 126
```

madara/crates/client/block_production/src/batcher.rs
Trantorian1, line 1: Cool stuff!

madara/crates/client/block_production/src/executor/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+            ]
+            .into_iter()
+            .collect(),
+        )
+        .await
+        .unwrap();
+
+    assert_matches!(setup.handle.replies.recv().await, Some(ExecutorMessage::StartNewBlock { .. }));
+    assert_matches!(setup.handle.replies.recv().await, Some(ExecutorMessage::BatchExecuted(res)) => {
+        assert_eq!(res.executed_txs.len(), 1); // only one transaction! not two
```
Trantorian1, line 175: Why does only 1 transaction get executed in this case?

Trantorian1: Duplicate nonce I assume? It would be nice to make that a bit clearer.

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -311,7 +350,24 @@ impl ExecutorThread {
                     WaitTxBatchOutcome::Exit => return Ok(()),
                 };
 
-                to_exec.extend(taken);
+                for (tx, additional_info) in taken {
+                    // Remove duplicate l1handlertxs. We want to be absolutely sure we're not duplicating them.
+                    if let Some(nonce) = tx.l1_handler_tx_nonce() {
+                        let nonce: u64 = nonce.to_felt().try_into().context("Converting nonce from felt to u64")?;
```
Trantorian1, line 316: > [!WARNING]
> _**bounds check**_

There is no case where this `Nonce` can fall out of the bounds of `u64`?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -2049,4 +2085,111 @@ pub(crate) mod tests {
...
+        .clone();
+        assert_eq!(receipt.execution_result(), ExecutionResult::Succeeded);
+        tracing::info!("Events = {:?}", receipt.events());
+        assert_eq!(receipt.events().len(), 1);
+
+        // pub struct CalledFromL1 {
+        //     pub from_address: felt252,
+        //     pub arg1: felt252,
+        //     pub arg2: felt252,
+        // }
```
Trantorian1:
```suggestion
```

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
Trantorian1, line 1: Should we add tests for those or are they being used in some e2e tests anyway?

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -701,148 +714,170 @@ mod l1_messaging_tests {
...
+    //     // assert_eq!(last_block.block_number, 0);
+    //     // let nonce = Nonce(Felt::from_dec_str("0").expect("Should parse the known valid test nonce"));
+    //     // // cancelled message nonce should be inserted to avoid reprocessing
+    //     // assert!(db.backend().has_l1_messaging_nonce(nonce).unwrap());
+    //     // assert!(logs_contain("Message was cancelled in block at timestamp: 0x66b4f105"));
+
+    //     todo!()
+
+    //     // worker_handle.abort();
+    // }
```
Trantorian1:
```suggestion
```

madara/crates/client/settlement_client/src/gas_price.rs
```diff
@@ -58,82 +56,67 @@ impl L1BlockMetrics {
...
-        )
-        .await
-        {
-            Ok(_) => {}
-            Err(e) => return Err(SettlementClientError::GasPrice(format!("Failed to update gas prices: {}", e))),
-        }
-    }
+impl L1ClientImpl {
+    pub(crate) async fn gas_price_worker(
+        self: Arc<Self>,
```
Trantorian1: This feels a bit strange. Couldn't we work off `&self` directly?

madara/crates/client/settlement_client/src/lib.rs
```diff
@@ -1,9 +1,146 @@
...
+    }
+
+    pub async fn new_starknet(
+        backend: Arc<MadaraBackend>,
+        rpc_url: Url,
+        core_contract_address: String,
+    ) -> anyhow::Result<Self> {
+        let provider = StarknetClient::new(StarknetClientConfig { rpc_url, core_contract_address })
+            .await
+            .context("Creating ethereum client")?;
```
Trantorian1: > [!Warning]
> _**Typo**_

```suggestion
            .context("Creating Starknet client")?;
```

madara/crates/client/settlement_client/src/lib.rs
```diff
@@ -1,9 +1,146 @@
...
+mod utils;
+
+/// Interface consumed by downstream crates for opeations that they may want to do with the settlement layer.
+pub trait SettlementClient: Send + Sync + 'static {
+    /// Create a stream consuming pending messages to l2.
+    fn create_message_to_l2_consumer(&self) -> BoxStream<'static, anyhow::Result<L1HandlerTransactionWithFee>>;
+    /// Get the messages to l2 that an l1 transaction sent to the core contract.
+    /// Mainly useful for the get_message_status rpc method.
+    fn get_transaction_messages(&self, transaction_hash: L1TransactionHash)
+        -> Option<Vec<L1HandlerTransactionWithFee>>;
```
Trantorian1: What is the point of this method if we are only returning `None`? Is this a placeholder for future use or do we have other implementations outside of this file?

Trantorian1: Just checked and couldn't find any other impls than those in the current file.

madara/crates/client/settlement_client/src/messages_to_l2_consumer.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            // We have no more messages in db.
+
+            if wait {
+                permit.as_mut().await; // Wait until we're notified.
+                permit.set(self.notify.notified());
+            }
+            wait = true;
+        }
```
Trantorian1, line 63: Why not just hoist the first iteration outside of the loop?

madara/crates/client/settlement_client/src/starknet/event.rs
```diff
@@ -1,258 +1,158 @@
...
+        Ok(Some((WatchBlockNEvent {
+            new,
+            previous,
+        }, (provider, interval, Some(new)))))
+    })
+    // only return events where the block_n changed
+    .try_filter(|&ev| async move {
+        ev.previous != Some(ev.new)
+    })
+}
```
Trantorian1, line 108: How does this work in a local block prod context, are we still polling a fixed interval?

madara/crates/client/settlement_client/src/starknet/test_utils.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Returns an error if deployment fails
+pub async fn deploy_contract(account: &StarknetAccount, sierra: &[u8]) -> Felt {
+    let contract_artifact: SierraClass = serde_json::from_slice(sierra).unwrap();
+    let flattened_class = contract_artifact.flatten().unwrap();
+
+    let (compiled_class_hash, _compiled_class) =
+        mp_class::FlattenedSierraClass::from(flattened_class.clone()).compile_to_casm().unwrap();
+
+    let result = account.declare_v2(Arc::new(flattened_class), compiled_class_hash).send().await.unwrap();
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
Trantorian1: Do we not have a better way to wait for contract deployment than to sleep for 5s? Using `getMessageStatus` in a loop with a polling interval of even just `subscribeTransactionStatus` when it becomes available #679 would be less error-prone.

madara/crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -98,6 +98,9 @@ fn default_pending_block_update_time() -> Option<Duration> {
 fn default_block_time() -> Duration {
     Duration::from_secs(30)
 }
+fn default_l1_messages_replay_max_duration() -> Duration {
+    Duration::from_secs(3 * 24 * 60 * 60)
```
Trantorian1, line 86: Why the choice of 3 days?

Trantorian1 (dismissed):
Some build issues related to the cairo artifacts. Other than that have left various comments and nits throughout the review.

Regarding PR size I don't mind reviewing large PRs, and this certainly was one, but it felt like this could have been scoped better. The code is very interesting and well implement as usual, but at times I was feeling like I was reviewing four different PRs: one for the event updates, one for the L1 messages changes, one for the block production update and another one for the new endpoints...

This ends up being quite a lot and it would make it easier to review these changes if they could be broken up into smaller, individual PRs. I get that it can be more convenient when developing these features to work off a single branch, but in my experience using multiple branches which are based off one another and just rebasing changes when needed is quite manageable as well.

Anyway, please let me know what you think about this, we can discuss this async if you want :)

build-artifacts/build.docker
cchudant: :(

madara/crates/client/block_production/src/executor/tests.rs
cchudant: Yes, sorry, this whole test suite is about testing duplicate nonce issues

madara/crates/client/block_production/src/executor/thread.rs
cchudant: The l1 handler transactions come from the madara types which are u64s so no, but i'll make that clearer i guess?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
cchudant: not yet but i figured it wasnt such a big deal to omit the tests for now
since they're admin rpcs

madara/crates/client/settlement_client/src/gas_price.rs
cchudant: haven't tried, this was to make the returned future 'static so that you can put it in a tokio spawn
but actually i think you're right i can just
let client_ = client.clone();
async move { gas_price_worker(&client) to make it 'static when calling that if i want to

madara/crates/client/settlement_client/src/lib.rs
cchudant: this is for future use; for get message status
do you think i should remove it? i wanted to include that here to make it clear what I think this trait should be
I think I'm gonna remove it for now yes

madara/crates/client/settlement_client/src/messages_to_l2_consumer.rs
cchudant: i didn't want to repeat myself
the alternative would be to hoist the db call in a separate closure, but that closure is an async closure and I won't be able to use `return` in it so I thought this was clearer.
I'm split, idk which one would be clearer

madara/crates/client/settlement_client/src/starknet/event.rs
cchudant: What do you mean local block prod context? You can't run a madara node that settles on itself. I mean you could I guess, but it would have to call itself on its own RPC api.

madara/crates/client/settlement_client/src/starknet/test_utils.rs
cchudant: yes i copied that code, this isnt mine
you're right this is slow and flaky
and probably explains why test setup takes sooo long. nice catch :) i was wondering why it was so slow

madara/crates/primitives/chain_config/src/chain_config.rs
cchudant: originally I wanted it to be less than 5 days, because a message can be cancelled after 5 days but my worker now supports even with replay durations >5days so any value would be fine I guess
it's very arbitrary yes I don't know what value to put here

madara/crates/client/block_production/src/executor/tests.rs
```diff
@@ -0,0 +1,302 @@
...
+    (tx, AdditionalTxInfo { declared_class })
+}
+
+fn make_l1_handler_tx(
+    backend: &MadaraBackend,
+    contract_address: Felt,
+    nonce: u64,
+    from_l1_address: Felt,
+    arg1: Felt,
+    arg2: Felt,
```
Mohiiit, line 39: this is because of the `test_account.cairo` setup?

madara/crates/client/block_production/src/handle.rs
```diff
@@ -0,0 +1,103 @@
...
+        executor_commands: mpsc::UnboundedSender<executor::ExecutorCommand>,
+        bypass_input: mpsc::Sender<ValidatedMempoolTx>,
+    ) -> Self {
+        Self {
+            executor_commands,
+            bypass_input: bypass_input.clone(),
+            tx_converter: TransactionValidator::new(
+                Arc::new(BypassInput(bypass_input)),
+                backend,
+                TransactionValidatorConfig::default().with_disable_validation(true),
```
Mohiiit, line 55: would there be a case when we would need validation post this step?

madara/crates/client/block_production/src/handle.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// Send a transaction through the bypass channel to bypass mempool and validation.
+    pub async fn send_tx_raw(&self, tx: ValidatedMempoolTx) -> Result<(), ExecutorCommandError> {
+        self.bypass_input.send(tx).await.map_err(|_| ExecutorCommandError::ChannelClosed)
+    }
+}
+
+// For convenience, we proxy the submit tx traits.
+
+#[async_trait]
```
Mohiiit, line 78: since we are using rust 1.85, maybe we can skip using async_trait as a dependency?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -383,6 +388,13 @@ impl BlockProductionTask {
 
     fn store_pending_block(&mut self) -> anyhow::Result<()> {
         if let TaskState::Executing(state) = self.current_state.as_mut().context("No current state")? {
+            for l1_nonce in &state.block.consumed_core_contract_nonces {
+                // This ensures we remove the nonces for rejected L1 to L2 message transactions. This avoids us from reprocessing them on restart.
+                self.backend
+                    .remove_pending_message_to_l2(*l1_nonce)
+                    .context("Removing pending message to l2 from database")?;
+            }
```
Mohiiit, line 401: on every pending tick we are removing the nonce from the hashset? so the values in hashset remains only when a pending tick starts, so it's just for keeping count for the reverted l1 txn? (not just, I get it that's imp)

madara/crates/client/block_production/src/lib.rs
```diff
@@ -749,7 +754,7 @@ pub(crate) mod tests {
             nonce,
             contract_class: flattened_class.into(),
             resource_bounds: ResourceBoundsMapping {
-                l1_gas: ResourceBounds { max_amount: 210000, max_price_per_unit: 10000 },
+                l1_gas: ResourceBounds { max_amount: 220000, max_price_per_unit: 10000 },
```
Mohiiit, line 759: curious about this change!

madara/crates/client/db/src/l1_db.rs
```diff
@@ -1,140 +1,117 @@
...
+        let pending_cf = self.db.get_column(Column::CoreContractNonceToPendingMsg);
+        self.db.get_cf(&pending_cf, core_contract_nonce.to_be_bytes())?;
+        let Some(res) = self.db.get_pinned_cf(&pending_cf, core_contract_nonce.to_be_bytes())? else { return Ok(None) };
+        Ok(Some(bincode::deserialize(&res)?))
     }
 
-    #[tracing::instrument(skip(self, nonce), fields(module = "L1DB"))]
-    pub fn set_l1_messaging_nonce(&self, nonce: Nonce) -> Result<(), DbError> {
-        let nonce_column = self.db.get_column(Column::L1MessagingNonce);
+    pub fn get_next_pending_message_to_l2(
```
Mohiiit, line 58: is this function any different from `get_pending_message_to_l2` with start_nonce could be what you are looking for, could be just start_nonce+1? what am I missing here

madara/crates/client/rpc/src/versions/admin/v0_1_0/api.rs
```diff
@@ -15,12 +18,42 @@ pub enum ServiceRequest {
...
+    #[method(name = "bypassAddInvokeTransaction")]
+    async fn bypass_add_invoke_transaction(
+        &self,
+        invoke_transaction: BroadcastedInvokeTxn,
+    ) -> RpcResult<AddInvokeTransactionResult>;
+
+    /// Force close a block.
+    /// Only works in block production mode.
+    #[method(name = "closeBlock")]
+    async fn close_block(&self) -> RpcResult<()>;
```
Mohiiit, line 56: 🙌

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -304,28 +292,72 @@ impl SettlementClientTrait for EthereumClient {
...
-    ) -> Result<Self::StreamType, SettlementClientError> {
-        let filter = self.l1_core_contract.event_filter::<LogMessageToL2>();
-        let event_stream = filter
-            .from_block(last_synced_event_block.block_number)
-            .to_block(BlockNumberOrTag::Finalized)
-            .watch()
+    /// Get cancellation status of an L1 to L2 message
+    ///
+    /// This function query the core contract to know if a L1->L2 exists in the contract. This is useful
+    /// for processing old L1 to L2 messages, because when a message is fully canceled we will find its old event
```
Mohiiit:
```suggestion
    /// for processing old L1 to L2 messages, because when a message is fully cancelled we will find its old event
```

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -1,353 +1,193 @@
...
-            )));
-        }
-        _ => {}
+async fn sync_inner(
+    settlement_client: Arc<dyn SettlementLayerProvider>,
+    backend: Arc<MadaraBackend>,
+    notify_consumer: Arc<Notify>,
+) -> Result<(), SettlementClientError> {
+    // Note: Reprocessing events.
+    // It's really important to make sure we don't mess up, we really want a strong guarantee we can't, in any circonstance, include an
```
Mohiiit:
```suggestion
    // It's really important to make sure we don't mess up, we really want a strong guarantee we can't, in any circumstance, include an
```

madara/crates/client/settlement_client/src/messaging/find_start_block.rs
```diff
@@ -0,0 +1,368 @@
...
+        loop {
+            let candidate = low.saturating_sub(step);
+            let candidate_timestamp = get_block_timestamp(candidate).await?;
+
+            low = candidate;
+            step *= 2; // Double step size
+
+            if candidate_timestamp <= target_timestamp || candidate == 0 {
+                break;
+            }
```
Mohiiit: value of low and step should be updated after the condition is checked right? if we are updating the value before checking, we might get an extra cycle of updation?

madara/crates/client/settlement_client/src/starknet/event.rs
```diff
@@ -1,258 +1,158 @@
...
-    processed_events: HashSet<Felt>,
+// Starknet event conversion
+impl TryFrom<EmittedEvent> for MessageToL2WithMetadata {
+    type Error = SettlementClientError;
 
-    /// The current future being polled (if any).
-    future: Option<FutureType>,
+    fn try_from(event: EmittedEvent) -> Result<Self, Self::Error> {
+        // https://github.com/keep-starknet-strange/piltover/blob/a7dc4141fd21300f6d7c23b87d496004a739f430/src/messaging/component.cairo#L86-L96
+        // keys: ['LogStateUpdate', message_hash, from_address, to_address]
```
Mohiiit:
```suggestion
        // keys: ['MessageSent', message_hash, from_address, to_address]
```

madara/crates/client/settlement_client/src/starknet/mod.rs
Mohiiit, line 1: do you think it would be a good idea to move important selector to a const?

Mohiiit (commented):
thanks for all the work, specifically those l1 txn nonce tests 🙇 

although please consider to have smaller PRs 😅

build-artifacts/build.docker
cchudant: fixed

madara/crates/client/settlement_client/src/gas_price.rs
cchudant: fixed

madara/crates/client/settlement_client/src/lib.rs
cchudant: fixed

madara/crates/client/settlement_client/src/lib.rs
cchudant: removed :)

madara/crates/client/settlement_client/src/starknet/test_utils.rs
cchudant: fixed :)

madara/crates/client/block_production/src/executor/tests.rs
cchudant: yes :)

madara/crates/client/block_production/src/handle.rs
cchudant: this is only for the bypass input, declare bootstrap txs wouldnt pass the validation so we disable it.
TransactionValidator is only used for its logic and error mapping regarding class compilation etc.

madara/crates/client/block_production/src/handle.rs
cchudant: I think we still need it for trait objects, dyn SubmitTransaction

madara/crates/client/block_production/src/lib.rs
cchudant: yes :) this is redundant as it is also done in backend `into_full_block_with_classes` but the backend operation won't see rejected l1handler txs and we would like to remove them from db too

madara/crates/client/block_production/src/lib.rs
cchudant: I changed the cairo contract a bit and had to bump it :)

madara/crates/client/db/src/l1_db.rs
cchudant: We use rocksdb iterators so that `get_next_pending_message_to_l2(start_nonce = 10)` will return the next event that has nonce >= 10

madara/crates/client/settlement_client/src/messaging.rs
cchudant: fixed :)

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: fixed :)

madara/crates/client/settlement_client/src/starknet/event.rs
cchudant: fixed :)

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -51,6 +56,23 @@ enum ExecutorThreadState {
...
 }
 
+impl ExecutorThreadState {
+    fn consumed_l1_to_l2_nonces(&mut self) -> &mut HashSet<u64> {
+        match self {
+            ExecutorThreadState::Executing(s) => &mut s.consumed_l1_to_l2_nonces,
+            ExecutorThreadState::NewBlock(s) => &mut s.consumed_l1_to_l2_nonces,
+        }
+    }
+    fn layered_state_adaptor(&mut self) -> &mut LayeredStateAdaptor {
```
jbcaron:
```suggestion
    fn layered_state_adapter(&mut self) -> &mut LayeredStateAdaptor {
```

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -51,6 +56,23 @@ enum ExecutorThreadState {
...
+        match self {
+            ExecutorThreadState::Executing(s) => {
+                &mut s.executor.block_state.as_mut().expect("State already taken").state
+            }
+            ExecutorThreadState::NewBlock(s) => &mut s.state_adaptor,
+        }
+    }
```
jbcaron: Did you think the function name should be more explicit about the fact that it returns a mutable reference?

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -184,9 +206,16 @@ impl ExecutorThread {
 
         let state_diff = cached_state.to_state_diff().context("Cannot make state diff")?.state_maps;
         let mut cached_adaptor = cached_state.state;
```
jbcaron:
```suggestion
        let mut cached_adapter = cached_state.state;
```

madara/crates/client/block_production/src/batcher.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            let l1_txs_stream = self.l1_message_stream.as_mut().map(|res| {
+                Ok(res?
+                    .into_blockifier(chain_id, sn_version)
+                    .map(|(btx, declared_class)| (btx, AdditionalTxInfo { declared_class }))?)
+            });
+
+            // Note: this is not hoisted out of the loop, because we don't want to keep the lock around when waiting on the output channel reserve().
+            let mempool_txs_stream = stream::unfold(self.mempool.clone(), |mempool| async move {
+                let consumer = mempool.clone().get_consumer_wait_for_ready_tx().await;
```
jbcaron, line 85:
```suggestion
            let mempool_txs_stream = stream::unfold(Arc::clone(self.mempool), |mempool| async move {
                let consumer = Arc::clone(mempool).get_consumer_wait_for_ready_tx().await;
```

madara/crates/client/db/src/l1_db.rs
```diff
@@ -1,140 +1,117 @@
...
-        let messaging_column = self.db.get_column(Column::L1Messaging);
+    /// If the message is already pending, this will overwrite it.
+    pub fn add_pending_message_to_l2(&self, msg: L1HandlerTransactionWithFee) -> Result<(), MadaraStorageError> {
+        let pending_cf = self.db.get_column(Column::CoreContractNonceToPendingMsg);
         self.db.put_cf_opt(
-            &messaging_column,
-            LAST_SYNCED_L1_EVENT_BLOCK,
-            bincode::serialize(&last_synced_event_block)?,
+            &pending_cf,
+            msg.tx.nonce.to_be_bytes(),
```
jbcaron, line 34: It’s u64, right?

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -94,47 +98,51 @@ impl LayeredStateAdaptor {
...
 }
 
 impl StateReader for LayeredStateAdaptor {
     fn get_storage_at(&self, contract_address: ContractAddress, key: StorageKey) -> StateResult<Felt> {
-        for s in &self.cached_states_by_block_n {
-            if let Some(el) = s.state_diff.storage.get(&(contract_address, key)) {
-                return Ok(*el);
-            }
+        if let Some(el) =
+            self.cached_states_by_block_n.iter().find_map(|s| s.state_diff.storage.get(&(contract_address, key)))
```
jbcaron, line 113: Shouldn’t you reverse the iterator to find the most recent change?

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -94,47 +98,51 @@ impl LayeredStateAdaptor {
...
+            return Ok(*el);
         }
         self.inner.get_storage_at(contract_address, key)
     }
     fn get_nonce_at(&self, contract_address: ContractAddress) -> StateResult<Nonce> {
-        for s in &self.cached_states_by_block_n {
-            if let Some(el) = s.state_diff.nonces.get(&contract_address) {
-                return Ok(*el);
-            }
+        if let Some(el) = self.cached_states_by_block_n.iter().find_map(|s| s.state_diff.nonces.get(&contract_address))
```
jbcaron, line 120: ditto

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -94,47 +98,51 @@ impl LayeredStateAdaptor {
...
         }
         self.inner.get_nonce_at(contract_address)
     }
     fn get_class_hash_at(&self, contract_address: ContractAddress) -> StateResult<ClassHash> {
-        for s in &self.cached_states_by_block_n {
-            if let Some(el) = s.state_diff.class_hashes.get(&contract_address) {
-                return Ok(*el);
-            }
+        if let Some(el) =
+            self.cached_states_by_block_n.iter().find_map(|s| s.state_diff.class_hashes.get(&contract_address))
```
jbcaron, line 128: ditto

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -94,47 +98,51 @@ impl LayeredStateAdaptor {
...
         }
         self.inner.get_class_hash_at(contract_address)
     }
     fn get_compiled_class(&self, class_hash: ClassHash) -> StateResult<RunnableCompiledClass> {
-        for s in &self.cached_states_by_block_n {
-            if let Some(el) = s.classes.get(&class_hash) {
-                return <ApiContractClass as TryInto<RunnableCompiledClass>>::try_into(el.clone())
-                    .map_err(StateError::ProgramError);
-            }
+        if let Some(el) = self.cached_states_by_block_n.iter().find_map(|s| s.classes.get(&class_hash)) {
```
jbcaron, line 135: ditto

madara/crates/client/exec/src/layered_state_adaptor.rs
```diff
@@ -94,47 +98,51 @@ impl LayeredStateAdaptor {
...
         }
         self.inner.get_compiled_class(class_hash)
     }
     fn get_compiled_class_hash(&self, class_hash: ClassHash) -> StateResult<CompiledClassHash> {
-        for s in &self.cached_states_by_block_n {
-            if let Some(el) = s.state_diff.compiled_class_hashes.get(&class_hash) {
-                return Ok(*el);
-            }
+        if let Some(el) =
+            self.cached_states_by_block_n.iter().find_map(|s| s.state_diff.compiled_class_hashes.get(&class_hash))
```
jbcaron, line 143: ditto

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -737,10 +737,10 @@ impl MempoolInner {
...
         self.nonce_cache_inner.insert(tx_mempool.contract_address(), tx_mempool.nonce_next);
         self.limiter.mark_removed(&TransactionCheckedLimits::limits_for(&tx_mempool));
-        debug_assert!(
-            self.tx_received.remove(&tx_mempool.tx_hash()),
-            "Tried to remove a ready transaction which had not already been marked as received"
-        );
+        // debug_assert!(
+        self.tx_received.remove(&tx_mempool.tx_hash());
+        // "Tried to remove a ready transaction which had not already been marked as received"
+        // );
```
jbcaron: clean up

madara/crates/client/mempool/src/lib.rs
```diff
@@ -216,10 +185,11 @@ impl Mempool {
     }
 
     async fn accept_tx(&self, tx: ValidatedMempoolTx) -> Result<(), MempoolError> {
+        if tx.tx.as_l1_handler().is_some() {
```
jbcaron: Use `tx_type()` instead of `as_l1_handler()`

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -251,31 +235,18 @@ impl SettlementClientTrait for EthereumClient {
...
-        let payload_vec = event.payload.iter().try_fold(Vec::with_capacity(event.payload.len()), |mut acc, felt| {
-            let u256 = felt_to_u256(*felt).map_err(|e| -> SettlementClientError {
-                EthereumClientError::Conversion(format!("Failed to convert payload element to U256: {}", e)).into()
-            })?;
-            acc.push(u256);
-            Ok::<_, SettlementClientError>(acc)
-        })?;
-
-        let from_address_start_index = event.from.to_bytes_be().as_slice().len().saturating_sub(20);
+    fn get_messaging_hash(&self, event: &L1HandlerTransactionWithFee) -> Result<Vec<u8>, SettlementClientError> {
```
jbcaron:
```suggestion
    fn calculate_messaging_hash(&self, event: &L1HandlerTransactionWithFee) -> Result<Vec<u8>, SettlementClientError> {
```

madara/crates/client/settlement_client/src/messaging/find_start_block.rs
```diff
@@ -0,0 +1,382 @@
...
+        if candidate_timestamp < target_timestamp || candidate == 0 {
+            break;
+        }
+        step *= 2; // Double step size
+        high = candidate;
+    }
+    // Special case when the latest block is very old
+    if step == 1 {
+        return Ok(latest_block_n);
+    }
```
jbcaron, line 69: Why not make an assumption about an average block time of 12 seconds instead of doing the exponential search at the beginning?

madara/crates/client/settlement_client/src/starknet/event.rs
```diff
@@ -1,258 +1,158 @@
...
+    })
+    // flatten: convert stream of `Vec<EmittedEvent>` chunks into individual `EmittedEvent`s.
+    .try_flatten()
+}
+
+pub struct WatchEventFilter {
+    pub address: Option<Felt>,
+    pub keys: Option<Vec<Vec<Felt>>>,
+}
+pub fn watch_events(
```
jbcaron, line 133: You can keep the same filter and use `eth_getFilterLogs` to get updates

madara/crates/client/settlement_client/src/gas_price.rs
jbcaron, line 1: This client was refactored in #668

madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -201,6 +207,14 @@ impl TransactionValidator {
     ) -> Result<(), SubmitTransactionError> {
         let tx_hash = tx.tx_hash().to_felt();
 
+        if tx.tx_type() == TransactionType::L1Handler {
```
jbcaron, line 211: Should we also remove the tx declare V0?

jbcaron (changes requested):

madara/crates/client/block_production/src/executor/thread.rs
cchudant: fixed :)

madara/crates/client/block_production/src/executor/thread.rs
cchudant: fixed :)

madara/crates/client/block_production/src/executor/thread.rs
cchudant: cached_adaptor

madara/crates/client/block_production/src/batcher.rs
cchudant: I think we should stop doing that, it's clearer sure but it's such a bother..

madara/crates/client/db/src/l1_db.rs
cchudant: yes :) I think it's fine to restrict ourselves to u64 core contract nonces

madara/crates/client/exec/src/layered_state_adaptor.rs
cchudant: new blocks are pushed on the front

madara/crates/client/exec/src/layered_state_adaptor.rs
cchudant: ditto

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/lib.rs
cchudant: fixed :)

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: fixed :)

madara/crates/client/settlement_client/src/messaging/find_start_block.rs
cchudant: I want it to work well even on L3s that have different (weird?) assumptions

madara/crates/client/settlement_client/src/gas_price.rs
cchudant: cool :)

madara/crates/client/settlement_client/src/starknet/event.rs
cchudant: this is starknet though

madara/crates/client/submit_tx/src/validation.rs
cchudant: fixed :)

jbcaron (approved):

Mohiiit (approved):

madara/crates/client/block_production/src/executor/thread.rs
cchudant: oh sorry i meant it's fixed

PR #679, opened 2025-06-04
feat(`starknet_subscribeTransactionStatus`)

Trantorian1:
## Pull Request type

- Feature

Madara currently does not support the `starknet_subscribeTransactionStatus` ws endpoint.

## What is the new behavior?

- Added support for `starknet_subscribeTransactionStatus`.
- Added update channels to follow the state of pending transactions and the latest confirmed block on L1.

## Implementation details

Supported transaction statuses are:

### `Received`

Transaction has been received in the inner mempool. **This status is not supported when forwarding to a remote gateway.**

### `AcceptedOnL2`

Transaction has been inserted into a _closed_ block. **Pending block inclusion does not count towards** `AcceptedOnL2`.

### `AcceptedOnL1`

Transaction has been finalized on L1.

> [!WARNING]
> Transaction status `Rejected` is not supported as of now.

The set of legal state transitions are

```                                                                            
            ┌────┐                                                                  
         ┌─►│None├─────────────────────────────────────────────────────────┐        
         │  └────┘                                                         │        
         │                                                                 │        
         │                                                                 │        
┌─────┐  │  ┌────────────┐                             ┌────────────────┐  └─►┌───┐ 
│START├──┼─►│WaitReceived│┬────────────────────────┬──►│WaitAcceptedOnL1│────►│END│ 
└─────┘  │  └────────────┘│                        │   └────────────────┘     └───┘ 
         │                │                        ▲                                
         │                └───►┌────────────────┐  │                                
         └────────────────────►│WaitAcceptedOnL2│──┘                                
                               └────────────────┘                                   
```

No.

### DIFF
--- a/madara/crates/client/db/src/watch.rs
+++ b/madara/crates/client/db/src/watch.rs
@@ -5,6 +5,7 @@ use tokio::sync::{broadcast, watch};
 
 pub type ClosedBlocksReceiver = broadcast::Receiver<Arc<MadaraBlockInfo>>;
 pub type PendingBlockReceiver = watch::Receiver<Arc<MadaraPendingBlockInfo>>;
+pub type LastConfirmedBlockReceived = watch::Receiver<Option<u64>>;
 
 fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<MadaraPendingBlockInfo> {
     let Some(parent_block) = parent_block else {
@@ -27,6 +28,7 @@ fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<Madara
 pub(crate) struct BlockWatch {
     closed_blocks: broadcast::Sender<Arc<MadaraBlockInfo>>,
     pending_block: watch::Sender<Arc<MadaraPendingBlockInfo>>,
+    last_confirmed_block: watch::Sender<Option<u64>>,
 }
 
 impl BlockWatch {
@@ -34,17 +36,24 @@ impl BlockWatch {
         Self {
             closed_blocks: broadcast::channel(100).0,
             pending_block: watch::channel(make_fake_pending_block(None)).0,
+            last_confirmed_block: watch::channel(None).0,
         }
     }
 
     pub fn init_initial_values(&self, db: &MadaraBackend) -> Result<(), MadaraStorageError> {
         let block = db.get_pending_block_info_from_db()?;
-        self.update_pending(block.into());
+        let latest_block = db.get_l1_last_confirmed_block()?;
+        self.pending_block.send_replace(block.into());
+        self.last_confirmed_block.send_replace(latest_block);
         Ok(())
     }
 
     pub fn update_pending(&self, block: Arc<MadaraPendingBlockInfo>) {
         self.pending_block.send_replace(block);
+    }
+
+    pub fn update_last_confirmed_block(&self, latest_block: u64) {
+        self.last_confirmed_block.send_replace(Some(latest_block));
     }
 
     pub fn clear_pending(&self, parent_block: Option<&MadaraBlockInfo>) {
@@ -62,6 +71,9 @@ impl BlockWatch {
     pub fn subscribe_pending_block(&self) -> PendingBlockReceiver {
         self.pending_block.subscribe()
     }
+    pub fn subscribe_last_confirmed_block(&self) -> LastConfirmedBlockReceived {
+        self.last_confirmed_block.subscribe()
+    }
     pub fn latest_pending_block(&self) -> Arc<MadaraPendingBlockInfo> {
         self.pending_block.borrow().clone()
     }
@@ -75,6 +87,10 @@ impl MadaraBackend {
     #[tracing::instrument(skip(self), fields(module = "MadaraBackendWatch"))]
     pub fn subscribe_pending_block(&self) -> PendingBlockReceiver {
         self.watch_blocks.subscribe_pending_block()
+    }
+    #[tracing::instrument(skip(self), fields(module = "MadaraBackendWatch"))]
+    pub fn subscribe_last_confirmed_block(&self) -> LastConfirmedBlockReceived {
+        self.watch_blocks.subscribe_last_confirmed_block()
     }
     #[tracing::instrument(skip(self), fields(module = "MadaraBackendWatch"))]
     pub fn latest_pending_block(&self) -> Arc<MadaraPendingBlockInfo> {
--- a/madara/crates/client/db/src/watch.rs
+++ b/madara/crates/client/db/src/watch.rs
@@ -1,11 +1,11 @@
 use crate::{MadaraBackend, MadaraStorageError};
 use mp_block::{header::PendingHeader, MadaraBlockInfo, MadaraPendingBlockInfo};
 use std::sync::Arc;
-use tokio::sync::{broadcast, watch};
 
-pub type ClosedBlocksReceiver = broadcast::Receiver<Arc<MadaraBlockInfo>>;
-pub type PendingBlockReceiver = watch::Receiver<Arc<MadaraPendingBlockInfo>>;
-pub type LastConfirmedBlockReceived = watch::Receiver<Option<u64>>;
+pub type ClosedBlocksReceiver = tokio::sync::broadcast::Receiver<Arc<MadaraBlockInfo>>;
+pub type PendingBlockReceiver = tokio::sync::watch::Receiver<Arc<MadaraPendingBlockInfo>>;
+pub type PendingTxsReceiver = tokio::sync::broadcast::Receiver<mp_block::TransactionWithReceipt>;
+pub type LastConfirmedBlockReceived = tokio::sync::watch::Receiver<Option<u64>>;
 
 fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<MadaraPendingBlockInfo> {
     let Some(parent_block) = parent_block else {
@@ -26,17 +26,19 @@ fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<Madara
 }
 
 pub(crate) struct BlockWatch {
-    closed_blocks: broadcast::Sender<Arc<MadaraBlockInfo>>,
-    pending_block: watch::Sender<Arc<MadaraPendingBlockInfo>>,
-    last_confirmed_block: watch::Sender<Option<u64>>,
+    closed_blocks: tokio::sync::broadcast::Sender<Arc<MadaraBlockInfo>>,
+    pending_block: tokio::sync::watch::Sender<Arc<MadaraPendingBlockInfo>>,
+    pending_txs: tokio::sync::broadcast::Sender<mp_block::TransactionWithReceipt>,
+    last_confirmed_block: tokio::sync::watch::Sender<Option<u64>>,
 }
 
 impl BlockWatch {
     pub fn new() -> Self {
         Self {
-            closed_blocks: broadcast::channel(100).0,
-            pending_block: watch::channel(make_fake_pending_block(None)).0,
-            last_confirmed_block: watch::channel(None).0,
+            closed_blocks: tokio::sync::broadcast::channel(100).0,
+            pending_block: tokio::sync::watch::channel(make_fake_pending_block(None)).0,
+            pending_txs: tokio::sync::broadcast::channel(100).0,
+            last_confirmed_block: tokio::sync::watch::channel(None).0,
         }
     }
 
@@ -60,6 +62,10 @@ impl BlockWatch {
         self.update_pending(make_fake_pending_block(parent_block));
     }
 
+    pub fn on_new_pending_tx(&self, tx: mp_block::TransactionWithReceipt) {
+        let _no_listener_error = self.pending_txs.send(tx);
+    }
+
     pub fn on_new_block(&self, block: Arc<MadaraBlockInfo>) {
         let _no_listener_error = self.closed_blocks.send(Arc::clone(&block));
         self.update_pending(make_fake_pending_block(Some(&block)));
@@ -67,6 +73,9 @@ impl BlockWatch {
 
     pub fn subscribe_closed_blocks(&self) -> ClosedBlocksReceiver {
         self.closed_blocks.subscribe()
+    }
+    pub fn subscribe_pending_txs(&self) -> PendingTxsReceiver {
+        self.pending_txs.subscribe()
     }
     pub fn subscribe_pending_block(&self) -> PendingBlockReceiver {
         self.pending_block.subscribe()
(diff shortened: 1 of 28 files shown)


madara/crates/client/gateway/client/src/submit_tx.rs
```diff
@@ -119,6 +119,16 @@ impl SubmitTransaction for GatewayProvider {
             .map_err(map_gateway_error)
             .map(|res| AddInvokeTransactionResult { transaction_hash: res.transaction_hash })
     }
+
+    async fn received_transaction(&self, _hash: starknet_types_core::felt::Felt) -> Option<bool> {
```
jbcaron, line 123: Can you add a comment or a TODO?

madara/crates/client/gateway/client/src/submit_tx.rs
```diff
@@ -129,4 +139,14 @@ impl mc_submit_tx::SubmitValidatedTransaction for GatewayProvider {
     ) -> Result<(), SubmitTransactionError> {
         self.add_validated_transaction(tx).await.map_err(map_gateway_error)
     }
+
+    async fn received_transaction(&self, _hash: starknet_types_core::felt::Felt) -> Option<bool> {
+        None
+    }
+
+    async fn subscribe_new_transactions(
```
jbcaron, line 156: ditto

madara/crates/client/gateway/client/src/submit_tx.rs
```diff
@@ -129,4 +139,14 @@ impl mc_submit_tx::SubmitValidatedTransaction for GatewayProvider {
     ) -> Result<(), SubmitTransactionError> {
         self.add_validated_transaction(tx).await.map_err(map_gateway_error)
     }
+
+    async fn received_transaction(&self, _hash: starknet_types_core::felt::Felt) -> Option<bool> {
```
jbcaron, line 148: ditto

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -180,6 +178,9 @@ pub struct MempoolInner {
     /// [Mempool]: super::Mempool
     limiter: MempoolLimiter,
 
+    /// Keeps track of transaction which are currently in the inner mempool by their hash
+    tx_received: HashSet<TransactionHash>,
```
jbcaron, line 182: Since this structure is accessible by clients, could it be outside of `mempoolInner`, in a `DashSet`?

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -568,6 +576,10 @@ impl MempoolInner {
             let limits = TransactionCheckedLimits::limits_for(nonce_mapping_entry.get());
             if self.limiter.tx_age_exceeded(&limits) {
                 let mempool_tx = nonce_mapping_entry.remove();
+                assert!(
```
jbcaron: Why not just a debug assert?

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -625,6 +637,11 @@ impl MempoolInner {
                 // tx_intent_queue_pending_by_timestamp
 
                 let mempool_tx = nonce_mapping_entry.remove(); // *- snip -*
+                assert!(
```
jbcaron: ditto

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -725,8 +742,12 @@ impl MempoolInner {
 
         #[cfg(any(test, feature = "testing"))]
         self.nonce_cache_inner.insert(tx_mempool.contract_address(), tx_mempool.nonce_next);
-
         self.limiter.mark_removed(&TransactionCheckedLimits::limits_for(&tx_mempool));
+        assert!(
```
jbcaron: ditto

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
```diff
@@ -0,0 +1,795 @@
...
+///
+/// We do not currently support the **Rejected** transaction status.
+///
+/// Note that it is possible to call this method on a transaction which has not yet been received by
+/// the node and this endpoint will send an update as soon as the tx is received.
+///
+/// ## DOS mitigation
+///
+/// To avoid a malicious attacker keeping connections open indefinitely on an nonexistent
+/// transaction hash, this endpoint will terminate the connection after a global timeout period.
```
jbcaron: 👍

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
jbcaron, line 1: This is a clean and well-structured implementation of a state machine for tracking transaction status.
As a potential enhancement, the state transitions could be made verifiable at compile time using the __phantom type state pattern__.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
```diff
@@ -0,0 +1,795 @@
...
+                        Ok(Self::WaitAcceptedOnL1(StateTransitionAcceptedOnL1 { common, block_number, channel }))
+                    }
+                }
+            }
+        } else {
+            // Local mempool is the only AddTransactionProvider which allows us to inspect the state
+            // of received transactions. For other providers (such as when forwarding to a remote
+            // gateway), we default to assuming that the transaction has been received and wait for
+            // it to be accepted on L2.
+            match common.starknet.add_transaction_provider.subscribe_new_transactions().await {
```
jbcaron: I think this subscription should happen before `find_tx_hash_block_info`

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
```diff
@@ -0,0 +1,795 @@
...
+    type TransitionTo;
+
+    async fn transition(
+        self,
+    ) -> Result<StateTransitionResult<Self, Self::TransitionTo>, crate::errors::StarknetWsApiError>;
+}
+impl<'a> StateTransition for StateTransitionReceived<'a> {
+    type TransitionTo = TransitionMatrixReceived<'a>;
+
+    async fn transition(
```
jbcaron: This should be possible to simplify if you apply my previous suggestion.

madara/crates/client/mempool/src/inner/mod.rs
Trantorian1: I don't like the idea of having this outside of the inner mempool since this makes it harder to keep the set up-to-date with tx removals and insertions. For now I would be in favor of keeping it there and having the inner mempool update it as necessary just because this seems like a less error-prone mechanism as compared to having the outer mempool and mempool consumer handle that. Curious to hear your thoughts on this though.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
Trantorian1: Thanks! I'm generally weary of generic-based state machines as I find they can be quite a pain to work with and especially extend. Linking here the [article you sent me](https://zerotomastery.io/blog/rust-typestate-patterns/) for future reference :)

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
Trantorian1: Indeed, this has been updated!

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_transaction_status.rs
Trantorian1: Well, we no longer query the db every 100 received transactions but not much else could be removed because we still have to handle the cases where the channel would lag, causing the stream to terminate, in which case we need to check against db to make sure we haven't missed any updates.

Trantorian1: We also need to be able to exit the stream if there are not transactions in the mempool in order to transition to `WaitAcceptedOnL2`.

madara/crates/client/db/src/watch.rs
```diff
@@ -25,32 +26,46 @@ fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<Madara
...
 }
 
 impl BlockWatch {
     pub fn new() -> Self {
         Self {
-            closed_blocks: broadcast::channel(100).0,
-            pending_block: watch::channel(make_fake_pending_block(None)).0,
+            closed_blocks: tokio::sync::broadcast::channel(100).0,
+            pending_block: tokio::sync::watch::channel(make_fake_pending_block(None)).0,
+            pending_txs: tokio::sync::broadcast::channel(100).0,
```
Mohiiit, line 40: is the limit of 100 as per spec?

madara/crates/client/gateway/client/src/submit_tx.rs
```diff
@@ -119,6 +119,21 @@ impl SubmitTransaction for GatewayProvider {
             .map_err(map_gateway_error)
             .map(|res| AddInvokeTransactionResult { transaction_hash: res.transaction_hash })
     }
+
+    async fn received_transaction(&self, _hash: starknet_types_core::felt::Felt) -> Option<bool> {
+        // The gateway cannot inform us about the status of transactions it has received since this
+        // is forwarded to a remote node which does not expose any endpoint to query this state. By
+        // default, all transactions which pass through the gateway will be automatically considered
+        // as received.
```
Mohiiit, line 127: this is for the case when madara is running as full-node right? because in case the madara is running as sequencer and if the gateway is enabled, it should be able to respond right?

madara/crates/client/gateway/client/src/submit_tx.rs
```diff
@@ -119,6 +119,21 @@ impl SubmitTransaction for GatewayProvider {
...
+        None
+    }
+
+    async fn subscribe_new_transactions(
+        &self,
+    ) -> Option<tokio::sync::broadcast::Receiver<starknet_types_core::felt::Felt>> {
+        // We cannot subscribe to new transactions from the gateway for the same reasons as above
```
Mohiiit, line 134: same question as above

madara/crates/client/mempool/src/inner/property_testing.rs
```diff
@@ -68,6 +68,11 @@ pub enum TxTy {
 
 impl TxTy {
     fn tx(self, contract_address: Felt) -> blockifier::transaction::transaction_execution::Transaction {
+        static HASH: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+
+        let ordering = std::sync::atomic::Ordering::AcqRel;
```
Mohiiit, line 73: just curious about the ordering being used here, any specific reason to use AcqRel?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -151,6 +163,7 @@ impl Mempool {
             backend,
             inner: MempoolInnerWithNotify::new(config.limits.clone()),
             metrics: MempoolMetrics::register(),
+            tx_sender: tokio::sync::broadcast::channel(100).0,
```
Mohiiit, line 166: same question regarding the buffer length here!

madara/crates/client/mempool/src/lib.rs
```diff
@@ -367,6 +380,11 @@ pub(crate) mod tests {
 
     #[rstest::fixture]
     pub fn tx_account_v0_valid(#[default(CONTRACT_ADDRESS)] contract_address: Felt) -> ValidatedMempoolTx {
+        static HASH: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
```
Mohiiit, line 383: we can also make this HASH const tho? even tx_hash?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1164,9 +1182,8 @@ pub(crate) mod tests {
     ///
     /// [ready]: inner::TransactionIntentReady;
     /// [pending]: inner::TransactionInentPending;
-    #[rstest::rstest]
-    #[timeout(Duration::from_millis(1_000))]
```
Mohiiit, line 1168: hmm we removed timeout? just curious why

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_transaction_status.rs
Mohiiit, line 1: nice test structure!

madara/crates/client/db/src/watch.rs
Trantorian1: Nope, this is just a limit we put into place as number which seems reasonable. The issue is that `tokio` channels won't drop sent values until they have been seen by all receivers, so it would be possible to DOS the node by keeping a receiver open which either does not poll the channel or polls it slower than values are sent to it.

To avoid that scenarios we:

1. Do not use unbounded channels.
2. Use bounded channels with a rather small capacity, assuming that we will never be lagging more than 100 items at once. If we do lag, we refer to db checks to avoid loosing any information.

Trantorian1: Generally speaking those channels are not meant to be a source of absolute truth and it is possible for them to loose information. We mostly just use them as an optimization to avoid having to do constant db calls (which is quite a bit slower).

madara/crates/client/gateway/client/src/submit_tx.rs
Trantorian1: Yep, that is the case!

Trantorian1: The way things work currently Madara abstracts write transaction behind the `TransactionProvider` trait. In full node mode the gateway acts as its implementor, forwarding write transactions to a remote endpoint. In sequencer mode the mempool acts as its implementor, in which case we can reasons about the received state locally.

Trantorian1: So in this case the gateway is referring to the gateway as `TransactionProvider`, not as a capability.

madara/crates/client/mempool/src/inner/property_testing.rs
Trantorian1: Using `std::sync::atomic::Ordering::AcqRel` here since we are applying a `fetch_add` to this `AtomicU64`. When using `AcqRel` on an atomic operation which performs both a load _and_ a store (such as `fetch_add`, `fetch_sub`, `fetch_xor`, ...) it will apply `Acquire` ordering on the load and `Release` ordering on the store. This and `SeqCst` ordering are the only way to guarantee proper ordering on _both_ the load _and_ store parts of the operation. 

If we were using only `Acquire` ordering then the store would be `Relaxed` and could be observed differently by different threads. Similarly, using `Release` ordering would mean that the load would be `Relaxed` and could be observed differently by different threads.

This is not much of a problem since we are using this in tests but something good to keep in mind :) Also we are using `cargo nextest` which does use a process-per test approach and I don't know how that could impact the sharing of `static` variables.

madara/crates/client/mempool/src/lib.rs
Trantorian1: Same as before, we are trying to avoid DOS where a slow listener could keep the channel full at all times. We are really just using those channels as a shortcut to avoid querying the DB all the time but if they fail (ie: they [lag](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/enum.RecvError.html#variant.Lagged)) then we default to slower db calls.

madara/crates/client/mempool/src/lib.rs
Trantorian1: Nope! The reason why this is here is that we want `tx_hash` to be incrementally different each time this fixture is called. This is to avoid hash collisions when inserting into a block. Secondly, moving this to a `const` will produce a compiler warning about using interior mutability on a `const` item so we use `static` instead. The only difference is that `static` will not drop its value at the end of execution but this fine anyway since we are using a struct without any dynamic allocation.

madara/crates/client/mempool/src/lib.rs
Trantorian1: Can add those back, I think I removed those by accident. Generally though timeouts can be a bad idea as they can lead to flaky tests depending on the power of the maching running it. It have been guilty of using them quite a bit in the past but they have been biting at me quite a bit in this PR because of this :sweat_smile:. Curious to hear your opinion on this.

madara/crates/client/rpc/src/versions/user/v0_7_1/methods/read/get_transaction_status.rs
Trantorian1: Thanks :) Eventually it would be better to move this to a common module but for now `Ctrl-C / Ctrl-V` was easier :sweat_smile:

madara/crates/client/mempool/src/inner/property_testing.rs
Trantorian1: Hello Ser, a quick correction: the ordering on an atomic fetch-add doesn't have any impact on how it is viewed by the rest of the program, mb!

Trantorian1: The reason for that is that the fetch-add really isn't a fetch+add but a single atomic operation so it can't be compartmentalized into load and store operation. Using even `Relaxed` ordering here wouldn't change anything.

madara/crates/client/db/src/watch.rs
jbcaron: to be consistent, I think the buffer value for closed blocks should be much smaller than for pending transactions.

jbcaron (approved):

madara/crates/client/mempool/src/inner/property_testing.rs
Mohiiit: thanks for digging deeper into this!

madara/crates/client/mempool/src/lib.rs
Mohiiit: thanks, I missed that point altogether!

madara/crates/client/mempool/src/lib.rs
Mohiiit: I am not sure myself, I was just worried if the test would be impacted by removing the timeout, if the test is testing a time-sensitive operation or something!

Mohiiit (approved):

PR #683, opened 2025-06-10
feat(`starknet_subscribePendingTransactions`)

Trantorian1:
## Pull Request type

- Feature

Madara currently does not support the `starknet_subscribePendingTransactions` ws endpoint.

## What is the new behavior?

This pr adds support for `starknet_subscribePendingTransactions`.

No

### DIFF
--- a/madara/crates/client/db/src/watch.rs
+++ b/madara/crates/client/db/src/watch.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 pub type ClosedBlocksReceiver = tokio::sync::broadcast::Receiver<Arc<MadaraBlockInfo>>;
 pub type PendingBlockReceiver = tokio::sync::watch::Receiver<Arc<MadaraPendingBlockInfo>>;
 pub type PendingTxsReceiver = tokio::sync::broadcast::Receiver<mp_block::TransactionWithReceipt>;
-pub type LastConfirmedBlockReceived = tokio::sync::watch::Receiver<Option<u64>>;
+pub type LastBlockOnL1Receiver = tokio::sync::watch::Receiver<Option<u64>>;
 
 fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<MadaraPendingBlockInfo> {
     let Some(parent_block) = parent_block else {
@@ -29,7 +29,7 @@ pub(crate) struct BlockWatch {
     closed_blocks: tokio::sync::broadcast::Sender<Arc<MadaraBlockInfo>>,
     pending_block: tokio::sync::watch::Sender<Arc<MadaraPendingBlockInfo>>,
     pending_txs: tokio::sync::broadcast::Sender<mp_block::TransactionWithReceipt>,
-    last_confirmed_block: tokio::sync::watch::Sender<Option<u64>>,
+    last_block_on_l1: tokio::sync::watch::Sender<Option<u64>>,
 }
 
 impl BlockWatch {
@@ -38,7 +38,7 @@ impl BlockWatch {
             closed_blocks: tokio::sync::broadcast::channel(100).0,
             pending_block: tokio::sync::watch::channel(make_fake_pending_block(None)).0,
             pending_txs: tokio::sync::broadcast::channel(100).0,
-            last_confirmed_block: tokio::sync::watch::channel(None).0,
+            last_block_on_l1: tokio::sync::watch::channel(None).0,
         }
     }
 
@@ -46,7 +46,7 @@ impl BlockWatch {
         let block = db.get_pending_block_info_from_db()?;
         let latest_block = db.get_l1_last_confirmed_block()?;
         self.pending_block.send_replace(block.into());
-        self.last_confirmed_block.send_replace(latest_block);
+        self.last_block_on_l1.send_replace(latest_block);
         Ok(())
     }
 
@@ -54,8 +54,8 @@ impl BlockWatch {
         self.pending_block.send_replace(block);
     }
 
-    pub fn update_last_confirmed_block(&self, latest_block: u64) {
-        self.last_confirmed_block.send_replace(Some(latest_block));
+    pub fn update_last_block_on_l1(&self, latest_block: u64) {
+        self.last_block_on_l1.send_replace(Some(latest_block));
     }
 
     pub fn clear_pending(&self, parent_block: Option<&MadaraBlockInfo>) {
@@ -80,8 +80,8 @@ impl BlockWatch {
     pub fn subscribe_pending_block(&self) -> PendingBlockReceiver {
         self.pending_block.subscribe()
     }
-    pub fn subscribe_last_confirmed_block(&self) -> LastConfirmedBlockReceived {
-        self.last_confirmed_block.subscribe()
+    pub fn subscribe_last_block_on_l1(&self) -> LastBlockOnL1Receiver {
+        self.last_block_on_l1.subscribe()
     }
     pub fn latest_pending_block(&self) -> Arc<MadaraPendingBlockInfo> {
         self.pending_block.borrow().clone()
@@ -106,8 +106,8 @@ impl MadaraBackend {
         self.watch_blocks.subscribe_pending_block()
     }
     #[tracing::instrument(skip_all, fields(module = "MadaraBackendWatch"))]
-    pub fn subscribe_last_confirmed_block(&self) -> LastConfirmedBlockReceived {
-        self.watch_blocks.subscribe_last_confirmed_block()
+    pub fn subscribe_last_block_on_l1(&self) -> LastBlockOnL1Receiver {
+        self.watch_blocks.subscribe_last_block_on_l1()
     }
     #[tracing::instrument(skip_all, fields(module = "MadaraBackendWatch"))]
     pub fn latest_pending_block(&self) -> Arc<MadaraPendingBlockInfo> {
--- a/madara/crates/client/db/src/watch.rs
+++ b/madara/crates/client/db/src/watch.rs
@@ -2,10 +2,9 @@ use crate::{MadaraBackend, MadaraStorageError};
 use mp_block::{header::PendingHeader, MadaraBlockInfo, MadaraPendingBlockInfo};
 use std::sync::Arc;
 
-pub type ClosedBlocksReceiver = tokio::sync::broadcast::Receiver<Arc<MadaraBlockInfo>>;
-pub type PendingBlockReceiver = tokio::sync::watch::Receiver<Arc<MadaraPendingBlockInfo>>;
-pub type PendingTxsReceiver = tokio::sync::broadcast::Receiver<mp_block::TransactionWithReceipt>;
-pub type LastConfirmedBlockReceived = tokio::sync::watch::Receiver<Option<u64>>;
+pub type ClosedBlocksReceiver = broadcast::Receiver<Arc<MadaraBlockInfo>>;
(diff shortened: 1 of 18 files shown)


madara/crates/client/db/src/watch.rs
```diff
@@ -25,32 +26,46 @@ fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<Madara
...
+            pending_block: tokio::sync::watch::channel(make_fake_pending_block(None)).0,
+            pending_txs: tokio::sync::broadcast::channel(100).0,
+            last_confirmed_block: tokio::sync::watch::channel(None).0,
         }
     }
 
     pub fn init_initial_values(&self, db: &MadaraBackend) -> Result<(), MadaraStorageError> {
         let block = db.get_pending_block_info_from_db()?;
-        self.update_pending(block.into());
+        let latest_block = db.get_l1_last_confirmed_block()?;
```
jbcaron, line 47: Is this really the last block on L1 that we want? If so, it would be useful to rename the `last_confirmed_block` channel differently with the arrival of the consensus.

madara/crates/client/db/src/watch.rs
```diff
@@ -25,32 +26,46 @@ fn make_fake_pending_block(parent_block: Option<&MadaraBlockInfo>) -> Arc<Madara
...
+        self.pending_block.send_replace(block.into());
+        self.last_confirmed_block.send_replace(latest_block);
         Ok(())
     }
 
     pub fn update_pending(&self, block: Arc<MadaraPendingBlockInfo>) {
         self.pending_block.send_replace(block);
     }
 
+    pub fn update_last_confirmed_block(&self, latest_block: u64) {
```
jbcaron: ditto

madara/crates/client/devnet/src/lib.rs
```diff
@@ -460,7 +460,8 @@ mod tests {
         Duration::from_secs(500000),
         true
     )]
-    #[case::should_work_across_block_boundary(true, true, None, Duration::from_secs(1), true)]
+    // FIXME: flaky
+    // #[case::should_work_across_block_boundary(true, true, None, Duration::from_secs(1), true)]
```
jbcaron: It would be a good idea to open an issue, IMO

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -39,6 +39,10 @@ pub trait SubmitTransaction: Send + Sync {
         &self,
         tx: BroadcastedInvokeTxn,
     ) -> Result<AddInvokeTransactionResult, SubmitTransactionError>;
+
+    async fn received_transaction(&self, hash: mp_convert::Felt) -> Option<bool>;
+
+    async fn subscribe_new_transactions(&self) -> Option<tokio::sync::broadcast::Receiver<mp_convert::Felt>>;
```
jbcaron: Is this really the best trait for these functions?

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
```diff
@@ -0,0 +1,707 @@
...
+                format!("SubscribePendingTransactions failed to retrieve pending block from db for tx {tx_hash:#x}")
+            })?;
+
+        for (tx, tx_hash) in pending.inner.transactions.iter().zip(pending_info.tx_hashes.iter()) {
+            send_tx_if_matching(tx.clone(), *tx_hash, &sender_address, transaction_details, &sink).await?;
+        }
+    }
+
+    loop {
+        tokio::time::timeout(TIMEOUT, async {
```
jbcaron: We should check whether timeouts are really relevant in WS subscriptions

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
```diff
@@ -0,0 +1,707 @@
...
+) -> Result<(), crate::errors::StarknetWsApiError> {
+    let sink = if sender_address.len() as u64 <= super::ADDRESS_FILTER_LIMIT {
+        subscription_sink.accept().await.or_internal_server_error("Failed to establish websocket connection")?
+    } else {
+        subscription_sink.reject(crate::errors::StarknetWsApiError::TooManyAddressesInFilter).await;
+        return Ok(());
+    };
+
+    let pending_info = std::sync::Arc::clone(&starknet.backend.latest_pending_block());
+    let mut channel = starknet.backend.subscribe_pending_txs();
```
jbcaron: Shouldn't these two lines be swapped? Or maybe we just shouldn't handle the pending ones that were already generated during the subscription

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
```diff
@@ -0,0 +1,707 @@
...
+            send_tx_if_matching(tx, tx_hash, &sender_address, transaction_details, &sink).await?;
+
+            Ok(())
+        })
+        .await
+        .or_internal_server_error("SubscribePendingTransactions timed out")??;
+    }
+}
+
+async fn send_tx_if_matching(
```
jbcaron: Can you add a TODO to improve the filtering by adding a pre-filtering step that sends to multiple channels, similar to how `subscribe_events` does it?

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
```diff
@@ -0,0 +1,707 @@
...
+    sender_address: &std::collections::HashSet<starknet_types_core::felt::Felt>,
+    transaction_details: bool,
+    sink: &jsonrpsee::server::SubscriptionSink,
+) -> Result<(), crate::errors::StarknetWsApiError> {
+    let tx = match tx {
+        mp_transactions::Transaction::Invoke(ref inner) if sender_address.contains(inner.sender_address()) => tx,
+        mp_transactions::Transaction::L1Handler(ref inner) if sender_address.contains(&inner.contract_address) => tx,
+        mp_transactions::Transaction::Declare(ref inner) if sender_address.contains(inner.sender_address()) => tx,
+        mp_transactions::Transaction::Deploy(ref inner)
+            if sender_address.contains(&inner.calculate_contract_address()) =>
```
jbcaron: We should check if this is an overhead. In that case, we could store this information in the DB

jbcaron (commented):
Great, I think it would be good to have the constants directly in constant vars, even if they aren't necessarily reused elsewhere in the codebase.

madara/crates/client/db/src/watch.rs
Trantorian1: Good idea, have changed this to `last_block_on_l1`

madara/crates/client/devnet/src/lib.rs
Trantorian1: I think this is being fixed by @cchudant in #678

madara/crates/client/submit_tx/src/lib.rs
Trantorian1: We're kind of obliged to use this since the `Mempool` and  `GatewayProvider` are wrapped inside an `Arc<dyn SubmitTransaction>`, so don't have any other option than to implement the logic there.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
Trantorian1: I guess I could remove this? I don't feel strongly about this either way.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
Trantorian1: Nope, we don't want to switch these two lines up because then we could end up with duplicate transactions which are both in the `pending_info` and the `channel`.

Trantorian1: There still is a little margin of time during which we can miss transactions (between the call to `last_pending_block` and `subscribe_pending_txs`) but at least this way we get all the txs which were already present in the pending block.

madara/crates/client/rpc/src/versions/user/v0_8_0/methods/ws/subscribe_pending_transactions.rs
Trantorian1: I don't think that would be worth it given this is one of the only places in the codebase where we are need this info. That would potentially be a lot of extra space taken up in db for data which we only very rarely use.

jbcaron (approved):

madara/crates/primitives/transactions/src/compute_hash.rs
```diff
@@ -431,19 +431,23 @@ impl DeployAccountTransactionV3 {
 
 impl DeployTransaction {
     pub fn compute_hash(&self, chain_id: Felt, legacy: bool) -> Felt {
-        let contract_address = calculate_contract_address(
-            self.contract_address_salt,
-            self.class_hash,
-            &self.constructor_calldata,
-            Default::default(),
-        );
+        let contract_address = self.calculate_contract_address();
```
heemankv, line 434: Curious : 
What's the reason to make these specific changes ?

madara/crates/primitives/rpc/src/v0_8_1/starknet_ws_api.rs
```diff
@@ -3,3 +3,10 @@ pub struct TxnStatus {
     pub transaction_hash: starknet_types_core::felt::Felt,
     pub status: crate::v0_7_1::TxnStatus,
 }
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Eq, Hash, PartialEq, serde::Serialize, serde::Deserialize, Clone, Debug)]
+pub enum PendingTxnInfo {
+    Hash(starknet_types_core::felt::Felt),
+    Full(crate::v0_7_1::Txn),
```
heemankv, line 11: Curious :
`starknet_api_openrpc.rs` and other files seems to have all the generalised declaration irrespective of the types,
is there any specific reason to keep them as a part of `v0_7_1` ?

heemankv (approved):
Looks good to me !

madara/crates/primitives/transactions/src/compute_hash.rs
Trantorian1: This just makes it a bit more idiomatic to get the contract address for a `DeployTransaction` outside of the struct itself, and mirrors the API available under `DeclareTransaction`.

madara/crates/primitives/rpc/src/v0_8_1/starknet_ws_api.rs
Trantorian1: In most cases the code actually transparently imports the `v0_7_1` version, which is exported by default. I like to avoid import statement and so use the name of the module explicitely.

Trantorian1: Other than that, those are different and conflictual structs as compared to what will be available in `v0.8.1` and as such need to be in their own module.

heemankv: So do we plan to move them soon ?

PR #684, opened 2025-06-10
fix(artifacts): invalid ci

Trantorian1:
## Pull Request type

- Bugfix

Current artifacts CI publishes an invalid image tag, leading it to always being one version behind that of `.artifacts-versions.yml`.

## What is the new behavior?

- Fixed artifacts release CI
- QUOL changes, artifacts are re-generated more often so it should no longer be possible to accidentally remove them in a way in which the build script cannot handle.

No.

### DIFF
--- a/.github/workflows/task-build-artifacts.yml
+++ b/.github/workflows/task-build-artifacts.yml
@@ -66,6 +66,7 @@ jobs:
         run: |
           IMAGE="${{ inputs.registry }}/${{ github.repository_owner }}/artifacts"
           VERSION=$(yq -e ".current_version" .artifact-versions.yml)
+          VERSION=$(($VERSION+1))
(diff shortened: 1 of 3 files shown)


Mohiiit (approved):

.github/workflows/task-build-artifacts.yml
```diff
@@ -66,6 +66,7 @@ jobs:
         run: |
           IMAGE="${{ inputs.registry }}/${{ github.repository_owner }}/artifacts"
           VERSION=$(yq -e ".current_version" .artifact-versions.yml)
+          VERSION=$(($VERSION+1))
```
heemankv, line 69: Oh, so this was the bug that made the re-write

heemankv (approved):

PR #686, opened 2025-06-12
Add/factory

byteZorvin:
## Pull Request type
- Feature

## What is the new behavior?
- To support the boostrapper v2 changes, a cairo factory is to be used to deploy and setup the l2 contracts
- This siginificantly reduces the time taken to setup
- Makes thing atomic 

 No


bootstrapper-v2/l2-factory/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,53 @@
+// An EIC contract to set the bridge as legacy (vs. multi).
```
apoorvsadana: We should probably add a more descriptive comment regarding legacy and new bridges in Starkgate and why we're writing our own EIC as we want to start from the new bridge directly.

bootstrapper-v2/l2-factory/src/EthBridgeEIC/contract.cairo
apoorvsadana, line 1: So, I noticed this thing that the `legacy_bridge_eic` in the new token bridge assigns the `UPGRADE_GOVERNOR` to the caller address [here](https://github.com/starknet-io/starkgate-contracts/blob/5a10fd263d29cd032b7229691d043520edae0737/src/cairo/legacy_bridge_eic.cairo#L67) along with some other role initalisation.

Now, the token bridge constructor also calls an `_initialize_roles` which is exactly the same as the EIC one with the different that token bridge doesn't assign `UPGRADE_GOVERNOR` to anyone. Am I missing something?

apoorvsadana: Oh, since `GOVERNANCE_ADMIN` is the master admin it can always add the `UPGRADE_GOVERNOR`?

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+                impl_hash: self.token_bridge_class_hash.read(),
+                eic_data: Some(
+                    EICData {
+                        eic_hash: self.eic_class_hash.read(),
+                        eic_init_data: ['eth', l2_eth_token.into()].span(),
+                    },
+                ),
+                final: false,
+            };
+            ethBridgeReplaceable.add_new_implementation(impl_data);
```
apoorvsadana: let's use snake case everywhere!

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+                self.erc20_class_hash.read(), 'Eth_salt'.into(), calldata.span(), false,
+            )
+                .unwrap_syscall();
+
+            let ethBridgeRole = IRolesDispatcher { contract_address: l2_eth_bridge };
+            // Start upgrading eth bridge
+            ethBridgeRole.register_upgrade_governor(provisional_gov_admin);
+
+            let ethBridgeReplaceable = IReplaceableDispatcher { contract_address: l2_eth_bridge };
+            // Add implementation
```
apoorvsadana: Is this how it's done in Cairo usually? You create two variables pointing to the same address with different generic interfaces (instead of one specific interface that has all the methods)

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+            let impl_data: ImplementationData = ImplementationData {
+                impl_hash: self.token_bridge_class_hash.read(),
+                eic_data: Some(
+                    EICData {
+                        eic_hash: self.eic_class_hash.read(),
+                        eic_init_data: ['eth', l2_eth_token.into()].span(),
```
apoorvsadana: So `l1_token` is set to `eth` because Eth doesn't have an address right? Maybe we can add a comment with this, might be better for future understanding.

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+        self.erc20_class_hash.write(erc20_class_hash);
+        self.l1_eth_bridge_address.write(l1_eth_bridge_address);
+        self.l1_erc20_bridge_address.write(l1_erc20_bridge_address);
+        self.owner.write(initial_owner);
+    }
+
+
+    #[abi(embed_v0)]
+    impl MadaraFactoryImpl of IMadaraFactory<ContractState> {
+        fn deploy_bridges(
```
apoorvsadana: A lot of things happening here. We should either split it into functions or add a very good comment explaining what's happening imo.

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+            ethBridgeReplaceable.replace_to(impl_data);
+
+            ethBridgeRole.remove_upgrade_governor(provisional_gov_admin);
+            // Ensure to remove this constract as the provisional_gov post the deployment setup
+            ethBridgeRole.register_governance_admin(self.owner.read());
+
+            // Deploy Token Bridge
+            let provisional_gov_admin = starknet::get_contract_address();
+            let mut calldata = ArrayTrait::new();
+            provisional_gov_admin.serialize(ref calldata);
```
apoorvsadana: I guess we're not removing the governance admin of the token bridge?

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+    impl InternalImpl of InternalTrait {
+        fn configure_bridge(
+            ref self: ContractState,
+            bridge_address: ContractAddress,
+            l1_bridge: EthAddress,
+            is_erc20: bool,
+        ) {
+            let contract = starknet::get_contract_address();
+            let bridgeRole = IRolesDispatcher { contract_address: bridge_address };
+            bridgeRole.register_app_role_admin(contract);
```
apoorvsadana: we should add a proper comment explaining why we're giving this the admin role and that we're removing it later in the end.

bootstrapper-v2/l2-factory/src/EthBridgeEIC/contract.cairo
byteZorvin: Yes, I feel this was unnecessary role grant while setting up the bridges. Setting up the bridge usually wont involve upgrading it as well.
On a side note, we would be doing exactly this, upgrading just after deploying as EIC will be called while upgrading only.

bootstrapper-v2/l2-factory/src/EthBridgeEIC/contract.cairo
byteZorvin: Added a comment

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
byteZorvin: I have seen people doing this, seems a common practice to me.

.gitmodules
Mohiiit, line 1: we are moving away from the sub-modules, so it would be ideal to add the flow to build the artifacts in the docker file inside `build-artifacts/build.docker` and it can create an artifacts when docker runs and it will push it to the ghcr as well, so that everyone can use it without building it from scratch

wdyt?

bootstrapper-v2/contracts/madara/Scarb.lock
```diff
@@ -0,0 +1,24 @@
+# Code generated by scarb DO NOT EDIT.
+version = 1
+
+[[package]]
+name = "l2_factory"
```
Mohiiit:
```suggestion
name = "appchain_factory"
```

bootstrapper-v2/contracts/madara/Scarb.toml
```diff
@@ -0,0 +1,52 @@
+[package]
+name = "madara-contracts"
```
Mohiiit:
```suggestion
name = "appchain-factory-contracts"
```

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,87 @@
...
+// 3. While deploying we want to directly deploy the latest version instead of deploying Legacy
+// first and then the upgrading them, as the initial Legacy bridge were in cairo 0, which cant be
+// declared/deployed post starknet 0.14.0
+//
+// # Approach
+// 4. Due to differences of nature of Eth token with other ERC20 on Ethereum, there are 2 different
+// Contracts for EthBridge and TokenBridge(Starkgate overrides the existing functions like `deposit`
+// and `acceptDeposit`)
+//
+// 5. Also `token_bridge.cairo` strores the `l1_bridge` which can have only single value(obviously)
```
Mohiiit:
```suggestion
// 5. Also `token_bridge.cairo` stores the `l1_bridge` which can have only single value(obviously)
```

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,87 @@
...
+//
+// 6. Combining 4. and 5. mandates that we will have to have 2 instances of `token_bridge.cairo` on
+// on L2 as well.
+//
+// 7. The EthBridge has the `enroll_token` disabled, while uses the `bridgeToken()` which can be set
```
Mohiiit:
```suggestion
// 7. The EthBridge has the `enroll_token` disabled, while MultiBridge uses the `bridgeToken()` which can be set
```

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,87 @@
...
+// using the EIC 8. The `l1_handler` for `deposit` on `token_bridge.cairo` will need to have the
```
Mohiiit:
```suggestion
// using the EIC.
// 8. The `l1_handler` for `deposit` on `token_bridge.cairo` will need to have the
```

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,87 @@
...
+// maps `l1_l2_token_map` and `l2_l1_token_map` being set (for Eth in this case)
+//
+// 8. Because of 7. this can't be done with existing flows, we will have to uuse an EIC to set them
```
Mohiiit:
```suggestion
// 9. Because of 8. this can't be done with existing flows, we will have to use an EIC to set them
```

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
```diff
@@ -0,0 +1,87 @@
...
+
+    #[storage]
+    struct Storage {
+        // --- Token Bridge ---
+        // Mapping from between l1<->l2 token addresses.
+        l1_l2_token_map: starknet::storage::Map<EthAddress, ContractAddress>,
+        l2_l1_token_map: starknet::storage::Map<ContractAddress, EthAddress>,
+        // `l2_token` is a legacy storage variable from older versions.
+        // It should be written to as well to prevent multiple init, making the bridge a single.
+        // and also to support legact L1-L2 msgs.
```
Mohiiit: this is not very clear, can you re-phrase it?

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+    #[abi(embed_v0)]
+    impl MadaraFactoryImpl of IMadaraFactory<ContractState> {
+        fn deploy_bridges(
+            ref self: ContractState,
+        ) -> (ContractAddress, ContractAddress, ContractAddress) {
+            // Deploy l2 eth bridge
+            let provisional_gov_admin = starknet::get_contract_address();
+            let mut calldata = ArrayTrait::new();
+            provisional_gov_admin.serialize(ref calldata);
+            0.serialize(ref calldata); // upgrade_delay
```
Mohiiit: curious, what is happening in this step?

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
```diff
@@ -0,0 +1,146 @@
...
+            let contract = starknet::get_contract_address();
+            let bridgeRole = IRolesDispatcher { contract_address: bridge_address };
+            bridgeRole.register_app_role_admin(contract);
+            bridgeRole.register_app_governor(contract);
+            let bridgeAdmmin = ITokenBridgeAdminDispatcher { contract_address: bridge_address };
+            bridgeAdmmin.set_l2_token_governance(self.owner.read());
+            bridgeAdmmin.set_l1_bridge(l1_bridge);
+            if is_erc20 {
+                bridgeAdmmin.set_erc20_class_hash(self.erc20_class_hash.read());
+            }
```
Mohiiit: I think the variable should be snake case, it's not very consistent in the current code!
```suggestion
            let bridge_admin = ITokenBridgeAdminDispatcher { contract_address: bridge_address };
            bridge_admin.set_l2_token_governance(self.owner.read());
            bridge_admin.set_l1_bridge(l1_bridge);
            if is_erc20 {
                bridge_admin.set_erc20_class_hash(self.erc20_class_hash.read());
            }
```

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
Mohiiit: +1

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
Mohiiit: +1

bootstrapper-v2/contracts/madara/src/EthBridgeEIC/contract.cairo
byteZorvin: updated

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
byteZorvin: - One can't renounce governance_admin
- Its only possible to remove gov_admin by another gov_admin

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
byteZorvin: Added a comment

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
byteZorvin: updated

bootstrapper-v2/l2-factory/src/madara_factory/contract.cairo
apoorvsadana: Oh ok, might be an important log to add after running the bootstrapper

bootstrapper-v2/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    // Copy build artifacts to the appropriate locations
+    let copy_dirs = [
+        ("../build-artifacts/bootstrapper/solidity/starkware", "./contracts/ethereum/src/starkware"),
+        ("../build-artifacts/bootstrapper/solidity/third_party", "./contracts/ethereum/src/third_party"),
+        ("../build-artifacts/bootstrapper/solidity/out", "./contracts/ethereum/out"),
+        ("../build-artifacts/bootstrapper/cairo/target", "./contracts/madara/target"),
+    ];
+
+    for (src, dst) in copy_dirs.iter() {
```
apoorvsadana, line 18: as discussed, can we move this to forge build?

byteZorvin: Forge does not support build time scripts, so there is no straightforward way to achieve this

bootstrapper-v2/build.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    // Copy build artifacts to the appropriate locations
+    let copy_dirs = [
+        (format!("{}/solidity/starkware", BOOTSTRAPPER_DIR), format!("{}/ethereum/src/starkware", CONTRACTS_DIR)),
+        (format!("{}/solidity/third_party", BOOTSTRAPPER_DIR), format!("{}/ethereum/src/third_party", CONTRACTS_DIR)),
+        (format!("{}/solidity/out", BOOTSTRAPPER_DIR), format!("{}/ethereum/out", CONTRACTS_DIR)),
+        (format!("{}/cairo/target", BOOTSTRAPPER_DIR), format!("{}/madara/target", CONTRACTS_DIR)),
+    ];
+
+    for (src, dst) in copy_dirs.iter() {
```
apoorvsadana, line 18: should we try to run the make command itself here?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,241 @@
...
+  function setupCoreContract(
+    address coreContractImplementation,
+    CoreContractInitData calldata coreContractInitData,
+    address operator,
+    address governor
+  ) public returns (address) {
+    _requireNotPaused();
+    // Deploying proxy with 0 upgradeActivationDelay
+    Proxy coreContractProxy = new Proxy(0);
+    // [sub_contracts_addresses[], eic address, initData].
```
apoorvsadana, line 90: I am still getting confused. What is `[sub_contracts_addresses[], eic address, initData]` here? Where are we sending sub_contracts_addresses?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,241 @@
...
+    registerAdmins(managerProxy, governor);
+  }
+
+  // Ensure to remove the governance admin role post setup,
+  // As this function can be rerun 
+  function set_l2_bridge(
+    uint256 l2_eth_bridge_address,
+    uint256 l2_erc20_bridge_address,
+    address ethTokenBridge,
+    address tokenBridge
```
apoorvsadana: camelCase?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: The sub_contracts_addresses[] is optional and is not present here as its empty. For us it is always empty and not present

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: updated

bootstrapper-v2/build.rs
byteZorvin: - I assume you are tallking about the `make artifacts` command,
- Running here would enforce building artifacts locally everytime instead of downloading from ghcr, hitting building times.
- It could make sense in the else case at line 32, but even there I would like to avoid as there are cases for ex: no internet which could be more easily solved instead of building locally. On side note for no internet even `make artifacts` wont work as git clones would fail 😅

.artifact-versions.yml
```diff
@@ -1,5 +1,9 @@
-current_version: 2
+current_version: 3
 versions:
+  - version: 3
```
Mohiiit: it should be version 4 right?

Mohiiit (approved):

apoorvsadana (approved):

PR #687, opened 2025-06-12
fix: nix env

jbcaron:
## Pull Request type

- Build-related changes

Resolves: #NA

## What is the new behavior?

- added `mold` linker
- enhanced runtime environment via explicit `LD_LIBRARY_PATH`
- increase the limit of open file descriptors

No

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -43,6 +43,7 @@
             yq
             scarb
             gnumake
+            mold
             wget
             git
           ];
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -58,6 +59,10 @@
             ];
 
           shellHook = ''
+          # Increase the limit of open file descriptors
+          echo "[INFO] Increasing open file descriptor limit to 65535 (ulimit -n)"
+          ulimit -n 65535 || echo "[WARN] Failed to set ulimit -n to 65535"
```
Trantorian1, line 64: Why are we doing this?

jbcaron: rocksdb is greedy with file descriptors

Trantorian1 (approved):

cchudant (approved):

PR #688, opened 2025-06-14
fix(`starknet_unsubscribe`)

Trantorian1:
## Pull Request type

- Bugfix

Currently, the madara ws API does not adhere to the specs by not allowing ws subscriptions to be terminated via `starknet_unsubscribe`. This is a limitation in the way in which `jsonrpsee` works which makes it impossible to have different ws methods with the same unsubscribe name.

## What is the new behavior?

Manual handling of ws method unsubscribe via a new `starknet_unsubscribe` endpoint. This works by tracking ws subscription ids for active subscriptions and notifying them they have been unsubscribed when this method is called, resulting in a cancel.

Yes. Ws RPC method items were not conforming to specs and did not return the subscription id number. This has been fixed but marks a breaking change in the format of ws responses.

## Other information

> [!NOTE]
> The way this works is a bit janky: jsonrpsee does not let a ws method close itself, and strangely even the method exiting will not close the sink. The only way to close a ws subscription this way is to return an error. In the future, I would really like us to move _away_ from jsonrpsee in favor of our own Json-RPC middleware which would give us much more flexibility in scenarios such as these

### DIFF
--- a/madara/crates/client/rpc/src/lib.rs
+++ b/madara/crates/client/rpc/src/lib.rs
@@ -160,16 +160,17 @@ pub(crate) struct WsSubscribeHandles {
     /// > Locking behaviour: May deadlock if called when holding any sort of reference into the map.
     ///
     /// This is fine in our case as we do not maintain references to a map in the same thread while
-    /// mutating it and instead operate directly on-value.
+    /// mutating it and instead operate directly on-value by sharing the map inside an [Arc].
     ///
     /// [DashMap]: dashmap::DashMap
     /// [DashMap::entry]: dashmap::DashMap::entry
-    handles: dashmap::DashMap<u64, std::sync::Arc<tokio::sync::Notify>>,
+    /// [Arc]: std::sync::Arc
+    handles: std::sync::Arc<dashmap::DashMap<u64, std::sync::Arc<tokio::sync::Notify>>>,
 }
 
 impl WsSubscribeHandles {
     pub fn new() -> Self {
-        Self { handles: dashmap::DashMap::new() }
+        Self { handles: std::sync::Arc::new(dashmap::DashMap::new()) }
     }
 
     pub async fn subscription_register(&self, id: jsonrpsee::types::SubscriptionId<'static>) -> WsSubscriptionGuard {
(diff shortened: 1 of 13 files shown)


cchudant (approved):
It's hard for me to see if this hack is fine unless I try to do it and find why we have no choice
I have to reluctantly approve this until I get the time to check that I guess

madara/crates/client/rpc/src/lib.rs
```diff
@@ -131,3 +133,64 @@ pub fn rpc_api_admin(starknet: &Starknet) -> anyhow::Result<RpcModule<()>> {
...
 }
+
+pub(crate) struct WsSubscribeHandles {
+    /// Keeps track of all ws connection handles.
+    ///
+    /// This can be used to request the closure of a ws connection.
+    ///
+    /// ## Preventing Leaks
+    ///
+    /// Since connections can be ended abruptly and the jsonrpsee api does not allow use to retrieve
```
heemankv: Nitpick : 
`does not allow us` instead of `use`

madara/crates/client/rpc/src/lib.rs
```diff
@@ -131,3 +133,64 @@ pub fn rpc_api_admin(starknet: &Starknet) -> anyhow::Result<RpcModule<()>> {
...
+
+        let id = match id {
+            jsonrpsee::types::SubscriptionId::Num(id) => id,
+            jsonrpsee::types::SubscriptionId::Str(_) => {
+                unreachable!("Jsonrpsee middleware has been configured to use u64 subscription ids")
+            }
+        };
+
+        let mut lock = self.handles.write().await;
+        lock.retain(|_, handle| std::sync::Arc::strong_count(&handle.0) != 1);
```
heemankv: If I understand this correctly, 
During every new subscription registration, the code iterates through ALL existing connections to check if they're stale ?
is this the most optimised way to do it, given we are taking O(n) to read through all the hanndles 

Are we confident that this will handle through high connection churn

heemankv: Should we look for periodic background cleanup or lazy cleanup approachs ?

heemankv (commented):
I notice that majority of the code here is from #683, re-basing should clear that now that that's merged, 
Left minor comments, rest looks good to me!

madara/crates/client/rpc/src/lib.rs
Trantorian1: Yea I can see how this could be a problem. Will consider how to solve this. To be honest the whole fix is quite a bit of a hack, it feels like the main issue is that we should be migrating out of `jsonrpsee` :/

heemankv (approved):

PR #690, opened 2025-06-16
fix: block production outputs normalized state diff

cchudant:
## Pull Request type

- Bugfix

implement the state diff normalization behavior:
```rs
/// The resulting state diff is normalized: when merging state maps together, if for example a transaction changed a
/// given storage key to new value, and a following transaction change it back to the original value - the normalized state diff
/// should not preserve the storage key, since it ends up not being changed in the end. This function does this normalization,
/// by looking up every key in the backend and only keep the changed ones.
```


madara/crates/client/exec/src/state_diff.rs
```diff
@@ -0,0 +1,301 @@
...
+            },
+            ContractStorageDiffItem {
+                address: Felt::from_hex_unchecked("3"),
+                storage_entries: vec![
+                    StorageEntry { key: Felt::from_hex_unchecked("1"), value: Felt::ONE },
+                    StorageEntry { key: Felt::from_hex_unchecked("2"), value: Felt::TWO },
+                    StorageEntry { key: Felt::from_hex_unchecked("3"), value: Felt::THREE },
+                ],
+            },
+        ];
```
Trantorian1: This doesn't check for storage diff deduplication no? We should probably add a test for that and deduplicated class hashes.

cchudant: yes sorry im adding that

cchudant: test was added :)

jbcaron (commented):
Since the `StateDiff` is now deterministically ordered, it makes sense to revisit the hash computation logic for the `StateDiff` struct to guarantee consistency. We should also ensure that when syncing, the `StateDiff` is produced in an ordered form right from its creation.

madara/crates/client/exec/src/state_diff.rs
```diff
@@ -0,0 +1,295 @@
...
+        StorageEntry,
+    };
+    use starknet_api::{
+        core::{ClassHash, CompiledClassHash, ContractAddress, Nonce},
+        state::StorageKey,
+    };
+    use std::{collections::HashMap, sync::Arc};
+
+    #[tokio::test]
+    async fn test_create_normalized_state_diff() {
```
Trantorian1, line 142: Neat!

Trantorian1 (approved):
Lgtm :+1:

cchudant:
> Since the `StateDiff` is now deterministically ordered, it makes sense to revisit the hash computation logic for the `StateDiff` struct to guarantee consistency. We should also ensure that when syncing, the `StateDiff` is produced in an ordered form right from its creation.

I haven't changed the sync logic for that, let's do it in a future pr.

jbcaron (approved):

PR #691, opened 2025-06-18
Orchestrator: Fixes Workers

### LINKED ISSUE #590: Orchestrator Prod Readiness  : Optimise stopping on Failed jobs.
Currently the when the orchestrator detects a Failed or VerificationTimeout job, it stops creation of ANY new job.
ANY being the key word here. 
 
Example : 
1. Block 10 : About to create state transition job.
2. Block 15 : Processing SnosRun.

Let's say that SnosRun for block 15 failed, it will also lead to hault the state transition job of block 10, 
Which is completely unrelated.

We would Ideally only want the next new jobs to be stopped from and not the jobs previous to the one that failed.

Note : Need to ensure that the minimum job out of the multiple that failed is taken as reference.

heemankv:
# Resolves #590: Optimize Orchestrator Chain Settlement Logic

## Problem
The Orchestrator currently halts the entire chain settlement process when any job reaches `Failed` or `Verification Timeout` status, leading to inefficient processing and unnecessary blocking of independent operations.

**Current problematic behavior:**
- Chain settlement stops completely on any job failure
- Independent jobs are blocked unnecessarily (e.g., `StateTransition` for block 5 is halted when `SnosRun` for block 10 fails)
- Cascading delays affect the entire pipeline

## Solution
This PR optimizes the worker logic to continue processing independent jobs even when unrelated jobs fail, improving overall system resilience and throughput.


orchestrator/src/tests/database/mod.rs
```diff
@@ -215,7 +215,8 @@ async fn database_test_update_job() {
             JobItemUpdates::new()
                 .update_status(JobStatus::LockedForProcessing)
                 .update_metadata(updated_job_metadata)
-                .build(),
+                .build()
+                .unwrap(),
```
heemankv, line 219: Need to remove unwrap from here

orchestrator/src/tests/workers/proving/mod.rs
```diff
@@ -1,123 +1,248 @@
...
-    if incomplete_runs {
-        ctx.expect().times(4).with(eq(JobType::ProofCreation)).returning(move |_| Arc::clone(&job_handler));
-    } else {
-        ctx.expect().times(5).with(eq(JobType::ProofCreation)).returning(move |_| Arc::clone(&job_handler));
-    }
+    // Run the Proving worker
+    crate::worker::event_handler::triggers::proving::ProvingJobTrigger.run_worker(services.config).await?;
 
-    ProvingJobTrigger.run_worker(services.config).await?;
+    println!(
```
prkpndy: we can remove this `println!` statement

orchestrator/src/tests/workers/data_submission/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    let mut queue = MockQueueClient::new();
+    let mut job_handler = MockJobHandlerTrait::new();
+
+    // Mock earliest_failed_block_number query
+    db.expect_get_earliest_failed_block_number().returning(move || Ok(earliest_failed_block));
+
+    // Create completed proving job items
+    let proving_job_items: Vec<_> = completed_proving_jobs
+        .iter()
+        .map(|(block_num, input_path, proof_hash, n_steps)| {
```
Mohiiit: just a correction, there is nothing like proof_hash, it's a proof path, and whenever the path is given we download the proof, although I think it doesn't matter in the current case!

orchestrator/src/tests/workers/data_submission/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    // Setup job handler context
+    let job_handler: Arc<Box<dyn JobHandlerTrait>> = Arc::new(Box::new(job_handler));
+    let ctx = get_job_handler_context();
+    ctx.expect().with(eq(JobType::DataSubmission)).returning(move |_| Arc::clone(&job_handler));
+
+    // Mock queue operations for successful job creations
+    queue
+        .expect_send_message()
+        .times(expected_created_count)
```
Mohiiit, line 216: given `expected_created_count` would always be equal to the length of the `expected_data_submission_jobs`, I think we can reduce one variable!

wdyt? would be there be a case when it won't hold?

orchestrator/src/worker/event_handler/triggers/mod.rs
Mohiiit, line 1: quite a few lines commented over here! 

also, we are not using run_worker_if_enabled anymore? what's the reasoning behind that?

orchestrator/src/worker/event_handler/triggers/proving.rs
```diff
@@ -1,86 +1,202 @@
...
+    }
+
+    /// Builds proving job metadata from SNOS metadata
+    fn build_proving_metadata(&self, snos_metadata: &SnosMetadata, snos_fact: String) -> JobMetadata {
+        JobMetadata {
+            common: CommonMetadata::default(),
+            specific: JobSpecificMetadata::Proving(ProvingMetadata {
+                block_number: snos_metadata.block_number,
+                input_path: snos_metadata.cairo_pie_path.as_ref().map(|path| ProvingInputType::CairoPie(path.clone())),
+                download_proof: None,
```
Mohiiit, line 164: I am thinking how will it change in case of L3, given the context contains the information about whether it's a L2 or L3, I think we would need to pass the context here?

and based on that, it would set the download_proof path and ensure_on_chain_registration as well!

wdyt?

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -17,171 +18,373 @@ pub struct UpdateStateJobTrigger;
...
             }
         }
 
-        tracing::trace!(log_type = "completed", category = "UpdateStateWorker", "UpdateStateWorker completed.");
-        Ok(())
+        Ok(None)
+    }
+
+    /// Determines which blocks should be processed based on completed DA jobs and previous state transitions
+    /// Determines which blocks should be processed based on completed DA jobs and previous state transitions
```
Mohiiit, line 216:
```suggestion
```

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -17,171 +18,373 @@ pub struct UpdateStateJobTrigger;
...
+    async fn determine_blocks_to_process(&self) -> color_eyre::Result<Option<Vec<u64>>> {
+        let latest_job = self.config.database().get_latest_job_by_type(JobType::StateTransition).await?;
+
+        let earliest_failed_block = self.config.database().get_earliest_failed_block_number().await?;
+
+        let (completed_da_jobs, last_block_processed) = match latest_job {
+            Some(job) => {
+                let last_block = self.get_last_processed_block(&job).await?;
+                let da_jobs = self.get_da_jobs_after_block(last_block).await?;
```
Mohiiit, line 225: I think we should add the limit to query like these in general, can be handled in different PR, given it's an optimization!

wdyt?

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -17,171 +18,373 @@ pub struct UpdateStateJobTrigger;
...
+                (da_jobs, Some(last_block))
+            }
+            None => {
+                tracing::warn!("No previous state transition job found, fetching latest data submission job");
+                let da_jobs = self.get_all_unprocessed_da_jobs().await?;
```
Mohiiit, line 230: same as above!

Mohiiit (commented):
can you clarify the difference between context implementation and the trigger implementation here? I see it's been used interchangeably or maybe I am missing something and do we really need that segregation?

orchestrator/src/tests/workers/proving/mod.rs
heemankv: alright

orchestrator/src/tests/workers/data_submission/mod.rs
heemankv: Valid

orchestrator/src/worker/event_handler/triggers/mod.rs
heemankv: `run_worker_if_enabled` is removed because it determines even before the worker has started weather to run it or not only based on `Failed` or `VerificationTimeout` jobs. 
we want each worker to understand weather they are to create a job or not based on their individual logic.
Hence the removal!

orchestrator/src/worker/event_handler/triggers/proving.rs
heemankv: Valid, we'll have to handle that on the rebase with l3 pr!

orchestrator/src/worker/event_handler/triggers/update_state.rs
heemankv: Valid, I believe there are many db calls that we should add limits to!
Marking as an issue under Prod Readiness 
Here : #715

orchestrator/src/worker/event_handler/triggers/update_state.rs
heemankv: Here : #715

orchestrator/src/worker/event_handler/triggers/proving.rs
Mohiiit: cc: @itsparser

orchestrator/src/tests/workers/data_submission/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+        let block_num_str = block_num.to_string();
+
+        // Only expect job creation for jobs that should actually be created
+        // (i.e., are not beyond failed block)
+        if earliest_failed_block.is_none() || block_num < earliest_failed_block.unwrap() {
+            let mut da_job_item = get_job_item_mock_by_id(block_num_str.clone(), uuid);
+            da_job_item.metadata.specific = JobSpecificMetadata::Da(DaMetadata {
+                block_number: block_num,
+                blob_data_path: Some(format!("{}/{}", block_num, BLOB_DATA_FILE_NAME)),
+                tx_hash: None, // Will be populated during processing
```
Mohiiit: the comment here is wrong, given the current structure, it won't be populated during processing, but yeah you can argue it could be the case in future when we are using celestia or avail for DA

orchestrator/src/tests/workers/proving/mod.rs
```diff
@@ -1,123 +1,243 @@
...
 #[rstest]
-#[case(true)]
-#[case(false)]
+// Scenario 1: No completed SNOS jobs exist
+// Expected result: no proving jobs created
+#[case(
+    None,     // earliest_failed_block
+    vec![],   // completed_snos_jobs (no completed SNOS jobs)
+    vec![],   // expected_proving_jobs (no jobs to create)
+    0         // expected_created_count
```
Mohiiit: this `expected_created_count` can be removed just like DA?

orchestrator/src/tests/workers/data_submission/mod.rs
heemankv: removed from all !, thanks

orchestrator/src/tests/workers/data_submission/mod.rs
heemankv: Valid, removed

orchestrator/src/tests/workers/proving/mod.rs
heemankv: yes!, removed!

orchestrator/src/worker/event_handler/triggers/data_submission.rs
```diff
@@ -0,0 +1,189 @@
...
+            .get_jobs_without_successor(JobType::ProofCreation, JobStatus::Completed, JobType::DataSubmission)
+            .await
+    }
+
+    /// Determines if a block number should be skipped due to failed block constraints
+    fn should_skip_block(&self, block_number: u64) -> bool {
+        match self.earliest_failed_block {
+            Some(failed_block) => block_number >= failed_block,
+            None => false,
+        }
```
Trantorian1:
```suggestion
        self.earliest_failed_block.is_some_and(|failed_block| block_number >= failed_block)
```

orchestrator/src/worker/event_handler/triggers/proving.rs
```diff
@@ -1,86 +1,202 @@
...
+            .await
+            .context("Failed to get Eligible SNOS Jobs")
+    }
+
+    /// Determines if a block number should be skipped due to failed block constraints
+    fn should_skip_block(&self, block_number: u64) -> bool {
+        match self.earliest_failed_block {
+            Some(failed_block) => block_number >= failed_block,
+            None => false,
+        }
```
Trantorian1:
```suggestion
        self.earliest_failed_block.is_some_and(|failed_block| block_number >= failed_block)
```

orchestrator/src/worker/event_handler/triggers/proving.rs
```diff
@@ -1,86 +1,202 @@
...
+            }),
+        }
+    }
+}
+
+/// Context for processing jobs, containing shared data and configuration
+pub struct ProcessingContext {
+    pub config: Arc<Config>,
+    pub earliest_failed_block: Option<u64>,
+}
```
Trantorian1, line 176: Duplicate struct?

Trantorian1: https://github.com/madara-alliance/madara/blob/1e2a427f282ce5b4130f757caf6461bc91b02c51/orchestrator/src/worker/event_handler/triggers/data_submission.rs#L161-L164

Trantorian1 (approved):
From my limited experience with orchestrator, lgtm :ok_hand:

Mohiiit (approved):

PR #692, opened 2025-06-18
Fix: introduce deterministic hasher interface for bloom-filter

jbcaron:
This PR introduces a deterministic hashing abstraction to ensure reproducible and stable hash values across Bloom filter instances.

## Pull Request type

- Bugfix

Resolves: #NA

## What is the new behavior?

Bloom filters are persisted in the database. Therefore, the hashing logic must be deterministic across executions and environments to maintain filter integrity. Without this, deserialization or lookups on previously stored filters may behave incorrectly due to mismatched hash calculations.

- added `DeterministicBuildHashe`r trait for constructing reproducible hashers
- provided implementations for `AHasher` (with fixed seeds) and `DefaultHasher`
- fix linter warnings

yes, DB written with non-deterministic hashers prior to this change may no longer be valid or function as expected

### DIFF
--- a/madara/crates/client/submit_tx/src/validation.rs
+++ b/madara/crates/client/submit_tx/src/validation.rs
@@ -101,7 +101,7 @@ impl From<TransactionExecutionError> for SubmitTransactionError {
             | E::TransactionFeeError(_)
             | E::TransactionPreValidationError(_)
             | E::TryFromIntError(_)
-            | E::TransactionTooLarge { .. }) => rejected(ValidateFailure, format!("{err:#}")),
+            | E::TransactionTooLarge) => rejected(ValidateFailure, format!("{err:#}")),
             err @ E::InvalidVersion { .. } => rejected(InvalidTransactionVersion, format!("{err:#}")),
             err @ E::InvalidSegmentStructure(_, _) => rejected(InvalidProgram, format!("{err:#}")),
             E::StateError(err) => err.into(),
--- a/madara/crates/primitives/bloom_filter/src/filter.rs
+++ b/madara/crates/primitives/bloom_filter/src/filter.rs
@@ -13,7 +13,7 @@
 
 use crate::storage::{AtomicBitStore, BitStore};
 use serde::{Deserialize, Serialize};
-use std::hash::{Hash, Hasher};
+use std::hash::Hasher;
 use std::marker::PhantomData;
 /// A cache of computed hash positions that can be reused across different filter sizes.
 ///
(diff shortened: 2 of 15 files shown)


madara/crates/client/submit_tx/src/validation.rs
```diff
@@ -101,7 +101,7 @@ impl From<TransactionExecutionError> for SubmitTransactionError {
             | E::TransactionFeeError(_)
             | E::TransactionPreValidationError(_)
             | E::TryFromIntError(_)
-            | E::TransactionTooLarge { .. }) => rejected(ValidateFailure, format!("{err:#}")),
+            | E::TransactionTooLarge) => rejected(ValidateFailure, format!("{err:#}")),
```
cchudant, line 104: is that intended?

cchudant (approved):
nice!

madara/crates/primitives/bloom_filter/src/filter.rs
```diff
@@ -13,8 +13,45 @@
...
+}
+
+impl DeterministicBuildHasher for ahash::AHasher {
+    type H = ahash::AHasher;
+
+    fn build_hasher() -> Self::H {
+        ahash::RandomState::with_seeds(0x03d04773fad45387, 0x39d4e9b0f45e97cc, 0x63a175b11b94edaf, 0x075c02fb32039c72)
+            .build_hasher()
+    }
+}
```
Trantorian1: Is it really a good idea to implement this for `ahash`? My worry is that as the `ahash` hash function is not guaranteed to be stable across version we would be relying on minor semver versions not implementing any breaking changes. From having read the project readme, it is not clear that this is a guarantee which the project is making. We could just drop support for it in favor of `highway` as we discussed internally.

madara/crates/primitives/bloom_filter/src/tests/utils.rs
```diff
@@ -14,7 +14,7 @@ const BITS_PER_ELEMENT: f64 = 9.6;
 pub const FALSE_POSITIF_RATE: f64 = 0.01;
 
 /// Helper function to create a Bloom filter with the given number of elements
-pub fn create_filter<H: Hasher + Default>(nb_elem: u64) -> BloomFilter<H, AtomicBitStore> {
+pub fn create_filter<H: Hasher + DeterministicBuildHasher>(nb_elem: u64) -> BloomFilter<H, AtomicBitStore> {
```
Trantorian1: Is the new bound really needed here?

madara/crates/client/submit_tx/src/validation.rs
jbcaron: yes, it's requested by clippy

madara/crates/primitives/bloom_filter/src/filter.rs
jbcaron: I've updated the filter to use SipHasher24

Trantorian1 (approved):

PR #699, opened 2025-06-25
v0.14 mempool support

cchudant:
## Changes

- Fix mempool storage in DB (transactions that were not removed properly from db)
- Refactor the mempool datastructure
  - Simplified it, made it much more readable and easy to review.
  - Added mempool Tip mode: when enabled, transactions are ordered by tip instead of arrived_at timestamp. In v0.14 Starkware will stop doing FCFS and their mempool will be tip based, to mirror this functionnality, I implemented here too. However, I have kept the old FCFS mode in case an appchain really wants it. FCFS mode is still enabled by default.
  - Optimize mempool time-to-live auto removal: it's now done in batch by a mempool task.
  - Mempool now listens to backend updates and will remove unnecessary transactions as the chain progresses.
  - Implemented an eviction policy: when the mempool is full, insertion will make room for the new transactions:
    - If there is a non-ready nonce chain that is longer than the nonce difference with the account, this will remove the last transaction from the nonce chain.
    - If it's the same difference, score will be compared instead.
    - This is very important for DoS prevention.
  - Support out of order insertion properly.
  - Declare transactions can only be added to the mempool if the nonce exactly matches the account nonce.
  - Added a lot of tests for the mempool datastructure for every case I could think of.

In summary, the mempool now supports most of the v0.14 changes including Tip mode while keeping its FCFS mode available, and the code is now much cleaner.
I plan on refactoring transaction statuse to support the new v0.14 pre-confirmed block next, but in the meantime I did a dirty thing in this PR to support at least Received status when a tx is consumed for block production.

## Important change!

When popping a transaction to create a batch for execution, the new inner mempool will **not** increase its account nonce. This is important in order to stop a few DoS vectors (that are quite important in tip mode), but this does mean that the block production batcher will not be able to schedule multiple transactions for the same account in the same batch. Instead, the following transactions will only be made available to the batcher once the mempool has received and applied the update of the batch that had the first transaction.
This adds a bit of latency for now :( but once we have support for the new blockifier ConcurrentTransactionExecutor this should be less of an issue.

In addition: even if we were to allow executing multiple transactions for the same account in the same batch, the executor would not be able to execute them in parallel (via optimistic parallel execution) because they update the same nonce and ERC20 balance. Meaning, them being in the same batch shouldn't actually improve perf.


madara/crates/client/mempool/src/inner/accounts.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+#[derive(Debug, Default)]
+#[cfg_attr(any(test, feature = "testing"), derive(PartialEq, Eq, Clone))]
+pub struct Accounts {
+    accounts: HashMap<ContractAddress, AccountState>,
+}
+
+#[cfg(any(any(test, feature = "testing"), feature = "testing"))]
```
Trantorian1:
```suggestion
#[cfg(any(test, feature = "testing"))]
```

madara/crates/client/mempool/src/inner/accounts.rs
```diff
@@ -0,0 +1,403 @@
...
+                assert_eq!(*nonce, tx.nonce(), "Account {contract_address:?} tx has bad nonce key");
+            }
+        }
+
+        // Invariant: every account must have queued txs that have nonce >= account nonce.
+        for (contract_address, account) in &self.accounts {
+            let bad_nonces =
+                account.queued_txs.iter().filter(|(nonce, _)| **nonce < account.current_nonce).collect::<Vec<_>>();
+            assert_eq!(bad_nonces, vec![], "Account {contract_address:?} has queued txs < account nonce");
+        }
```
Trantorian1, line 196: Can't this be checked as part of the above loop?

madara/crates/client/mempool/src/inner/accounts.rs
Trantorian1, line 1: Cool changes! Much better compartmentalization and easier to understand, this is dope :)

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
 pub enum TxInsertionError {
-    #[error("A transaction with this nonce already exists in the transaction pool")]
+    #[error("Nonce mismatch: account nonce is {account_nonce:?}")]
+    NonceTooLow { account_nonce: Nonce },
+    #[error("A transaction with this nonce already exists in the mempool")]
     NonceConflict,
-    #[error("A transaction with this hash already exists in the transaction pool")]
+    #[error("A transaction with this hash already exists in the mempool")]
     DuplicateTxn,
+    #[error("Replacing a transaction requires at least a tip bump of at least {min_tip_bump} units")]
```
Trantorian1:
```suggestion
    #[error("Replacing a transaction requires a tip bump of at least {min_tip_bump} units")]
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
-                //
-                // In practice this is ok as l1 handler transactions are few and
-                // far between. Note that removing this check will result in an
-                // infinite loop if ever an l1 transaction is encountered.
-                ready_no_age_check.push(self.tx_intent_queue_ready.pop_first().expect("Already inside loop"));
+        // If we're replacing another transactions,
+        if let Some(previous_tx) = entry.replaced_tx() {
+            // If it's the same tx, show a nicer error message.
+            if previous_tx.tx_hash() == mempool_tx.tx_hash() {
+                return Err(TxInsertionError::DuplicateTxn);
```
Trantorian1, line 200: Shouldn't we support replacing existing txs through fee bumping?

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
-        #[cfg(any(test, feature = "testing"))]
-        self.nonce_cache_inner.insert(tx_mempool.contract_address(), tx_mempool.nonce_next);
-        self.limiter.mark_removed(&TransactionCheckedLimits::limits_for(&tx_mempool));
-        debug_assert!(
-            self.tx_received.remove(&tx_mempool.tx_hash()),
-            "Tried to remove a ready transaction which had not already been marked as received"
+        assert_eq!(
+            account_update.removed_txs.len(),
+            1,
+            "remove_ready_tx should remove exactly one tx from the mempool"
```
Trantorian1:
```suggestion
            "pop_next_ready should remove exactly one tx from the mempool"
```

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
```diff
@@ -0,0 +1,133 @@
...
+            tx_hash: Pedersen::hash_array(&[
+                nonce,
+                contract_address,
+                arrived_at.into(),
+                tip.map(Felt::from).unwrap_or_default(),
+                (is_declare as u32).into(),
+            ]),
+            is_declare,
+        })
+}
```
Trantorian1: Consider using the [`prop_compose!`](https://proptest-rs.github.io/proptest/proptest/tutorial/macro-prop-compose.html) macro to simplify the syntax here.

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
```diff
@@ -0,0 +1,133 @@
...
+fn arb_mempool_op() -> impl Strategy<Value = MempoolOp> {
+    prop_oneof![
+        (arb_tx(), (1u64..=5).prop_map(Felt::from)).prop_map(|(tx, nonce)| MempoolOp::InsertTx(tx, nonce)),
+        ((1u64..=3).prop_map(Felt::from), (1u64..=5).prop_map(Felt::from))
+            .prop_map(|(addr, nonce)| MempoolOp::UpdateAccountNonce(addr, nonce)),
+        Just(MempoolOp::PopNextReady),
+        Just(MempoolOp::RemoveTtlExceeded),
+        (0u128..=10000).prop_map(MempoolOp::SetCurrentTime),
+    ]
+}
```
Trantorian1: Ditto.

Trantorian1: Might be a good idea to break down each `prop_oneof` variation into its own generator for clarity

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
Trantorian1, line 1: Lol this is much better than using `proptest_state_machine`

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
```diff
@@ -0,0 +1,133 @@
...
+        max_transactions in 1usize..=10,
+        max_declare_transactions in prop::option::of(1usize..=5),
+        ttl_millis in 0u64..=5000,
+        ops in prop::collection::vec(arb_mempool_op(), 0..50)
+    ) {
+        let mut mempool = fcfs_mempool(
+            max_transactions,
+            max_declare_transactions,
+            Duration::from_millis(ttl_millis)
+        );
```
Trantorian1, line 106: I think you can use `rstest` fixtures inside the proptest macro.
````rust
#[rstest::rstest]
fn prop_fcfs_invariants_always_hold(
    max_transactions in 1usize..=10,
    max_declare_transactions in prop::option::of(1usize..=5),
    ttl_millis in 0u64..=5000,
    ops in prop::collection::vec(arb_mempool_op(), 0..50),
    #[with(
        max_transactions,
        max_declare_transactions,
        Duration::from_millis(ttl_millis)
    )]
    fcfs_mempool: MempoolTester
) {
    // ...
}
```

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
```diff
@@ -0,0 +1,133 @@
...
+        max_declare_transactions in prop::option::of(1usize..=5),
+        ttl_millis in 0u64..=5000,
+        ops in prop::collection::vec(arb_mempool_op(), 0..50)
+    ) {
+        let mut mempool = tip_mempool(
+            max_transactions,
+            min_tip_bump,
+            max_declare_transactions,
+            Duration::from_millis(ttl_millis)
+        );
```
Trantorian1, line 64: See comment bellow on rstest fixtures in proptest test.

madara/crates/client/mempool/src/inner/tx.rs
```diff
@@ -1,86 +1,192 @@
...
-            .field("tx_hash", &self.tx_hash().hex_display())
-            .field("nonce", &self.nonce.hex_display())
-            .field("nonce_next", &self.nonce_next.hex_display())
-            .field("contract_address", &self.contract_address().hex_display())
-            .field("tx_type", &self.tx.tx_type())
-            .field("arrived_at", &self.arrived_at)
-            .finish()
+impl Ord for EvictionScore {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.chain_nonce_len.cmp(&other.chain_nonce_len).then(self.score.cmp(&other.score).reverse())
```
Trantorian1, line 28: We reverse the eviction score to keep high eviction scores at the tip of a `BTree`?

madara/crates/client/mempool/src/inner/tx.rs
```diff
@@ -1,86 +1,192 @@
...
-        Self {
-            tx: self.tx.clone_blockifier_transaction(),
-            arrived_at: self.arrived_at,
-            converted_class: self.converted_class.clone(),
-            nonce: self.nonce,
-            nonce_next: self.nonce_next,
-        }
+impl EvictionScore {
+    pub fn new(tx: &MempoolTransaction, account_nonce: Nonce) -> Self {
+        Self { chain_nonce_len: Nonce(tx.nonce().0 - account_nonce.0), score: tx.score }
```
Trantorian1, line 39: We can't operate on `Nonce` directly? :///

madara/crates/client/mempool/src/inner/tx.rs
```diff
@@ -1,86 +1,192 @@
...
-        let nonce_next = nonce.try_increment()?;
-
-        Ok(Self { tx, arrived_at, converted_class, nonce, nonce_next })
+    pub fn new(
+        inner: ValidatedMempoolTx,
+        score_function: &ScoreFunction,
+    ) -> Result<MempoolTransaction, TxInsertionError> {
+        let _: ContractAddress =
+            inner.contract_address.try_into().map_err(|_| TxInsertionError::InvalidContractAddress)?;
+        Ok(Self { score: score_function.get_score(&inner).ok_or(TxInsertionError::NoTip)?, inner })
```
Trantorian1: Maybe we should have `get_score` return an error directly.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -49,30 +48,36 @@ pub enum MempoolError {
...
-
 #[derive(Debug, Clone)]
 pub struct MempoolConfig {
     /// Mempool limits
-    pub limits: MempoolLimits,
+    pub score_function: ScoreFunction,
+    pub max_transactions: usize,
+    pub max_declare_transactions: Option<usize>,
+    pub ttl: Duration,
     pub no_saving: bool,
```
Trantorian1: I suggest changing this to something like `save_txs`. Having to deal with the double negative of `no_saving = false` can make it a bit confusing.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -120,249 +146,229 @@ impl From<MempoolError> for SubmitTransactionError {
...
-        let tx_hash = tx.tx_hash;
         self.accept_tx(tx).await?;
-        let _ = self.tx_sender.send(tx_hash);
         Ok(())
     }
 
-    async fn received_transaction(&self, hash: mp_convert::Felt) -> Option<bool> {
-        Some(self.inner.read().await.has_transaction(&starknet_api::transaction::TransactionHash(hash)))
+    async fn received_transaction(&self, hash: Felt) -> Option<bool> {
+        Some(self.received_txs.read().expect("Poisoned lock").contains(&hash))
```
Trantorian1, line 129: Why are we using a `sync` lock in an `async` context?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -120,249 +146,229 @@ impl From<MempoolError> for SubmitTransactionError {
...
-            .map_err(SubmitTransactionError::Internal)?;
-        let res = L1HandlerTransactionResult { transaction_hash: tx.tx_hash().to_felt() };
-        self.accept_tx(ValidatedMempoolTx::from_blockifier(tx, arrived_at, converted_class)).await?;
-        Ok(res)
+        _tx: mp_transactions::L1HandlerTransaction,
+        _paid_fees_on_l1: u128,
+    ) -> Result<mp_transactions::L1HandlerTransactionResult, SubmitTransactionError> {
+        unimplemented!()
     }
 }
```
Trantorian1: Why do we still impl this then?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -120,249 +146,229 @@ impl From<MempoolError> for SubmitTransactionError {
...
+        let ret = self.inner.write().await.insert_tx(now, tx.clone(), Nonce(account_nonce), &mut removed_txs);
+        self.on_txs_removed(&removed_txs);
+        if ret.is_ok() {
+            self.on_tx_added(&tx, is_new_tx);
         }
+        ret.map_err(Into::into)
+    }
+
+    /// Update secondary state when a new transaction has been successfully added to the mempool.
+    /// Use `is_new_tx: true` when loading transactions from db, so that we skip saving in db and updating metrics.
```
Trantorian1:
```suggestion
    /// Use `is_new_tx: false` when loading transactions from db, so that we skip saving in db and updating metrics.
```

madara/crates/client/mempool/src/lib.rs
```diff
@@ -120,249 +146,229 @@ impl From<MempoolError> for SubmitTransactionError {
...
+            tracing::debug!("Removed transaction tx_hash={:#x}", tx.tx_hash);
+            lock.remove(&tx.tx_hash);
+        }
+        // TODO: tell self.tx_sender about the removal
+    }
 
-        Ok(())
+    /// Update secondary state when a new transaction has been consumed for block production.
+    fn on_tx_consumed(&self, _removed: &ValidatedMempoolTx) {
+        // nothing here yet.
```
Trantorian1: What is this planned for?

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -658,58 +657,55 @@ mod l1_messaging_tests {
...
-            .messaging_last_synced_l1_block_with_event()
-            .expect("Should successfully retrieve the last synced L1 block with messaging event")
-            .unwrap();
-        assert_ne!(last_block.block_number, 0);
-        // TODO: Assert that the transaction has been executed successfully
-        assert!(db.backend().has_l1_messaging_nonce(expected_nonce).unwrap());
-
-        // Explicitly cancel the listen task, else it would be running in the background
-        worker_handle.abort();
+        todo!();
```
Trantorian1: Waiting for #678 ?

madara/crates/client/mempool/proptest-regressions/inner/proptest.txt
Trantorian1, line 1: Please make sure to commit the new regressions file to vcs!

Trantorian1:
Awesome changes all around! :)

madara/crates/client/mempool/src/inner/accounts.rs
cchudant: true, but i wanted to separate the invariants, since this is test code we dont care about perf

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
cchudant: I think I prefer this syntax over prop_compose actually, wdyt?

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
cchudant: ..maybe ahah
I'll look into how proptest_state_machine works then

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
cchudant: crazy

madara/crates/client/mempool/src/inner/tx.rs
cchudant: So, I added a test for that but basically:
For two accounts with differing chain nonce len, higher eviction score means the one with the biggest len
For two accounts with same chain nonce len, the one with the lowest score is higher eviction score (we want to get rid of low score txs first when nonce len is matching)

Ill add a comment about that

madara/crates/client/mempool/src/inner/tx.rs
cchudant: i dont think so :(

madara/crates/client/mempool/src/lib.rs
cchudant: it's fine to use sync locks in async context if you never have a .await point while the lock is taken
this function shouldn't be async though

madara/crates/client/mempool/src/lib.rs
cchudant: so that the project compiles until #678 lands

madara/crates/client/mempool/src/lib.rs
cchudant: I'll remove it, I wanted to keep this in in case we changed our mind about the account nonces being updated only when the block production update arrives

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: yes :)

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
Trantorian1: Idk, I tend to find the `prop_compose` syntax easier to read but I guess this is down to personal preference

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
+//  datastructure.
+//
+// With this design, any mutation to the InnerMempool can follow these specific steps:
+// 1) Without modifying the InnerMempool, figure out the update you want to make and perform every pre-check needed.
+// 2) Update the backing `Accounts` datastructure with your change. Once it is updated, you are not allowed to early-return or error from this point on.
+// 3) Update all of the secondary datastructures to reflect the primary change. This is made totally generic of the specifics of the change thanks to the
+//   `AccountUpdate` struct.
+// By keeping this `AccountUpdate` generic, every invariant is local to its datastructure and this simplifies their handling.
+
+/// The in-memory datastructure containing all of the mempool transactions. Mutations to this datastructure is synchronous, via &mut referemces to ensure
```
jbcaron:
```suggestion
/// The in-memory datastructure containing all of the mempool transactions. Mutations to this datastructure are synchronous, via &mut references to ensure
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
-            *tx_counts.entry(intent.contract_address).or_insert(0) += 1;
-        }
+// Implementation details:
+//
+// The inner mempool is a complex datastructure which can be indexed in a number of ways, of which:
+// - Insertion requires indexing by a (account_address, nonce) tuple, and requires keeping track of an account nonce for every account.
+// - A ready queue, sorted by score (tip/timestamp), is needed for picking transactions for block building.
+// - A timestamp queue is needed for removing transactions that have exceeded their TTL (time-to-live).
+// - A mapping from transaction hash to transaction is required for getting the status of a transaction by hash.
+// - When inserting a transaction in a full mempool, we need to evict less desirable transaction to make space.
```
jbcaron:
```suggestion
// - When inserting a transaction in a full mempool, we need to evict less desirable transactions to make space.
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
+    // Secondary indices
+    /// Limiter: apply mempool limits.
+    limiter: MempoolLimiter,
+    /// Queue of ready accounts sorted by front tx score.
+    ready_queue: ReadyQueue,
+    /// Queue of all transactions sorted by arrived_at timestamp.
+    timestamp_queue: TimestampQueue,
+    /// Index TxHash => TxKey.
+    by_tx_hash: ByTxHashIndex,
+    /// Queue of all acounts sorted by eviction score of the the last queued tx of each account.
```
jbcaron:
```suggestion
    /// Queue of all acounts sorted by eviction score of the last queued tx of each account.
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
 
-    /// When `force` is `true`, this function should never return any error.
-    /// `update_limits` is `false` when the transaction has been removed from
-    /// the mempool in the past without updating the limits.
+    /// Insert a transaction into the mempool.
+    ///
+    /// ## Arguments
+    ///
+    /// * `now`: current time.
+    /// * `account_nonce`: the current nonce for the contract_address. Caller is responsible to get it from the backend. If the account already exists
```
jbcaron:
```suggestion
    /// * `account_nonce`: the current nonce for the contract_address. Caller is responsible for getting it from the backend. If the account already exists
```

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,860 +1,334 @@
...
+        self.apply_update(account_update, removed_txs);
+    }
 
-                // This works like a NonceMapping: if a pending intent queue is
-                // empty, we remove the mapping.
-                if queue.is_empty() {
-                    entry.remove();
-                }
+    /// Pop the next ready transaction for block building, or `None` if the mempool has no ready transaction.
+    /// This does not increment the nonce of the account, meaning the next transactions for the accounts will not be ready until a `update_account_nonce` is issued.
```
jbcaron:
```suggestion
    /// This does not increment the nonce of the account, meaning the next transactions for the accounts will not be ready until an `update_account_nonce` is issued.
```

madara/crates/client/mempool/src/inner/tx.rs
jbcaron: I don’t get why `NoTip` should be an error.

madara/crates/client/mempool/src/inner/tx.rs
```diff
@@ -1,86 +1,192 @@
...
+/// Key to index a transaction in the mempool.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TxKey(pub ContractAddress, pub Nonce);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AccountKey(pub ContractAddress);
+
+#[derive(Debug)]
+#[cfg_attr(any(test, feature = "testing"), derive(PartialEq, Eq))]
+pub struct TxInfo {
```
jbcaron: `TxInfo` is already a trait for transactions. Wouldn’t it make more sense to pick a different name for this one since it’s public, or to rename the trait?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -120,249 +146,229 @@ impl From<MempoolError> for SubmitTransactionError {
...
-                std::cmp::Ordering::Equal => Ok(NonceInfo::ready(nonce, nonce_next)),
-                std::cmp::Ordering::Greater => nonce_prev_check,
+    pub async fn run_mempool_task(&self, mut ctx: ServiceContext) -> anyhow::Result<()> {
+        self.load_txs_from_db().await.context("Loading transactions from db on mempool startup.")?;
+        let mut interval = tokio::time::interval(Duration::from_secs(5));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        loop {
+            tokio::select! {
+                _ = ctx.cancelled() => return Ok(()),
+                _ = interval.tick() => self.remove_ttl_exceeded_txs().await.context("Removing TTL-exceeded txs.")?,
```
jbcaron, line 287: I think it could be useful to make this feature optional.

jbcaron (approved):

madara/crates/client/mempool/src/lib.rs
cchudant: done :)

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/tx.rs
cchudant: renamed TxSummary

madara/crates/client/db/src/mempool_db.rs
```diff
@@ -101,13 +35,13 @@ impl MadaraBackend {
...
     #[tracing::instrument(skip(self, tx), fields(module = "MempoolDB"))]
-    pub fn save_mempool_transaction(&self, tx: &ValidatedMempoolTx, nonce_info: &NonceInfo) -> Result<()> {
+    pub fn save_mempool_transaction(&self, tx: &ValidatedMempoolTx) -> Result<()> {
         // Note: WAL is used here
         // This is because we want it to be saved even if the node crashes before the next flush
 
         let hash = tx.tx_hash;
         let col = self.db.get_column(Column::MempoolTransactions);
-        let tx_with_class = DbMempoolTxInfoEncoder { tx, nonce_info };
+        let tx_with_class = tx;
```
Mohiiit: seems like an extra variable assignment

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -705,8 +706,8 @@ mod l1_messaging_tests {
         // TODO: Assert that the transaction has been executed successfully
         assert!(db.backend().get_pending_message_to_l2(0).unwrap().is_some());
 
-        // Explicitly cancel the listen task, else it would be running in the background
-        worker_handle.abort();
+        // // Explicitly cancel the listen task, else it would be running in the background
+        // worker_handle.abort();
```
Mohiiit: hmm interesting, what's the reasoning behind this?

madara/crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -157,13 +164,17 @@ pub struct ChainConfig {
     #[serde(skip)]
     pub private_key: ZeroingPrivateKey,
 
+    #[serde(default)]
+    pub mempool_mode: MempoolMode,
+    #[serde(default)]
+    pub mempool_min_tip_bump: u128,
```
Mohiiit: interesting, is this part of the spec as well? or something from our end?

madara/crates/client/mempool/src/inner/accounts.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    /// New account status once this entry is inserted.
+    pub fn updated_data(&self) -> &AccountUpdateData {
+        &self.update_data
+    }
+
+    pub fn insert(self, tx: MempoolTransaction) -> AccountUpdate {
+        let info = tx.info();
+        assert_eq!(self.tx_key, tx.tx_key());
```
Mohiiit, line 144: I noticed we are using assert throughout the PR, I am not sure what's the best practice is but shouldn't we have proper error handling instead of assertion?

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,851 +1,338 @@
...
-            } else {
-                pending_no_age_check
-                    .push(self.tx_intent_queue_pending_by_timestamp.pop_first().expect("Already inside loop"));
+                // We made room!
+
+                // Reborrow the insertion `entry`, as `try_make_room_for` had to borrow it mutably to make its modifications.
+                entry = self
+                    .accounts
+                    .tx_entry_for_insertion(&mempool_tx, account_nonce)
+                    .expect("Insertion should not fail after making room");
```
Mohiiit, line 227: we can remove this expect right? insert txn returns Result

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -1,851 +1,338 @@
...
+                assert!(entry.replaced_tx().is_none(), "Entry should be the same");
```
Mohiiit, line 228: same as above

madara/crates/client/mempool/src/lib.rs
```diff
@@ -109,209 +93,246 @@ impl From<MempoolError> for SubmitTransactionError {
...
-            .insert_tx(
-                MempoolTransaction { tx, arrived_at, converted_class, nonce, nonce_next },
-                force,
-                /* update_limits */ true,
-                nonce_info,
-            )
-            .await?;
+        new_nonce_updates: impl IntoIterator<Item = NonceUpdate>,
+        executed_txs: impl IntoIterator<Item = Felt>,
+    ) -> anyhow::Result<()> {
```
Mohiiit, line 241: we moved away from MempoolError, what's the thought behind that?

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: nothing
although, a new runtime is created for every tokio::test so we shouldn't have to do this, but this change wasn't intentional

madara/crates/primitives/chain_config/src/chain_config.rs
cchudant: spec :) you can't micro-bump, for DoS prevention reasons
i think we want to change that to a percentage though, it should be 10% and not a fixed value I think

madara/crates/client/mempool/src/inner/accounts.rs
cchudant: asserts are only for tests and for the invariants of the mempool inner datastructure here.
My thinking is that the inner mempool is a datastructure. If its invariants are violated, it's not because another service gave it garbage data, but because the mempool itself got itself in an invalid state.
But you're right, it's debatable.

In this particular case it's a bit different I guess, it's because it's theorically possible to create the insertion entry using a tx, and then use a completely different one for the insert call. Which would kind of make no sense to use the API this way but I added this assert to be sure. It's different IMO because this is a panic indicating a wrong API usage, and not an unexpected invalid state. (This api is only used internally within the inner mempool datastructure)

cchudant: Also: this code is prop-tested. This is the main thing that gives us confidence that the invalid states are actually not possible to reach during normal execution, and all panics are unreachable in prod

madara/crates/client/mempool/src/inner/mod.rs
cchudant: true but if it should never be an error, if it's an error it's probably because the mempool has entered an invalid state

madara/crates/client/mempool/src/lib.rs
cchudant: I think we should just use anyhow::Result for any internal error, it's simpler and I don't think thiserror makes sense for internal errors
we don't really want the caller to `match` on them anyway, since it's the kind of error that should stop the node. using anyhow the caller can also use `.context` to add context to the error, and the error will also have a backtrace which is neat.

madara/crates/client/db/src/mempool_db.rs
cchudant: fixed :)

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: fixed :)

madara/crates/primitives/chain_config/src/chain_config.rs
cchudant: fixed :) it's now percentage based, with the default value being 10%
idk how I missed that x) but 10% is what was written in the v0.14 change document IIRC

madara/crates/client/mempool/src/inner/accounts.rs
cchudant: added a comment about this panic

madara/crates/client/mempool/src/inner/accounts.rs
cchudant: fixed :)

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/mod.rs

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
cchudant: i don't think you can prop compose a prop oneof like that though

madara/crates/client/mempool/src/inner/tests/mempool_proptest.rs
cchudant: hmm it seems you can't :(

madara/crates/client/mempool/src/inner/tx.rs
cchudant: added :)

madara/crates/client/mempool/src/inner/tx.rs
cchudant: NoTip is not an error anymore :) by default, tip will be 0.

madara/crates/client/mempool/src/lib.rs
cchudant: fixed :)

madara/crates/client/mempool/src/lib.rs
cchudant: oh it's a trait function, yea this one is expected to be async sorry

madara/crates/client/mempool/src/lib.rs
cchudant: it landed :)

madara/crates/client/mempool/src/lib.rs
cchudant: it's fixed

madara/crates/client/settlement_client/src/eth/mod.rs
cchudant: it has landed :)

madara/crates/client/mempool/proptest-regressions/inner/proptest.txt
cchudant: i did not get any regression :)

Mohiiit (approved):

PR #704, opened 2025-06-30
Bootstrapper 14.0 structuring PR

byteZorvin:
## Pull Request type

- Feature

- Add initial structuring code for Bootstrapper v2

Resolves: #NA

No


bootstrapper-v2/configs/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "base_layer": {
+    "layer": "Ethereum",
```
apoorvsadana: should we use all upper case for this? (i usually prefer screaming snake case but that's same as uppercase if it's just one word)

bootstrapper-v2/configs/config.json
```diff
@@ -0,0 +1,10 @@
...
+    "rpc_url": "http://localhost:8545",
+    "implementation_address": {}
```
apoorvsadana:
```suggestion
    "implementation_addresses": {}
```

bootstrapper-v2/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn main() -> Result<()> {
+    color_eyre::install()?;
+
+    dotenvy::from_filename_override("../.env")?;
+    let args = CliArgs::parse();
+
+    match args.command {
+        Commands::SetupBase(setup_base) => {
+            let config: Config = serde_json::from_reader(File::open(setup_base.config_path)?)?;
```
apoorvsadana: isn't it better to deserialize directly to base config here (and same for Madara)? that way, you don't need to check for optional stuff

bootstrapper-v2/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+            config.validate()?;
+
+            let private_key = dotenvy::var("BASE_LAYER_PRIVATE_KEY")?;
```
apoorvsadana: clap will read the private key already and make it available inside `setup_base`, you don't need to read from env again

bootstrapper-v2/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+            let base_layer_setup =
+                config.get_base_layer_setup(private_key).map_err(|e| color_eyre::eyre::eyre!("{}", e))?;
```
apoorvsadana: you can make main return a color_eyre Result and then just do `?` instead of `map_err`. ideally we should throw custom errors wherever possible but because bootstrapper is a script we can skip enum errors?

however, use `.context` from color_eyre to add context to the error, like "Failed to setup base layer"

bootstrapper-v2/src/setup/base_layer/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use anyhow::Result;
+
+#[allow(dead_code)]
+pub trait BaseLayerSetupTrait {
+    // This function does prerequisite setup for running the base layer setup.
+    // It should be called before the base layer setup.
+    fn init(&self) -> Result<()>;
+    fn setup(&self) -> Result<()>;
+    fn post_madara_setup(&self) -> Result<()>;
```
apoorvsadana: let's add doc for all functions.

doc starts with `///` and comments start with `//`

bootstrapper-v2/configs/config.json
byteZorvin: Changed. I am not aware of the standard practices here

bootstrapper-v2/src/main.rs
byteZorvin: Thanks for pointing this out, 100% agreed

bootstrapper-v2/src/main.rs
byteZorvin: Got it, updated

bootstrapper-v2/configs/config.json
byteZorvin: Applied

bootstrapper-v2/src/main.rs
byteZorvin: Will add few Concrete enum errors, as I write the implementations.
Added context to errors for now

apoorvsadana (approved):

bootstrapper-v2/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn main() -> Result<()> {
+    color_eyre::install()?;
+
+    dotenvy::from_filename_override("../.env")?;
+    let args = CliArgs::parse();
+
+    match args.command {
+        Commands::SetupBase(setup_base) => {
+            let config: BaseConfigOuter = serde_json::from_reader(File::open(setup_base.config_path)?)?;
```
0xvasanth, line 22: Hey @byteZorvin, this is interesting! You might want to check out the [Cargo config documentation](https://doc.rust-lang.org/cargo/reference/config.html) — it provides a flexible and idiomatic way to handle configurations in Rust.

We’re planning to use a similar approach in the orchestrator as well.

Why I recommend this config format:
	• Out-of-the-box extensibility
	• Built-in linker support
	• Environment-specific overrides (great for local/dev/prod setup)
	• Clean separation of code and configuration
	• Easy to compose and merge configs from multiple sources

Let me know your thoughts!

bootstrapper-v2/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+pub mod config;
+mod setup {
+    pub mod base_layer;
+    pub mod madara;
+}
+
+use clap::Parser;
+use cli::{CliArgs, Commands};
+
+use color_eyre::{eyre::WrapErr, Result};
```
0xvasanth: Since we’re reworking this module, I recommend switching to anyhow here. As part of the Karnot team, we aligned with Kasar to standardize on using anyhow across all subcrates. This helps eliminate redundancy, as both anyhow and color_eyre serve similar purposes. Keeping just anyhow simplifies our dependency footprint and aligns with our shared convention.

bootstrapper-v2/src/main.rs
byteZorvin: I would say this just another format to store config(in toml) instead of json. I agree that toml is bit better although I couldnt find any built in support for it.
Instead of serde_json we would have to use serde_toml/toml
For the environment specific overrides, the decisions on this would remain irrespective of toml or json being used.

bootstrapper-v2/src/main.rs
byteZorvin: Makes sense, I went initially with color_eyre for its colored well formatted output as this is CLI application. Although no strong choices, updated to anyhow

Makefile
```diff
@@ -2,6 +2,8 @@
 #                              STARKNET NODE RUNNER                            #
 # ============================================================================ #
 
+SHELL := /bin/zsh
```
0xvasanth: Just a minor suggestion — explicitly setting the shell to /bin/zsh could lead to issues for developers who don’t have zsh installed or use a different default shell. Would it be more portable to default to bash or detect the shell dynamically?

0xvasanth (approved):
LGFM!

build-artifacts/cairo_artifacts/openzeppelin_AccountUpgradeable.contract_class.json
Trantorian1, line 1: Is there no way to store these artifacts outside the repo? We could just create a new `artifacts` repo inside of the Madara org, upload artifacts there and clone that in.

Trantorian1 (approved):
lgtm!

PR #708, opened 2025-07-01
fix: fetch_l1_to_l2_messaging_events method to fetch with retry

prkpndy:
Update the fetch l1 to l2 message polling logic to add retries

## Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Currently, the application panic if it fails to fetch events from the RPC even once

Resolves: #NA

## What is the new behavior?

Now, if it fails, it'll try to refetch the events `MAX_RETRIES` number of times waiting for `RETRY_TIMEOUT` seconds after every failure.


Mohiiit (commented):
can we add a test for this as well? which insures that it's retrying 3 times before failing and stuff? we can create a mock server for this I think!

wdyt?

heemankv (approved):
LGTM !, Lets just add the tests as @Mohiiit has mentioned

jbcaron:
I think using a middleware like `tower` directly on the client would be a good approach. This allows us to have a timeout layer, an error handling layer (rate limit, bad request, etc.), and a retry layer. For example, with the current implementation, wouldn’t we want to retry more than 3 times if it’s a too_many_requests error? Do we want to retry more quickly if the response is fast?
WDYT?

prkpndy:
> I think using a middleware like `tower` directly on the client would be a good approach. This allows us to have a timeout layer, an error handling layer (rate limit, bad request, etc.), and a retry layer. For example, with the current implementation, wouldn’t we want to retry more than 3 times if it’s a too_many_requests error? Do we want to retry more quickly if the response is fast? WDYT?

Yeah. Makes sense.
The way I am thinking, it'll require changes changes at a higher level of the application, such that the whole application would be using it. Probably will require changes at other places as well (tests is one that I can think of right now). Should I make these changes in the current PR only?

Mohiiit (approved):

jbcaron (approved):

PR #714, opened 2025-07-05
Eth factory

byteZorvin:
## Pull Request type

- Feature

- No


jbcaron:
#643

.gitmodules
Trantorian1, line 1: > [!WARNING]
> Why does this PR re-add submodules to the repo? These were explicitely removed as part of #643 in favor of a dockerized build to avoid relying on too many external deps.

bootstrapper-v2/contracts/ethereum/.github/workflows/test.yml
Trantorian1, line 1: > [!NOTE]
> What is this workflow used for? Please move it to the root-level `.github` to maintain consistency.

bootstrapper-v2/contracts/ethereum/lib/forge-std
Trantorian1, line 1: > [!CAUTION]
> Please, please, please move this into the [build docker](https://github.com/madara-alliance/madara/blob/main/build-artifacts/build.docker). This is precisely the kind of dependency hell which we were trying to solve by moving external artifacts to a docker build.

bootstrapper-v2/contracts/ethereum/script/Counter.s.sol
Trantorian1, line 1: > [!WARNING]
> Why are these contracts not part of the root `build-artifacts`?

e2e-tests/artifacts/contracts/Starknet.json
Trantorian1, line 1: > [!CAUTION]
> Please do not store artifacts in the git repo proper as they pollute the git tree irreversibly. Artifacts should instead be built at compile-time or downloaded from a remote cache when possible.

Trantorian1 (changes requested):

.gitignore
```diff
@@ -1,14 +1,25 @@
 .env*
```
byteZorvin, line 1: Theese changes were done to address issues faced with .gitignore in the build-artifacts directory.
From git documentation. Ignoring first the build-artifacts/ would cause this issue

byteZorvin: image

Makefile
```diff
@@ -199,14 +203,20 @@ artifacts:
...
 	@rm -rf "$(ARTIFACTS)/js_tests"
 	@rm -rf "$(ARTIFACTS)/orchestrator_tests"
 	@rm -rf "$(ARTIFACTS)/starkgate_latest"
 	@rm -rf "$(ARTIFACTS)/starkgate_legacy"
-	@docker build -f $(ARTIFACTS)/build.docker -t contracts .
-	@ID=$$(docker create contracts do-nothing) && docker cp $${ID}:/artifacts/. $(ARTIFACTS) && docker rm $${ID} > /dev/null
+	@rm -rf "$(BOOTSTRAPPER_DIR)/ethereum/out"
+	@rm -rf "$(BOOTSTRAPPER_DIR)/ethereum/src/starkware"
+	@rm -rf "$(BOOTSTRAPPER_DIR)/ethereum/src/third_party"
+	@docker build --platform=linux/amd64 -f $(ARTIFACTS)/build.docker -t contracts .
```
byteZorvin, line 211: Added `platform` to support macos rossetta error

.dockerignore
```diff
@@ -6,3 +6,9 @@ scripts
 target/
 *Dockerfile
 .npm-global
+
+# Bootstrapper
+bootstrapper-v2/contracts/ethereum/cache/
+bootstrapper-v2/contracts/ethereum/.vscode/
+bootstrapper-v2/contracts/ethereum/.github/
+bootstrapper-v2/contracts/ethereum/out/
```
apoorvsadana: `bootstrapper-v2/contracts/*` doesn't work? also, we don't have `build-artifacts` in this ignore?

.github/workflows/task-build-artifacts.yml
```diff
@@ -58,8 +58,12 @@ jobs:
...
             --exclude cairo_artifacts \
             --exclude Cargo.toml \
-            --exclude src \
-            -czvf artifacts.tar.gz build-artifacts/
+            --exclude build-artifacts/src \
+            -czvf artifacts.tar.gz \
+            build-artifacts/ \
+            bootstrapper-v2/contracts/ethereum/out/ \                     # TODO: to change to bootstrapper/ post v2 release of bootstrapper
+            bootstrapper-v2/contracts/ethereum/src/starkware/ \
+            bootstrapper-v2/contracts/ethereum/src/third_party/
```
apoorvsadana: not needed as we will put all artifacts inside the artifacts folder itself as discussed

Makefile
```diff
@@ -99,6 +99,7 @@ DOCKER_TAG     := madara:latest
 DOCKER_IMAGE   := ghcr.io/madara-alliance/$(DOCKER_TAG)
 DOCKER_GZ      := image.tar.gz
 ARTIFACTS      := ./build-artifacts
+BOOTSTRAPPER_DIR := ./bootstrapper-v2/contracts
```
apoorvsadana: changes to artifacts folder as discussed offline

bootstrapper-v2/contracts/ethereum/.github/workflows/test.yml
apoorvsadana, line 1: Do we need this now? Also, this seems to run on the root level instead of the bootstrapper level.

bootstrapper-v2/contracts/ethereum/README.md
apoorvsadana, line 1: can probably delete this or change it to be specific to our code

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+    // In case of Starknet.sol the initData looks like
+    //
+    // (
+    //    uint256 programHash_,
+    //    uint256 aggregatorProgramHash_,
+    //    address verifier_,
+    //    uint256 configHash_,
+    //    StarknetState.State memory initialState
+    // ) = abi.decode(data, (uint256, uint256, address, uint256, StarknetState.State));
+    bytes memory initData = abi.encode(address(0), coreContractInitData);
```
apoorvsadana: maybe we should call this `upgradeToData`? i thought this `initData` is same as the one on like 89 (`[sub_contracts addresses, eic address, initData]`) but I don't think it is. This is the whole array and sub contracts are always 0?

Also a more elaborate comment on subcontract part will help. Maybe have function called `buildUpgradeToData`, call that from everywhere and add a comment on the function on why we use `address(0)`

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+    ProxySetup.addImplementationAndUpgrade(
+      address(coreContractProxy),
+      coreContractImplementation,
+      initData
+    );
+
+    IOperator(address(coreContractProxy)).registerOperator(operator);
+    IGovernor(address(coreContractProxy)).starknetNominateNewGovernor(governor);
+    IProxyRoles(address(coreContractProxy)).registerGovernanceAdmin(governor);
```
apoorvsadana, line 109: I noticed in Rust code, we also nominate the governor for the proxy contract itself. Do we need to do that here as well?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+
+    return address(coreContractProxy);
+  }
+
+  function deployManagerRegistry() public returns (address, address) {
```
apoorvsadana:
```suggestion
  function deployManagerAndRegistry() public returns (address, address) {
```

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana, line 1: Overall, this is so much more readable than the current bootstrapper 👏

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  function setupManager(
+    address managerProxy,
+    address managerImplementation,
+    address bridgeProxy,
+    address registryProxy,
+    address governor
+  ) public whenNotPaused {
+    bytes memory initData = abi.encode(address(0), bridgeProxy, registryProxy);
```
apoorvsadana: [is it registry first and then bridge](https://github.com/madara-alliance/madara/blob/c063107fec9d8b45bf7878ec277cff6e1e1a2718/bootstrapper/src/contract_clients/token_bridge.rs#L163)

bootstrapper-v2/contracts/ethereum/src/factory/Implementations.sol
apoorvsadana, line 1: is this being used anywhere?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+
+  function registerAdmins(
+    address proxyContract,
+    address governanceAdmin
+  ) private {
+    IRoles(proxyContract).registerGovernanceAdmin(governanceAdmin);
+    IRoles(proxyContract).registerSecurityAdmin(governanceAdmin);
+    IRoles(proxyContract).registerUpgradeGovernor(governanceAdmin);
+    IRoles(proxyContract).registerAppRoleAdmin(governanceAdmin);
+    IRoles(proxyContract).registerAppGovernor(governanceAdmin);
```
apoorvsadana, line 228: upgrade governor?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -0,0 +1,216 @@
...
+
+  function deployManagerRegistry() public returns (address, address) {
+    _requireNotPaused();
+    Proxy managerProxy = new Proxy(0);
+    Proxy registryProxy = new Proxy(0);
+
+    return (address(managerProxy), address(registryProxy));
+  }
+
+  function setupMultiBridge(
```
apoorvsadana, line 122: what about the function to set values on the bridge (once L2 is deployed). that will be in a separate PR?

bootstrapper-v2/contracts/ethereum/src/factory/Implementations.sol
byteZorvin: Yes, the `Factory.sol` inherits this and uses the storage.
This pattern is commonly used to separate the storage in a separate abstract contract

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: Agreed, lets go 🚀

.dockerignore
byteZorvin: It wont work as we want to include the src/ folders inside ethereum/

.github/workflows/task-build-artifacts.yml
byteZorvin: Updated

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: What do you mean by the proxy contract itself? The governors set in proxy storage will be used for the implementation. There is just one place(contract storage) where you set the roles as implementation storage is never touched

byteZorvin: Also I have not nominated the upgrade governor anywhere as initialise also does not during the constructor call for msg.sender(Factory contract).

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: It can be later nominated by governance admin. I have not done it here as the initialise call also does not do it, and tried to only replicate it

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: It will be just 2 calls, so I was thinking to do it in scripts. On another note if we do it through the factory contract, we would have to either delay renouncing the roles from the factory contract till then(as calling `set_l2_bridge` needs appGovernor role) or add and then remove the call

.github/workflows/task-build-artifacts.yml
```diff
@@ -58,8 +58,9 @@ jobs:
             --exclude build.docker \
             --exclude cairo_artifacts \
             --exclude Cargo.toml \
-            --exclude src \
-            -czvf artifacts.tar.gz build-artifacts/
+            --exclude build-artifacts/src \
```
apoorvsadana: we went from `--exclude src` to `build-artifacts/src`?

Makefile
apoorvsadana: what is the error exactly? the docker won't run on my local arm machine if I build it on amd right?

bootstrapper-v2/build.rs
```diff
@@ -0,0 +1,30 @@
...
+        ("../build-artifacts/bootstrapper/solidity/out", "./contracts/ethereum/out"),
+    ];
+
+    for (src, dst) in copy_dirs.iter() {
+        let src_path = Path::new(src);
+        let dst_path = Path::new(dst);
+
+        // Skipping if the src_path does not exist
+        // Run `make artifacts` before to ensure artifacts are present
+        if src_path.exists() {
```
apoorvsadana, line 23: shouldn't we log an error if it doesn't exist?

bootstrapper-v2/build.rs
```diff
@@ -0,0 +1,30 @@
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    build_version::get_or_compile_artifacts(1).expect("Failed to load artifacts");
+
+    // Copy build artifacts to the appropriate locations
+    let copy_dirs = [
+        ("../build-artifacts/bootstrapper/solidity/starkware", "./contracts/ethereum/src/starkware"),
```
apoorvsadana: `build-artifacts/bootstrapper/solidity` and `/contracts/ethereum` can be a constant

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: So if you check [this](https://github.com/madara-alliance/madara/blob/67a2ac59af4c2d98216b8fb2ef1e538020cbf462/bootstrapper/src/setup_scripts/core_contract.rs#L91-L96).

It has a method called `nominate_governor_core_contract_proxy` which actually calls `register_app_governor`. Maybe its worded incorrectly now that I think of it. But we aren't calling that, probably governance admin gives master control anyways?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: If its two transactions, I think one function in the factory would be better. If its one transaction, then script works too.

Regarding the renouncing part, even if you do a script, you won't be able to renounce because script will work with a hot wallet key and app governon will be a multisig?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: It does here [right](https://github.com/starknet-io/starkgate-contracts/blob/5a10fd263d29cd032b7229691d043520edae0737/src/cairo/token_bridge.cairo#L959)

build-artifacts/build.docker
```diff
@@ -210,6 +213,20 @@ RUN mkdir scarb && wget -c $(cat /tmp/scarb_url) -O - | tar -xz -C scarb --strip
...
 RUN ./scarb/bin/scarb build
 
+# =========================================================================== #
+#                                  BOOTSTRAPPER                               #
+# =========================================================================== #
+
+FROM ghcr.io/foundry-rs/foundry AS bootstrapper
+WORKDIR /contracts
+
+ENV STARKGATE_DIR="/contracts/starkgate-contracts"
```
apoorvsadana, line 223: this env `STARKGATE_DIR` is defined again in artifacts below, anyway to reuse it?

build-artifacts/build.docker
```diff
@@ -236,6 +254,11 @@ COPY --from=starkgate-latest $ARTIFACTS_CAIRO/ERC20.casm ERC20_070.casm.json
 COPY --from=starkgate-latest $ARTIFACTS_SOLID/StarknetEthBridge.json eth_bridge_upgraded.json
 COPY --from=starkgate-latest $ARTIFACTS_SOLID/StarkgateUpgradeAssistExternalInitializer.json eic_eth_bridge.json
 
+WORKDIR ../bootstrapper
+COPY --from=starkgate-latest $STARKGATE_DIR/starkware ./solidity/starkware/
+COPY --from=starkgate-latest $STARKGATE_DIR/third_party ./solidity/third_party/
+COPY --from=bootstrapper "/contracts/out" ./solidity/out/
```
apoorvsadana: `/contracts/out` should be BOOTSTRAPPER_DIR?

.github/workflows/task-build-artifacts.yml
byteZorvin: Oh nice catch, its not needed now so have removed it. Before had other `src` deeper inside `build-artifacts` which were not to be ignored, but after we decided we will copy artifacts separately this is unnecessary.

build-artifacts/build.docker
byteZorvin: Ah idts, this is the directory of the previous stage from which we are copying. `BOOTSTRAPPER_DIR` should be the final directory for boostrapper aritfacts in `./build-artifacts`, like `WORKDIR ../bootstrapper` here

build-artifacts/build.docker
byteZorvin: ENV and ARG vars are scoped per stage in multi stage builds. So I dont see a way

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: This is only setting the admin and not granting them, but as discussed f2f have added the upgrade governor role as well.

Makefile
byteZorvin: It should now, after this. I am able to post adding the above fix

bootstrapper-v2/build.rs
byteZorvin: Oh yes, updated

bootstrapper-v2/build.rs
byteZorvin: CI would fail to build this, untill the solidity artifacts are published on github, which will be done post the merge only

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: Updated with a new function

bootstrapper-v2/build.rs
apoorvsadana: ok, maybe we print a warning then? or add a TODO to throw an error later.

build-artifacts/build.docker
apoorvsadana: Got it but do we need to hardcode `contracts` or should we use the relative path from `BASE_CHAIN_FACTORY_DIR` or create another env like for all other paths?

Makefile
apoorvsadana: I didn't understand. How will the docker work if its being built for the amd platform?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: https://github.com/madara-alliance/madara/pull/686/files#r2261138543

build-artifacts/build.docker
byteZorvin: Updated

Mohiiit (approved):

PR #716, opened 2025-07-07
fix: prettier check

byteZorvin:
## Pull Request type

- Bugfix

- The github code action does not use the same prettier version as from the local package.json
- This might cause different styles being used locally and during action

## What is the new behavior?
- Installs the same version of prettier as from package.json

No

### DIFF
--- a/.github/workflows/task-lint-code-style.yml
+++ b/.github/workflows/task-lint-code-style.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run prettier
-        run: npx prettier --check .
+        run: npm install && npx prettier --check .
 
   # Check markdown files for style consistency
   markdown-lint:


Mohiiit (approved):

.github/workflows/task-lint-code-style.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run prettier
-        run: npx prettier --check .
+        run: npm install && npx prettier --check .
```
0xvasanth, line 25: Nitpick: If version locking via package.json or local installation isn’t required, you could simplify this to 
`npx prettier@3.5.3 --check . `
Since npx can fetch the specified version directly, the separate npm install step may not be necessary here.

byteZorvin: This would be hardcoding the version, instead I would prefer that there is single source of the version directly from package.json present at madara root.

npx does not check the package.json version to install. Although if you do npm install first it will use the locally installed version [here](https://arc.net/l/quote/xcfkwxol)

wdyt?

0xvasanth (approved):

PR #719, opened 2025-07-10
fix: gateway url

cchudant:
v0.14 sepolia is live today, we don't support it yet but update the gateway URLs so that our CI continues to pass.

## Pull Request type

- Bugfix

No


madara/test.sh
Mohiiit, line 1: I think this is being added by mistake!

Mohiiit (approved):

PR #720, opened 2025-07-12
Optimize: base chain requests

Mohiiit:
We were hitting more than expected requests on the base chain for data retrieval (mainly events), this is mainly happening for the starknet (currently we have chains running as L3 only tho, so this is expected) we will have information on the number of requests for ethereum when we run one instance of that for a while!

## Pull Request type

Madara is hitting a lot of requests which is expensive for the user, given they will be paying for the l1 endpoint!

Resolves: #NA

## What is the new behavior?

We are limiting the number of requests by using good enough buffer and sleep intervals!

No


madara/crates/client/settlement_client/src/starknet/mod.rs
```diff
@@ -79,6 +79,11 @@ impl StarknetClient {
     }
 }
 
+const POLL_INTERVAL: Duration = Duration::from_secs(5); // Interval between event polling attempts
+const EVENT_SEARCH_BLOCK_RANGE: u64 = 6000; // Number of blocks to search backwards for l1->l2 events
+const STATE_UPDATE_EVENT_SEARCH_BLOCK_RANGE: u64 = 1000;// Number of blocks to search backwards for state update events
```
apoorvsadana, line 84: Should these be a part of `StarknetClientConfig`? I can imagine we set these values from Cli at some time although can be pushed to another PR when needed.

madara/crates/client/settlement_client/src/starknet/mod.rs
```diff
@@ -79,6 +79,11 @@ impl StarknetClient {
     }
 }
 
+const POLL_INTERVAL: Duration = Duration::from_secs(5); // Interval between event polling attempts
```
apoorvsadana, line 82:
```suggestion
const STATE_UPDATE_POLL_INTERVAL: Duration = Duration::from_secs(5); // Interval between event polling attempts
```

Mohiiit: this is also being used for message to L2 events as well

madara/crates/client/settlement_client/src/starknet/mod.rs
Mohiiit: surely! https://github.com/madara-alliance/madara/issues/738

apoorvsadana (approved):

madara/crates/client/settlement_client/src/starknet/mod.rs
jbcaron: Shouldn't this be dynamic and based on time, in order to find the corresponding block_n for a past timestamp? This would allow it to adapt to the chain regardless of the block time.

jbcaron (approved):

madara/crates/client/settlement_client/src/starknet/mod.rs
Mohiiit: that's true, but I am assuming in case of appchains the block_n itself could change

PR #722, opened 2025-07-15
feat: implement self-healing mechanism for orphaned jobs

0xvasanth:
This prevents jobs from being permanently stuck when the application crashes during processing, enabling automatic recovery without manual intervention.

## Pull Request type

- Bugfix

Currently, when the orchestrator application crashes or stops unexpectedly while processing jobs, those jobs remain permanently stuck in `LockedForProcessing` status. This creates several problems:

- **Orphaned jobs**: Jobs cannot be picked up by any worker since they're locked
- **Processing delays**: New jobs for the same block cannot proceed until stuck jobs are manually resolved
- **System degradation**: Accumulation of stuck jobs can impact overall system performance

## What is the new behavior?

- Add job processing timeout configuration (default: 30 minutes)
- Implement database method to find jobs stuck in LockedForProcessing status
- Integrate self-healing into all job triggers (SNOS, Proving, ProofRegistration, DataSubmission, StateTransition)
- Add backup mechanism to auto_heal_job collection for audit trail
- Reset orphaned jobs from LockedForProcessing to Created status
- Include comprehensive logging and error handling

Note: This is a straightforward solution to address the immediate issue. That said, we have a broader vision for handling lock-based problems more robustly. You can find more details about the long-term approach in [this issue](https://github.com/madara-alliance/madara/issues/737).


Mohiiit (approved):

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -963,6 +963,40 @@ impl DatabaseClient for MongoDbClient {
...
+        let start = Instant::now();
+
+        // Calculate the cutoff time (current time - timeout)
+        let cutoff_time = Utc::now() - chrono::Duration::seconds(timeout_seconds as i64);
+
+        // Query for jobs of specific type in LockedForProcessing status with process_started_at older than cutoff
+        let filter = doc! {
+            "job_type": mongodb::bson::to_bson(job_type)?,
+            "status": mongodb::bson::to_bson(&JobStatus::LockedForProcessing)?,
+            "metadata.common.process_started_at": {
```
jbcaron, line 978: Does this correctly retrieve the jobs before process_started_at? Because I don’t find it very clear.

jbcaron (approved):

orchestrator/src/core/client/database/mongodb.rs
0xvasanth: Hey @jbcaron,Apologies for missing this comment before merging

To clarify: yes, the query is working as intended. The process_started_at timestamp is initialized at the moment a job transitions into the LockedForProcessing state, so it accurately reflects the time the job was locked for processing. Since we’re filtering specifically by that state and using this timestamp, it ensures the correct jobs are retrieved. This behavior has also been tested and verified.

PR #723, opened 2025-07-16
E2E : 2 : Init + Server

heemankv:
## Pull Request type

- Feature
- Testing

Resolves: #NA

## What is the new behavior?

- Created a new `e2e` crate for end-to-end testing infrastructure
- Implemented a generic `Server` struct to manage test service processes
- Added RPC helper methods for interacting with Starknet nodes
- Set up error handling and logging for test services

No

## Other information

This PR adds a new e2e testing library that can be used across the project. It includes:

- Process management for starting/stopping test services
- TCP connection validation to ensure services are ready
- Stdout/stderr capture with proper logging
- Starknet RPC methods for block queries
- Proper error handling with custom error types

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "e2e"
+version = "0.1.0"
+edition = "2021"
+authors = ["Heemank Verma <heemankv@gmail.com>"]
+
+[lib]
+name = "e2e"
+path = "src/mod.rs"
+
+[dependencies]
+
+async-trait = { workspace = true }
+
+dotenvy = { workspace = true }
+env_logger = { workspace = true }
+httpmock = { version = "0.8.0-alpha.1", features = ["proxy", "remote"] }
+log = { workspace = true }
+thiserror = { workspace = true }
+reqwest = { workspace = true, features = ["json"] }
+rstest = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
+testcontainers = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
+url = { workspace = true }
+uuid = { workspace = true }
--- a/e2e/src/services/helpers.rs
+++ b/e2e/src/services/helpers.rs
@@ -1,0 +1,70 @@
+use url::Url;
+use serde_json::json;
+use async_trait::async_trait;
+
+#[derive(Debug, thiserror::Error)]
+pub enum NodeRpcError{
+    #[error("Invalid response")]
+    InvalidResponse,
+}
+
+#[async_trait]
+pub trait NodeRpcMethods: Send + Sync {
+    fn get_endpoint(&self) -> Url;
+
+    /// Fetches the latest block number from the Starknet RPC endpoint.
+    ///
+    /// Returns:
+    /// - `Ok(-1)` when no blocks have been mined yet (RPC returns "Block not found" error with code 24)
+    /// - `Ok(block_number)` when blocks exist
+    /// - `Err(NodeRpcError::InvalidResponse)` for other RPC errors or parsing failures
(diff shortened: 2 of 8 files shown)


e2e/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "e2e"
+version = "0.1.0"
+edition = "2021"
+authors = ["Heemank Verma <heemankv@gmail.com>"]
```
Mohiiit, line 5:
```suggestion
[package]
name = "e2e"
description = ""
authors.workspace = true
edition.workspace = true
license.workspace = true
repository.workspace = true
version.workspace = true
homepage.workspace = true
```

e2e/src/services/server.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// Wait until the server is ready to accept connections
+    async fn wait_till_started(&mut self) -> Result<(), ServerError> {
+        let mut attempts = self.config.connection_attempts;
+        let addr = self.endpoint();
+
+        if let Some(addr) = addr {
+
+            // Extract just the socket address from the URL
+            let socket_addr = if let Some(host) = addr.host_str() {
+                let port = addr.port().unwrap_or(80); // Default to 80 if no port
```
Mohiiit, line 203: let's move this to const above! and does it assign default port 80 for every service?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,267 @@
...
+                        }
+
+                        if attempts == 0 {
+                            return Err(ServerError::ConnectionTimeout(self.config.connection_attempts));
+                        }
+                    }
+                }
+
+                attempts -= 1;
+                tokio::time::sleep(Duration::from_millis(self.config.connection_delay_ms)).await;
```
Mohiiit, line 225: I think it's cleaner to have config.connection_delay_ms as Duration type, wdyt?

e2e/src/services/helpers.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Fetches the latest block number from the Starknet RPC endpoint.
+    ///
+    /// Returns:
+    /// - `Ok(-1)` when no blocks have been mined yet (RPC returns "Block not found" error with code 24)
+    /// - `Ok(block_number)` when blocks exist
+    /// - `Err(NodeRpcError::InvalidResponse)` for other RPC errors or parsing failures
+    ///
+    /// Note: The -1 return value indicates that the blockchain is in an initial state
+    /// with no blocks mined, which is common when first starting a node or testnet.
+    async fn get_latest_block_number(&self) -> Result<i64, NodeRpcError> {
```
Mohiiit, line 24: interesting choice of i64, why not Option<u64>? Because in orch and madara we do so

e2e/src/services/server.rs
```diff
@@ -0,0 +1,267 @@
...
+            command.stdout(Stdio::piped());
+
+            let stdout = process.stdout.take().ok_or(ServerError::StartupFailed(
+                std::io::Error::new(std::io::ErrorKind::Other, "Failed to capture stdout")
+            ))?;
+
+            let service_name = config.service_name.clone();
+            let stdout_task_inner = task::spawn(async move {
+                // Keeping a large buffer capacity for stdout, to not have buffer overflow
+                let reader = BufReader::with_capacity(65536, stdout);
```
Mohiiit, line 99: is this capacity random? let's move the const to the top

e2e/src/services/server.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub fn is_running(&mut self) -> bool {
+       self.has_exited().is_none()
+    }
+
+    /// Get a free port
+    fn get_free_port() -> u16 {
+        std::net::TcpListener::bind("127.0.0.1:0")
+            .and_then(|listener| listener.local_addr())
+            .map(|addr| addr.port())
+            .unwrap_or(8080) // Fallback port
```
Mohiiit, line 191: same question here, for different services, it will return the same default port? shall we have a util function which finds a free port and assign that?

PR #730, opened 2025-07-16
Better node logging

cchudant:
Upstream the old quaza logging, I forgot to do that for the longest time. Some of the OTEL fields have been changed a tiny bit, I may revert that if it's a problem.

It looks like this:
Screenshot 2025-07-16 at 14 15 28

- time at the end of RPC lines is now correct
- logs gateway calls too
- better format (nicer imo)

Target info has been added to DEBUG entries (this was the main reason for me touching these parts of the code, the fact that `target` log field was never shown for debug logs really made debugging stuff a lot harder) 

There is also a prometheus endpoint mode that i had implemented, it's in this PR too.

The new format is also nice when running e2e tests with multiple madara nodes:
`cargo run -p 'mc-e2e-tests' --nocapture transaction_flow::declare_sierra_then_deploy::case_6_
gateway_and_sequencer_no_pending`
Screenshot 2025-07-16 at 14 17 43

here a gateway node and sequencer output, it's way easier to understand what's going on


Mohiiit (approved):

antiyro (approved):
lgtm!

PR #733, opened 2025-07-17
feat: add transaction finality validation in StateUpdateJobHandler

0xvasanth:
## Pull Request type

- enhancement
- Build-related changes

State update transactions are submitted without waiting for finality confirmation, which could lead to issues if transactions fail or are reverted.

Resolves: #NA

## What is the new behavior?

- Added transaction finality check after submitting state updates to ensure transactions are confirmed
- Extracted hardcoded values as constants for better maintainability
- Simplified block processing logic by removing layer-specific hardcoded limits

No


orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -111,16 +110,7 @@ impl JobTrigger for UpdateStateJobTrigger {
...
-
-        // TODO: Remove this once we have a proper way to handle L3 blocks with use of receipt
-        let max_blocks = match config.layer() {
-            Layer::L2 => 10,
-            Layer::L3 => 1,
-        };
-        if blocks_to_process.len() >= max_blocks {
-            blocks_to_process = blocks_to_process.into_iter().take(max_blocks).collect();
-        }
+        let blocks_to_process = find_successive_blocks_in_vector(blocks_to_process, Some(10));
```
Mohiiit: the limit would be 10 for L3s as well?

0xvasanth: Yes ser, since we are checking for finality possibility of nounce issue is zero so we are having 10 state update at a time

Mohiiit (approved):

orchestrator/src/worker/event_handler/triggers/update_state.rs
heemankv: SInce we are refactoring it, it makes sense to move this 10 to the constants.rs file!

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -199,11 +189,13 @@ impl JobTrigger for UpdateStateJobTrigger {
...
 /// Eg : input_vec : [1,2,3,4,7,8,9,11]
 /// We will take the first 4 block numbers and send it for processing
-pub fn find_successive_blocks_in_vector(block_numbers: Vec<u64>) -> Vec<u64> {
+pub fn find_successive_blocks_in_vector(block_numbers: Vec<u64>, limit: Option<usize>) -> Vec<u64> {
     block_numbers
         .iter()
         .enumerate()
-        .take_while(|(index, block_number)| **block_number == (block_numbers[0] + *index as u64))
+        .take_while(|(index, block_number)| {
+            **block_number == (block_numbers[0] + *index as u64) && (limit.is_none() || *index <= limit.unwrap())
```
heemankv: Shouldn't it be 
```rust
index < limit 
```
or else we'll have an extra element !

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -219,7 +211,7 @@ mod test_update_state_worker_utils {
     #[case(vec![1, 2, 3, 4, 7, 8, 9, 11], vec![1, 2, 3, 4])]
     #[case(vec![1, 3, 5, 7, 9], vec![1])]
     fn test_find_successive_blocks(#[case] input: Vec<u64>, #[case] expected: Vec<u64>) {
-        let result = super::find_successive_blocks_in_vector(input);
+        let result = super::find_successive_blocks_in_vector(input, None);
```
heemankv: if the above comment about < instead of <= is valid then ideally we should have caught that in this testcase if we were passing a limit, 
maybe we should spend time and add that testcase ?
wdyt ?

heemankv (changes requested):

orchestrator/src/worker/event_handler/triggers/update_state.rs
0xvasanth: updated some new test case

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -213,13 +207,17 @@ mod test_update_state_worker_utils {
...
-    #[case(vec![1, 2, 3, 4, 5], vec![1, 2, 3, 4, 5])]
-    #[case(vec![1, 2, 3, 4, 7, 8, 9, 11], vec![1, 2, 3, 4])]
-    #[case(vec![1, 3, 5, 7, 9], vec![1])]
-    fn test_find_successive_blocks(#[case] input: Vec<u64>, #[case] expected: Vec<u64>) {
-        let result = super::find_successive_blocks_in_vector(input);
+    #[case(vec![], Some(3), vec![])]
+    #[case(vec![1], Some(3), vec![1])]
+    #[case(vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], Some(10), vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10])]
+    #[case(vec![1, 2, 3, 4, 5], Some(3), vec![1, 2, 3])] // limit smaller than available
+    #[case(vec![1, 2, 3], Some(5), vec![1, 2, 3])] // limit larger than available
```
heemankv, line 215: Nice ! 
can we also add a limit : None case ? 
that would complete it!

package-lock.json
```diff
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "madara",
```
heemankv, line 7: Hmm, this is new, are we sure of this ? 
if yes, what's the reason ?

package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "madara",
```
heemankv, line 2: Same as above

heemankv (changes requested):

package.json
0xvasanth: When the code is checked out in a folder with a name different from madara, the package name in the lock file keeps getting updated. To avoid this unwanted change, I explicitly added the name field in package.json, so the package manager consistently uses it instead of relying on the folder name.

package-lock.json
0xvasanth: same as below

package.json
heemankv: yeah, I've seen that happening locally, 
makes sense thanks !

heemankv (approved):
LGTM

PR #734, opened 2025-07-17
E2E : 1 : Mock Atlantic, Dummy Verifier

heemankv:
# Add Mock Atlantic Server for Testing

## Pull Request type

- Feature
- Testing

## What is the current behavior?

Resolves: #NA

## What is the new behavior?

- Added a mock Atlantic server implementation for testing and development
- Provides a complete simulation of the Atlantic API endpoints
- Supports configurable behavior including failure simulation and processing delays
- Implements realistic job lifecycle with status transitions
- Includes comprehensive documentation and usage examples

## Does this introduce a breaking change?

No

## Other information

The mock server can be used in two ways:
1. As a standalone server for testing Atlantic client integration
2. Integrated with the orchestrator for faster testing without external dependencies

The implementation includes:
- Full API compatibility with the real Atlantic service
- Configurable failure rates and processing times
- Health check endpoints
- Realistic proof generation

PR #740, opened 2025-07-22
E2E: 2 : Init + Server

heemankv:
# Add E2E Testing Framework

## Pull Request type

- Feature
- Testing

Resolves: #NA

## What is the new behavior?

- Created a new `e2e` crate for end-to-end testing infrastructure
- Implemented a generic `Server` abstraction for managing test services with configurable logging
- Added RPC helper methods for interacting with Starknet nodes
- Established foundation for service management in tests (process control, endpoint handling)

No

## Other information

This PR introduces a structured approach to E2E testing with proper process management, logging, and RPC interaction capabilities. The framework handles process lifecycle, connection verification, and provides clean abstractions for test implementations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,13 @@ test-artifacts/madara
 
 # global npm package installation directory for nix
 .npm-global
+
+
+# e2e test
+e2e_data/
+/e2e_data.zip
+/e2e_data/*
+/e2e_data*
+
+out
+cache
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "e2e"
-description = ""
+description = "End-to-end testing for Madara Services"
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -20,7 +20,6 @@ dotenvy = { workspace = true }
 env_logger = { workspace = true }
 httpmock = { version = "0.8.0-alpha.1", features = ["proxy", "remote"] }
 log = { workspace = true }
-thiserror = { workspace = true }
 reqwest = { workspace = true, features = ["json"] }
 rstest = { workspace = true }
 serde = { workspace = true }
@@ -28,6 +27,7 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 testcontainers = { workspace = true }
+thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tokio-stream = { workspace = true }
 tokio-util = { workspace = true }
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "e2e"
+version = "0.1.0"
+edition = "2021"
+authors = ["Heemank Verma <heemankv@gmail.com>"]
+
+[lib]
+name = "e2e"
+path = "src/mod.rs"
+
+[dependencies]
+
+async-trait = { workspace = true }
+
+dotenvy = { workspace = true }
+env_logger = { workspace = true }
+httpmock = { version = "0.8.0-alpha.1", features = ["proxy", "remote"] }
+log = { workspace = true }
+thiserror = { workspace = true }
+reqwest = { workspace = true, features = ["json"] }
+rstest = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
+testcontainers = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
+tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
+url = { workspace = true }
+uuid = { workspace = true }
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -1,8 +1,12 @@
 [package]
 name = "e2e"
-version = "0.1.0"
-edition = "2021"
-authors = ["Heemank Verma <heemankv@gmail.com>"]
+description = ""
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+homepage.workspace = true
 
 [lib]
 name = "e2e"
--- a/e2e/src/services/constants.rs
+++ b/e2e/src/services/constants.rs
@@ -1,2 +1,87 @@
+// =============================================================================
+// CONSTANTS
+// =============================================================================
+
+use std::path::PathBuf;
+use std::sync::LazyLock;
+use tokio::time::Duration;
+
+// =============================================================================
+// GENERAL / SHARED
+// =============================================================================
+pub static REPO_ROOT: LazyLock<PathBuf> = LazyLock::new(|| {
+    PathBuf::from(env!("CARGO_MANIFEST_DIR")).parent().expect("Failed to get workspace root").to_path_buf()
+});
 pub const DEFAULT_SERVICE_HOST: &str = "127.0.0.1";
-pub const DEFAULT_MONGODB_DIR: &str = "data";
+pub const BINARY_DIR: &str = "target/release";
+pub const DATA_DIR: &str = "e2e_data";
+
+// =============================================================================
+// ANVIL SERVICE
+// =============================================================================
+pub const ANVIL_DATABASE_FILE: &str = "anvil.json";
+pub const ANVIL_BLOCK_TIME: f64 = 1_f64;
+pub const ANVIL_PORT: u16 = 8545;
+pub const ANVIL_PRIVATE_KEY: &str = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+
+// =============================================================================
+// BOOTSTRAPPER SERVICE
+// =============================================================================
+pub const BOOTSTRAPPER_BINARY: &str = "bootstrapper";
+pub static BOOTSTRAPPER_SETUP_L1_TIMEOUT: LazyLock<Duration> = LazyLock::new(|| Duration::from_secs(120));
+pub static BOOTSTRAPPER_SETUP_L2_TIMEOUT: LazyLock<Duration> = LazyLock::new(|| Duration::from_secs(1200));
+pub const BOOTSTRAPPER_CONFIG: &str = "e2e/config/bootstrapper.json";
+pub const BOOTSTRAPPER_ADDRESSES_FILE: &str = "addresses.json";
+
+// =============================================================================
+// LOCALSTACK SERVICE
+// =============================================================================
+pub const LOCALSTACK_IMAGE: &str =
+    "localstack/localstack@sha256:763947722c6c8d33d5fbf7e8d52b4bddec5be35274a0998fdc6176d733375314";
+pub const LOCALSTACK_PORT: u16 = 4566;
+pub const LOCALSTACK_CONTAINER: &str = "localstack-service";
+
+// =============================================================================
+// MONGODB SERVICE
+// =============================================================================
+pub const MONGODB_IMAGE: &str = "mongo:latest";
+pub const MONGODB_PORT: u16 = 27017;
+pub const MONGODB_CONTAINER: &str = "mongodb-service";
+
+// =============================================================================
+// MADARA SERVICE
+// =============================================================================
+pub const MADARA_NAME: &str = "madara";
+pub const MADARA_DATABASE_DIR: &str = "madara-db";
+pub const MADARA_BINARY: &str = "madara";
+pub const MADARA_CONFIG: &str = "e2e/config/madara.yaml";
+pub const MADARA_RPC_PORT: u16 = 9944;
+pub const MADARA_RPC_ADMIN_PORT: u16 = 9943;
+pub const MADARA_GATEWAY_PORT: u16 = 9944;
+
+// =============================================================================
+// MOCK PROVER SERVICE
+// =============================================================================
+pub const MOCK_PROVER_BINARY: &str = "mock-atlantic-server";
+pub const MOCK_PROVER_VERIFIER_ADDRESS_FILE: &str = "verifier_address.txt";
+pub const MOCK_PROVER_PORT: u16 = 3001;
+
+// =============================================================================
+// PATHFINDER SERVICE
+// =============================================================================
+pub const PATHFINDER_PORT: u16 = 9545;
+pub const PATHFINDER_BINARY: &str = "pathfinder";
+
+// =============================================================================
+// ORCHESTRATOR SERVICE
+// =============================================================================
+pub const ORCHESTRATOR_BINARY: &str = "orchestrator";
+pub const ORCHESTRATOR_DATABASE_NAME: &str = "orchestrator";
+
+// =============================================================================
+// MOCK VERIFIER DEPLOYER SERVICE
+// =============================================================================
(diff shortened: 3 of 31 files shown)


heemankv:
* **#740** <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#741](https://github.com/madara-alliance/madara/pull/741) <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/741?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#748](https://github.com/madara-alliance/madara/pull/748) <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/748?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#734** <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/734?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#723](https://github.com/madara-alliance/madara/pull/723) <a href="https://app.graphite.dev/github/pr/madara-alliance/madara/723?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

e2e/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "e2e"
+description = ""
```
Mohiiit: let's add description?

e2e/src/services/helpers.rs
```diff
@@ -0,0 +1,67 @@
...
+            .await
+            .map_err(|_| NodeRpcError::InvalidResponse)?;
+
+        let json = response.json::<serde_json::Value>().await.map_err(|_| NodeRpcError::InvalidResponse)?;
+
+        // Check if there's an error in the JSON-RPC response
+        if let Some(error) = json.get("error") {
+            // Check for specific "Block not found" error (code 24)
+            if let (Some(code), Some(message)) = (error.get("code"), error.get("message")) {
+                if code.as_u64() == Some(24) && message.as_str().map(|s| s.contains("Block not found")).unwrap_or(false)
```
Mohiiit: let's move 24 to const at top

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            let stdout = process.stdout.take().ok_or(ServerError::StartupFailed(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Failed to capture stdout",
+            )))?;
+
+            let service_name = config.service_name.clone();
+            let stdout_task_inner = task::spawn(async move {
+                // Keeping a large buffer capacity for stdout, to not have buffer overflow
+                let reader = BufReader::with_capacity(65536, stdout);
```
Mohiiit: same here for 65536

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn is_running(&mut self) -> bool {
+        self.has_exited().is_none()
+    }
+
+    /// Get a free port
+    fn get_free_port() -> u16 {
+        std::net::TcpListener::bind("127.0.0.1:0")
+            .and_then(|listener| listener.local_addr())
+            .map(|addr| addr.port())
+            .unwrap_or(8080) // Fallback port
```
Mohiiit: if 8080 is gonna be default port, let's move to const

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    /// Wait until the server is ready to accept connections
+    async fn wait_till_started(&mut self) -> Result<(), ServerError> {
+        let mut attempts = self.config.connection_attempts;
+        let addr = self.endpoint();
+
+        if let Some(addr) = addr {
+            // Extract just the socket address from the URL
+            let socket_addr = if let Some(host) = addr.host_str() {
+                let port = addr.port().unwrap_or(80); // Default to 80 if no port
```
Mohiiit: same here

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        // Start the process
+        let mut process = command.spawn().map_err(ServerError::StartupFailed)?;
+
+        let mut stdout_task = None;
+        let mut stderr_task = None;
+
+        // Extract stdout and stderr for log monitoring
+        if config.logs.0 {
+            command.stdout(Stdio::piped());
```
Mohiiit: isn't this already done in line 66?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+        if !self.config.rpc_port.is_some() {
+            return Ok(());
+        }
+        if self.has_exited().is_some() {
+            return Ok(());
+        }
+
+        // Try to terminate gracefully first
+        let pid = self.process.id();
+        let kill_result = Command::new("kill").args(["-s", "TERM", &pid.unwrap().to_string()]).spawn();
```
Mohiiit: this can be handled more gracefully, if process.id() is none, simply return Ok() else proceed with the kill? wdyt?

I am thinking about the case when has_exited() return some and process get killed in between the check (pid would be none in that case) and actual kill

e2e/src/services/server.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub logs: (bool, bool),
+}
+
+impl Default for ServerConfig {
+    fn default() -> Self {
+        Self {
+            service_name: String::from(""),
+            rpc_port: None,
+            connection_attempts: 30,
+            connection_delay_ms: 1000,
```
Mohiiit: let's move all magic numbers to const

e2e/src/services/server.rs
heemankv: Valid removed, 
Side note : We need to register `command.stdout(Stdio::piped());` before we do `command.spawn()`
And we can get the logs only after the process is available, using `process.stdout.take()`
Hence we have two if cases, one before command.spawn() and one after.

e2e/src/services/server.rs
heemankv: Done Done

.gitignore
```diff
@@ -77,4 +77,14 @@ image.tar.gz
...
 
 # global npm package installation directory for nix
-.npm-global
\ No newline at end of file
+.npm-global
+
+
+# e2e test
+e2e_data/
+/e2e_data.zip
```
0xvasanth, line 85: Just wanted to clarify — I noticed that e2e_data.zip is present in the code but appears to be ignored in the second PR. Was this intentional, or might it have been missed by accident?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+                    }
+                }
+
+                attempts -= 1;
+                tokio::time::sleep(Duration::from_millis(self.config.connection_delay_ms)).await;
+            }
+        } else {
+            Err(ServerError::EndpointNotAvailable)
+        }
+    }
```
0xvasanth: What happens if the process is spawned successfully by the application, but fails internally (e.g., crashes silently or hangs)? In that case, it seems wait_till_started could loop indefinitely without detecting a failure. Should we add an internal health check or a timeout safeguard?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
+use std::collections::HashMap;
+use std::process::{ExitStatus, Stdio};
+use std::time::Duration;
+use tokio::net::TcpStream;
+use url::Url;
```
0xvasanth: I’d recommend considering [tokio::process] for handling the child process lifecycle, since it allows non-blocking, async-aware process management. This might help better integrate with the rest of the async context and provide more fine-grained control over the process state

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+}
+
+impl ServerConfig {
+    /// Helper method to inherit current process environment variables
+    pub fn with_current_env(mut self) -> Self {
+        let current_env: HashMap<String, String> = std::env::vars().collect();
+        self.env_vars = Some(current_env);
+        self
+    }
```
0xvasanth, line 66: If possible, it might be better to avoid pulling environment variables from multiple places, as it increases the risk of inconsistencies—one block might have access while another might not. Instead, consider passing the environment map as a parameter from the parent or initializer function for better control and predictability.

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    /// Get the process ID if the process is still running
+    pub fn pid(&self) -> Option<u32> {
+        self.process.id()
+    }
+
+    /// Check if the process has exited
+    pub fn has_exited(&mut self) -> Option<ExitStatus> {
+        match self.process.try_wait() {
+            Ok(status) => status,
```
0xvasanth: try_wait() can return Ok(None) when the process has started but hasn’t exited yet — this is different from an actual error. However, both Ok(None) and all Err(_) cases are currently treated the same (None). Would it make sense to distinguish these two scenarios for better observability or error handling?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+                    Ok(_) => return Ok(()),
+                    Err(_) => {
+                        // Check if process has exited
+                        if let Some(status) = self.has_exited() {
+                            return Err(ServerError::ProcessExited(status));
+                        }
+
+                        if attempts == 0 {
+                            return Err(ServerError::ConnectionTimeout(self.config.connection_attempts));
+                        }
```
0xvasanth, line 212:
```suggestion
                        if attempts <= 0 {
                            return Err(ServerError::ConnectionTimeout(self.config.connection_attempts));
                        }
``` always consider using less than so if we screen somewhere is the attempt it will not bereak the system

File: e2e/src/services/server.rs [hunk 15]
```diff
@@ -0,0 +1,269 @@
...
+
+                while let Ok(Some(line)) = lines.next_line().await {
+                    println!("[STDOUT] [{}] {}", service_name, line);
+
+                    // Flush immediately to prevent backing up
+                    use std::io::Write;
+                    let _ = std::io::stdout().flush();
+                }
+            });
+            stdout_task = Some(stdout_task_inner);
```
0xvasanth: can we consider using circular buffer

e2e/src/services/server.rs
0xvasanth: I would prefer timeout here with fix time insead of max retry, since each api response time can be varied

e2e/src/services/constants.rs
0xvasanth, line 1: Several hardcoded values should be configurable for different test scenarios:

  - BUFFER_CAPACITY (65536) - Different services may need different buffer sizes
  - CONNECTION_DELAY_MS (1000) - May need faster retries for unit tests
  - CONNECTION_ATTEMPTS (30) - Different services may need different retry counts

  Consider adding these to ServerConfig with sensible defaults, allowing per-service customization:
  ```
  impl ServerConfig {
      pub fn with_buffer_size(mut self, size: usize) -> Self {
          self.buffer_capacity = size;
          self
      }
  }
  ```

e2e/src/services/server.rs
0xvasanth: The current implementation uses attempt count rather than actual time, which could lead to:
  - Very long waits if connection attempts are slow
  - Inconsistent behavior across different network conditions

  Consider using a combination of timeout and retries:
  ```
  const CONNECTION_TIMEOUT: Duration = Duration::from_secs(30);
  let start_time = Instant::now();

  loop {
      if start_time.elapsed() > CONNECTION_TIMEOUT {
          return Err(ServerError::ConnectionTimeout(self.config.connection_attempts));
      }

      match timeout(Duration::from_secs(2), TcpStream::connect(&socket_addr)).await {
          Ok(Ok(_)) => return Ok(()),
          _ => {
              // Check process status and continue
          }
      }
  }
  ```

.gitignore
prkpndy: we can put the test artifacts in a separate repo now and download them when needed

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+        let mut server = Self { process, config, stdout_task, stderr_task };
+
+        // We wait & validate only if the service has an API endpoint
+        // e.g : Skips for Bootstrapper and Orchestrator setup
+        if has_rpc_endpoint {
+            println!("🔔 Waiting for {} server to be ready", service_name);
+            server.wait_till_started().await?;
+        }
+
+        println!("😁 {} Server is ready", service_name);
```
prkpndy, line 160: can we use info/debug instead of println?

e2e/src/services/server.rs
```diff
@@ -0,0 +1,269 @@
...
+            task.abort();
+        }
+
+        Ok(())
+    }
+}
+
+impl Drop for Server {
+    fn drop(&mut self) {
+        let _ = self.stop();
```
prkpndy, line 275: we can log the result of stop method informing the user if graceful shutdown happened or not

prkpndy (commented):
Minor Comments. gtg

.gitignore
heemankv: Irrelevant now, this file was removed all-together

e2e/src/services/server.rs
heemankv: Since we have tokio command in place, all of the code halts if any application fails

e2e/src/services/server.rs
heemankv: utilised

e2e/src/services/server.rs
heemankv: Was reading through this code and realised it's not properly written! 
and docs for tokio::Command and realised that the child process inherits the parent's envs by default on it's own
So I have now removed this, as this is of no use!

e2e/src/services/server.rs
heemankv: Alright, changed the return statement to : Result<Option<ExitStatus>, ServerError>

e2e/src/services/server.rs
heemankv: I believe your second comment is for another comment : [here](https://github.com/madara-alliance/madara/pull/740#discussion_r2255769252) ?? @itsparser

e2e/src/services/constants.rs
heemankv: Alright, added buffer_capacity to server_config

e2e/src/services/server.rs
heemankv: There's no benefit actually, since this is testing we only care about printing the logs, and not trace them !

heemankv: lemme know if your thoughts diverge!

e2e/src/services/server.rs
heemankv: Oh, so each servicer has it's `stop` impl that print's the services that were stopped
see this :  
```rust
pub fn stop(&mut self) -> Result<(), MockProverError> {
        println!("☠️ Stopping Mock Prover");
        self.server.stop().map_err(|err| MockProverError::Server(err))
    }
```
Screenshot 2025-08-12 at 3 16 25 PM

e2e/src/services/server.rs
heemankv: Yes, marked and refactoring all the timeouts in the last PR!

prkpndy (approved):

0xvasanth (approved):

PR #751, opened 2025-07-24
fix: bootstrapper environment variables

Mohiiit:
## Pull Request type

- Bugfix

There was a PR earlier which had a bug, ideally we want some of the fields optional in config file and these fields should be env driven too.

## What is the new behavior?

Now the variables are optional for the config files.

No


heemankv (approved):

byteZorvin (approved):
Lgtm !!

PR #755, opened 2025-07-26
feat: Update dependencies and clean up unused variables in Starknet client implementation

0xvasanth:
## Pull Request type

- Feature

Currently we are passing onchain data hash as part of the stateupdate call to zaun, which we have removed from the core contract

## What is the new behavior?

- Removed the onchnain data hash and size

### DIFF
--- a/orchestrator/crates/settlement-clients/starknet/src/lib.rs
+++ b/orchestrator/crates/settlement-clients/starknet/src/lib.rs
@@ -15,14 +15,14 @@ use mockall::automock;
 use mockall::predicate::*;
 use orchestrator_settlement_client_interface::{SettlementClient, SettlementVerificationStatus};
 use starknet::accounts::{ConnectedAccount, ExecutionEncoding, SingleOwnerAccount};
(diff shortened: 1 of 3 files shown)


orchestrator/crates/settlement-clients/starknet/src/lib.rs
```diff
@@ -117,8 +117,8 @@ impl SettlementClient for StarknetSettlementClient {
         &self,
         snos_output: Vec<[u8; 32]>,
         program_output: Vec<[u8; 32]>,
-        onchain_data_hash: [u8; 32],
```
heemankv, line 120: Why haven't we removed it completely ? like from the parent functions as well ?

0xvasanth: @heemankv, This parameter is still required for ethereum, and since we’re following a generic function structure, it may not be ideal to remove it at this point.

heemankv (approved):

Mohiiit (approved):

PR #759, opened 2025-07-31
Preconfirmed block (v0.14 support)

cchudant:
This pull request is big and took a long time to make. However, it was kind of impossible to break up.
- Backend has been rewritten from scratch to support the pre-confirmed block. We needed to do that because contrary to the pending block, the pre-confirmed block can be cleared and remplaced.
  - DbBlockId does not exist anymore. Instead, we have new view types.
- New subscription helpers in the backend - follow the chain tip, follow new heads, etc. These helpers build on one another, 
- All subscription RPCs have been disabled for now :( but we will re-add them in a later PR.
- Mempool now subscribes to chain tip. This means that it can now be updated via sync. Transaction statuses tracking has been reimplemented in the mempool, but this is not hooked up to get_transaction_status yet. It will be in a later PR once we enable the baby mempool.
- The backend is now separated from the DB through a MadaraStorage trait. I thought it was necessary since the backend got way bigger, and the rocksdb part of the backend didn't fit with the rest anymore. This also means that it is now possible to use other database implementations other than rocksdb, as long as you implement that trait.
  - I think we should move the rocksdb part in a crate called mc-rocksdb or mc-db, and the backend will finally be totally free of the horrible coupling we had earlier.
- "Pending block update time" is a thing of the past. Now, the backend will be updated live with every new executed preconfirmed transaction.
- The preconfirmed block can be saved to DB using a (for now, enabled by default) CLI flag. Appending to the preconfirmed block is now cheap as it will only save the new transactions instead of rewriting the whole block to db.
- Getting a single transaction or receipt from DB is now cheaper as transactions are now all saved as separate DB keys.

## Motivation fo Backend Views

The new backend API looks like this:

```rs
// rpc v0.7.1 get_class implementation as an example.
pub fn get_class(
    starknet: &Starknet,
    block_id: BlockId,
    class_hash: Felt,
) -> StarknetRpcResult<MaybeDeprecatedContractClass> {
    let view = starknet.backend.view_on(&block_id)?; // state view
    let class_info = view.get_class_info(&class_hash)?.ok_or(StarknetRpcApiError::class_hash_not_found())?;

    Ok(class_info.contract_class().into())
}
```

Instead of querying the backend, you now create a view and perform your queries on it. This view gives you an immutable view of the current chain state, where, if the preconfirmed block / latest confirmed block changes in the backend, you will always get the state from when the view was created.

You can think of the view as kind of a "snapshot" of the current backend. Behind the scenes, it is not a snapshot ; instead, all queries will "censor out" values from later confirmed blocks. For pre-confirmed block, it gets more interesting: the view will clone an Arc of the current in-memory preconfirmed block, and every view has a "n_txs_visible" field which dictates which transactions are be visible from this view. This means that the pre-confirmed block is able to be appended to, even there might be currently a RPC endpoint running concurrently with a view on it, without requiring any copy of the whole block. In addition, the view will still be valid even if the preconfirmed block is cleared.
This is very important for execution RPC endpoints which may take some time to execute! Before, appending to the pending block could completely break any execution currently happening concurrently on it. This wasn't much of an issue because we were cloning pending blocks a lot of the time, but now, with the in-memory pre-confirmed block, this issue would have been much more of a concern and it had to be addressed and this was the motivation for the rewrite.

## A few types of backend views

There are two main types of backend views: block views, and state views.

Block views are views that relate to a specific block, and it allows you to get state diffs, transactions, header etc. specific to the targetted block in question. There are two kinds of block views: MadaraPreconfirmedBlockView, and MadaraConfirmedBlockView. MadaraBlockView also exists to abstract over the two when you need to.

State views are views that allow you to query the backend over all of the previous blocks. It provides queries such as `get_contract_storage` (kv storage), `get_contract_class`, `get_class`, etc.
A big difference between block views and state views is that state views can be created on top of the empty pre-genesis state. In this case, all queries will return `None` (not found). Block views can never target the pre-genesis empty state, since they have to target a block that exists. (MadaraBlockView .get_block_info cannot return None, you can't make a block view on a block that does not exist.)

There are a multitude of methods to get from state views to block views, preconfirmed/confirmed block views and back. For example, you can get the latest block visible from a state view and you will get a block view. You can get a state view on the state visible at a given block view etc.

One of the neat things about views is that they are decoupled from the rest of the backend. It's an abstraction built on top of the `chain tip` abstraction and db (MadaraStorageRead) queries.

## The backend Writer

All write-related methods (at least, for blocks) are under backend.write_access().xxx now.
The supported methods are
```rs
    /// Append transactions to the current preconfirmed block. Returns an error if there is no preconfirmed block.
    /// Replaces all candidate transactions with the content of `replace_candidates`.
    pub fn append_to_preconfirmed(
        &self,
        executed: &[PreconfirmedExecutedTransaction],
        replace_candidates: impl IntoIterator<Item = Arc<ValidatedTransaction>>,
    ) -> Result<()> {
```
```rs
    /// Clears the current preconfirmed block. Does nothing when the backend has no preconfirmed block.
    pub fn clear_preconfirmed(&self) -> Result<()> {
```
```rs
    /// Start a new preconfirmed block on top of the latest confirmed block. Deletes and replaces the current preconfirmed block if present.
    /// Warning: Caller is responsible for ensuring the block_number is the one following the current confirmed block.
    pub fn new_preconfirmed(&self, block: PreconfirmedBlock) -> Result<()> {
```
```rs
    /// Add a block. Returns the block hash.
    /// Warning: Caller is responsible for ensuring the block_number is the one following the current confirmed block.
    pub fn add_full_block_with_classes(
        &self,
        block: &FullBlockWithoutCommitments,
        classes: &[ConvertedClass],
        pre_v0_13_2_hash_override: bool,
    ) -> Result<AddFullBlockResult> {
```

as well as some additional low-level methods for sync. It is very similar to the previous implementation, except that changing the chain tip, aka
- importing a new block
- clearing the preconfirmed block
- adding / replacing the preconfirmed block

is now completely atomic. This is done via the new ChainTip abstraction.


madara/crates/client/block_production/src/executor/mod.rs
```diff
@@ -1,19 +1,15 @@
...
-use mp_convert::Felt;
-use std::{any::Any, collections::HashMap, panic::AssertUnwindSafe, sync::Arc};
+use std::{any::Any, panic::AssertUnwindSafe, sync::Arc};
 use tokio::sync::{
     mpsc::{self, UnboundedReceiver},
     oneshot,
 };
 
-mod tests;
+// mod tests;
```
jbcaron: add TODO

madara/crates/client/block_production/src/executor/tests.rs
```diff
@@ -7,7 +7,7 @@ use blockifier::transaction::transaction_execution::Transaction;
 use mc_db::MadaraBackend;
 use mc_exec::execution::TxInfo;
 use mp_chain_config::StarknetVersion;
-use mp_convert::ToFelt;
+use mp_convert::{Felt, ToFelt};
```
jbcaron, line 10: import `starknet_types_core::felt::Felt` instead

madara/crates/client/block_production/src/lib.rs
```diff
@@ -530,1692 +362,1700 @@ impl BlockProductionTask {
...
-            receipt.events()[0],
-            Event {
-                from_address: contract_address,
-                keys: vec![get_selector_from_name("CalledFromL1").unwrap()],
-                data: vec![/* from_address */ Felt::THREE, /* arg1 */ Felt::ONE, /* arg2 */ Felt::TWO]
-            }
-        );
-    }
-}
+// #[cfg(test)]
```
jbcaron: why not use `#[cfg(FALSE)]` ?

madara/crates/client/db/src/view/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /// Returns a view on the preconfirmed block. This view is used to query content and listen for changes in that block.
+    /// This returns a fake preconfirmed block if there is not currently one in the backend.
+    /// The preconfirmed block view will not include candidate transactions.
+    pub fn block_view_on_preconfirmed_or_fake(self: &Arc<Self>) -> MadaraPreconfirmedBlockView<D> {
+        MadaraPreconfirmedBlockView::new(
+            self.clone(),
+            self.preconfirmed_block().unwrap_or_else(|| {
+                // fake preconfirmed block.
```
jbcaron: Wouldn’t it be better to have our own BlockId type, separate from the RPC types, and provide two conversions/implementations depending on the RPC BlockId version (0.7.1 vs 0.9.0)?

madara/crates/client/db/src/rocksdb/backup.rs
```diff
@@ -0,0 +1,85 @@
...
+};
+use std::{fs, path::Path, sync::Arc};
+
+#[derive(Debug, Clone)]
+pub struct BackupManager(Option<std::sync::mpsc::Sender<BackupRequest>>);
+
+impl BackupManager {
+    pub fn start_if_enabled(db_path: &Path, config: &RocksDBConfig) -> Result<Self> {
+        // when backups are enabled, a thread is spawned that owns the rocksdb BackupEngine (it is not thread safe) and it receives backup requests using a mpsc channel
+        // There is also another oneshot channel involved: when restoring the db at startup, we want to wait for the backupengine to finish restoration before returning from open()
```
jbcaron:
```suggestion
        // when backups are enabled, a thread is spawned that owns the rocksdb BackupEngine (it is not thread safe) and it receives backup requests using an mpsc channel
        // There is also another oneshot channel involved: when restoring the db at startup, we want to wait for the backupengine to finish restorating before returning from open()
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -38,155 +39,88 @@ mod util;
...
-            .collect(),
-        state_diff,
-    };
-
-    Ok((block, classes))
+/// Used for listening to state changes in tests.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BlockProductionStateNotification {
+    ClosedBlock,
+    UpdatedPendingBlock,
```
jbcaron: Is Pending really appropriate?

madara/crates/client/db/src/view/block_preconfirmed.rs
```diff
@@ -0,0 +1,396 @@
...
+            storage_diffs.entry(Felt::ONE).or_default().insert(Felt::from(block_number), block_hash);
+        }
+
+        // Aggregate the transaction state diffs.
+        for ((addr, key), value) in borrow.executed_transactions().flat_map(|tx| &tx.state_diff.storage) {
+            let previous_value = view_on_parent_block.get_contract_storage(addr, key)?.unwrap_or(Felt::ZERO);
+            if &previous_value != value {
+                // only keep changed keys.
+                storage_diffs.entry(*addr).or_default().insert(*key, *value);
+            }
```
jbcaron: If tx1 applies V0 → V1 and tx2 applies V1 → V0, then V1 will not be deleted.

madara/crates/client/block_production/src/executor/tests.rs
cchudant: i think I prefer having our own reexport here actually? I think it just makes more sense

i kind of wanted to switch the whole codebase to this import, actually

madara/crates/client/block_production/src/lib.rs
cchudant: closing as outdated

madara/crates/client/db/src/view/mod.rs
cchudant: that is true. However, it turns out we don't need a blockid type at all for internal use - it's only used in RPC for now.
And yes, I think we should have separate block id types for separate rpc versions. Maybe one even for gateway..

madara/crates/client/db/src/view/block_preconfirmed.rs
cchudant: nice catch! fixed, and a big test has been added.

jbcaron (approved):

antiyro (approved):

madara/node/src/service/l2.rs
```diff
@@ -47,13 +48,15 @@ impl SyncService {
         db: &Arc<MadaraBackend>,
         l1_head_recv: L1HeadReceiver,
         warp_update: Option<WarpUpdateConfig>,
+        unsafe_starting_block_enabled: bool,
```
Mohiiit, line 51: given this could break the database, what could be the usage for this flag?

madara/node/src/cli/l2.rs
```diff
@@ -17,10 +17,6 @@ pub struct L2SyncParams {
     #[clap(env = "MADARA_SYNC_DISABLED", long, alias = "no-sync")]
     pub l2_sync_disabled: bool,
 
-    /// The block you want to start syncing from. This will most probably break your database.
-    #[clap(env = "MADARA_UNSAFE_STARTING_BLOCK", long, value_name = "BLOCK NUMBER")]
-    pub unsafe_starting_block: Option<u64>,
-
```
Mohiiit, line 23: given we are removing this from the cli, does that mean in the l2_service it's for testing only?

madara/node/src/cli/l1.rs
```diff
@@ -24,22 +24,68 @@ impl fmt::Display for MadaraSettlementLayer {
 
 #[derive(Clone, Debug, clap::Args, Deserialize, Serialize)]
 pub struct L1SyncParams {
-    /// Disable L1 sync.
+    /// Disable L1 sync. This is handy when:
+    /// - You are running the full-node or gateay presets (sync-only node), but you do not have access to an L1 RPC node. You only want to sync
```
Mohiiit, line 28:
```suggestion
    /// - You are running the full-node or gateway presets (sync-only node), but you do not have access to an L1 RPC node. You only want to sync
```

madara/node/src/cli/l1.rs
```diff
@@ -61,6 +107,15 @@ pub struct L1SyncParams {
...
 
+    /// The settlement layer type. This specifies the type of API `--l1-endpoint <RPC url>` uses: Ethereum RPC, or Starknet RPC.
+    /// The usual cases for this is:
+    /// - When settling on Ethereum, which is the case for Starknet Mainnet, we are usually running an L2 - because Ethrereum is an L1 and doesn't settle on any other chain.
+    /// - When settling on Starknet, we are usually running an L3; because Starknet is already an L2 because it settles on Ethereum.
+    ///
+    /// This language can be quite confusing at times, because it doesn't really matter for madara whether the settlement layer itself settles on anything else. As
+    /// a simplification, all of the madara node documentation will usually use the term "L1" when talking about the *settlement layer* ("the layer above") and L2 when
+    /// talking about the current chain.
+    /// For even more confusion, you can actually run a madara chain as a standalone L1 using `--no-l1-sync`.
```
Mohiiit, line 118: hahah facts

madara/crates/primitives/state_update/src/normalize.rs
Mohiiit, line 1: empty file

madara/crates/primitives/state_update/src/lib.rs
```diff
@@ -20,18 +173,12 @@ pub struct PendingStateUpdate {
...
-pub enum DeclaredClassCompiledClass {
-    Sierra(/* compiled_class_hash */ Felt),
-    Legacy,
-}
-
 #[derive(Clone, Debug, Default, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct StateDiff {
     /// Changed storage values. Mapping (contract_address, storage_key) => value.
     pub storage_diffs: Vec<ContractStorageDiffItem>,
     /// New declared classes. List of class hashes.
```
Mohiiit, line 180: wrong comment here

madara/crates/primitives/state_update/src/lib.rs
```diff
@@ -20,18 +173,12 @@ pub struct PendingStateUpdate {
...
-    pub deprecated_declared_classes: Vec<Felt>,
+    pub old_declared_contracts: Vec<Felt>,
```
Mohiiit, line 181: isn't `deprecated_declared_classes` better name tho?

madara/crates/client/block_production/src/executor/tests.rs
```diff
@@ -27,7 +27,7 @@ fn make_tx(backend: &MadaraBackend, tx: impl IntoStarknetApiExt) -> (Transaction
         .unwrap()
         .into_blockifier_for_sequencing()
         .unwrap();
-    (tx, AdditionalTxInfo { declared_class })
+    (tx, AdditionalTxInfo { declared_class, arrived_at: Default::default() })
```
Mohiiit: can't we use the `ts` from above?

madara/crates/primitives/class/src/class_hash.rs
```diff
@@ -87,20 +87,19 @@ impl CompressedLegacyContractClass {
...
+    use starknet_providers::{Provider, SequencerGatewayProvider};
     use starknet_types_core::felt::Felt;
 
-    use crate::ContractClass;
-
     #[tokio::test]
     async fn test_compute_sierra_class_hash() {
         let provider = SequencerGatewayProvider::new(
-            Url::parse("https://alpha-mainnet.starknet.io/gateway").unwrap(),
+            Url::parse("https://gatway.alpha-mainnet.starknet.io/gateway").unwrap(),
```
Mohiiit:
```suggestion
            Url::parse("https://gateway.alpha-mainnet.starknet.io/gateway").unwrap(),
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -38,166 +33,128 @@ mod util;
...
+                                    *class.class_hash(),
+                                    class
+                                        .as_sierra()
+                                        .map(|class| DeclaredClassCompiledClass::Sierra(class.info.compiled_class_hash))
+                                        .unwrap_or(DeclaredClassCompiledClass::Legacy),
+                                )
+                            })
+                            .collect(),
+                    },
+                    declared_class,
```
Mohiiit, line 122: we are not updating this `declared_class`?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -362,7 +270,10 @@ impl BlockProductionTask {
                 let TaskState::Executing(state) = current_state else {
                     anyhow::bail!("Invalid executor state transition: expected current state to be Executing")
                 };
-                state.append_batch(batch_execution_result);
+
+                state.append_batch(batch_execution_result).await?;
+
+                self.send_state_notification(BlockProductionStateNotification::BatchExecuted);
             }
             ExecutorMessage::EndBlock => {
```
Mohiiit, line 278: something that we found out while replaying the 0.13.5 transaction on the madara is that in the current codebase we are not calling the executor.finalize which gives us the aliases for the keys (populating the 0x2 contract), ideally we should call that in every batch or when we are closing it

madara/crates/client/db/Cargo.toml
```diff
@@ -41,13 +42,16 @@ anyhow.workspace = true
...
+dashmap.workspace = true
 futures = { workspace = true }
+itertools.workspace = true
 librocksdb-sys = { workspace = true }
 num-traits = { workspace = true }
 rayon = { workspace = true }
 rocksdb.workspace = true
 serde = { workspace = true }
 siphasher.workspace = true
+smallvec = { version = "1.15.0", features = ["write"] }
```
Mohiiit: can we move this to the root cargo.toml?

madara/crates/client/db/src/preconfirmed.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub arrived_at: TxTimestamp,
+}
+
+impl PreconfirmedExecutedTransaction {
+    /// Creates a ValidatedTransaction from this PreconfirmedExecutedTransaction. This is used to put
+    /// transactions back into mempool.
+    pub fn to_validated(&self) -> ValidatedTransaction {
+        ValidatedTransaction {
+            transaction: self.transaction.transaction.clone(),
+            paid_fee_on_l1: None,
```
Mohiiit, line 24: we are assuming that there won't be any L1 txn in mempool?

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
```diff
@@ -0,0 +1,83 @@
...
+};
+use mp_convert::Felt;
+use mp_receipt::L1HandlerTransactionReceipt;
+use mp_transactions::{L1HandlerTransaction, L1HandlerTransactionWithFee};
+use rocksdb::ReadOptions;
+
+/// <core_contract_nonce 8 bytes> => bincode(tx hash)
+pub const L1_TO_L2_PENDING_MESSAGE_BY_NONCE: Column =
+    Column::new("l1_to_l2_pending_message_by_nonce").set_point_lookup();
+/// <core_contract_nonce 8 bytes> => bincode(class_info)
```
Mohiiit: in the value, it won't be class_info right? transaction_hash?

madara/crates/client/block_production/src/lib.rs
cchudant: fixed

madara/crates/client/rpc/src/versions/user/v0_8_1/methods/ws/subscribe_transaction_status.rs
```diff
@@ -452,8 +452,8 @@ mod test {
     }
 
     #[rstest::fixture]
-    fn pending(tx_with_receipt: mp_block::TransactionWithReceipt) -> mp_block::PendingFullBlock {
-        mp_block::PendingFullBlock {
+    fn pending(tx_with_receipt: mp_block::TransactionWithReceipt) -> mp_block::PreconfirmedFullBlock {
```
Mohiiit, line 455:
```suggestion
    fn pre_confirmed(tx_with_receipt: mp_block::TransactionWithReceipt) -> mp_block::PreconfirmedFullBlock {
```

madara/crates/client/settlement_client/src/gas_price.rs
```diff
@@ -122,7 +122,8 @@ impl GasPriceProviderConfigBuilder {
     pub fn build(self) -> anyhow::Result<GasPriceProviderConfig> {
         let needs_oracle = self.fix_strk_per_eth.is_none();
         if needs_oracle && self.oracle_provider.is_none() {
-            return Err(anyhow::anyhow!("Oracle provider must be set if no fix_strk_per_eth is provided"));
+            // TODO: fix this messagee.
+            tracing::debug!("Oracle provider must be set if no fix_strk_per_eth is provided. Gas prices may be incorrect when producing blocks.");
```
Mohiiit, line 126: if we are not throwing an error, does that mean we are taking some default value for eth to strk?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -132,108 +128,142 @@ impl PipelineSteps for GatewaySyncSteps {
...
-                            .block
-                            .transaction_receipts
-                            .iter()
-                            .map(|tx| &tx.transaction_hash)
-                            .eq(db_block.tx_hashes.iter());
-
-                    if block_has_not_changed {
+                    if in_backend.block_number() != block_number {
+                        return Ok(None);
+                    }
```
Mohiiit, line 190: does this mean that we are importing a new preconfirmed block?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -132,108 +128,142 @@ impl PipelineSteps for GatewaySyncSteps {
...
+                    // True if this gateway block should be considered as a new pre-confirmed block entirely.
+                    let new_preconfirmed = in_backend.header() != &header
+                        || in_backend.num_executed_transactions() > n_executed
+                        ||
+                        // Compare hashes
+                        // TODO: should we compute these hashes? probably not?
+                        Iterator::ne(
+                            in_backend.borrow_content().executed_transactions().map(|tx| tx.transaction.receipt.transaction_hash()),
+                            block.transactions[..n_executed].iter().map(|tx| tx.transaction_hash())
+                        );
```
Mohiiit, line 201: we can just replace all the txns too right?

madara/crates/primitives/class/src/compile.rs
```diff
@@ -444,19 +444,19 @@ fn felt_to_big_uint(value: &Felt) -> BigUint {
...
-    use starknet_providers::{Provider, SequencerGatewayProvider, Url};
+    use starknet_providers::Url;
+    use starknet_providers::{Provider, SequencerGatewayProvider};
     use starknet_types_core::felt::Felt;
 
     #[tokio::test]
     async fn test_compressed_legacy_class_to_blockifier() {
         let provider = SequencerGatewayProvider::new(
-            Url::parse("https://alpha-mainnet.starknet.io/gateway").unwrap(),
+            Url::parse("https://gatway.alpha-mainnet.starknet.io/gateway").unwrap(),
```
Mohiiit:
```suggestion
            Url::parse("https://gateway.alpha-mainnet.starknet.io/gateway").unwrap(),
```

Mohiiit (commented):
> Mempool now subscribes to chain tip. This means that it can now be updated via sync.

didn’t get this point, is this for future when we have the p2p? or is this something we need currently too? 

> It will be in a later PR once we enable the baby mempool.

baby mempool? is this mempool in p2p would be gossiping? 

> "Pending block update time" is a thing of the past.

but pending tick was more of an internal tick right? for madara to send a batch to the blockifier? how would that change now? is this changed by the batches directly now? the async batches we have now?

> the view will still be valid even if the preconfirmed block is cleared.

does this means that if I have queried at time t1 and there was a preconfirmed block but by the time t2 when the rpc is sending the response, although at t2 the block was confirmed, given the view was created at t1, it will still show the block as preconfirmed? just for my understanding

cchudant:
> > Mempool now subscribes to chain tip. This means that it can now be updated via sync.
> 
> didn’t get this point, is this for future when we have the p2p? or is this something we need currently too?

we don't currently need it, but it will be useful when we have a baby mempool or when we'll have p2p yes.

> > It will be in a later PR once we enable the baby mempool.
> 
> baby mempool? is this mempool in p2p would be gossiping?

in v0.14 there is no get_transaction_status endpoint in the feeder anymore, as full node we're supposed to keep track of a baby mempool. The baby mempool is just an internal list of transactions the node received by RPC, and it's supposed to drop them when they arrive into a block or the time to live has exceeded. The RPC get_transaction_status is supposed to query this baby mempool and output Received or TransactionNotFound based on that.
In madara, we intend on using the mempool as the baby mempool directly. When you send a transaction via RPC, it'll be inserted in the internal mempool of the full node, and `get_transaction_status` will see it just as well as when we're running in a sequencer. The time to live timer is handled just as well as the sequencer mempool. It'll also work nicely with p2p mempool in the far future.
As of now, this baby mempool is not hooked up and this PR uses the gateway `get_transaction_status` like before, which means it doesn't work for sepolia/mainnet. It'll be in another PR.

> > "Pending block update time" is a thing of the past.
> 
> but pending tick was more of an internal tick right? for madara to send a batch to the blockifier? how would that change now? is this changed by the batches directly now? the async batches we have now?

Before, this pending update time was a timer so that we don't have to re-write the pending block in DB too fast. Now, the preconfirmed block transactions are stored all as different keys in the db, which means we can directly put a transaction in the backend as soon as we receive the executor batch response.
The executor executes transactions in batch, so the pre-confirmed block will now be updated in the backend and database every batch, which is much faster.
It would be cool to remove this restriction too - in fact, there is a new blockifier executor for that: the [ConcurrentTransactionExecutor](https://github.com/starkware-libs/sequencer/blob/main/crates/blockifier/src/blockifier/concurrent_transaction_executor.rs). It would unfortunately make a lot of my executor code obselete, but if we hooked that up, we could have a much faster latency between the execution and the update being available in the backend and RPCs/gateway.

> > the view will still be valid even if the preconfirmed block is cleared.
> 
> does this means that if I have queried at time t1 and there was a preconfirmed block but by the time t2 when the rpc is sending the response, although at t2 the block was confirmed, given the view was created at t1, it will still show the block as preconfirmed? just for my understanding

yes, it'll still show as preconfirmed, nothing will change whatever happens to the chain tip.

madara/node/src/service/l2.rs
cchudant: I often use it to confirm that we can import newer sepolia / mainnet blocks to check version compatibility without having to sync a full node all over again.

madara/node/src/cli/l2.rs
cchudant: it's now a backend param

madara/crates/primitives/state_update/src/lib.rs
cchudant: no, old means legacy, it's the list of new cairo 0 classes declared in this block

madara/crates/primitives/state_update/src/lib.rs
cchudant: it doesn't match with the name in the official gateway

madara/crates/client/block_production/src/lib.rs
cchudant: nope, it's our field, we didn't get it from blockifier we got it from our additional info struct

madara/crates/client/db/src/preconfirmed.rs
cchudant: yes :)

madara/crates/client/rpc/src/versions/user/v0_8_1/methods/ws/subscribe_transaction_status.rs
cchudant: will be fixed in later pr

madara/crates/client/sync/src/gateway/blocks.rs
cchudant: it's a new preconfirmed block if the current preconfirmed block does not match the beginning of the gateway preconfirmed block.
if it matches, and headers are the same, and tx hashes match, it's the same block and we just append on our current db preconfirmed block

madara/crates/client/sync/src/gateway/blocks.rs
cchudant: we should make a new preconfirmed block even if some txs match, the preconfirmed block in its implementation can't go back it can only be appended on.

PR #764, opened 2025-08-05
feat: Add Atlantic mock server and dummy verifier for L3 testing

0xvasanth:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

Currently, the orchestrator requires a real Atlantic prover service connection for testing proof generation functionality. This creates several issues:
- Dependency on external services for local development and testing
- Inability to test prover interactions in CI/CD pipelines without real Atlantic credentials
- Difficulty simulating edge cases and failure scenarios in the prover service

Resolves: #NA

## What is the new behavior?

This PR introduces a **Mock Atlantic Server** that simulates the Atlantic prover service for testing and development purposes:

### Key Changes:
- **New Mock Atlantic Server** (`test_utils/crates/mock-atlantic-server/`): A complete mock implementation of the Atlantic API
  - Implements all Atlantic API endpoints (submit job, check status, etc.)
  - Supports configurable behavior including failure rates and processing delays
  - Provides realistic job lifecycle simulation with proper status transitions
  - Includes comprehensive Swagger documentation for API reference
  
- **Orchestrator Integration**: Modified the orchestrator to support mock mode
  - Added `atlantic_mock_fact_hash` configuration option to enable mock mode (set to "true" to enable)
  - When enabled, automatically starts the mock Atlantic server
  - Relaxed parameter validation in mock mode to allow testing without real credentials
  
- **Dummy Verifier Contract**: Added deployment script and contract for testing L3 verification
  - `MockGPSVerifier.sol`: Simple verifier contract for testing
  - `deploy_dummy_verifier.sh`: Script to deploy the mock verifier on local Anvil instance

- **Additional Improvements**:
  - Fixed project name and reorganized Cargo dependencies for better maintainability
  - Enhanced test infrastructure for E2E testing scenarios
  - Added support for simulating various proof verification scenarios

### Benefits:
- Enables comprehensive testing of prover integration without external dependencies
- Allows simulation of various failure scenarios for robust error handling
- Speeds up development by removing network latency to external services
- Enables CI/CD testing of proof generation flows
- Provides a controlled environment for testing L3 proof verification

**No**

This change is backward compatible. The mock mode is opt-in via the `atlantic_mock_fact_hash` configuration option. When not enabled, the orchestrator behaves exactly as before, connecting to the real Atlantic service.

### Usage Example:
```bash
# Enable mock mode in orchestrator config
# Set the environment variable or pass as CLI argument
export MADARA_ORCHESTRATOR_ATLANTIC_MOCK_FACT_HASH="true"

# The orchestrator will automatically start the mock server
make run-orchestrator-l3

# Deploy dummy verifier for testing (optional)
./test_utils/scripts/deploy_dummy_verifier.sh
```

### Testing:
- The mock server can be run standalone: `cargo run --bin mock-atlantic-server`
- API documentation available in `test_utils/crates/mock-atlantic-server/swagger.json`
- Supports configurable failure rates and delays for comprehensive testing
- Mock server runs on port 8080 by default (configurable)

### Configuration Options:
- `MADARA_ORCHESTRATOR_ATLANTIC_MOCK_FACT_HASH`: Set to "true" to enable mock mode
- Mock server environment variables:
  - `FAILURE_RATE`: Probability of job failure (0.0 to 1.0)
  - `PROCESSING_DELAY_MS`: Simulated processing time for jobs
  - `PORT`: Server port (default: 8080)

This PR significantly improves the testability of the L3 prover integration and enables local development without external dependencies.


orchestrator/src/core/config.rs
```diff
@@ -199,13 +200,66 @@ impl Config {
...
+
+                    // Create modified Atlantic params that point to the mock server
+                    let mock_atlantic_params = AtlanticValidatedArgs {
+                        atlantic_api_key: atlantic_params.atlantic_api_key.clone(),
+                        atlantic_service_url: format!("http://127.0.0.1:{}", mock_port)
+                            .parse()
+                            .expect("Failed to parse mock server URL"),
+                        atlantic_rpc_node_url: atlantic_params.atlantic_rpc_node_url.clone(),
+                        atlantic_verifier_contract_address:
+                            "0x007a9a039a9471a7bea2962695cebc03f01a702ffa848763a6749052b0396ccc".to_string(), // Hard-coded for mock mode
```
Mohiiit: why is this hardcoded tho? are we sure this would be the address all the time? is this on starknet sepolia? if yes then let's add the link to it as well (voyager link)

orchestrator/src/core/config.rs
```diff
@@ -199,13 +200,66 @@ impl Config {
...
     /// # Returns
     /// * `Box<dyn ProverClient>` - The proving service
-    pub(crate) fn build_prover_service(prover_params: &ProverConfig) -> Box<dyn ProverClient + Send + Sync> {
+    pub(crate) async fn build_prover_service(prover_params: &ProverConfig) -> Box<dyn ProverClient + Send + Sync> {
         match prover_params {
             ProverConfig::Sharp(sharp_params) => Box::new(SharpProverService::new_with_args(sharp_params)),
-            ProverConfig::Atlantic(atlantic_params) => Box::new(AtlanticProverService::new_with_args(atlantic_params)),
+            ProverConfig::Atlantic(atlantic_params) => {
+                // If mock_fact_hash is "true", start the mock Atlantic server
+                if atlantic_params.atlantic_mock_fact_hash.eq("true") {
```
Mohiiit: but this is true for L3s only right? when we have L2s mocked, we can still make call to the chain right? or maybe not in case of AR

I think it makes sense to have it for both L2 and L3

orchestrator/src/types/params/prover.rs
```diff
@@ -54,43 +53,80 @@ impl TryFrom<RunCmd> for ProverConfig {
...
+                if is_mock_mode {
+                    // Mock mode: provide sensible defaults for missing parameters
+                    Ok(Self::Atlantic(AtlanticValidatedArgs {
+                        atlantic_api_key: atlantic_args.atlantic_api_key.unwrap_or_else(|| "mock-api-key".to_string()),
+                        atlantic_service_url: atlantic_args
+                            .atlantic_service_url
+                            .unwrap_or_else(|| "http://unused-in-mock-mode.com".parse().unwrap()),
+                        atlantic_rpc_node_url: atlantic_args
+                            .atlantic_rpc_node_url
+                            .unwrap_or_else(|| "http://unused-in-mock-mode.com".parse().unwrap()),
```
Mohiiit: url to const?

0xvasanth: this is dummy URL

orchestrator/src/core/config.rs
0xvasanth: Re-frame this in better way with new CLI

orchestrator/src/core/config.rs
0xvasanth: this code is removed and made better now

orchestrator/src/cli/prover/atlantic.rs
```diff
@@ -23,7 +23,7 @@ pub struct AtlanticCliArgs {
     #[arg(required_if_eq("atlantic", "true"))]
     pub atlantic_rpc_node_url: Option<Url>,
 
-    /// Whether to use mock fact registry.
+    /// Enable mock mode for Atlantic prover (set to "true" to enable).
     #[arg(env = "MADARA_ORCHESTRATOR_ATLANTIC_MOCK_FACT_HASH", long)]
```
heemankv, line 28: In the earlier implementations, this variable was used to be sent to Atlantic.
Now that we won't be mocking anything to the actual Atlantic service, we can rename this variable while we are re-purposing it 
say : MADARA_ORCHESTRATOR_MOCK_ATLANTIC_SERVICE ?
wdyt @itsparser

orchestrator/src/core/config.rs
```diff
@@ -257,16 +263,52 @@ impl Config {
     ///
     /// # Arguments
     /// * `prover_params` - The proving service parameters
-    /// * `params` - The config parameters
     /// # Returns
     /// * `Box<dyn ProverClient>` - The proving service
-    pub(crate) fn build_prover_service(prover_params: &ProverConfig) -> Box<dyn ProverClient + Send + Sync> {
+    pub(crate) async fn build_prover_service(prover_params: &ProverConfig) -> Box<dyn ProverClient + Send + Sync> {
```
heemankv: Unsure as to why we need this `async` ?
What am I missing ?

orchestrator/src/core/config.rs
```diff
@@ -257,16 +263,52 @@ impl Config {
...
+        match prover_params {
+            ProverConfig::Atlantic(atlantic_params) => {
+                // Start mock Atlantic server if flag is enabled and we're in testnet
+                if allow_mock_hash_server && atlantic_params.atlantic_network == "TESTNET" {
+                    info!("Mock Atlantic server flag is enabled, starting mock server...");
+
+                    // Start the mock server in a background task
+                    tokio::spawn(async move {
+                        info!("Starting mock Atlantic server on port 4001");
+                        if let Err(e) = utils_mock_atlantic_server::start_mock_atlantic_server().await {
```
heemankv, line 317: Since we are spawning the service from within code, should we not allow to set the port from CLI as well ?
The user isn't made aware about this port requirement while they are starting the app, 
like they will not know about it because it's not an env
as well as it will not be printed at the start of the orchestrator 
wdyt ?

heemankv: we can always default to 4001 as you have written

orchestrator/src/core/config.rs
```diff
@@ -257,16 +263,52 @@ impl Config {
...
+                            error!("Failed to start mock Atlantic server: {}", e);
+                        }
+                    });
+
+                    // Give the mock server time to start
+                    tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
heemankv, line 323: Nitpick : 
if you have bandwidth, we can use `is-alive` endpoint 
else we can skip since it will only be used for testing, totally your call

orchestrator/src/tests/config.rs
```diff
@@ -395,10 +395,10 @@ pub mod implement_client {
         }
     }
 
-    pub(crate) fn init_prover_client(service: ConfigType, params: &EnvParams) -> Box<dyn ProverClient> {
+    pub(crate) async fn init_prover_client(service: ConfigType, params: &EnvParams) -> Box<dyn ProverClient> {
```
heemankv: if `build_prover_service` will not be async then this will also be removed

package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "madara",
```
heemankv, line 2: already done in other pr, might face conflicts

test_utils/crates/mock-atlantic-server/src/server.rs
```diff
@@ -314,7 +302,37 @@ pub async fn add_job_handler(
...
+                .iter()
+                .filter(|(_, job)| matches!(job.query.status, AtlanticQueryStatus::Done | AtlanticQueryStatus::Failed))
+                .min_by_key(|(_, job)| job.created_at);
+
+            if let Some((id_to_remove, _)) = oldest_done {
+                let id_to_remove = id_to_remove.clone();
+                jobs.remove(&id_to_remove);
+                debug!("Removed old job {} to maintain memory limit", id_to_remove);
+            } else {
+                // If no completed jobs, remove the oldest job regardless of status
```
heemankv, line 334: Thinking out loud
Will this not hamper the load with with E2E sends the jobs, 
currently it sends 100 ProofCreation jobs together to Atlantic server !

orchestrator/src/cli/prover/atlantic.rs
0xvasanth: I believe there’s a slight misunderstanding of the implementation. All the existing code remains unchanged — the only addition here is a new CLI flag that allows users to enable the mock service when needed.

test_utils/crates/mock-atlantic-server/src/server.rs
0xvasanth: never though of that but this is bottle-necking local memory, so better option is to make it as a env variable of the server which will allow use have the flexibility to customize the resource size

package.json
0xvasanth: it is gone already when you pull

orchestrator/src/core/config.rs
0xvasanth: yes that is right was used async in previous code change, will revert this back

orchestrator/src/core/config.rs
0xvasanth: Since this is test infrastructure, the simple 2-second sleep is probably fine - it's predictable and won't hang. The is-alive approach is "nicer" but not crucial for a mock server used only in tests.

orchestrator/src/core/config.rs
0xvasanth: I think adding CLI support here might be unnecessary overhead. If someone needs a custom port, they can already run the server manually, 

`PORT=8080 make run-mock-atlantic-server`

Including this directly in the orchestrator doesn’t add much value and could complicate things unnecessarily.

Mohiiit (approved):

test_utils/crates/mock-atlantic-server/src/server.rs
heemankv: Is this converted to env var ?

heemankv (approved):
LGTM

PR #765, opened 2025-08-06
add rpc primitives for v0.8.1 and v0.9.0

jbcaron:
## Pull Request type

- Feature

Resolves: #NA

## What is the new behavior?

-
-
-


madara/crates/client/submit_tx/src/error.rs
```diff
@@ -67,6 +67,10 @@ pub enum RejectedTransactionErrorKind {
     TransactionLimitExceeded,
     #[error("InvalidTransactionNonce")]
     InvalidTransactionNonce,
+    #[error("Replacement transaction is underpriced")]
+    ReplacementTransactionUnderpriced,
+    #[error("Transaction fee below minimum")]
+    FeeBelowMinimum,
```
cchudant, line 73: 👍

madara/crates/primitives/block/src/lib.rs
```diff
@@ -15,13 +14,14 @@ use starknet_types_core::felt::Felt;
...
+pub mod to_rpc;
 
 pub use event_with_info::EventWithInfo;
 pub use header::Header;
 pub use primitive_types::{H160, U256};
 
-pub type BlockId = mp_rpc::BlockId;
-pub type BlockTag = mp_rpc::BlockTag;
+pub type BlockId = mp_rpc::v0_7_1::BlockId;
+pub type BlockTag = mp_rpc::v0_7_1::BlockTag;
```
cchudant, line 24: i'd like to have our own types for that tbh, maybe on the v0.9 pr

madara/crates/primitives/rpc/src/v0_8_1/mod.rs
```diff
@@ -1,3 +1,40 @@
...
+
+mod starknet_api_openrpc;
+mod starknet_trace_api_openrpc;
 mod starknet_ws_api;
 
+pub use self::starknet_api_openrpc::*;
+pub use self::starknet_trace_api_openrpc::*;
 pub use self::starknet_ws_api::*;
+
+pub use crate::v0_7_1::{
```
cchudant, line 14: why don't you use `*` here?

cchudant: oh i get it sorry 👍

madara/crates/primitives/rpc/src/v0_9_0/starknet_ws_api.rs
```diff
@@ -0,0 +1,20 @@
+use serde::{Deserialize, Serialize};
+
+use super::{EmittedEvent, TxnFinalityStatus};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize, Default)]
+pub enum FinalityStatus {
```
cchudant, line 6: we could just use the other FinalityStatus struct here TBH.

cchudant (changes requested):
very good, nothing to say :)

except a small thing that I'd like you do: now that we have the `ReplacementTransactionUnderpriced` and `FeeBelowMinimum` error variants, I would like you to change the mempool error => submit tx error conversion [here in mc_mempool lib.rs](https://github.com/madara-alliance/madara/blob/9778245c92c43f7ad688fe35773728f6f71ea2c7/madara/crates/client/mempool/src/lib.rs#L86-L114)

and change
```rs
            E::InnerMempool(TxInsertionError::MinTipBump { min_tip_bump }) => rejected(
                ValidateFailure,
                format!("Replacing a transaction requires increasing the tip by at least {}%", min_tip_bump * 10.0),
            ),
```
to
```rs
            E::InnerMempool(TxInsertionError::MinTipBump { min_tip_bump }) => rejected(
                ReplacementTransactionUnderpriced,
                format!("Replacing a transaction requires increasing the tip by at least {}%", min_tip_bump * 100.0),
            ),
```
(also notice the `* 100.0`, this was an error on my part it's supposed to be 100 not 10)

----

As for the new `FeeBelowMinimum`, we also don't return it yet, but it's not clear what this error actually means. So, I guess let's not do anything with it for now.

I already asked the spec authors beginning of july what `FeeBelowMinimum` means, and I haven't received an answer yet. Bumping it now on the slack spec channel.

Since it's talking about **fee**s and not **tip**s it could be blockifier's [`MaxGasAmountTooLow`](https://github.com/starkware-libs/sequencer/blob/c2813b7bfdd82aa9addb4ead6c27a8718df5f19c/crates/blockifier/src/transaction/errors.rs#L20-L35), but it seems weird to make an error type just for that while leaving everything else generic `ValidateError`s.
I suspect they made a typo and they meant "tip below minimum", aka, the tip of your transaction is below the minimum tip of the whole mempool at its current state. In other words, this specific case:
- the mempool is full, meaning, adding a transaction will trigger [the eviction policy](https://github.com/madara-alliance/madara/blob/main/madara/crates/client/mempool/src/inner/mod.rs#L211-L230)
- you try to add a transaction, but the `tip` field of that transaction is too low, it can't replace any current transaction in the mempool. (in our implementation, it means the eviction score of our transaction is higher than the highest eviction score of any current mempool transaction)
- we currently return a `TransactionLimitExceeded` with message `The mempool has reached the limit of {max} transactions` in this case which is probably a UX issue since the user might not understand they just need to bump the tip of their tx and try again
I *think* that's the kind of case that they wanted to capture with this new error message, since I would assume wallets and other API users want to handle this case separately from a regular validate failure. That's also the only way for them to know that they need to bump the tip I think.

----

(As an aside,
it'd be great to move away from using mp_rpc types inside of mc_exec, but since i've revamped mc_exec a bit in my PR, this could be done later 👍)

cchudant (approved):
actually given that we don't know what FEE_BELOW_MINIMUM means i think it's fine to merge as-is and worry about returning the new errors later

madara/crates/primitives/rpc/src/v0_9_0/starknet_ws_api.rs
jbcaron: Yes, but care will have to be taken in the implementation of the methods because some variants are forbidden in the responses.

jbcaron: I stayed as close as possible to the specs

PR #769, opened 2025-08-13
Add dummy nightly build and publish task which is used in PR to be merged in main

prkpndy:
## Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/.github/workflows/pull-request-main-ci.yml
+++ b/.github/workflows/pull-request-main-ci.yml
@@ -169,11 +169,14 @@ jobs:
   #                                MERGE QUEUE                                #
   # ========================================================================= #
 
-  build-nightly:
-    uses: ./.github/workflows/task-do-nothing-build-nightly.yml
+  build-nightly-and-publish-madara:
+    uses: ./.github/workflows/task-do-nothing-build-nightly-and-publish.yml
 
   # test-hive:
   #   uses: ./.github/workflows/task-do-nothing-test-hive.yml
 
-  publish-nightly:
-    uses: ./.github/workflows/task-do-nothing-publish-image.yml
+  build-nightly-and-publish-orchestrator:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-main-ci.yml
```diff
@@ -169,11 +169,14 @@ jobs:
...
-    uses: ./.github/workflows/task-do-nothing-build-nightly.yml
+  build-nightly-and-publish-madara:
+    uses: ./.github/workflows/task-do-nothing-build-nightly-and-publish.yml
 
   # test-hive:
   #   uses: ./.github/workflows/task-do-nothing-test-hive.yml
 
-  publish-nightly:
-    uses: ./.github/workflows/task-do-nothing-publish-image.yml
+  build-nightly-and-publish-orchestrator:
```
byteZorvin, line 178: This would run on every pr? I hope this does not publish images for branches that are yet to be merged, and publish may be incorrect images?

prkpndy: It's using dummy job defined in `task-do-nothing-***.yml` as written in `uses` field. So yes, it won't build the image. We have added this just to satisfy the branch protection rules

Mohiiit (approved):

heemankv (approved):
Looks good, although do we really need a 'do nothing publish file,
Can we not create a general do nothing file and use that ?

prkpndy:
> Looks good, although do we really need a 'do nothing publish file, Can we not create a general do nothing file and use that ?

We can't do that because the job name needs to match the name mentioned in the "real" workflow which also matches the name in branch protection rules

byteZorvin (approved):

PR #774, opened 2025-08-22
feat: Implement SQS message version filtering with receive message filtering

0xvasanth:
## Pull Request type

- Feature
- Testing

The orchestrator uses SQS queues without version filtering, which can lead to message processing conflicts when multiple orchestrator versions are running concurrently. Messages from incompatible versions may be consumed by the wrong orchestrator instance.

Resolves: #NA

## What is the new behavior?

This PR implements version-based message filtering using SQS message attributes and receive message attribute filtering to ensure orchestrator instances only process messages compatible with their version.

Key changes:
- **Version Attribute Tagging**: All messages sent to SQS are tagged with `OrchestratorVersion` message attribute
- **Receive Message Filtering**: Consumer uses SQS ReceiveMessageAttributeNames filtering to only receive messages matching the current orchestrator version
- **Backward Compatibility**: Messages without version attributes (from older orchestrators) are filtered out by the receive filtering mechanism
- **Version Management**: Added `ORCHESTRATOR_VERSION` constant and `ORCHESTRATOR_VERSION_ATTRIBUTE` for consistent version tracking
- **Comprehensive Testing**: 
  - Test for version attribute attachment on message send
  - Integration test validating version-based filtering with incompatible messages
  - Queue message count validation to verify filtering behavior

## Technical Implementation

The solution uses AWS SQS's built-in message attribute filtering at the receive level:
1. Messages are sent with `OrchestratorVersion` attribute set to current version
2. When receiving messages, consumers specify message attribute names filter that matches only compatible versions
3. Incompatible messages remain in the queue and are never delivered to incompatible consumers
4. This approach is more efficient than consuming and nacking incompatible messages

No - This change is backward compatible. Messages without version attributes will simply not be consumed by version-aware consumers.

**Migration Path:**
1. Deploy the new version which will automatically tag all new messages with version attributes
2. Older messages without version attributes will remain in queue until they expire or are processed by older orchestrator versions
3. Gradual rollout is supported as version filtering is applied at receive time

## Testing

Added comprehensive test coverage:
- `test_send_message_with_version_attribute`: Verifies version attribute is added to messages
- `test_version_filtering_consumes_only_compatible_messages`: Integration test validating that only compatible messages are received and incompatible messages remain in queue
- Queue fixture for reusable test setup

## Other information

This enhancement improves system reliability when running multiple orchestrator versions in parallel environments, preventing version mismatch issues and ensuring proper message routing based on orchestrator versions.


orchestrator/src/types/constant.rs
```diff
@@ -40,3 +40,19 @@ lazy_static! {
...
+
+/// Version of the Madara client
+pub const MADARA_VERSION: &str = "0.1.0";
+
+/// Version of the Starknet client
+pub const STARKNET_VERSION: &str = "0.1.0";
+
+/// Version of the Orchestrator
+pub const ORCHESTRATOR_VERSION: &str = "1.0.0";
+
```
heemankv: What is 'Madara Version' and 'Starknet Version' ?
If it's not related to the code-commit of Madara.
Then let's merge these two ?

heemankv: Similarly what is 'Orchestrator Version' here ?

orchestrator/src/types/constant.rs
```diff
@@ -40,3 +40,19 @@ lazy_static! {
     pub static ref TWO: BigUint = 2u32.to_biguint().unwrap();
     pub static ref ONE: BigUint = 1u32.to_biguint().unwrap();
 }
+
+
+/// Version of the Madara client
+pub const MADARA_VERSION: &str = "0.1.0";
+
+/// Version of the Starknet client
+pub const STARKNET_VERSION: &str = "0.1.0";
```
heemankv: Also just madara_version might be enough, since orchestrator is an externsion to madara, we can prefer madara nomenclature IMO

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -205,6 +239,56 @@ impl QueueClient for SQS {
     /// TODO: this should not be need remove this after reviewing the code access for usage
```
heemankv: Is this TODO relevant now ? 
I am unable to understand what this means

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -205,6 +239,56 @@ impl QueueClient for SQS {
...
+            let delivery = consumer.receive().await?;
+
+            // Check if message version matches our orchestrator version
+            if let Err(version_error) = self.verify_message_version(&delivery) {
+                tracing::warn!("Version mismatch detected: {:?}. NACKing message back to queue.", version_error);
+
+                // NACK the message - this will return it to the queue for other consumers
+                if let Err(nack_error) = delivery.nack().await {
+                    tracing::error!("Failed to NACK message: {:?}", nack_error);
+                    // Even if NACK fails, we shouldn't process this mismatched message
```
heemankv: I am interested to know what will happen if NACK fails, I assume the queue will re-init the message to be consumed after a timeout ?

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -205,6 +239,56 @@ impl QueueClient for SQS {
...
+                if let Some(message_group_id) = attributes.get("MessageGroupId").and_then(|v| v.as_str()) {
+                    if message_group_id != our_version {
+                        return Err(QueueError::Generic(format!(
+                            "Version mismatch: expected={}, got={}",
+                            our_version, message_group_id
+                        )));
+                    }
+                    tracing::debug!("Message version verified via MessageGroupId: {}", message_group_id);
+                } else {
+                    tracing::debug!("No MessageGroupId found - accepting message for backward compatibility");
```
heemankv: Nice, let's add a todo here to remove this acceptance after we are sure that all running instances are migrated!

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -1,4 +1,5 @@
 use crate::core::client::queue::QueueError;
+use crate::types::constant::generate_version_string;
```
heemankv: IMO, we should not let the version to be directly accesible, rather have it coming from configs 
so that we can a consistent flow of configurations
Wdyt?

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -87,9 +108,34 @@ impl Resource for InnerSQS {
...
+                        dlq_creation_attributes
+                            .insert(QueueAttributeName::ContentBasedDeduplication, "true".to_string());
+
+                        // Set deduplication scope to message group for DLQ as well
+                        dlq_creation_attributes
+                            .insert(QueueAttributeName::DeduplicationScope, "messageGroup".to_string());
+
+                        // Set FIFO throughput limit to per message group ID for DLQ
+                        dlq_creation_attributes
+                            .insert(QueueAttributeName::FifoThroughputLimit, "perMessageGroupId".to_string());
```
heemankv: Can we extract the `CreationAttributes` to a helper function so as to not have repetation ?

heemankv (changes requested):
I also have some architectural questions, before which I need clarity on usage of and intent behind
- Madara_version
- Starknet_version
- Orchestrator_version

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -167,14 +168,47 @@ impl SQS {
...
+
+        // Get the orchestrator version
+        let version = generate_version_string();
+
+        // Generate a unique deduplication ID using timestamp and payload hash
+        let deduplication_id = format!(
+            "{}_{}",
+            std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_millis(),
+            // Use a simple hash of the payload for additional uniqueness
+            payload.len()
```
apoorvsadana: its using `len` instead of a hash, is that by mistake?

orchestrator/src/types/constant.rs
```diff
@@ -40,3 +40,19 @@ lazy_static! {
...
+
+/// Version of the Orchestrator
+pub const ORCHESTRATOR_VERSION: &str = "1.0.0";
+
+/// Generate a version string combining Starknet and Orchestrator versions
+/// This format allows for future extensibility with additional dependencies
+pub fn generate_version_string() -> String {
```
apoorvsadana:
```suggestion
pub fn get_version_string() -> String {
```

orchestrator/src/core/client/queue/sqs.rs
apoorvsadana: Yes, after the visibility timeout

apoorvsadana (commented):
We will also need another PR to ensure worker doesn't handle SN blocks that it doesn't support?

orchestrator/src/types/constant.rs
apoorvsadana: Madara version can change independent of Starknet version. Imagine a change in admin endpoints in Madara that needs changes in orchestrator. Starknet version remains same.

orchestrator/src/types/constant.rs
```diff
@@ -40,3 +40,19 @@ lazy_static! {
...
+    format!("starknet-{}::orchestrator-{}", STARKNET_VERSION, ORCHESTRATOR_VERSION)
```
apoorvsadana: Will a check on just orchestrator version suffice for us? An orchestrator can only consume a job if it was produced by another orchestrator of the same version.

And, during job creation, we can have a check if the orchestrator version is compatible w/ Starknet block version...

orchestrator/src/types/constant.rs
0xvasanth: As discussed yesterday, there is a possibility where a user might want to switch between different categories of jobs, so in that case, the `ORCHESTRATOR_VERSION` might come in handy.

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: Nope

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -166,15 +167,38 @@ impl SQS {
...
+            .build()
+            .map_err(|e| QueueError::MessageAttributeError(format!("Failed to build message attribute: {}", e)))?;
+
+        let mut send_message_request = self
+            .inner
+            .client()
+            .send_message()
+            .queue_url(&queue_url)
+            .message_body(&payload)
+            .message_attributes("OrchestratorVersion", version_attribute);
```
apoorvsadana: `OrchestratorVersion` should be a constant

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -202,9 +226,101 @@ impl QueueClient for SQS {
...
+        loop {
+            // Receive message using AWS SDK to access attributes
+            let receive_result = self
+                .inner
+                .client()
+                .receive_message()
+                .queue_url(&queue_url)
+                .message_attribute_names("OrchestratorVersion")
+                .max_number_of_messages(1)
+                .wait_time_seconds(20)
```
apoorvsadana: what does `wait_time_seconds` do?

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -202,9 +226,101 @@ impl QueueClient for SQS {
...
+                None => {
+                    tracing::debug!("No message attributes, accepting message (backward compatibility)");
+                    true // Accept messages without any attributes
+                }
+            };
+
+            if should_accept {
+                // Make the message immediately available again by setting visibility to 0
+                // so omniqueue can consume it
+                self.inner
```
apoorvsadana: need to remove this as discussed, we can directly consume the message

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
 
+use crate::types::constant::get_version_string;
+use crate::types::queue::QueueType;
+use aws_sdk_sqs::types::MessageAttributeValue;
+
+/// Test that send_message adds the OrchestratorVersion message attribute
+#[rstest]
+#[tokio::test]
+#[ignore] // Requires AWS credentials and actual queue
+async fn test_send_message_with_version_attribute() {
```
apoorvsadana: we can't do it w/ localstack as an integration test? can push to later if we want.

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
+    assert_ne!(our_version, different_version);
+}
+
+/// Test queue type enumeration
 #[rstest]
 #[tokio::test]
-async fn test_queue() {
-    // TODO: write test case
+async fn test_queue_types() {
+    use strum::IntoEnumIterator;
```
apoorvsadana: i don't see the point of these tests? why do we need to test string assertions for example?

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
+}
+
+#[cfg(test)]
+mod integration_tests {
+    use super::*;
+    use aws_config::{BehaviorVersion, Region};
+    use aws_sdk_sqs::types::MessageAttributeValue as AwsMessageAttributeValue;
+
+    /// Helper to create localstack AWS config
+    async fn get_localstack_config() -> aws_config::SdkConfig {
```
apoorvsadana: should make this a fixture?

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
+            .endpoint_url("http://localhost:4566")
+            .load()
+            .await
+    }
+
+    /// Integration test for full message flow (requires localstack or AWS)
+    #[rstest]
+    #[tokio::test]
+    #[ignore] // Run with: cargo test --lib tests::queue::integration_tests -- --ignored --test-threads=1
+    async fn test_full_message_flow_with_version() {
```
apoorvsadana: the test seems to be testing AWS sqs client and not really our code logic, am I missing something?

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
+
+        // Cleanup
+        sqs_client.delete_queue().queue_url(&queue_url).send().await.ok();
+    }
+
+    /// Integration test for backward compatibility - messages without version attribute
+    #[rstest]
+    #[tokio::test]
+    #[ignore] // Run with: cargo test --lib tests::queue::integration_tests -- --ignored --test-threads=1
+    async fn test_backward_compatibility_no_version_attribute() {
```
apoorvsadana: same as above

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,246 @@
...
+        let version2 = get_version_string();
+
+        // Same process should generate same version
+        assert_eq!(version1, version2);
+    }
+
+    /// Test version string format
+    #[rstest]
+    #[tokio::test]
+    async fn test_version_format() {
```
apoorvsadana: why is this different from test_version_string_generation?

orchestrator/src/types/constant.rs
apoorvsadana: bump

orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -34,6 +34,11 @@ pub struct CommonMetadata {
     pub verification_completed_at: Option<DateTime<Utc>>,
     /// Reason for job failure if any
     pub failure_reason: Option<String>,
+    /// Starknet protocol version associated with the block being processed.
+    /// This version is extracted from the block header returned by the sequencer/full node.
+    /// Format: "0.13.2", "0.13.4", etc.
+    /// Used to ensure version-specific processing logic is applied correctly across the pipeline.
+    pub starknet_version: Option<String>,
```
apoorvsadana: why not an enum for the version?

orchestrator/src/types/constant.rs
0xvasanth: This function is already been updated

orchestrator/src/types/jobs/metadata/mod.rs
0xvasanth: done

orchestrator/src/tests/queue/mod.rs
0xvasanth: reverted this test case again

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: done

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: is timeout basically

orchestrator/src/tests/queue/mod.rs
0xvasanth: can you breif this

orchestrator/src/tests/queue/mod.rs
0xvasanth: done

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -166,15 +167,38 @@ impl SQS {
...
+            .build()
+            .map_err(|e| QueueError::MessageAttributeError(format!("Failed to build message attribute: {}", e)))?;
+
+        let mut send_message_request = self
+            .inner
+            .client()
+            .send_message()
+            .queue_url(&queue_url)
+            .message_body(&payload)
+            .message_attributes(ORCHESTRATOR_VERSION_ATTRIBUTE, version_attribute);
```
apoorvsadana, line 190: shouldn't `ORCHESTRATOR_VERSION_ATTRIBUTE` live inside the sqs file? types/constants should have application specific constants imo, this seems like a sqs specific constant

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,172 @@
 use rstest::*;
 
+/// Test that send_message adds the OrchestratorVersion message attribute using Omni Queue
 #[rstest]
 #[tokio::test]
-async fn test_queue() {
-    // TODO: write test case
+#[ignore]
+async fn test_send_message_with_version_attribute() {
```
apoorvsadana: this is an integration test too right?

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,172 @@
 use rstest::*;
 
+/// Test that send_message adds the OrchestratorVersion message attribute using Omni Queue
 #[rstest]
 #[tokio::test]
-async fn test_queue() {
-    // TODO: write test case
+#[ignore]
```
apoorvsadana: why ignore tests? we've tests in CI w/ localstack right.

orchestrator/src/tests/queue/mod.rs
apoorvsadana: bump, it can be reused in `test_send_message_with_version_attribute`

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,172 @@
...
+            .load()
+            .await
+    }
+
+    /// Integration test for concurrent message production and consumption using Omni Queue
+    /// Tests multiple producer threads sending messages and consumer threads receiving them
+    #[rstest]
+    #[tokio::test]
+    #[ignore]
+    async fn test_concurrent_version_monitoring() {
```
apoorvsadana: is the test name incorrect? 

i thought you will check sending messages w/ different versions and ensure only the right ones get consumed (and incorrect ones get pushed back into the queue). regardless, i think we should've a test case which does this.

orchestrator/src/types/constant.rs
```diff
@@ -40,3 +40,83 @@ lazy_static! {
...
+
+impl std::fmt::Display for StarknetVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+/// Supported Starknet protocol versions by this orchestrator build.
+/// This list must be kept in sync with the SNOS library version (currently v0.13.3).
+pub const SUPPORTED_STARKNET_VERSIONS: &[&str] = &["0.13.0", "0.13.1", "0.13.2", "0.13.3"];
```
apoorvsadana: @Mohiiit @prkpndy, does the current orchestrator version support all these?

orchestrator/src/types/constant.rs
apoorvsadana: Hmm. So what I was actually referring to was something like this

```
pub fn get_version_string() -> String {
    format!("orchestrator-{}", ORCHESTRATOR_VERSION)
}
```

We only need to ensure orchestrator version matches when consuming a message. Is there any advantage of adding a starknet version too?

orchestrator/src/types/constant.rs
heemankv: Only 0.13.2, 0.13.3

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -156,9 +156,15 @@ impl JobTrigger for UpdateStateJobTrigger {
             Layer::L3 => self.collect_paths_l3(&config, &to_process, &mut state_update_metadata).await?,
         }
 
-        // Create StateTransition job metadata
+        let starknet_version = if let Some(first_job) = jobs_to_process.first() {
```
heemankv: Doubt: Are we sure that all the jobs to process will be of the same version as the first job? 
IMO jobs_to_process will be filled with parent process's job and we are fetching it from db which doesn't adhear to version of the orchestrator, so it will give all the jobs irrespective of the orchestrator's version.

This talks about a bigger problem that might arrise whenever we fetch from the db, without specifying the orchestrator version, the orchestrator might be presented with jobs it can't process!

So, I believe to fix this problem properly we should add a filter to the db to give jobs that are only compatible with the orchestrator that are fetching for them!

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -532,11 +569,58 @@ async fn create_jobs_snos(config: Arc<Config>, block_numbers_to_pocesss: Vec<u64
...
+
+    let starknet_version = match block {
+        starknet::core::types::MaybePendingBlockWithTxHashes::Block(block) => block.starknet_version,
+        starknet::core::types::MaybePendingBlockWithTxHashes::PendingBlock(_) => {
+            return Err(eyre!("Block {} is still pending, cannot determine final Starknet version", block_number));
+        }
+    };
+
+    Ok(starknet_version)
+}
```
heemankv, line 610: Can move this to the utils crate.

orchestrator/src/types/constant.rs
heemankv: IMO we should try to utilise the version specified in cargo.toml of `madara` and `orchestrator`.
And enforce updates to them with every merge to main / release.  
Discussed with @apoorvsadana as well.
Let's try this out, if it's not a lot of changes we should use this, else let's add a todo.

orchestrator/src/worker/event_handler/triggers/aggregator.rs
```diff
@@ -47,9 +47,31 @@ impl JobTrigger for AggregatorJobTrigger {
...
+                    warn!(
+                        batch_id = %batch.id,
+                        batch_index = %batch.index,
+                        "Could not fetch Starknet version from child jobs"
+                    );
+                    None
+                }
+            };
+
+            // Construct aggregator job metadata, propagating Starknet version from child proving jobs
```
heemankv: The reason this is safe is because of `batching.rs` making batches of the blocks that this one uses,
Please do checkout : `batching.rs`'s `self.assign_batch_to_blocks(start_block, end_block, config.clone()).await?;` 
and ensure that all the blocks being sent to the batch are of the same version!

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -112,20 +111,16 @@ impl EventWorker {
...
+    /// get_message - Get the next message from the queue with version-based filtering
+    /// This function blocks until a compatible message is available or an error occurs
+    /// Messages with incompatible versions are re-enqueued for other orchestrator instances
     /// It returns a Result<Delivery, EventSystemError> - never returns None
     pub async fn get_message(&self) -> EventSystemResult<Delivery> {
-        let mut consumer = self.consumer().await?;
         loop {
             debug!("Polling for message from queue {:?}", self.queue_type);
-            match consumer.receive().await {
+            match self.config.clone().queue().consume_message_from_queue(self.queue_type.clone()).await {
```
heemankv, line 121: Can you explain these changes once ? 
Why have we removed `fn consumer`
And why are we not using QueueError::NoData anymore ?

orchestrator/src/setup/aws/sqs.rs
```diff
@@ -59,13 +59,23 @@ impl Resource for InnerSQS {
...
+                        .insert(QueueAttributeName::VisibilityTimeout, queue.visibility_timeout.to_string());
+
+                    let res = self
+                        .client()
+                        .create_queue()
+                        .queue_name(&queue_name)
+                        .set_attributes(Some(creation_attributes))
+                        .send()
+                        .await
+                        .map_err(|e| {
```
heemankv, line 73:
```
insert(QueueAttributeName::FifoQueue, "true".to_string());
.insert(QueueAttributeName::ContentBasedDeduplication, "true".to_string());

// Set deduplication scope to message group for DLQ as well
.insert(QueueAttributeName::DeduplicationScope, "messageGroup".to_string());

// Set FIFO throughput limit to per message group ID for DLQ
.insert(QueueAttributeName::FifoThroughputLimit, "perMessageGroupId".to_string());
```

I am confused, what happened to these attributes ?
are we initializing them somewhere else ?

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -185,8 +209,9 @@ impl QueueClient for SQS {
     async fn get_producer(&self, queue: QueueType) -> Result<SqsProducer, QueueError> {
         let queue_name = self.get_queue_name(&queue)?;
```
heemankv, line 208: Is `get_producer` not being used anymore ?

heemankv (changes requested):

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: I am using the constant in single place which is harmless

orchestrator/src/tests/queue/mod.rs
0xvasanth: No, this is just unit testcase which focus on the unit case

orchestrator/src/types/constant.rs
0xvasanth: let me remove the remaining version then

0xvasanth (commented):
I think code have been change since the version logic have been changed , might need to re-review  again

orchestrator/src/worker/event_handler/triggers/update_state.rs
0xvasanth: based on the previous comment conv  i have change this to staknet version that might align, with this comment too

orchestrator/src/worker/event_handler/triggers/snos.rs
0xvasanth: this is not going to been used anywhere to make this inside Utils

0xvasanth (commented):
I think code have been change since the version logic have been changed , might need to re-review again

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: yes , not been used

orchestrator/src/setup/aws/sqs.rs
0xvasanth: No been used since Fifo is not need only need incase if you use Single, other two attribute are only used in Fifo Queue

orchestrator/src/worker/controller/event_worker.rs
0xvasanth: Consumer builder does not support Attribute mapping in Omniqueue

orchestrator/src/tests/queue/mod.rs
0xvasanth: rename to more appropriate

orchestrator/src/tests/queue/mod.rs
0xvasanth: removed

orchestrator/src/tests/queue/mod.rs
0xvasanth: renamed integration_test into test this might be confusing

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,172 @@
...
+        use aws_sdk_sqs::types::MessageAttributeValue;
+        use std::sync::Arc;
+        use tokio::time::{sleep, Duration};
+
+        let config = localstack_config.await;
+        let queue_name = format!("test-snos-job-processing-{}", uuid::Uuid::new_v4());
+
+        // Create queue using InnerSQS (needed for test setup only)
+        let inner_sqs = InnerSQS::new(&config);
+        inner_sqs.client().create_queue().queue_name(&queue_name).send().await.expect("Failed to create queue");
```
apoorvsadana: nitpick, but i would also put this part in a fixture as its common across the test cases

orchestrator/src/tests/queue/mod.rs
```diff
@@ -1,7 +1,172 @@
...
+        let queue_url = inner_sqs.get_queue_url_from_client(&queue_name).await.expect("Failed to get queue URL");
+        inner_sqs.client().delete_queue().queue_url(&queue_url).send().await.ok();
+    }
+
+    /// Integration test for version-based message filtering
+    /// Tests that consumers only receive messages matching their orchestrator version
+    /// and incompatible messages are NOT consumed (remain in queue or go to DLQ)
+    #[rstest]
+    #[tokio::test]
+    async fn test_version_filtering_consumes_only_compatible_messages(
```
apoorvsadana, line 81: fix this test as discussed

0xvasanth: done

orchestrator/src/tests/queue/mod.rs
0xvasanth: done

apoorvsadana (approved):

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -198,13 +221,95 @@ impl QueueClient for SQS {
...
-    /// TODO: this should not be need remove this after reviewing the code access for usage
+    /// Consume a message from the queue with version filtering.
+    ///
+    /// This function filters messages based on the OrchestratorVersion attribute:
+    /// - Accepts messages with matching orchestrator version
+    /// - Accepts messages without version attribute (backward compatibility)
+    /// - Re-enqueues messages with incompatible versions for other orchestrator instances
+    ///
+    /// The filtering ensures that only compatible messages are processed by this
+    /// orchestrator instance, preventing version conflicts in multi-version deployments.
```
heemankv: Can we explain the approach here, + why can we not use omniqueue ?

> Consumer builder does not support Attribute mapping in Omniqueue

heemankv (commented):
Do we not have to ensure that the batcher worker is picking up only compatible snos and batching them ? 
can we batch snos or two sn versions or orchestrator versions ? 

https://github.com/madara-alliance/madara/pull/774#discussion_r2405692050

orchestrator/src/core/client/queue/sqs.rs
0xvasanth: Resolve both comment

    1. **Attribute-Level Filtering Requirements**: We need to filter messages based on the
       `OrchestratorVersion` message attribute to ensure version compatibility. The default
       OmniQ consumer builder does not provide extensibility for attribute-level filtering
       during message consumption.
    
    2. **Server-Side Filtering Limitation**: AWS SQS does not support server-side filtering
       based on message attributes. All filtering must be implemented on the client-side,
       requiring direct access to message metadata before consumption.
    
    3. **Re-enqueue Incompatible Messages**: When a message has an incompatible version, we need to:
       - Delete the message from the queue
       - Re-enqueue it with the same attributes for other orchestrator instances
       - Continue polling for compatible messages
       This workflow requires fine-grained control over message lifecycle that isn't available
       through the standard OmniQ abstraction.

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -198,13 +221,107 @@ impl QueueClient for SQS {
...
+
+        let messages = self
+            .inner
+            .client()
+            .receive_message()
+            .queue_url(&queue_url)
+            .max_number_of_messages(1)
+            .message_attribute_names("All")
+            .visibility_timeout(30)
+            .send()
```
heemankv, line 292: Do we want to make these constants ?

heemankv (approved):

orchestrator/src/tests/workers/batching/mod.rs
```diff
@@ -112,13 +249,82 @@ async fn test_batching_worker(#[case] has_existing_batch: bool) -> Result<(), Bo
...
+        "New batch should start at block 4 (first block after the closed batch), but starts at {}",
+        new_batch.start_block
+    );
+
+    assert_eq!(
+        new_batch.index, 2,
+        "New batch should have index 2 (next after existing batch with index 1), but has index {}",
+        new_batch.index
+    );
+
```
apoorvsadana, line 298: isn't this missing the check that the batch should only have blocks of the same version and that there should be two batches?

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -243,6 +300,28 @@ impl BatchingTrigger {
...
+            if cfg!(test) {
+                tracing::warn!(
+                    start_block = %start_block,
+                    error = %e,
+                    "Failed to fetch Starknet version in test environment, using default 0.13.2"
+                );
+                "0.13.2".to_string()
+            } else {
+                panic!("Failed to fetch Starknet version for block {}: {}", start_block, e);
+            }
```
apoorvsadana: wouldn't mocking be the right approch here as we do it already in tests?

0xvasanth: Yep, this is not needed anymore, since i was trying out different option for detect match in mock instead of something like this `"block_number\":5"`,

orchestrator/src/tests/workers/batching/mod.rs
0xvasanth: I think, creating the bucket, which means it was switching from existing to new, since you have question here, i have created a clear test case, better then before

PR #778, opened 2025-09-01
update: aggregator logic

prkpndy:
## Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

* Update Atlantic Prover Client to add `mockProof` field
* Add a field to toggle storing of audit artifacts


orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -339,4 +347,8 @@ impl AtlanticProverService {
             ))
         }
     }
+
+    fn should_mock_proof(&self) -> bool {
+        self.fact_checker.is_none()
```
Mohiiit: we should be using `atlantic_mock_fact_hash` here from the cli arg? just a clean way I think

orchestrator/src/tests/config.rs
```diff
@@ -644,6 +644,7 @@ pub(crate) fn get_env_params() -> EnvParams {
         server_config,
         snos_layout_name: LayoutName::all_cairo,
         prover_layout_name: LayoutName::dynamic,
+        store_audit_artifacts: false,
```
Mohiiit: let's take this from the env as well

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
prkpndy: done

orchestrator/src/tests/config.rs
prkpndy: done

Mohiiit (approved):

byteZorvin (approved):
lgtm !!

PR #780, opened 2025-09-02
Adding documentation to all `mc-*` crates

Trantorian1:
## Pull Request type

Resolves: #NA

No


madara/crates/client/block_production/src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+//!
+//! # Execution Model
+//!
+//! Block production works by processing transactions which are streamed to it from the [mempool].
+//! For performance reasons, this is separated into a **batching phase**, an **aggregation phase**
+//! and a **pending phase**, which are optimistically parallelized. Effectively, this allows block
+//! production to lag behind on certain tasks (on large blocks for example) while still being able
+//! to make progress wherever it can.
+//!
+//! ## Aggregation Phase
```
Mohiiit:
```suggestion
//! ## Batching Phase
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+//!   as per the constraints set it the chain config, else the block time has elapsed as per the
+//!   constraints set in the chain config. In any of these cases, whenever the [`ExecutorThread`]
+//!   receives this message it will proceed to finalize (seal) the pending block and store it to db
+//!   as a full block.
+//!
+//! ## Pending Phase
+//!
+//! One important detail to note is that the [`PendingBlockState`] kept in the
+//! [`BlockProductionTask`] is stored in **RAM**. We periodically flush this value to db at an
+//! interval defined by the `pending tick` as set in the chain config.
```
Mohiiit, line 65: this I think would be updated now, as per the latest changes by @cchudant

madara/crates/client/devnet/src/lib.rs
```diff
@@ -1,3 +1,47 @@
+//! The Madara Devnet is a local, single node blockchain network which mirrors the functionality of
+//! the Starknet mainnet. This allows you to produce your own local state and experiment on it. It
+//! is also very useful in setting up a sample chain for running tests.
```
Mohiiit, line 3: heads would be nice here that given we are pushing contracts directly to the db and not in the global trie root, it will fail in the snos

although to think about it, I think it shouldn't be hard to push the state-diff to the trie 🤔

madara/crates/client/gateway/client/src/lib.rs
```diff
@@ -1,3 +1,226 @@
...
+//!   "id": 1
+//! }
+//! ```
+//!
+//! ### `get_contract_addresses`
+//!
+//! Returns the on-chain address of the `Starknet` core contract and the `GpsStatementVerifier`
+//! contract on the layer this node's state is settling on (most likely Ethereum). Note that even if
+//! the node is not _producing_ state and is only _synchronizing_ it then this refers to the layer
+//! to which that state is being finalized by the block producers the node in synchronizing from.
```
Mohiiit: the starknet mainnet endpoint now returns:

```
{
    "GpsStatementVerifier": "0xf294781D719D2F4169cE54469C28908E6FA752C1",
    "MemoryPageFactRegistry": "0x5628E75245Cc69eCA0994F0449F4dDA9FbB5Ec6a",
    "MerkleStatementContract": "0xd414f8f535D4a96cB00fFC8E85160b353cb7809c",
    "FriStatementContract": "0x55d049b4C82807808E76e61a08C6764bbf2ffB55",
    "Starknet": "0xE2Bb56ee936fd6433DC0F6e7e3b8365C906AA057",
    "HybridGpsFactAdapter": "0x68cb84164E27cbf65222F604BAef58CC4149FCFC",
    "strk_l2_token_address": "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
    "eth_l2_token_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"
}
```

madara/crates/client/gateway/client/src/lib.rs
```diff
@@ -1,3 +1,226 @@
...
+//!   "method": "trusted_add_validated_transactions",
+//!   "params": {
+//!      // Transaction info goes here
+//!   },
+//!   "id": 1
+//! }
+//! ```
+//!
+//! [`legacy`]: mp_class::ContractClass::Legacy
+//! [`sierra`]: mp_class::ContractClass::Legacy
```
Mohiiit: not sure what's right but it must be something else
```suggestion
//! [`sierra`]: mp_class::ContractClass::Sierra
```

madara/crates/client/gateway/server/src/lib.rs
Mohiiit, line 1: I think the docs here are save as it was in the `gateway/client/src/lib.rs`

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//! 1. **We look for an eviction candidate**: we check the `eviction_queue` for the account with the
+//!    highest eviction score (accounts with transactions furthest from being executable).
+//! 2. **If the new transaction has lower priority** than the worst transaction in the mempool, it
+//!    is rejected.
+//! 3. **Make room**: remove the last transaction from account chosen in step 1.
+//! 4. **Insertion**: The transaction is added to the primary accounts structure
+//! 5. **Update propagation**: All queues are updated to reflect the change
+//!
+//! The eviction score considers both nonce distance (how far from executable) and transaction
+//! score within that distance tier.
```
Mohiiit, line 64: in case of a high tip, does the eviction score reduces?

madara/crates/client/sync/src/lib.rs
```diff
@@ -1,3 +1,141 @@
+//! Madara sync. This crate manages state synchronization for Madara nodes. This includes fetching
+//! blocks, transactions, and state updates from external sources and applying them to the local
+//! database.
+//!
+//! # Overview
+//!
+//! The Sync module orchestrates is responsible for downloading, verifying, and applying blocks
```
Mohiiit:
```suggestion
//! The Sync module is responsible for downloading, verifying, and applying blocks
```

madara/crates/client/rpc/src/lib.rs
```diff
@@ -1,6 +1,789 @@
...
+//!   },
+//!   "id": 1
+//! }
+//! ```
+//!
+//! #### `starknet_getStorageAt`
+//!
+//! [`versions::user::v0_7_1::StarknetReadRpcApiV0_7_1Server::get_storage_at`]
+//!
+//! Returns the value of a storage variable at a given address and key.
```
Mohiiit:
```suggestion
//! Returns the value of a storage variable at a given block, address and key.
```

madara/crates/client/rpc/src/lib.rs
```diff
@@ -1,6 +1,789 @@
...
+//!   },
+//!   "id": 1
+//! }
+//! ```
+//!
+//! #### `starknet_getNonce`
+//!
+//! [`versions::user::v0_7_1::StarknetReadRpcApiV0_7_1Server::get_nonce`]
+//!
+//! Returns the nonce of a contract account.
```
Mohiiit:
```suggestion
//! Returns the nonce of a contract account at given Block ID.
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1,4 +1,78 @@
...
+//! (TODO(mohit 13/10/2025): update this when 0.14.0 merges)
```
heemankv, line 66: We still need to update the doc to mention Pre-Confirmed block instead of PendingBlockState right?

madara/crates/client/block_production/Cargo.toml
```diff
@@ -17,6 +17,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 [dev-dependencies]
 
 rstest = { workspace = true }
+mc-sync = { workspace = true }
```
heemankv: Why do we need this here ?

madara/crates/client/rpc/src/lib.rs
```diff
@@ -1,6 +1,789 @@
-//! Starknet RPC server API implementation
+//! Starknet RPC module. Implements the JSON-RPC server for Madara, providing a standardized
+//! interface for interacting with the Starknet node. This module implements the official Starknet
```
heemankv, line 2: IMO adding docs to all the rpc fns at the top is unnescessarily over-populating the file and it doesn't match the flow of documentation as well, I would rather prefer this to be on top of each fn

heemankv: We can also consideer dumping this to RPC.md and add a linker to that here ?

madara/node/Cargo.toml
```diff
@@ -85,3 +85,36 @@ tracing = { workspace = true }
...
+mp-gateway = { workspace = true }
+mp-oracle = { workspace = true }
+mp-receipt = { workspace = true }
+mp-rpc = { workspace = true }
+mp-state-update = { workspace = true }
+mp-transactions = { workspace = true }
+mp-utils = { workspace = true }
+# Helper
+m-cairo-test-contracts = { workspace = true }
+m-proc-macros = { workspace = true }
```
heemankv: What's the reason to have the dev-deps that are already included in the deps ?

madara/crates/client/block_production/Cargo.toml
heemankv: Removed, Cargo build passes after removing this

madara/node/Cargo.toml
heemankv: Removed, Cargo build passes after removing this

madara/crates/client/db/src/lib.rs
```diff
@@ -1,39 +1,115 @@
...
-//! allowing it to see the partial data it has saved beyond the latest block marked as full.
+//! ## Low-level API
+//!
+//! For the low-level API, there are a few responsibilities to follow. The database can store
+//! partial blocks: blocks are divided into _headers_, _transactions & receipts_, _classes_,
+//! _state diffs_ and _events_. These can be stored individually, so that for example if the node
+//! can store a block's header faster than its other components, it can move on to the next block
+//! and start storing _its_ header. Partial block storage allows the node to make progress in block
+//! sync while minimizing the churn induced by certain heavy operations such as state root
+//! computation.
```
prkpndy, line 38: Didn't exactly understood the use of low level APIs. When and why will we store partial block info in the DB (like just the block header as mentioned in the example)? Also, when we do store partial info, will we store the full info later or we won't do it at all?

madara/crates/client/db/src/lib.rs
```diff
@@ -1,39 +1,115 @@
...
+//! Each of the low-level API functions can be called in parallel, however, [`apply_to_global_trie`]
+//! needs to be called _sequentially_. This is because we cannot update the global trie across
+//! multiple blocks at once. However, parallelism is still used inside of that function -
+//! parallelism within a single block.
+//!
+//! ### Head Status
+//!
+//! Because each block component can be written to at different speeds, we need to keep track of the
+//! advancement of each component stored this way. For example, we might have stored block headers
+//! until block 6 but only have all block transactions and receipts until block 3.
```
prkpndy, line 62: Noob question: Reading this, it seems like storing the full block info is a very heavy write operation and takes a significant amount of time. Is that the case? If yes, what part of it takes time and how much?

madara/crates/client/block_production/src/executor/mod.rs
```diff
@@ -35,8 +35,13 @@ pub enum ExecutorCommand {
 }
 
 #[derive(Debug)]
-/// Executor thread => master task
+/// Actor model messages, sent between the block production and itself to drive the production of
+/// new blocks.
+///
+/// We use this since the block production is parallelized and message passing allows for easy
```
byteZorvin, line 43: Being curious, can you point me where can I read more this parallelization, and how we handle conflicts b/w txns of different blocks.

madara/crates/client/gateway/server/src/lib.rs
```diff
@@ -1,3 +1,367 @@
+//! # Feeder Gateway Server
```
prkpndy, line 1: This part seems to have several duplication from gateway client's doc

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//! - [`eviction_queue`]: Orders accounts by eviction priority when the mempool is full
+//!
+//! # Dynamic Scoring
+//!
+//! Transaction priority is determined by a [`ScoreFunction`] enum which supports two modes:
+//!
+//! - **FCFS mode**: First-come-first-served based on timestamp
+//! - **Tip mode**: Prioritizes by tip amount with minimum bump requirements
+//!
+//! While the scoring function is an enum that could theoretically be swapped at runtime, this
```
prkpndy, line 31: By this you mean that the priority (FCFS or Tip) can be theoretically changed at runtime?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//! 5. **Update propagation**: All queues are updated to reflect the change
+//!
+//! The eviction score considers both nonce distance (how far from executable) and transaction
+//! score within that distance tier.
+//!
+//! # Mempool Updates
+//!
+//! After any change (insertion, removal, or nonce update), the mempool maintains consistency
+//! through a modular update system. This is backed by the [`AccountUpdate`] struct, which describes
+//! what changes took place (which account was affect, which transactions were added/removed).
```
prkpndy, line 70:
```suggestion
//! what changes took place (which account was affected, which transactions were added/removed).
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1,4 +1,78 @@
...
+//! production. The aggregation phase is used to drive updates to the block production state through
+//! an actor model implemented via message passing, where the block production task drives itself to
+//! completion by messaging itself across threads, state updates and method calls. This is handled
+//! by [`process_reply`], which needs to handle the following messages:
+//!
+//! - [`StartNewBlock`]: this message is sent whenever the [`ExecutorThread`] starts a new block and
+//!   it instructs the [`BlockProductionTask`] to clear its [`PendingBlockState`] in preparation for
+//!   the next batch.
+//!
+//! - [`BatchExecuted`]: this message is sent whenever the [`ExecutorThread`] has finished executing
```
byteZorvin, line 50: Can there be multiple batches in a signle block?

prkpndy (commented):
GG work. Very easy to understand. Thanks for this! LGTM, just some questions for my clarity.

madara/crates/client/db/src/lib.rs
```diff
@@ -1,39 +1,115 @@
...
+//! [`BlockNStatus`]: chain_head::BlockNStatus
+//! [`BlockNStatus::set_current`]: chain_head::BlockNStatus::set_current
+//! [`head_status`]: MadaraBackend::head_status
+//! [`headers`]: ChainHead::headers
+//! [`state diffs`]: ChainHead::state_diffs
+//! [`classes`]: ChainHead::classes
+//! [`transactions`]: ChainHead::transactions
+//! [`events`]: ChainHead::events
+//! [`global trie`]: ChainHead::global_trie
+//! [`full block`]: ChainHead::full_block
```
byteZorvin, line 109: I guess the ChainHead mentioned here should be ChainTip ?

madara/crates/client/devnet/src/lib.rs
```diff
@@ -1,3 +1,50 @@
...
+//!
+//! PS: Important point to note here that we are not updating the tries here, hence SNOS won't work
+//! as we won't have storage proofs of the declaration and other transactions here!!!
+//!
+//! Below is a list of the core responsibilities of the Madara devnet.
+//!
+//! ## UDC Deployment
+//!
+//! Madara Devnet is responsible for deploying the **Universal Deployer** smart contract. This
```
byteZorvin, line 12: Is this the new Cairo 1 Udc, or the cairo zero one?

madara/crates/client/analytics/Cargo.toml
```diff
@@ -40,3 +40,6 @@ tracing-core = { workspace = true, default-features = false }
 tracing-opentelemetry = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 url = { workspace = true }
+
+[dev-dependencies]
+mp-utils = { workspace = true }
```
byteZorvin, line 45: Being curious, why was this added in the docs pr. I could not figure it out.

madara/crates/client/gateway/client/src/lib.rs
```diff
@@ -1,3 +1,292 @@
...
+//! - **Full nodes** to forward user transactions to sequencer nodes for block production
+//!
+//! If you're looking for the server-side implementation that **serves** these endpoints, see the
+//! gateway server crate under `gateway/server`.
+//!
+//! # Feeder Gateway
+//!
+//! The **Feeder Gateway**, which you will sometimes see referred to as _fgw_, is an unspecified
+//! endpoint of a Starknet node. It is used for synchronization during block sync from a centralized
+//! source, predating Starknet's upgrade to p2p block sync.
```
byteZorvin, line 18: Does madara have p2p block sync?

madara/crates/client/gateway/server/src/lib.rs
byteZorvin: I guess that is intentional, so that hover docs work, whether you land at `server/lib.rs` or `client/lib.rs` you are able to see the relevant doc

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//! 4. **Space check**: If the mempool is full, we attempt eviction (see next section)
+//! 5. **Insertion**: The transaction is added to the primary accounts structure
+//! 6. **Update propagation**: All queues are updated to reflect the change
+//!
+//! # Transaction Eviction
+//!
+//! When the mempool reaches capacity:
+//!
+//! 1. **We look for an eviction candidate**: we check the `eviction_queue` for the account with the
+//!    highest eviction score (accounts with transactions furthest from being executable).
```
byteZorvin, line 56: This is a bit confusing to me, what do we exactly mean by `furthest from being executable`

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//! 2. **If the new transaction has lower priority** than the worst transaction in the mempool, it
```
byteZorvin, line 57: Is this priority same as the eviction score talked about above?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+//!    is rejected.
+//! 3. **Make room**: remove the last transaction from account chosen in step 1.
+//! 4. **Insertion**: The transaction is added to the primary accounts structure
+//! 5. **Update propagation**: All queues are updated to reflect the change
+//!
+//! The eviction score considers both nonce distance (how far from executable) and transaction
```
byteZorvin, line 63: How is the transaction score calculated?

madara/crates/client/block_production/src/lib.rs
Mohiiit: yes, one block can have multiple batches

madara/crates/client/db/src/lib.rs
Mohiiit: > Didn't exactly understood the use of low level APIs.

low level APIs are usually used to create separation of concern between the business logic and interaction with db, something that I would like to do in the orch code too (there is no separation of concern as of now)

> When and why will we store partial block info in the DB (like just the block header as mentioned in the example)?

when: madara will store the partial information while syncing to db (could be full node or sequencer)
why: imagine you have to store 10 things related to the block, you can either make sure all 10 get stored and then only stores the next block or you can store the next block's header as soon as the current one is done, that way you are making sure your system is not ideal for anything (Ik not a huge optimization but nonetheless one)

> Also, when we do store partial info, will we store the full info later or we won't do it at all?

partial here means some part of the same block and yes we store all the components of the block, partial here doesn't mean we store half the headers or something like that!

madara/crates/client/db/src/lib.rs
Mohiiit: > storing the full block info is a very heavy

it's very relative to be honest, might take a few ms at worst case (given rocks db is on disk) but still this is some optimization done by the kasar team (and a good one)

> what part of it takes time and how much

we know headers are a const size, but txn and txn receipt could be huge (imagine multiple internal calls), I don't have number on me and I never benchmarked it but this makes sense

madara/crates/client/gateway/server/src/lib.rs
Mohiiit: I have tried to create the separation tho, making sure it's mentioned in both server and client side about how it's being used (let me know if this is missing)

madara/crates/client/mempool/src/lib.rs
Mohiiit: yup, seems correct

madara/crates/client/gateway/client/src/lib.rs
Mohiiit: not yet, hence predating

madara/crates/client/mempool/src/lib.rs
Mohiiit: for example you have a txn with nonce 100, although your current nonce is 1, hence the eviction score for this txn would be significant

madara/crates/client/mempool/src/lib.rs
Mohiiit: yeah, imagine every txn getting a score, if queue is full, depending on the score the txn is accepted or rejected 

and when accepted, we have to kick some txn out

madara/crates/client/analytics/Cargo.toml
Mohiiit: not sure, might be a mistake, will remove it if not needed

madara/crates/client/block_production/src/executor/mod.rs
Mohiiit: yup, here: https://github.com/madara-alliance/madara/blob/main/madara/crates/client/block_production/src/lib.rs#L357

Here’s how the flow currently works:
1. The batcher creates batches and sends them to a channel.
2. The executor reads from this channel, executes the batch, and sends a reply to another channel.
3. `process_reply` reads from this reply channel and processes the message accordingly.

While `process_reply` is running (which can take time due to trie calculation and other operations), the execution of the next batch already starts. The state changes from the last batch are stored in the cache, and the next executor run will use this cache to fetch the current state. Any missing data is fetched from the DB during execution.

Since the executor generally runs faster than process_reply, there’s a cap on how far ahead execution can get currently, the maximum difference between batch execution and reply processing is around 10.

madara/crates/client/mempool/src/lib.rs
Mohiiit: you can read more about it here: https://github.com/madara-alliance/madara/blob/main/madara/crates/client/mempool/src/inner/tx.rs#L106

byteZorvin (approved):

madara/crates/client/db/src/lib.rs
prkpndy: gotcha

madara/crates/client/db/src/lib.rs
prkpndy: okay

madara/crates/client/mempool/src/lib.rs
prkpndy: cool

prkpndy (approved):
gg

PR #782, opened 2025-09-04
bootstrapper v2: complete e2e flow

byteZorvin:
## Pull Request type
- Feature
- Adds the connecting structure for the 2 factories in the ongoing bootstrapper-v2 restructure.
- Adds the cli commands that can be run on them which will do the entire setup for the bootstrapping madara chain as an L2


bootstrapper-v2/.env.example
```diff
@@ -0,0 +1,3 @@
+BASE_LAYER_PRIVATE_KEY=abcdefghi
```
apoorvsadana:
```suggestion
BASE_LAYER_PRIVATE_KEY=0xabcd
```

bootstrapper-v2/Cargo.toml
```diff
@@ -1,24 +1,40 @@
...
+serde = { version = "1.0.210", features = ["derive"] }
+serde_json = "1.0.135"
+starknet = "=0.13.0"
+thiserror = "2.0.10"
+tokio = { version = "1.40.0", features = ["full"] }
 
-[lints]
-workspace = true
+# [lints]
+# workspace = true
```
apoorvsadana: do we need to uncomment this?

bootstrapper-v2/Cargo.toml
```diff
@@ -1,24 +1,40 @@
...
 
 [build-dependencies]
-build-version.workspace = true
+build-version = { path = "../build-artifacts" }
```
apoorvsadana, line 50: does this need to change to?

bootstrapper-v2/contracts/ethereum/src/eic/ORIGINAL_AUTHORS.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Modifications Made
+
+### 1. Import Path Adjustments
+
+Import paths have been modified throughout the files to suit the project structure and dependency management.
+
+### 2. Role Initialization Fix
+
+In `ConfigureSingleBridge.sol`, the `RolesLib.initialize()` call has been commented out to
```
apoorvsadana: The current file is camelCase (first `C` is small). Can you fix that. Also, should the file be `ConfigureSingleBridgeEIC.sol`?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -2,32 +2,53 @@
...
+import {IProxy} from "./interfaces/IProxy.sol";
 
 import {IOperator} from "./interfaces/IOperator.sol";
 import {IBridge} from "./interfaces/IBridge.sol";
 import {IRoles} from "./interfaces/IRoles.sol";
 import {IStarknetGovernor} from "./interfaces/IStarknetGovernor.sol";
 import {IProxyRoles} from "./interfaces/IProxyRoles.sol";
 
-contract Factory is Ownable, Pausable, Implementations {
+import {console} from "forge-std/console.sol";
```
apoorvsadana: I think when you create production build, this is automatically skipped. But what's the standard here, should we remove console imports and prints everywhere?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -127,14 +150,14 @@ contract Factory is Ownable, Pausable, Implementations {
     address messagingContract, // coreContract
     address governor
   ) public returns (address) {
-    _requireNotPaused();
```
apoorvsadana, line 130: why did we remove this from everywhere?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -227,15 +258,7 @@ contract Factory is Ownable, Pausable, Implementations {
     IRoles(proxyContract).registerSecurityAdmin(governanceAdmin);
     IRoles(proxyContract).registerUpgradeGovernor(governanceAdmin);
     IRoles(proxyContract).registerAppRoleAdmin(governanceAdmin);
+    IRoles(proxyContract).registerAppRoleAdmin(address(this));
```
apoorvsadana, line 263: don't we need to renounce app role admin give to factory for eth brige, token brige, manager, registry and core contract?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -2,32 +2,53 @@
...
+    address implementation,
+    bytes memory initData
+  ) public {
+    console.log(
+      "Adding implementation: ",
+      implementation,
+      " to proxy: ",
+      proxy
+    );
+    IRoles(proxy).registerUpgradeGovernor(address(this));
```
apoorvsadana, line 36: when is this revoked?

bootstrapper-v2/contracts/ethereum/src/factory/Implementations.sol
```diff
@@ -3,18 +3,16 @@ pragma solidity ^0.8.13;
 
 import "./libraries/DataTypes.sol";
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
-import {Pausable} from "@openzeppelin/contracts/utils/Pausable.sol";
 
-abstract contract Implementations is Ownable, Pausable {
+abstract contract Implementations is Ownable  {
```
apoorvsadana, line 7: why did we remove pausable?

bootstrapper-v2/contracts/ethereum/src/factory/interfaces/IRoles.sol
```diff
@@ -23,4 +23,6 @@ interface IRoles {
   function revokeAppRoleAdmin(address account) external;
```
apoorvsadana:
```suggestion

```

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
```diff
@@ -0,0 +1,38 @@
...
+  function setUp() public {
+    ConfigureSingleBridgeEIC ethBridgeEIC = new ConfigureSingleBridgeEIC();
+    ImplementationContracts
+      memory implementationContracts = ImplementationContracts({
+        ethBridgeEIC: address(ethBridgeEIC),
+        registry: address(0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0),
+        multiBridge: address(0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9),
+        coreContract: address(0x5FbDB2315678afecb367f032d93F642f64180aa3),
+        manager: address(0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512),
+        ethBridge: address(0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9)
```
apoorvsadana, line 30: how do you get these addresses?

bootstrapper-v2/contracts/madara/src/account/contract.cairo
apoorvsadana, line 1: should we add a comment/original_authors.md that the only thing we changed here is the `deploy_contract`?

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
```diff
@@ -105,45 +126,49 @@ mod MadaraFactory {
...
             18.serialize(ref calldata); // decimals
-            0.serialize(ref calldata); // initial_supply
-            0.serialize(ref calldata); // initial_supply_recipient 
+            initial_supply.serialize(ref calldata); // initial_supply
+
+            self.owner.read().serialize(ref calldata); // initial_supply_recipient
             l2_eth_bridge.serialize(ref calldata); // permitted_minter
-            self.owner.read().serialize(ref calldata); // provisional_governance_admin 
+            self.owner.read().serialize(ref calldata); // provisional_governance_admin
             0.serialize(ref calldata); // upgrade_delay
```
apoorvsadana, line 169: which contract can i look at to see if this calldata is right?

bootstrapper-v2/src/config.rs
```diff
@@ -19,8 +20,19 @@ pub struct MadaraConfigOuter {
...
-    Ethereum { rpc_url: String, implementation_addresses: HashMap<String, String> },
-    Starknet { rpc_url: String },
+    Ethereum {
+        rpc_url: String,
+        // Addresses of the implementation contracts behind proxies
+        // Idea is that these can be reused by just deploying the proxy contract
+        // and pointing to the same implementation contract.
+        // This would save gas and cost, as the heavy implementation contracts are only deployed once.
+        // This a map of the implementation contract name to the implementation contract address.
+        implementation_addresses: HashMap<String, String>,
```
apoorvsadana: the key should be an enum instead of a string.

apoorvsadana: and let's stick to one case. our config is split between snake and camel at the moment.

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+                log::info!("Deployed ConfigureSingleBridgeEIC at address: {:?}", address);
+                self.implementation_address.insert(contract.to_string(), address.to_string());
+            } else {
+                let artifact_path = format!("../build-artifacts/starkgate_latest/solidity/{}.json", contract);
+                let address = self
+                    .deploy_contract_from_artifact(&artifact_path)
+                    .await
+                    .with_context(|| format!("Failed to deploy {}", contract))?;
+                log::info!("Deployed {} at address: {:?}", contract, address);
+                self.implementation_address.insert(contract.to_string(), address.to_string());
```
apoorvsadana: only

```
let artifact_path = format!("../build-artifacts/starkgate_latest/solidity/{}.json", contract);
```

needs to be inside the if right? remaining can move outside.

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+
+        // Write the addresses to a JSON file
+        let addresses_json = serde_json::to_string_pretty(&self.implementation_address)?;
+        save_addresses_to_file(addresses_json, &self.addresses_output_path)?;
```
apoorvsadana, line 182: we did this in the previous bootstrapper but its not clean. better to return the results in a struct and the main.rs file writes the final struct to the location.

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+            // let address = self.implementation_address.get(contract).unwrap();
+            if contract == "coreContract" {
+                let artifact_path = "../build-artifacts/cairo_lang/Starknet.json";
+                let address = self
+                    .deploy_contract_from_artifact(artifact_path)
+                    .await
+                    .with_context(|| format!("Failed to deploy {}", contract))?;
+
+                log::info!("Deployed coreContract at address: {:?}", address);
+                self.implementation_address.insert(contract.to_string(), address.to_string());
```
apoorvsadana: aren't we supposed to check if the contact is already present in `implementation_address`? if yes, we DON'T deploy

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    async fn setup(&mut self) -> anyhow::Result<()> {
+        let provider = ProviderBuilder::new()
+            .wallet(self.signer.clone())
+            .connect_http(self.rpc_url.parse().expect("Invalid RPC URL"));
+
+        // Convert implementation addresses to the required format
+        // (to be passed to the factory constructor)
+        let implementations_addresses = serde_json::to_string_pretty(&self.implementation_address)?;
```
apoorvsadana:
```suggestion
```

can this not be skipped?

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        let madara_addresses: serde_json::Value =
+            serde_json::from_str(&madara_addresses_content).context("Failed to parse madara_addresses.json")?;
+
+        let l2_eth_bridge_address = madara_addresses["addresses"]["l2_eth_bridge"]
+            .as_str()
+            .context("ethTokenBridge address not found in madara_addresses.json")?;
+        let l2_erc20_bridge_address = madara_addresses["addresses"]["l2_token_bridge"]
+            .as_str()
+            .context("tokenBridge address not found in madara_addresses.json")?;
+
```
apoorvsadana: `.context` is more acceptablel because this is a script at the end of the day. but i might ask AI to replace all .context w/ `thiserror` library errors, might make the code cleaner and more readable.

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
```diff
@@ -0,0 +1,63 @@
...
+        *self.factory_contract.address()
+    }
+
+    pub async fn setup(
+        &self,
+        core_contract_init_data: CoreContractInitData,
+        operator: Address,
+        governor: Address,
+    ) -> Result<BaseLayerContracts> {
+        let contract_deployed_filter = self.factory_contract.BaseLayerContractsDeployed_filter().watch().await?;
```
apoorvsadana, line 71: rust analyzer is failing for me here but this feels like `.watch().await` will wait for the event before we call `.setup`? is `.watch` just returning a event listener?

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
```diff
@@ -0,0 +1,63 @@
...
+        let tx_hash =
+            self.factory_contract.setup(core_contract_init_data, operator, governor).send().await?.watch().await?;
+        log::info!("Factory setup transaction hash: {:?}", tx_hash);
+
+        // Wait for the BaseLayerContractsDeployed event
+        let mut event_stream = contract_deployed_filter.into_stream().take(1);
+        let event = event_stream
+            .next()
+            .await
+            .ok_or_else(|| anyhow::anyhow!("Failed to receive BaseLayerContractsDeployed event"))?;
```
apoorvsadana: if txn fails then `.send().await?.watch().await?` throws an error?
is it possible to have some timeout to the event stream? if yes, can add a 5min timeout perhaps

bootstrapper-v2/src/setup/base_layer/starknet.rs
```diff
@@ -14,24 +15,27 @@ pub struct StarknetSetup {
...
     }
 }
 
+#[allow(unused_variables)]
+#[async_trait]
 impl BaseLayerSetupTrait for StarknetSetup {
-    fn init(&self) -> anyhow::Result<()> {
+    #[allow(unused_variables)]
+    async fn init(&mut self) -> anyhow::Result<()> {
         Ok(())
```
apoorvsadana: better do make it `unimplemented!("Starknet base layer is not implemented yet")`

bootstrapper-v2/src/utils.rs
```diff
@@ -0,0 +1,148 @@
...
+    account: &SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>,
+) -> Felt {
+    log::info!("sierra_path: {:?}", sierra_path);
+    log::info!("casm_path: {:?}", casm_path);
+    let contract_artifact: SierraClass = serde_json::from_reader(std::fs::File::open(sierra_path).unwrap()).unwrap();
+
+    let contract_artifact_casm: CompiledClass =
+        serde_json::from_reader(std::fs::File::open(casm_path).unwrap()).unwrap();
+    let class_hash = contract_artifact_casm.class_hash().unwrap();
+    let sierra_class_hash = contract_artifact.class_hash().unwrap();
```
apoorvsadana: reverse terminology to normal SN. `class_hash` by default refers to sierra and we say `compiled_class_hash` to specify casm

bootstrapper-v2/src/setup/madara/mod.rs
```diff
@@ -1,25 +1,352 @@
...
+        )
+        .await;
+        let token_bridge_class_hash = token_bridge_class;
+        log::info!("Token bridge contract declared successfully!");
+
+        log::info!("Declaring ERC20 contract...");
+        let erc20_class = declare_contract(
+            "../build-artifacts/starkgate_latest/cairo/ERC20_070.sierra.json",
+            "../build-artifacts/starkgate_latest/cairo/ERC20_070.casm.json",
+            account,
```
apoorvsadana: move sierra <> casm paths to a HashMap and have a function that loops over them and declares all

bootstrapper-v2/src/setup/madara/mod.rs
apoorvsadana, line 1: avoid unwraps if the return type is Result AND if it can actually panic (example of what cannot panic is `Felt::from("0x1").unwrap)`

bootstrapper-v2/src/setup/madara/account_helpers.rs
apoorvsadana, line 1: can this be merged w/ utils? not sure why is an account helper.

bootstrapper-v2/src/setup/madara/bootstrap_account.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    pub fn bootstrap_address() -> Felt {
+        Felt::from_hex("0x424f4f545354524150").unwrap()
+    }
+
+    // A felt representation of the string 'BOOTSTRAP'.
+    pub async fn bootstrap_declare(&self) -> anyhow::Result<()> {
+        let contract_artifact: SierraClass = serde_json::from_reader(
+            std::fs::File::open("contracts/madara/target/dev/madara_factory_contracts_Account.contract_class.json").unwrap(),
```
apoorvsadana: all file paths to constants pls!

bootstrapper-v2/src/setup/madara/bootstrap_account.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        // Get the class hash
+        let class_hash = contract_artifact.class_hash()?;
+
+        // Create a signer from the private key
+        let signer = LocalWallet::from(SigningKey::from_secret_scalar(
+            Felt::from_hex(private_key).context("Invalid private key format")?,
+        ));
+
+        let salt = Felt::from(123); // Salt for deployment
```
apoorvsadana: might be nice to use a readable salt, something like `bootstrap_account_salt`

apoorvsadana (commented):
Some points

- Let's use types for keys of Maps wherever possible
- I think some of the issues could've been caught in a self review
- Code flow is def much better compared to V1!

bootstrapper-v2/contracts/ethereum/src/eic/ORIGINAL_AUTHORS.md
byteZorvin: Updated

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: I agree on removal actually. Updated

bootstrapper-v2/Cargo.toml
byteZorvin: Yes, as `bootstrapper-v2` is not part of the root level workspace, and does not have any workspace.

bootstrapper-v2/Cargo.toml
byteZorvin: `bootstrapper-v2` is explicity excluded from the root workspace. So we cant use this. But now I have replicated the same lint config here too.

bootstrapper-v2/src/utils.rs
byteZorvin: This was a miss. Thanks for pointing out.

bootstrapper-v2/src/setup/base_layer/starknet.rs
byteZorvin: Agreed, updated

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: This wasnt being revoked, updated to revoke just below

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: I had started to face contract size issue while deploying, and then had mulitple optimizations, where I removed this considering the code is stateless, ideal case we should have it anyways.
I have tried to re-add it back right now, and it is working.

bootstrapper-v2/contracts/ethereum/src/factory/Implementations.sol
byteZorvin: Had removed for size issue, considering we had ownable

bootstrapper-v2/contracts/madara/src/account/contract.cairo
byteZorvin: Yes makes sense, added

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
byteZorvin: [This File](https://github.com/starknet-io/starkgate-contracts/blob/45941888479663ac93e898cd7f8504fa9066c54c/src/openzeppelin/token/erc20_v070/erc20.cairo#L110) can be referred

bootstrapper-v2/src/setup/madara/bootstrap_account.rs
byteZorvin: Starknet rs does not have `Felt::from` implemented for `&str` or `String`. It would not be straight forward to implement and even if we achive through some transformation might even beat the point.

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
byteZorvin: Since we dont have the files registry, multibridge etc, we cant really deploy them here. The only workaround I saw was to run the deployment on anvil through rust code and run foundry test with hardcoded adddress and a forked url of the anvil. 
More importantly this is definitely not the kind of tests we would like to have, had been using this testcase mostly to debug the solidity code we wrote, although kept the testcase in the pr too considering this could be useful later on as well.

bootstrapper-v2/src/setup/madara/mod.rs
byteZorvin: This should have been caught by me 😓, updated now

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
byteZorvin: To return to main, the return type would leak from Ethereum implementation to Starknet implementation vice versa. One ugly way is to return Serde::Value in both cases, but it does not seem clean to me either, wdyt ?

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
byteZorvin: It just returns the EventPoller, and the polling is not started yet.
the polling starts with `contract_deployed_filter.into_stream()`

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
byteZorvin: Yes the txn fail will throw an error. 
Added a 5 min timeout

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
byteZorvin: Update with this_error `errors`

bootstrapper-v2/src/setup/madara/account_helpers.rs
byteZorvin: Merged, had it to reduce file len

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: Added a renounce role ✅

bootstrapper-v2/.env.example
apoorvsadana: pending

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: missing in core contract?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
```diff
@@ -100,21 +117,24 @@ contract Factory is Ownable, Pausable, Implementations {
...
       upgradeData
     );
 
-    IOperator(address(coreContractProxy)).registerOperator(operator);
-    IStarknetGovernor(address(coreContractProxy)).starknetNominateNewGovernor(governor);
+    IStarknetGovernor(address(coreContractProxy)).starknetNominateNewGovernor(
+      governor
+    );
     IProxyRoles(address(coreContractProxy)).registerGovernanceAdmin(governor);
+    IProxyRoles(address(coreContractProxy)).registerAppRoleAdmin(address(this));
```
apoorvsadana, line 131: this revoke is still pending?

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
apoorvsadana: added a seperate comment

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
apoorvsadana: fair, can you add a TODO explaining the ideal flow and how we can use this test case for now?

bootstrapper-v2/contracts/madara/src/account/contract.cairo
apoorvsadana: can't find this, can you confirm?

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
```diff
@@ -105,45 +126,49 @@ mod MadaraFactory {
...
             'ETH'.serialize(ref calldata); // symbol
             18.serialize(ref calldata); // decimals
-            0.serialize(ref calldata); // initial_supply
-            0.serialize(ref calldata); // initial_supply_recipient 
+            initial_supply.serialize(ref calldata); // initial_supply
+
+            self.owner.read().serialize(ref calldata); // initial_supply_recipient
             l2_eth_bridge.serialize(ref calldata); // permitted_minter
-            self.owner.read().serialize(ref calldata); // provisional_governance_admin 
+            self.owner.read().serialize(ref calldata); // provisional_governance_admin
```
apoorvsadana, line 168: who's the owner supposed to be for contracts deployed from bridge 🤔

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
apoorvsadana: would be nice to link this too!

bootstrapper-v2/src/config.rs
apoorvsadana: enum seems resolved. on the mix between camel and snake, is it because alloy rs needs it in camel?

bootstrapper-v2/src/setup/base_layer/ethereum/implementation_contracts.rs
```diff
@@ -0,0 +1,47 @@
...
+    BaseLayerFactory,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct ImplementationContractInfo {
+    pub implementation_contract: ImplementationContract,
+    pub artifact_path: &'static str,
+}
+
+pub static IMPLEMENTATION_CONTRACTS_DATA: [ImplementationContractInfo; 6] = [
```
apoorvsadana: why does't this have `BaseLayerFactory`? also, why not have the address path as a property of the enum then

```
pub enum ImplementationContract {
    CoreContract(path),
}
```

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
apoorvsadana: Its an address, so we can define a [u8;32] type and convert to that maybe?i

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
apoorvsadana: pending?

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
apoorvsadana: hmm, so even though this has thiserror now, it doesn't actually solve the purpose I imagined. the code is still filled w/ big strings of error messages and not easy to read. what you ideally want is

- good #[from] implementations so that a `?` is enough and you don't need to map
- the error message being defined in the enum itself using `#[error(`, such that, you don't need to add extra context when you throw it. if you've to, you probably don't have the write enum types. (it's fine in one two types, like `UnexpectedError` enum, but not in most)

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        // Wait for the BaseLayerContractsDeployed event with 5-minute timeout
+        let mut event_stream = contract_deployed_filter.into_stream().take(1);
+        let event = tokio::time::timeout(
+            std::time::Duration::from_secs(300), // 5 minutes
+            event_stream.next()
+        )
+        .await
+        .map_err(|_| anyhow::anyhow!("Timeout waiting for BaseLayerContractsDeployed event after 5 minutes"))?
+        .ok_or_else(|| anyhow::anyhow!("Failed to receive BaseLayerContractsDeployed event"))?;
```
apoorvsadana: nitpick: does `.context` work here? empty `|_|` seems like we can use some other func

bootstrapper-v2/src/setup/madara/class_contracts.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct MadaraClassInfo {
+    pub madara_class: MadaraClass,
+    pub sierra_path: &'static str,
+    pub casm_path: &'static str,
+}
+
+pub static MADARA_CLASSES_DATA: [MadaraClassInfo; 5] = [
```
apoorvsadana: same as the comment on implementation contracts prev

bootstrapper-v2/src/setup/madara/bootstrap_account.rs
apoorvsadana: Can hardcode the hex and write a comment to say what the hex is, wdyt?

bootstrapper-v2/contracts/madara/src/account/contract.cairo
byteZorvin: Added the comment [here](https://github.com/madara-alliance/madara/blob/7d3fc48e8e484410ed57b8b33ac1fcd6fdbaf02a/bootstrapper-v2/contracts/madara/src/account/contract.cairo#L9-L12)

bootstrapper-v2/.env.example
byteZorvin: Oh yes, although with you get invalid hex length with `0xabac`, but anyways the user will have change the key in anycase.

bootstrapper-v2/contracts/ethereum/src/factory/Factory.sol
byteZorvin: Updated

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
byteZorvin: The owner is being used in the calldata for `initial_supply_recipient` and `provisional_governance_admin`. The `initial_supply` is hardcoded to be 0 so `initial_supply_recipient` does not matter. The bridge as `l2_governnance_admin` var which must be set before any enrollment which is used for `provisional_governance_admin`.

bootstrapper-v2/src/config.rs
byteZorvin: The issue is because solidity convention is camel case, while rust is snake. I want to serde the types directly in solidity.

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
byteZorvin: Added

bootstrapper-v2/src/setup/base_layer/ethereum/mod.rs
byteZorvin: Oh yes, added the skipping logic

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
byteZorvin: Currently to use this test, 1. Run anvil. 2. Run the `setup-base` layer command from the rust cli 3. Run the test by using `forge test` (the hardcoded addresses are what will be deployed on anvil). Ideally it would have been better that we could do implementation contract deployment too in foundry test, but anyways we are not writing the test for the implementation contract but for the factory, so this works somehow.

byteZorvin: One otherway to skip running anvil and setup-base command is to test using sepolia fork too, where we can have the implementations pre-deployed

bootstrapper-v2/src/setup/base_layer/ethereum/implementation_contracts.rs
byteZorvin: Did not have the BaselayerFactory here as it was being deployed [here](https://github.com/madara-alliance/madara/blob/d993065c14cad4792cd0c85a58eb9f2661da7259/bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs#L8-L14) which has the several advantages like binding, and function types.

byteZorvin: There is a struct in solidity factory which looks like this.
``` 
struct ImplementationContracts {
  address coreContract;
  address manager;
  address registry;
  address multiBridge;
  address ethBridge;
  address ethBridgeEIC;
}
``` 

In the current structure, the rust types and solidity really work interchangeably. I tried this with suggestion of CoreContract(path), but in that case lot of structure was breaking and was increasing complexity and needed more boilerplate code.

bootstrapper-v2/src/setup/madara/class_contracts.rs
byteZorvin: Trying the other way was increasing complexity, and needed more boilerplate code

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
byteZorvin: Updated

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
```diff
@@ -0,0 +1,94 @@
...
+// Factory contract for deploying other contracts
+sol!(
+    #[allow(missing_docs)]
+    #[sol(rpc)]
+    #[derive(serde::Serialize, serde::Deserialize, Debug)]
+    Factory,
+    "contracts/ethereum/out/Factory.sol/Factory.json"
+);
+
+pub struct FactoryDeploy<P> {
```
apoorvsadana: can we rename this to `FactoryClient` or `DeployedFactory`? and rename `new` to `deploy_new` so that its explicit that this functions deploys a new factory. `new` functions don't do this normally afaik.

bootstrapper-v2/src/setup/base_layer/ethereum/implementation_contracts.rs
apoorvsadana: As discussed, lets

- have `EnumIter`
- impl a function `get_contract_path` on the enum
- remove `BaseLayerFactory` from the enum

bootstrapper-v2/README.md
apoorvsadana, line 1: might need to update this as per latest discussion with mohit

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
apoorvsadana: we should add this in the file!

bootstrapper-v2/contracts/madara/src/madara_factory/contract.cairo
apoorvsadana: do you think the term `owner` here is a little confusing? usually we use it for the contract owner so it seems we're talking about the factory owner. however, here, it's the owner of the deployed token?

bootstrapper-v2/src/error/madara.rs
```diff
@@ -0,0 +1,73 @@
...
+    ClassHashNotFound(String),
+
+    #[error("Required field '{0}' not found in base layer addresses")]
+    MissingBaseLayerAddress(String),
+
+    #[error("Provider error: {0}")]
+    ProviderError(#[from] starknet::providers::ProviderError),
+
+    #[error("Anyhow error: {0}")]
+    AnyhowError(#[from] anyhow::Error),
```
apoorvsadana: check this once - https://chatgpt.com/share/68f22b6c-0d88-8001-ba66-48d5a9063c29

bootstrapper-v2/src/setup/madara/class_contracts.rs
apoorvsadana: same as implementation contracts comment?

bootstrapper-v2/contracts/ethereum/test/Factory.t.sol
byteZorvin: Updated

bootstrapper-v2/src/setup/base_layer/ethereum/factory.rs
byteZorvin: Update

bootstrapper-v2/src/setup/base_layer/ethereum/implementation_contracts.rs
byteZorvin: Updated

bootstrapper-v2/src/error/madara.rs
byteZorvin: Makese sense, removed.

bootstrapper-v2/README.md
byteZorvin: Based on latest discussion, keeping this for later

bootstrapper-v2/src/setup/madara/class_contracts.rs
byteZorvin: Updated similart to EthereumSetup

bootstrapper-v2/src/setup/madara/class_contracts.rs
```diff
@@ -9,38 +16,36 @@ pub enum MadaraClass {
     MadaraFactory,
 }
 
+pub trait GetMadaraArtifacts {
```
apoorvsadana: nitpick: a trait seems to be an overkill for get artifacts method?

apoorvsadana (approved):

apoorvsadana:
closing this in favor of #810

PR #785, opened 2025-09-15
Fix/orchestrator bugs

prkpndy:
## Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-


orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -228,90 +247,68 @@ impl SettlementClient for EthereumSettlementClient {
...
-        let tx_envelope: TxEnvelope = tx_signed.into();
+        #[cfg(not(feature = "testing"))]
+        if let Some(last_settled_block) = self.get_last_settled_block().await? {
+            if last_settled_block >= to_block_num {
+                warn!(
+                    "Contract state ({}) already ahead of the block to be settled ({}). Skipping update state call.",
+                    last_settled_block, to_block_num
+                );
+                // Send a dummy txn hash
+                return Ok(format!("0x{:x}", B256::from_str("0")?));
```
Mohiiit: instead of dummy txn, maybe we can return this statement itself, that there was indeed state already there!

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -449,6 +446,117 @@ impl EthereumSettlementClient {
...
+        // Prepare input bytes for transaction
+        let input_bytes = Self::build_input_bytes(program_output, state_diff).await?;
+
+        // Prepare EIP4844 transaction
+        let tx = TxEip4844 {
+            chain_id,
+            nonce,
+            // we noticed Starknet uses the same limit on the mainnet
+            // https://etherscan.io/tx/0x8a58b936faaefb63ee1371991337ae3b99d74cb3504d73868615bf21fa2f25a1
+            gas_limit: 5_500_000,
```
Mohiiit: let's move this to const too

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -449,6 +446,117 @@ impl EthereumSettlementClient {
...
+    fn add_safety_margin(&self, value: u128, mul_factor: f64) -> u128 {
+        (value as f64 * mul_factor) as u128
+    }
+
+    fn get_next_mul_factor(&self, mul_factor: f64) -> Result<f64> {
+        let min_mul_factor = GAS_PRICE_MIN_INCREMENT_PERCENTAGE * mul_factor;
+        let max_mul_factor = GAS_PRICE_INCREMENT_PERCENTAGE * mul_factor;
+
+        if min_mul_factor > self.max_gas_price_mul_factor {
+            bail!("Gas price multiplier is too high")
```
Mohiiit, line 521: this will panic the orch right?  shall we do that?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -449,6 +446,117 @@ impl EthereumSettlementClient {
...
+        } else {
+            Ok(self.max_gas_price_mul_factor.min(max_mul_factor))
+        }
+    }
+
+    /// Method to send transaction with a retry mechanism
```
Mohiiit: I don't see the retry mechanism in the below function?

orchestrator/src/core/client/lock/mongodb.rs
```diff
@@ -255,15 +255,11 @@ impl LockClient for MongoLockClient {
...
-                        Ok(Some((expires_at - now).num_seconds()))
-                    } else {
-                        Ok(Some(0))
-                    }
+                let now = Utc::now();
+                if lock.expires_at > now {
+                    Ok(Some((lock.expires_at - now).num_seconds()))
                 } else {
-                    Ok(None)
+                    Ok(Some(0))
```
Mohiiit, line 262: just curious, is there a difference in the some(0) or none?

orchestrator/src/types/queue_control.rs
```diff
@@ -64,7 +64,9 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
         QueueConfig {
             visibility_timeout: 300,
             dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
-            queue_control: QueueControlConfig::default_with_message_count(200),
+            queue_control: QueueControlConfig::default_with_message_count(
+                std::env::var("MADARA_ORCHESTRATOR_MAX_CONCURRENT_SNOS_JOBS").unwrap().parse().unwrap(),
```
Mohiiit: are we sure about the unwraps here?

orchestrator/src/types/queue_control.rs
```diff
@@ -82,7 +84,9 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
         QueueConfig {
             visibility_timeout: 300,
             dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
-            queue_control: QueueControlConfig::new(10),
+            queue_control: QueueControlConfig::default_with_message_count(
+                std::env::var("MADARA_ORCHESTRATOR_MAX_CONCURRENT_PROVING_JOBS").unwrap().parse().unwrap(),
```
Mohiiit: ditto

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -224,6 +226,15 @@ impl JobHandlerTrait for StateUpdateJobHandler {
             SettlementContext::Batch(data) => data.to_settle,
         };
 
+        // Return the status from the settlement contract if the layer is L2
+        if config.layer() == &Layer::L2 {
```
Mohiiit, line 239: this can be done for the L3 as well no?

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -415,20 +442,31 @@ impl StateUpdateJobHandler {
...
+        batch_num: u64,
         nonce: u64,
         artifacts: StateUpdateArtifacts,
     ) -> Result<String, JobError> {
+        // Get the batch details for the batch to settle
+        let batches = config.database().get_batches_by_indexes(vec![batch_num]).await?;
+        if batches.is_empty() {
+            return Err(JobError::Other(OtherError(eyre!("Failed to fetch batch {} from database", batch_num))));
+        }
+        let batch = batches.first().unwrap();
```
Mohiiit: let's remove this unwrap? and if it's safe, let's mention that

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: You mean a string saying `state already present`?
I think it's better to keep all in the same format, in case we try to deserialize it into a Hex object or something. `0x000...` anyways signify that the state is already present, no?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: done

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: No. Just sending an error

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: yeah, bad doc string. updated and put the explanation at the right place - `update_state_with_blobs`

orchestrator/src/core/client/lock/mongodb.rs
prkpndy: `Some(number > 0)` - Specifies the time left in seconds before expiry
`Some(0)` - Specifies that time has finished (expiry < now)
`None` - Data is not present

Technically, Some(0) will not happen because it should not be present if expiry < now. But since mongo takes some time to delete thing after expiry (around 1 min), it can happen

orchestrator/src/types/queue_control.rs
prkpndy: Yeah, this will be constructed at the start only. So, in case it panics (because of wrong value or value not present, we should be good)
I have added error messages and also added a default value if the value is not present

orchestrator/src/types/queue_control.rs
prkpndy: ditto

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: Didn't want to touch L3s actually. It can mess up the retry mechanism there since many state updates are present in 1 state job

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: it's safe. have added a comment

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
Mohiiit: hmm, edge case yes, then I will argue if we really need to store this anymore? we can get rid of it right?

orchestrator/src/worker/event_handler/jobs/state_update.rs
Mohiiit: got it, let's add a TODO then?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -223,95 +245,82 @@ impl SettlementClient for EthereumSettlementClient {
...
 
-        match res {
-            Some(_) => {
-                log::info!("Txn hash : {:?} Finalized ✅", pending_transaction.tx_hash().to_string());
-            }
-            None => {
-                log::error!("Txn hash not finalised");
+            match res {
+                Some(_) => {
+                    tracing::info!("✅ Txn hash : {:?} finalized", pending_transaction.tx_hash().to_string());
```
Mohiiit, line 307: is there any difference between log info and tracing info? just curious

orchestrator/src/types/queue_control.rs
```diff
@@ -64,7 +64,9 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
         QueueConfig {
             visibility_timeout: 300,
             dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
-            queue_control: QueueControlConfig::default_with_message_count(200),
+            queue_control: QueueControlConfig::default_with_message_count(
+                get_env_var_or_default("MADARA_ORCHESTRATOR_MAX_CONCURRENT_SNOS_JOBS", "1").parse().expect("MADARA_ORCHESTRATOR_MAX_CONCURRENT_SNOS_JOBS does not have correct value. Should be a whole number"),
```
Mohiiit: given we have a default value, this expect would never happen right?

orchestrator/src/types/queue_control.rs
```diff
@@ -82,7 +84,9 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
         QueueConfig {
             visibility_timeout: 300,
             dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
-            queue_control: QueueControlConfig::new(10),
+            queue_control: QueueControlConfig::default_with_message_count(
+                get_env_var_or_default("MADARA_ORCHESTRATOR_MAX_CONCURRENT_PROVING_JOBS", "10").parse().expect("MADARA_ORCHESTRATOR_MAX_CONCURRENT_PROVING_JOBS does not have correct value. Should be a whole number"),
```
Mohiiit, line 90: ditto

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -223,95 +245,82 @@ impl SettlementClient for EthereumSettlementClient {
...
-        let signature = self.wallet.default_signer().sign_transaction(&mut variant).await?;
-        let tx_signed = variant.into_signed(signature);
-        let tx_envelope: TxEnvelope = tx_signed.into();
+        #[cfg(not(feature = "testing"))]
+        if let Some(last_settled_block) = self.get_last_settled_block().await? {
+            if last_settled_block >= to_block_num {
+                warn!(
+                    "Contract state ({}) already ahead of the block to be settled ({}). Skipping update state call.",
+                    last_settled_block, to_block_num
+                );
```
heemankv: Adding this check here doesn't seem like the best solution IMO 
if we see how this fn is being called 
process_job of state_update.rs calls update_state_for_block which calls update_state_with_blobs or update_state_calldata based on kzg settings

This is an internal handler function that is called after preparing all the data (We fetch snos, program_output, blob_data and process params)
If we get to know that we are supposed to NOT process the block here, that means we have over-computed, 
I would urge for a solution that does this check before processing the data, is that possible ? 
What am I missing ?

heemankv: Ideally this will also solve the problem of having to return back a dummy txn

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -56,6 +59,14 @@ const MAX_TX_FINALISATION_ATTEMPTS: usize = 30;
 const REQUIRED_BLOCK_CONFIRMATIONS: u64 = 3;
 const TX_WAIT_SLEEP_DELAY_SECS: u64 = 60;
 
+// Ethereum Gas Price Estimation
+const GAS_PRICE_MULTIPLIER_START: f64 = 1.2;
```
heemankv, line 62: INITIAL_GAS_PRICE_MULTIPLIER

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Yessir, we are supposed to use tracing only, it preserves span information 
log will not preserve it while we send the event to OTEL

orchestrator/src/cli/settlement/ethereum.rs
```diff
@@ -23,4 +23,7 @@ pub struct EthereumSettlementCliArgs {
     /// The address of the Starknet operator.
     #[arg(env = "MADARA_ORCHESTRATOR_STARKNET_OPERATOR_ADDRESS", long)]
     pub starknet_operator_address: Option<String>,
+
+    #[arg(env = "MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR", long)]
```
heemankv: Can we keep a default value, else we'll have justify this everytime (for e.g I don't think this will be needed in testing but we'll still have to provide it)

orchestrator/src/types/params/settlement.rs
```diff
@@ -27,6 +27,7 @@ impl TryFrom<RunCmd> for SettlementConfig {
                     Address::from_str(&run_cmd.ethereum_settlement_args.l1_core_contract_address.clone().ok_or(
                         OrchestratorError::SetupCommandError("L1 core contract address is required".to_string()),
                     )?)?;
+                let max_gas_price_mul_factor = run_cmd.ethereum_settlement_args.max_gas_price_mul_factor;
```
heemankv: Let's move this to line 59 directly ?

orchestrator/src/types/queue_control.rs
heemankv: Yes, we should have default value and there should not be any need of expect !

heemankv: Also if we really think about it deep down, this is per pod, 
could we name better ? 
Or should we mention in this env's description that this is per instance of orchestrator

heemankv: also 1 is a very low value, we can keep it 5

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -307,37 +318,65 @@ impl JobHandlerTrait for StateUpdateJobHandler {
             }
         }
 
+        // Finally, check the status of the settlement-contract
+        Self::verify_through_contract(&config, &nums_settled, &job.id, &internal_id).await
```
heemankv, line 331: Why are we doing it at the end again ?

orchestrator/src/types/queue_control.rs
prkpndy: Default is for the case when env is not present.
In case it's present but the value is in the wrong format (not a whole number) then the code will panic (which is expected)

orchestrator/src/types/queue_control.rs
prkpndy: ditto

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: We can for L2s. For L3s we still need it as I have not updated it. Let's remove it in the L3s as well and then remove it entirely from the collection. Will add a TODO for it and can pick up later

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: done

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: yes

orchestrator/src/types/queue_control.rs
prkpndy: Made it 5

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: makes sense. updated the code accordingly

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: hmm.. wanted all constants related to gas price to start with `GAS_PRICE`

orchestrator/src/cli/settlement/ethereum.rs
prkpndy: done. default is 1.5

orchestrator/src/types/params/settlement.rs
prkpndy: done

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: for L3s

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: alright, that works too

orchestrator/src/worker/event_handler/jobs/state_update.rs
heemankv: More context would be appreciated. also can we add the justification in code as comment ?

orchestrator/src/types/queue_control.rs
heemankv: The resons we could have a parsing issue here is because we are not passing this from clip provided config, clap already uses usize to define this variable.
Can we scope out the effort required to change the QUEUES variable to be defined from clap provided config ?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Much better, thanks!

heemankv (approved):

heemankv (approved):

Mohiiit (approved):

heemankv (commented):
Need to re-review after queue_control changes in logging pr

heemankv (changes requested):
Need to re-review some parts (queue_control) before we merge this PR

heemankv (approved):

PR #791, opened 2025-09-19
feat: Add blockchain reorganization detection and handling

0xvasanth:
## Summary

This PR implements comprehensive blockchain reorganization (reorg) detection and handling capabilities for Madara, enabling the node to detect and respond to chain reorganizations during synchronization.

### Core Features
- **Reorganization Detection** (`madara/crates/client/sync/src/reorg.rs`)
  - Added `detect_reorg()` function to compare local and gateway block hashes
  - Implemented `find_common_ancestor()` to identify divergence points
  - Enhanced parent hash consistency checks

### Gateway Synchronization Improvements
- **Enhanced Block Fetching** (`madara/crates/client/sync/src/gateway/`)
  - Fixed block fetching issues with improved error handling
  - Added reorg detection integration during sync process
  - Improved logging for block hash comparisons

### Testing Infrastructure  
- **Test Script** (`madara/test_reorg.sh`)
  - Comprehensive bash script to simulate and test reorg scenarios
  - Supports multi-sequencer setup with divergent chains
  - Automated verification of reorg detection and recovery

## Implementation Details

The reorg detection works by:
1. Comparing local block hashes with gateway provider hashes at each sync step
2. When a mismatch is detected, searching backwards to find the common ancestor
3. Logging detailed information about detected reorganizations
4. Providing a foundation for future reorg recovery mechanisms


madara/crates/client/db/src/lib.rs
```diff
@@ -239,11 +241,17 @@ impl<D: MadaraStorage> MadaraBackend<D> {
         }
 
         // Init chain_tip and set starting block
-        let chain_tip = ChainTip::from_storage(if let Some(starting_block) = self.starting_block {
+        let storage_tip = if let Some(starting_block) = self.starting_block {
+            tracing::info!("🔧 Using unsafe_starting_block: {}", starting_block);
```
Mohiiit: having this as warning would be better I think!

madara/crates/client/db/src/lib.rs
```diff
@@ -318,6 +326,96 @@ impl MadaraBackend<RocksDBStorage> {
...
+            if target_block_n >= current {
+                tracing::info!("Target block {} is >= current block {}, nothing to rollback", target_block_n, current);
+                return Ok(());
+            }
+        } else {
+            tracing::warn!("No blocks found in database, cannot rollback");
+            return Ok(());
+        }
+
+        let current_block = current_block.unwrap();
```
Mohiiit: we won't  need to unwrap the current_block again!
```suggestion
       let Some(current_block) = current_block else {
            tracing::warn!("No blocks found in database, cannot rollback");
            return Ok(());
       };

       if target_block_n >= current_block {
           tracing::info!(
                  "Target block {} is >= current block {}, nothing to rollback",
                    target_block_n, current_block
            );
           return Ok(());
      }
```

madara/crates/client/db/src/lib.rs
```diff
@@ -318,6 +326,96 @@ impl MadaraBackend<RocksDBStorage> {
...
+        
+        // Use the revert_to function from RocksDBStorage
+        let state_diffs = self.db.revert_to(target_block_n)?;
+        
+        tracing::info!("📊 Reverted {} blocks, updating chain tip", state_diffs.len());
+        
+        // CRITICAL: Revert the Bonsai tries to maintain state consistency
+        // This is the key fix for the trie log issue
```
Mohiiit:
```suggestion
```

madara/crates/client/db/src/lib.rs
```diff
@@ -318,6 +326,96 @@ impl MadaraBackend<RocksDBStorage> {
...
+        
+        // CRITICAL: Commit all tries after reverting to ensure consistency
+        self.db.contract_trie().commit(target_block_id)
+            .map_err(|e| anyhow::anyhow!("Failed to commit contract trie after revert: {}", e))?;
+        self.db.contract_storage_trie().commit(target_block_id)
+            .map_err(|e| anyhow::anyhow!("Failed to commit contract storage trie after revert: {}", e))?;
+        self.db.class_trie().commit(target_block_id)
+            .map_err(|e| anyhow::anyhow!("Failed to commit class trie after revert: {}", e))?;
+        
+        tracing::debug!("✓ All tries committed at block #{}", target_block_n);
```
Mohiiit: would it be possible to get the time taken in the revert process?

madara/crates/client/db/src/lib.rs
```diff
@@ -318,6 +326,96 @@ impl MadaraBackend<RocksDBStorage> {
...
+        let contract_root = self.db.contract_trie().root_hash(bonsai_identifier::CONTRACT)
+            .map_err(|e| anyhow::anyhow!("Failed to get contract root after rollback: {}", e))?;
+        let class_root = self.db.class_trie().root_hash(bonsai_identifier::CLASS)
+            .map_err(|e| anyhow::anyhow!("Failed to get class root after rollback: {}", e))?;
+        
+        tracing::info!("📊 After rollback to block #{}: contract_root={:#x}, class_root={:#x}",
+            target_block_n, contract_root, class_root);
+        
+        // Update the chain tip to the target block
+        let new_chain_tip = if target_block_n == 0 {
```
Mohiiit: when we say target_block, it means we will retain the state of the target_block right? 

if yes, how do we differentiate between reverting to empty chain vs reverting to block 0?

madara/crates/client/db/src/rocksdb/blocks.rs
```diff
@@ -225,6 +225,117 @@ impl RocksDBStorageInner {
...
+                };
+                if self.db.get_pinned_cf(&block_info_col, block_n_u32.to_be_bytes())?.is_none() {
+                    break;
+                }
+                latest = next_block;
+            }
+            latest
+        };
+
+        let mut state_diffs = Vec::with_capacity((latest_block_n - revert_to) as usize);
```
Mohiiit:
```suggestion
        let mut block_n_to_state_diffs = Vec::with_capacity((latest_block_n - revert_to) as usize);
```

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -205,6 +205,66 @@ impl RocksDBStorage {
...
+
+        // Then revert classes using the collected state diffs
+        if !state_diffs.is_empty() {
+            self.inner.class_db_revert(&state_diffs)
+                .with_context(|| format!("Reverting classes for {} blocks", state_diffs.len()))?;
+        }
+
+        // Optionally revert state trie changes
+        // Note: This may need to be implemented based on your state management needs
+        // self.inner.state_revert(&state_diffs)?;
```
Mohiiit: is this needed?

madara/crates/client/sync/src/gateway/mod.rs
```diff
@@ -147,17 +221,84 @@ impl ForwardPipeline for GatewayForwardSync {
...
 
             let start_next_block = self.pipeline_status().min().map(|n| n + 1).unwrap_or(0);
 
             tokio::select! {
                 Some(res) = self.apply_state_pipeline.next() => {
+                    // Check if we got a GlobalStateRoot mismatch error which might indicate a reorg
+                    if let Err(err) = &res {
+                        let err_msg = format!("{:?}", err);
+                        // Check both the error message and its cause for state root mismatch
+                        if err_msg.contains("Global state root mismatch") || err_msg.contains("GlobalStateRoot") {
```
Mohiiit: do we need an "OR" condition here? Global state root mismatch is more than enough? 

I am worries that re-org will start even when the error is something else 😓

madara/crates/client/sync/src/reorg.rs
```diff
@@ -0,0 +1,164 @@
...
+        let local_info = local_view.get_block_info()?;
+        let local_hash = local_info.block_hash;
+        tracing::debug!(
+            "Checking block #{}: local_hash={:#x}, gateway_hash={:#x}",
+            block_to_check,
+            local_hash,
+            gateway_block.block_hash
+        );
+        
+        if local_hash != gateway_block.block_hash {
```
Mohiiit:
```suggestion
        if local_block_hash != gateway_block.block_hash {
```

madara/crates/client/sync/src/reorg.rs
```diff
@@ -0,0 +1,164 @@
...
+            // This can occur when the local node hasn't synced this far yet
+            continue;
+        }
+
+        let gateway_block = client
+            .get_block(BlockId::Number(check_block))
+            .await
+            .context("Failed to fetch block for common ancestor search")?;
+
+        let local_info = local_block_view.unwrap().get_block_info()?;
```
Mohiiit: let's handle this unwrap gracefully!

madara/crates/client/sync/src/reorg.rs
```diff
@@ -0,0 +1,164 @@
...
+pub async fn find_common_ancestor(
+    start_block: u64,
+    backend: &Arc<MadaraBackend>,
+    client: &Arc<GatewayProvider>,
+) -> anyhow::Result<u64> {
+    let mut check_block = start_block;
+
+    tracing::info!("🔍 Finding common ancestor starting from block {}", start_block);
+
+    while check_block > 0 {
```
Mohiiit: what if the block 0 is also not matching?

madara/node/src/cli/backend.rs
```diff
@@ -62,7 +62,7 @@ pub struct BackendParams {
     /// This is the number of blocks for which you can get storage proofs using the storage proof endpoints.
     /// Blocks older than this limit will not be stored for retrieving historical merkle trie state. By default,
     /// no historical merkle trie state access is allowed.
-    #[clap(env = "MADARA_DB_MAX_SAVED_TRIE_LOGS", long)]
+    #[clap(env = "MADARA_DB_MAX_SAVED_TRIE_LOGS", long, default_value = Some("100"))]
```
Mohiiit, line 65: let's update the comment above as well!

Mohiiit (commented):
We should really add some test cases which test out the different reorg scenarios we have:

1. when sequencer has less number of blocks than the full node
2. when the block hash is not matching

madara/crates/client/db/src/lib.rs
0xvasanth: reformatted

madara/crates/client/db/src/lib.rs
```diff
@@ -316,7 +326,146 @@ impl MadaraBackend<RocksDBStorage> {
...
+        };
+
+        if target_block_n >= current_block {
+            tracing::info!(
+                "Target block {} is >= current block {}, nothing to rollback",
+                target_block_n,
+                current_block
+            );
+            return Ok(());
+        }
```
Mohiiit: what if the target_block_n has the right block hash and the current block hash is not matching?

madara/crates/client/db/src/lib.rs
```diff
@@ -316,7 +326,146 @@ impl MadaraBackend<RocksDBStorage> {
...
+        tracing::info!("📊 Reverted {} blocks, updating chain tip", state_diffs.len());
+
+        self.chain_tip.send_replace(ChainTip::Confirmed(target_block_n));
+        tracing::debug!("✓ Chain tip updated to block #{}", target_block_n);
+
+        tracing::info!("🔨 Rebuilding Bonsai trie state from database after rollback");
+
+        // Force flush and rebuild the trie state by clearing internal Bonsai state
+        // This ensures the tries are completely rebuilt from the database
+        self.rebuild_bonsai_tries_from_database(target_block_n)?;
```
Mohiiit: completely rebuilding? not using the trie logs? maybe this comment is wrong or misleading

madara/crates/client/db/src/lib.rs
```diff
@@ -316,7 +326,146 @@ impl MadaraBackend<RocksDBStorage> {
...
+        self.db.flush()?;
+
+        tracing::info!("✅ Bonsai tries rebuilt successfully for block #{}", target_block_n);
+
+        Ok(())
+    }
+
+    /// Initialize Bonsai trie state for existing data during backend startup
+    /// This prevents state root mismatches when starting with existing blocks
+    fn init_bonsai_tries_for_existing_data(&self, latest_block_n: u64) -> Result<()> {
```
Mohiiit: I don't get the purpose of this function 😓

madara/crates/client/db/src/lib.rs
```diff
@@ -316,7 +326,146 @@ impl MadaraBackend<RocksDBStorage> {
...
+        tracing::info!("🧹 Clearing all Bonsai column families for fresh rebuild...");
+        self.db.clear_bonsai_tries()?;
+
+        // Step 2: Collect all state diffs from genesis to target block
+        let mut state_diffs = Vec::new();
+        for block_n in 0..=target_block_n {
+            if let Some(state_diff) = self.db.get_block_state_diff(block_n)? {
+                state_diffs.push(state_diff);
+            }
+        }
```
Mohiiit: are we sure about this? instead of re-building this, we can simply use the revert_to right? or is this happening in a specific case?

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -205,6 +205,137 @@ impl RocksDBStorage {
...
+    /// # Safety
+    ///
+    /// This method deletes ALL Bonsai trie data. It should only be called:
+    /// 1. During a re-org recovery process
+    /// 2. When you plan to immediately rebuild the tries from state diffs
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the database operation fails.
+    pub fn clear_bonsai_tries(&self) -> Result<()> {
```
Mohiiit: is this the permanent way or is this for a special case when nothing is matching?

Mohiiit (commented):
I think the fundamental approach here is wrong, and we shouldn't reorg the way we are doing it here! we are not using trie logs at all

madara/crates/client/db/src/rocksdb/blocks.rs
```diff
@@ -258,4 +258,106 @@ impl RocksDBStorageInner {
...
+        let block_info_col = self.get_column(BLOCK_INFO_COLUMN);
+        let block_txs_col = self.get_column(BLOCK_TRANSACTIONS_COLUMN);
+        let tx_hash_to_index_col = self.get_column(TX_HASH_TO_INDEX_COLUMN);
+        let block_n_to_state_diff = self.get_column(BLOCK_STATE_DIFF_COLUMN);
+
+        let mut latest_block_n = revert_to_block_n + 1;
+        while self.get_block_info(latest_block_n)?.is_some() {
+            latest_block_n += 1;
+        }
+        latest_block_n -= 1;
```
Mohiiit: given in the `revert_to` function we get the `current_tip` of the chain, if we pass along that info here, we won't need this loop right?

madara/crates/client/db/src/rocksdb/blocks.rs
```diff
@@ -258,4 +258,106 @@ impl RocksDBStorageInner {
...
+            let l1_handler_count = transactions.iter().filter(|v| v.transaction.as_l1_handler().is_some()).count();
+            if l1_handler_count > 0 {
+                tracing::debug!("📦 REORG [block_db_revert]: Removing {} L1->L2 messages from block {}", l1_handler_count, block_n);
+            }
+
+            // TODO: No sure how to implement the same for the L2 Network
+            // self.message_to_l2_remove_txns(
+            //     transactions.iter().filter_map(|v| v.transaction.as_l1_handler()).map(|tx| tx.nonce),
+            //     &mut batch,
+            // )?;
```
Mohiiit, line 340: what's this about? is this still required?

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -470,4 +472,179 @@ impl MadaraStorageWrite for RocksDBStorage {
...
+    /// 10. **Database Flush**: Ensures all changes are persisted to disk
+    ///
+    /// # Notes
+    ///
+    /// * After calling this function, the caller MUST refresh the backend's chain_tip cache
+    ///   by reading from the database, as this function only updates the database state.
+    /// * This is a destructive operation - all blocks after the target block are permanently removed.
+    /// * The function is atomic - if any step fails, the database may be in an inconsistent state.
+    /// ```
+    fn revert_to(&self, new_tip_block_hash: &Felt) -> Result<(u64, Felt)> {
```
Mohiiit, line 517: this function is relevant for major of the cases, although I think it's worth metioning the point that in case nothing in the current chain matches with upstream and we need to revoke everything (ever re-creation of genesis block is needed) this function won't be relevant in that case!

madara/crates/client/db/src/rocksdb/state.rs
```diff
@@ -222,4 +222,86 @@ impl RocksDBStorageInner {
...
+    /// to be reverted in that block.
+    ///
+    /// **Warning:** While not enforced, the following should be true:
+    ///  * Each `StateDiff` should include the entire state for its block
+    ///  * `state_diffs` should form a contiguous range of blocks
+    ///  * that range should end with the current blockchain tip
+    ///
+    /// If this isn't the case, the db could end up storing inconsistent state for some blocks.
+    #[tracing::instrument(skip(self, state_diffs))]
+    pub(super) fn contract_db_revert(&self, state_diffs: &[(u64, StateDiff)]) -> Result<()> {
```
Mohiiit, line 238: we can reuse `state_remove` fn here tho? seems very same tbh!

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -41,6 +41,91 @@ pub struct GatewaySyncSteps {
...
+        loop {
+            if probe_block_n == 0 {
+                tracing::warn!("🔍 Reached genesis block, using it as common ancestor");
+                let genesis_view = self._backend.block_view(&BlockId::Number(0))?;
+                let genesis_info = genesis_view.get_block_info()?;
+                return Ok(genesis_info
+                    .as_closed()
+                    .ok_or_else(|| anyhow::anyhow!("Genesis must be confirmed"))?
+                    .block_hash);
+            }
```
Mohiiit, line 121: ideally we should check even for block 0 if the block hash is matching or not?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -41,6 +41,91 @@ pub struct GatewaySyncSteps {
...
+
+            tracing::debug!("🔍 Probing block {} for common ancestor", probe_block_n);
+
+            // Get what we have stored for this block
+            if let Ok(block_view) = self._backend.block_view(&BlockId::Number(probe_block_n)) {
+                let block_info = block_view.get_block_info()?;
+                if let Some(closed_info) = block_info.as_closed() {
+                    let our_hash = closed_info.block_hash;
```
Mohiiit:
```suggestion
                    let local_block_hash = closed_info.block_hash;
```

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -41,6 +41,91 @@ pub struct GatewaySyncSteps {
...
+    /// * No common ancestor is found (should never happen in practice)
+    ///
+    /// 1. Start from `starting_block_n` and walk backwards
+    /// 2. For each block:
+    ///    - Get our local block hash
+    ///    - Fetch the same block from gateway
+    ///    - Compare the hashes
+    /// 3. If hashes match, we found the common ancestor
+    /// 4. If we reach genesis (block 0), use it as the common ancestor
+    async fn find_common_ancestor(&self, starting_block_n: u64) -> anyhow::Result<mp_convert::Felt> {
```
Mohiiit, line 82: it's worth mentioning that the current algo can be optimized, and currently we are doing O(n) which can be changed to a faster O(logn)

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -63,6 +149,47 @@ impl PipelineSteps for GatewaySyncSteps {
...
+                if block_n > 0 {
+                    // Try to get the parent block's info (only confirmed blocks during gateway sync)
+                    match self._backend.block_view(&BlockId::Number(block_n - 1)) {
+                        Ok(parent_view) => {
+                            let parent_info = parent_view.get_block_info()?;
+                            let incoming_parent_hash = gateway_block.header.parent_block_hash;
+
+                            // For gateway sync, we only deal with confirmed blocks, so use as_closed()
+                            if let Some(closed_parent_info) = parent_info.as_closed() {
+                                let our_parent_hash = closed_parent_info.block_hash;
```
Mohiiit:
```suggestion
                                let local_parent_hash = closed_parent_info.block_hash;
```

madara/crates/client/sync/src/gateway/mod.rs
```diff
@@ -86,7 +89,19 @@ impl GatewayForwardSync {
...
         config: ForwardSyncConfig,
     ) -> Self {
-        let starting_block_n = backend.latest_confirmed_block_n().map(|n| n + 1).unwrap_or(/* genesis */ 0);
+        tracing::info!("🌐 Initializing sync from gateway");
+        // Ensure we flush any pending writes before reading head status
+        if let Err(e) = backend.db.flush() {
+            tracing::warn!("Failed to flush database before reading head status: {}", e);
+        }
+
+        let latest_full_block = backend.latest_block_n().unwrap_or(0);
```
Mohiiit: what if there is no block in db, it will return 0 as the latest full block 👀

madara/crates/client/sync/src/tests/reorg.rs
```diff
@@ -0,0 +1,418 @@
...
+/// 1. Reorg from genesis - Tests reverting back to block 0
+/// 2. Empty orphan chain - Tests reorg when no blocks need to be orphaned
+/// 3. Standard reorg - Tests typical reorg with one orphaned block
+/// 4. Multiple passes - Tests repeated reorgs to ensure robustness
+/// 5. Deep reorg - Tests reorging multiple blocks
+async fn test_reorg(reorg_ctx: ReorgTestContext, #[case] args: ReorgTestArgs) {
+    assert!(args.original_chain_length > 0, "Cannot create an empty chain, we always need at least genesis");
+    assert!(args.passes > 0, "Must have at least one pass");
+    assert!(
+        args.original_chain_length <= 2,
```
Mohiiit: there are 0, 1 and 2 blocks, which means max length would be 3?

I noticed we are doing -1 to the original_chain_length below to get the block anyway, so this makes sense!

```suggestion
        args.original_chain_length <= 3,
```

madara/crates/client/sync/src/tests/reorg.rs
```diff
@@ -0,0 +1,418 @@
...
+/// by simulating various fork scenarios where the chain needs to revert
+/// to a common ancestor and rebuild from an alternative fork.
+///
+/// Test scenarios:
+/// 1. Reorg from genesis - Tests reverting back to block 0
+/// 2. Empty orphan chain - Tests reorg when no blocks need to be orphaned
+/// 3. Standard reorg - Tests typical reorg with one orphaned block
+/// 4. Multiple passes - Tests repeated reorgs to ensure robustness
+/// 5. Deep reorg - Tests reorging multiple blocks
+async fn test_reorg(reorg_ctx: ReorgTestContext, #[case] args: ReorgTestArgs) {
```
Mohiiit, line 117: I didn't get this test, here is the flow as per me:

1. there is an original chain test build (thinking it like starknet sepolia/mainnet)
2. there would be an another chain (its called orphaned chain?)
3. test is syncing the original chain to a certain length
4. then in the loop (with number of passes), test would try to revert but my doubt is, given we are using the same backend (which means the same db) so technically we are reverting the original chain only no?

PS: I think you can ignore the original chain altogether I think, the flow could be simply:

1. you have n blocks which would match in your chain (let's say total blocks are t), and there are x blocks you can fetch from the gateway and ofc (x>=t)
2. the matching n blocks would have the right hashes, the rest of (t-n) blocks data won't match
3. you can then try reverting on these t-n blocks and see if things are working fine

madara/crates/client/db/src/rocksdb/blocks.rs
0xvasanth: Yes, you’re right — I wasn’t denying that part. The original objective of this function is to revert the chain to a specific node, regardless of which node it was on.

However, the function’s namespace might not fully align with that objective if we start passing the current_tip. It would make the function indirectly dependent on revert_to, which isn’t ideal for its intended purpose.

That said, if the goal is simply to avoid the loop, we can definitely refactor it accordingly — we just need to ensure we’re not breaking the function’s isolation or introducing tight coupling.

madara/crates/client/db/src/rocksdb/blocks.rs
0xvasanth: In the previous version, there wasn’t a distinction between L1 and L2 transaction removals. In the latest version, we now have two separate functions:
	1.	For removing transactions from L2
	2.	For the L1 handler

I’ve implemented the L1 part, but for L2, I’m not entirely sure when or whether the remove_txns should be triggered. So for now, I’ve left a comment as a reminder to confirm how we want to handle and validate L2 transaction removal before proceeding.

madara/crates/client/sync/src/gateway/mod.rs
0xvasanth: yes ser

madara/crates/client/db/src/rocksdb/state.rs
0xvasanth: Yes ser I have used in the initial commit reverted back, while debugging another issue, will do the change

madara/crates/client/sync/src/tests/reorg.rs
0xvasanth: removed

madara/crates/client/sync/src/tests/reorg.rs
0xvasanth: got it

madara/crates/client/sync/src/tests/reorg.rs
0xvasanth: this is not valid anymore

madara/crates/client/sync/src/gateway/mod.rs
Mohiiit: we shouldn't do that right? because seem contradicting info to me, we are saying starting block is 0 and the latest block in the db is also 0?

or is there a diff between latest block and latest confirmed block?

Mohiiit (approved):
LGTM!

build-artifacts/src/lib.rs
```diff
@@ -131,7 +131,6 @@ fn get_artifacts(root: &RootDir, artifacts: &VersionFileArtifacts) -> Result<(),
 
     let version = get_version(artifacts)?;
     let image = format!("ghcr.io/madara-alliance/artifacts:{version}");
-    println!("cargo::warning=fetching artifacts from image: {}", image);
```
heemankv, line 134: Let's keep this, helps in insuring the correct artifacts are being used!

madara/crates/client/sync/src/gateway/mod.rs
```diff
@@ -86,7 +89,19 @@ impl GatewayForwardSync {
         client: Arc<GatewayProvider>,
         config: ForwardSyncConfig,
     ) -> Self {
-        let starting_block_n = backend.latest_confirmed_block_n().map(|n| n + 1).unwrap_or(/* genesis */ 0);
+        tracing::info!("🌐 Initializing sync from gateway");
+        // Ensure we flush any pending writes before reading head status
+        if let Err(e) = backend.db.flush() {
+            tracing::warn!("Failed to flush database before reading head status: {}", e);
+        }
```
heemankv: Which specific case has this been handled for ?

madara/crates/client/db/src/rocksdb/meta.rs
```diff
@@ -139,7 +139,10 @@ impl RocksDBStorageInner {
             }
         };
 
-        self.db.write_opt(batch, &self.writeopts_no_wal)?;
+
+        let mut writeopts = rocksdb::WriteOptions::default();
```
heemankv: An explaination on this would be appreciated

madara/crates/client/sync/src/import.rs
```diff
@@ -522,13 +531,21 @@ impl BlockImporterCtx {
...
             self.backend.write_access().apply_to_global_trie(block_range.start, state_diffs).map_err(|error| {
                 BlockImportError::InternalDb { error, context: "Applying state diff to global trie".into() }
             })?;
 
         self.backend.write_latest_applied_trie_update(&block_range.end.checked_sub(1))?;
 
+        tracing::info!(
+            "✅ State diff applied successfully for blocks {:?}, latest_applied_trie_update set to {}",
+            block_range,
+            last_block_n
```
heemankv: Do we really need these as info logs, these will clutter the syncing logging process, let's change to debug

madara/crates/client/sync/src/gateway/mod.rs
heemankv: We have a code that removes all the pending blocks when madara starts, hence unsure what this will flush

madara/crates/client/sync/src/gateway/mod.rs
heemankv: What's the latest on this ?

madara/crates/client/sync/src/gateway/mod.rs
```diff
@@ -112,7 +127,64 @@ impl GatewayForwardSync {
...
+            self.config.classes_batch_size,
+        );
+        self.apply_state_pipeline = super::apply_state::apply_state_pipeline(
+            self.backend.clone(),
+            self.importer.clone(),
+            starting_block_n,
+            self.config.apply_state_parallelization,
+            self.config.apply_state_batch_size,
+            self.config.disable_tries,
+        );
```
heemankv: This seems to be written twice now, can we make a inner fn out of this ?

heemankv (approved):

PR #793, opened 2025-09-23
Fix/graceful shutdown

byteZorvin:
## Pull Request type

- Bugfix

- Fixes graceful shutdown timeout issue
- No self healing for Prooving jobs
- Atlantic error propagation

Resolves: #NA

## What is the new behavior?

-
-
-

 - No


orchestrator/src/main.rs
```diff
@@ -90,6 +91,10 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
 
     info!("Initiating orchestrator shutdown sequence (triggered by: {})", shutdown_signal);
 
+    let shutdown_timeout = get_env_var_or_default(
+        "MADARA_ORCHESTRATOR_SHUTDOWN_TIMEOUT",
+        DEFAULT_SHUTDOWN_TIMEOUT_SECS.to_string().as_str(),
+    );
```
0xvasanth: Could we consider exposing this through CLI arguments instead of only via environment variables?

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -114,7 +114,9 @@ impl AtlanticClient {
             let response_text = response.bytes().await.map_err(AtlanticError::GetJobArtifactsFailure)?;
             Ok(response_text.to_vec())
         } else {
-            Err(AtlanticError::AtlanticService(response.status()))
+            Err(AtlanticError::AtlanticService(
+                response.text().await.unwrap_or_else(|_| "No response text available".to_string()),
+            ))
```
0xvasanth: Unwrapping the empty response with a contextual message is a good improvement . That said, using a generic message like `No response text available` might make it harder to trace where this error originated. Could we add more context to the message for clarity?

0xvasanth: can you update your error message in all other message also

orchestrator/src/main.rs
```diff
@@ -19,7 +20,7 @@ static A: jemallocator::Jemalloc = jemallocator::Jemalloc;
 /// Default graceful shutdown timeout in seconds
 /// This matches Docker's default graceful shutdown period and provides sufficient time
 /// for workers to complete their current tasks before forcing termination
-const DEFAULT_SHUTDOWN_TIMEOUT_SECS: u64 = 300;
+const DEFAULT_SHUTDOWN_TIMEOUT_SECS: u64 = 120;
```
0xvasanth:
```suggestion
const DEFAULT_SHUTDOWN_TIMEOUT_SECS: &str = "120";
```

orchestrator/src/main.rs
```diff
@@ -90,6 +91,10 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
 
     info!("Initiating orchestrator shutdown sequence (triggered by: {})", shutdown_signal);
 
+    let shutdown_timeout = get_env_var_or_default(
+        "MADARA_ORCHESTRATOR_SHUTDOWN_TIMEOUT",
+        DEFAULT_SHUTDOWN_TIMEOUT_SECS.to_string().as_str(),
```
0xvasanth:
```suggestion
    let shutdown_timeout = get_env_var_or_default(
        "MADARA_ORCHESTRATOR_SHUTDOWN_TIMEOUT",
        DEFAULT_SHUTDOWN_TIMEOUT_SECS,
```

0xvasanth (changes requested):

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -77,6 +77,11 @@ pub struct AtlanticClient {
 }
 
 impl AtlanticClient {
+    /// Helper method to extract error text from response with context
+    async fn extract_error_text(response: reqwest::Response, operation: &str) -> String {
+        response.text().await.unwrap_or_else(|_| format!("No response text available for {} operation", operation))
+    }
+
```
heemankv: Seems like a generalized function, IMO it would not be the best solution to keep it here, can we shift it to utils ?

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -77,6 +77,11 @@ pub struct AtlanticClient {
 }
 
 impl AtlanticClient {
+    /// Helper method to extract error text from response with context
+    async fn extract_error_text(response: reqwest::Response, operation: &str) -> String {
+        response.text().await.unwrap_or_else(|_| format!("No response text available for {} operation", operation))
```
heemankv: Let's just say "Failed with Status : XYZ" instead of "No response text available for" WDYT ?

orchestrator/crates/prover-clients/atlantic-service/src/error.rs
```diff
@@ -17,7 +16,7 @@ pub enum AtlanticError {
     SubmitL2QueryFailure(#[source] reqwest::Error),
 
     #[error("Atlantic service returned an error {0}")]
-    AtlanticService(StatusCode),
+    AtlanticService(String),
```
heemankv: Ideally we would want to see both right ?

heemankv: `AtlanticService(StatusCode, Option<StatusMsg>)`

heemankv: PS. Ignore the ":"

orchestrator/src/utils/signal_handler.rs
```diff
@@ -131,13 +131,25 @@ impl SignalHandler {
...
-        match tokio::time::timeout(timeout_duration, shutdown_future).await {
+        // Calculate remaining time to ensure we wait at least the full timeout duration
+        let elapsed = start_time.elapsed();
+        let remaining_time = timeout_duration.saturating_sub(elapsed);
+
+        match shutdown_result {
             Ok(Ok(())) => {
                 info!("✅ Graceful shutdown completed successfully");
+                if remaining_time > tokio::time::Duration::from_secs(0) {
+                    info!("Waiting remaining {:?} to ensure full timeout duration", remaining_time);
```
heemankv: full timeout duration -> graceful shutdown

heemankv (changes requested):

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
0xvasanth: The key point here is that when Atlantic doesn’t return an error message, this unwrap gets invoked. In that case, showing only "Failed with Status: XYZ" might not provide enough clarity about which function or operation triggered the issue. Adding more context would make debugging easier.

0xvasanth (approved):

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
heemankv: Yeah let's do "Failed with Status: XYZ, for ABC operation"
WDYT ?

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
byteZorvin: Makes sense, updating it

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
byteZorvin: Since we are going to add the status code back as the tuple field, `(Status code, "Atlantic request failing while during {} operation, no error response returned")` should be fine

orchestrator/crates/prover-clients/atlantic-service/src/error.rs
```diff
@@ -17,7 +17,7 @@ pub enum AtlanticError {
     SubmitL2QueryFailure(#[source] reqwest::Error),
 
     #[error("Atlantic service returned an error {0}")]
-    AtlanticService(StatusCode),
+    AtlanticService(StatusCode, String),
```
heemankv: Nitpick 
But I would recommend writing it like this : 

```
type StatusMessage = String;

#[error("Atlantic service returned an error {0}")]
AtlanticService(StatusCode, StatusMessage),
```

orchestrator/crates/utils/src/http_client.rs
```diff
@@ -323,6 +324,23 @@ impl<'a> RequestBuilder<'a> {
...
+/// * `operation` - The operation name for context in error messages
+///
+/// # Returns
+/// * `String` - The error text from the response or a fallback message
+pub async fn extract_http_error_text(response: Response, operation: &str) -> (String, StatusCode) {
+    let status = response.status();
+    let text = response
+        .text()
+        .await
+        .unwrap_or_else(|_| format!("Atlantic request failing while during {} operation", operation));
```
heemankv: "Atlantic request fails during {} operation - no error response provided"

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
heemankv: "Atlantic request fails during {} operation - no error response provided"

heemankv (changes requested):

heemankv (approved):

PR #805, opened 2025-10-07
feat: add SNOS batching support for L3s to make it 0.14.0 compatible

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, No batching happens for L3s. PIEs/Proofs are created for individual blocks.

Resolves: #NA

## What is the new behavior?

Updated the batching worker to create SNOS batches in case of L3s

### Batching Worker

Batching worker works differently for L2s and L3s

For L2s, we create both aggregator and snos batches.
Snos batches are then used by the snos job to create PIEs for batches instead of single blocks.
Aggregator batches (which essentially contains multiple snos batches and hence many blocks)
are used in proof creation job to create a single aggregator pie and a single proof which is
then registered on chain and the state update can be done for multiple blocks at once.

For L3s, we create only the snos batch.
Snos jobs are created for the generated snos batches.
Then these PIEs are sent for proof generation and registration on chain.
Finally, we can do the state update for multiple blocks (which are present in the snos batch)
at once.

Yes


Mohiiit (commented):
let's add some test for the helper function that we have added, apart from it, looks good!!

heemankv:
@claude  can you review this PR extensively ?

heemankv (approved):

heemankv:
Please do resolve the issues from Claude review tho!

heemankv (dismissed):
Lots of changes have been made after my approval, I'll re-review

.github/workflows/pull-request-merge.yml
```diff
@@ -44,16 +44,7 @@ jobs:
...
-    with:
-      image-name: bootstrapper
-      image-file: ./bootstrapper/Dockerfile
-    permissions:
-      contents: read
-      packages: write
-      attestations: write
-      id-token: write
-    secrets: inherit
+    uses: ./.github/workflows/task-do-nothing-build-nightly-and-publish.yml
```
apoorvsadana: let's keep this, we need bootstrapper v1 for now

.github/workflows/task-build-binary.yml
```diff
@@ -50,6 +50,12 @@ jobs:
         with:
           python-version: ${{ env.BUILD_PYTHON_VERSION }}
 
+      - name: Python Cairo setup (for apollo_starknet_os_program dependencies)
+        if: inputs.is-standalone == false
+        uses: ./.github/actions/setup-python-cairo
+        with:
+          python-version: ${{ env.BUILD_PYTHON_VERSION }}
+
```
apoorvsadana: duplicate?

Makefile
```diff
@@ -432,12 +432,12 @@ git-hook:
...
 setup-l2:
 	@echo -e "$(DIM)Setting up orchestrator with L2 layer...$(RESET)"
-	@cargo run --package orchestrator -- setup --layer l2 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type rule
+	@cargo run --package orchestrator -- setup --layer l2 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type schedule
 
 .PHONY: setup-l3
 setup-l3:
 	@echo -e "$(DIM)Setting up orchestrator with L3 layer...$(RESET)"
-	@cargo run --package orchestrator -- setup --layer l3 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type rule
+	@cargo run --package orchestrator -- setup --layer l3 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type schedule
```
apoorvsadana, line 447: is this because of a localstack change?

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -48,21 +48,28 @@ pub struct AtlanticBucketInfo {
     pub bucket_job_index: Option<u64>,
 }
 
+struct CustomizationParams {
```
apoorvsadana:
```suggestion
struct ProvingParams {
```

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -48,21 +48,28 @@ pub struct AtlanticBucketInfo {
...
+    /// Layout to be used
+    layout: LayoutName,
+}
+
 trait ProvingLayer: Send + Sync {
-    fn customize_request<'a>(&self, request: RequestBuilder<'a>) -> RequestBuilder<'a>;
+    fn customize_request<'a>(&self, request: RequestBuilder<'a>, params: CustomizationParams) -> RequestBuilder<'a>;
```
apoorvsadana:
```suggestion
    fn add_proving_params<'a>(&self, request: RequestBuilder<'a>, params: CustomizationParams) -> RequestBuilder<'a>;
```

orchestrator/src/utils/constants.rs
```diff
@@ -1,5 +1,5 @@
 /// State Update
 ///
 /// Number of max state update happen in a single job
-pub const STATE_UPDATE_MAX_NO_BLOCK_PROCESSING: usize = 10;
+pub const STATE_UPDATE_MAX_NO_BLOCK_PROCESSING: usize = 1;
```
apoorvsadana, line 4: was the the previous value outdated?

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -322,18 +322,29 @@ impl StateUpdateJobHandler {
...
             Layer::L2 => {
                 // Get the batch details for the last-settled batch
                 let batch = config.database().get_aggregator_batches_by_indexes(vec![*last_settled]).await?;
                 if batch.is_empty() {
                     Err(JobError::Other(OtherError(eyre!("Failed to fetch batch {} from database", last_settled))))
                 } else {
                     // Return the end block of the last batch
-                    Ok((batch[0].end_block, batch.first().cloned()))
+                    // unwrap is safe since we checked that it's not empty
+                    Ok((batch[0].end_block, batch.first().unwrap().index))
```
apoorvsadana: not specific to this commit but usually cleaner to implement this type of stuff with

```
if let Ok(batch) = batch {} else { // throw error }
```

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -75,56 +97,20 @@ impl JobTrigger for BatchingTrigger {
...
-        // Check if any existing batch needs to be closed
-        if let Some(aggregator_batch) = &latest_aggregator_batch {
-            if let Some(snos_batch) = &latest_snos_batch {
-                self.check_and_close_batches(&config, aggregator_batch, snos_batch).await?;
-            } else {
-                return Err(eyre!("Aggregator and SNOS batches are out of sync. We have an Aggregator batch ({}) in the DB but no SNOS batch", aggregator_batch.index));
+        // Invoking method to assign batches to all the blocks in the range
+        if start_block < end_block {
+            match config.layer() {
+                Layer::L2 => self.assign_batch_to_blocks(start_block, end_block, &config).await?,
```
apoorvsadana: let's do a good to have linear task with things we discussed on call

orchestrator/src/worker/event_handler/jobs/state_update.rs
byteZorvin: But the `batch.empty()` does not return an option 🤔

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -322,18 +322,29 @@ impl StateUpdateJobHandler {
...
+            }
+            Layer::L3 => {
+                // Get the batch details for the last-settled batch
+                let batch = config.database().get_snos_batches_by_indices(vec![*last_settled]).await?;
+                if batch.is_empty() {
+                    Err(JobError::Other(OtherError(eyre!("Failed to fetch batch {} from database", last_settled))))
+                } else {
+                    // Return the end block of the last batch
+                    // unwrap is safe since we checked that it's not empty
+                    Ok((batch[0].end_block, batch.first().unwrap().snos_batch_id))
```
byteZorvin: `batch.first().unwrap()` is same as batch[0] right ? If yes, it was a bit confusing to me, may be we can use same impl for both values

orchestrator/src/utils/constants.rs
prkpndy: since we can do multiple blocks in a single batch now, having multiple batches in a single state update job does complicate things in practice. we have done a similar thing for L2s as well after we faced issues while doing the load test

Makefile
prkpndy: no. i think it's better to use schedule since it's used with AWS. have to update this everytime when trying to setup aws for testing. wdyt?

.env.test
```diff
@@ -98,6 +98,7 @@ ETHEREUM_SETTLEMENT_TXN_WAIT_SLEEP_DELAY_SECS=60
 #### SERVICE ####
 
 MADARA_ORCHESTRATOR_MADARA_RPC_URL=https://pathfinder-madara-ci.d.karnot.xyz
+MADARA_ORCHESTRATOR_MADARA_ADMIN_RPC_URL=https://pathfinder-madara-ci.d.karnot.xyz
```
byteZorvin, line 101: Why do we need the admin rpc also now?

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -116,22 +116,10 @@ impl JobTrigger for UpdateStateJobTrigger {
...
-                        let min_block_to_process = config.service_config().min_block_to_process;
-                        if to_process[0] != min_block_to_process {
-                            warn!("DA job for the first block is not yet completed. Returning safely...");
-                            return Ok(());
-                        }
-                    }
+                // if the last processed batch is not there, (i.e., this is the first StateTransition job), check if the batch being processed is equal to 1
+                if to_process[0] != 1 {
+                    warn!("Parent job (Aggregator/DA) for the first batch is not yet completed. Can't proceed with batch {}, Returning safely...", to_process[0]);
+                    return Ok(());
```
byteZorvin, line 122: This might be unrelated still asking, Do we now only pickup on blocks of matching starknet version, like 14.0 blocks only?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -131,28 +133,21 @@ pub async fn fetch_block_starknet_version(config: &Arc<Config>, block_number: u6
...
-/// # Arguments
-/// * `start_block` - The starting block number of the batch
-/// * `end_block` - The ending block number of the batch
-/// * `full_output` - Set to true if layer is L3, false otherwise
-///
-/// # Returns
-/// * `JobMetadata` - Complete job metadata with common and SNOS-specific fields
-fn create_job_metadata(start_block: u64, end_block: u64, full_output: bool) -> JobMetadata {
+// create_job_metadata is a helper function to create job metadata for a given block number and layer
+// set full_output to true if layer is L3, false otherwise
```
byteZorvin: iirc we need to make l3s now work with full output false right?

infra/k8s/base/configmap.yaml
```diff
@@ -0,0 +1,63 @@
...
+data:
+  RUST_LOG: "info"
+  ALERT: "sns"
+  AWS_DEFAULT_PROFILE: "default-mfa"
+  AWS_PROFILE: "default-mfa"
+  AWS_REGION: "ap-south-1"
+  MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR: "2"
+  JEMALLOC_SYS_WITH_MALLOC_CONF: "background_thread:true,narenas:1,tcache:false,dirty_decay_ms:0,muzzy_decay_ms:0,abort_conf:true"
+  LOG_FORMAT: "json"
+  MADARA_ORCHESTRATOR_ATLANTIC_CAIRO_VM: "python"
```
byteZorvin, line 15: Curious: We usually prefer rust vm right?

byteZorvin (commented):
GG, code looks quite modular for l2/l3 batches now 🙏

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: you mean that we should throw error in the else statement? done that!

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
prkpndy: done

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
prkpndy: done

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: yeah. updated it!

.env.test
prkpndy: for fetching builtin weights

orchestrator/src/worker/event_handler/triggers/update_state.rs
prkpndy: you mean for state update? haven't added any such code myself

orchestrator/src/worker/event_handler/triggers/snos.rs
prkpndy: yes

prkpndy: the comment is wrong. will update

infra/k8s/base/configmap.yaml
prkpndy: yes, but 14.0 doesn't work with rustvm as of now. that's why using pythonvm

.github/workflows/pull-request-merge.yml
apoorvsadana: pendign

.github/workflows/task-build-binary.yml
apoorvsadana: pending

Makefile
apoorvsadana: i see. but this fails w/ localstack then? i remember something like this some time back, where schedule works with one but not with the other.

orchestrator/src/utils/constants.rs
apoorvsadana: right. but i am curious how it worked with 10 until now then? we're doing AR for a while becaise.

orchestrator/src/worker/event_handler/jobs/state_update.rs
apoorvsadana: Ya I meant throwing the error in the else and capturing the value inside the if, so there's no unwrap in the code.

The unwrap is removed now but that's because you're directly accessing the index (which would unwrap internally) instead of doing `fist`. It should be something like 

```
if let Some(value) = arr.get(2) {
    println!("{}", value);
} else {
    println!("No value at index 2");
}
```

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -783,25 +968,47 @@ impl BatchingTrigger {
...
+        for block_number in batch.start_block..=(batch.end_block + 1) {
             // Get the bouncer weights for this block
             info!("getting the bouncer weights for block number: {:?}", block_number);
             let bouncer_weights = self
                 .get_block_builtin_weights(config, block_number)
                 .await
                 .map_err(|e| JobError::ProviderError(e.to_string()))?;
-            current_builtins.checked_add(bouncer_weights);
+            if let Some(cb) = &mut current_builtins {
+                current_builtins = cb.checked_add(bouncer_weights);
```
apoorvsadana, line 987: so if `checked_add` in iteration 5 overflows, it panics in iteration 6? from the code, it seems like we do this because we don't want to panic if it overflows on the last block? however, not clear to me why do we want it to work that way.

also, we do want to panic here or throw an error?

Makefile
prkpndy: yes, schedule doesn't work with localstack. but rule also don't work with aws afaik.

orchestrator/src/utils/constants.rs
prkpndy: it works with 10 as well. just that it's more stable with 1. The only issue with 1 earlier (before AR) was that it was slower which won't be the case now

orchestrator/src/worker/event_handler/jobs/state_update.rs
prkpndy: git it

.github/workflows/pull-request-merge.yml
prkpndy: reverted

.github/workflows/task-build-binary.yml
prkpndy: remvoved

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: > so if checked_add in iteration 5 overflows, it panics in iteration 6?

yes

> it seems like we do this because we don't want to panic if it overflows on the last block?

yeah, that's one reason. other reason is that if we create a batch whose combined builtins overflows the max limits, we have done something wrong in the code. that's the reason i thought it's best to panic instead on returning an error (this shouldn't happen in ideal flow).

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: https://github.com/madara-alliance/madara/issues/824

Makefile
apoorvsadana: I see, so essentially this changes makes the makefile compatible with AWS instead of localstack. I am not sure if that's what everyone wants but if everyone is testing w/ AWS, sure let's do thsi

orchestrator/src/utils/constants.rs
apoorvsadana: got it

orchestrator/src/worker/event_handler/triggers/batching.rs
apoorvsadana: and why do we want to process end_block+1? why not process end_blocks only and panic if it overflows at any points. (asking because that would be much more readable)

apoorvsadana (approved):

infra/k8s/base/configmap.yaml
byteZorvin: Got it

orchestrator/src/worker/event_handler/triggers/update_state.rs
byteZorvin: Ok, I meant for snos processing.

byteZorvin (approved):

Makefile
prkpndy: added new make commands for setting up localstack

PR #809, opened 2025-10-10
[Feature] Madara Full Node Snap Sync

heemankv:
# Pull Request: Implement Snap Sync Optimization

## Pull Request Type
- [x] Feature
- [ ] Bugfix
- [ ] Build-related changes
- [ ] Testing

---

**Resolves:** #NA

The current synchronization process experiences significant performance bottlenecks due to the `apply_state_pipeline` mechanism, which performs `apply_to_global_trie` operations after processing each individual block. This approach, while ensuring state root consistency at every step, becomes prohibitively slow in scenarios where:
- The node is significantly behind the network tip
- Intermediate state roots are not required
- Rapid synchronization to the latest state is the primary goal

Additionally, the current implementation suffers from:
- Inaccurate status reporting that shows the last block range end rather than the actual last trie application point
- Frequent gateway call failures that interrupt synchronization without retry mechanisms, causing Madara to halt syncing operations

---

### **Snap Sync Implementation**
* Introduces an optimized batch synchronization mode activated via the `MADARA_SNAP_SYNC` environment variable
* Processes blocks in configurable batches of 1,000 blocks (empirically determined for optimal performance)
* Maintains block data in-memory during batch processing, deferring trie application until the full batch is collected
* Dramatically improves synchronization speed by reducing the frequency of expensive trie operations from once-per-block to once-per-batch

### **Intelligent Sync Mode Switching**
* Implements automatic mode detection that switches between snap sync and normal sync based on the node's proximity to the chain tip
* When the node is >1,000 blocks behind: uses snap sync for rapid catch-up
* When the node is <1,000 blocks behind: automatically switches to normal sync mode for maximum compatibility and real-time block processing
* Ensures seamless transitions between modes without manual intervention

### **Enhanced Reliability & Status Reporting**
* Adds retry logic for gateway calls to handle transient network failures gracefully, preventing synchronization interruptions
* Corrects status reporting to accurately reflect the `last_trie_applied` rather than the misleading `last_block_range_end`
* Improves observability of the actual synchronization state

### **Performance Characteristics**
* Significantly reduces I/O operations and state computation overhead during catch-up synchronization
* Maintains full compatibility with existing sync mechanisms when near the chain tip
* Operates transparently in the background, requiring no changes to existing workflows

---

**No.** This feature is designed to be backward compatible and non-breaking:
- Snap sync is opt-in via environment variable (`MADARA_SNAP_SYNC`)
- Automatically falls back to normal sync behavior when not enabled or when near chain tip
- Does not modify existing APIs or data structures
- Existing nodes can continue operating without any changes

---

### **Configuration**
- Enable snap sync by setting the `MADARA_SNAP_SYNC` environment variable
- Batch size: 1,000 blocks (hardcoded based on performance testing)
- Automatic mode switching threshold: 1,000 blocks from tip

### **Testing Recommendations**
- Verify snap sync performance on nodes significantly behind the network
- Confirm smooth transitions between snap sync and normal sync modes
- Test retry mechanism behavior under poor network conditions
- Validate status reporting accuracy during both sync modes

### **Performance Impact**
- Expected synchronization speed improvement: significant (exact metrics depend on network conditions and hardware)
- Memory usage: increased during batch processing (holds up to 1,000 blocks in memory)
- Recommended for nodes performing initial sync or recovering from extended downtime

snap_vs_normal_sync_madara_2

image


heemankv (commented):
Self review done, lgtm
please note addition of get_state_root_hash, this simplifies testing and debugging locally whenever we want to see the state root out of calculation code, there was no other fn like this and this was very useful while debugging, hence added

prkpndy (approved):
lgtm

madara/crates/client/db/src/rocksdb/update_global_trie/mod.rs
```diff
@@ -57,6 +57,18 @@ fn calculate_state_root(contracts_trie_root: Felt, classes_trie_root: Felt) -> F
     }
 }
 
+pub fn get_state_root(backend: &RocksDBStorage) ->  Result<Felt> {
+    let contract_trie = backend.contract_trie();
+    let contract_trie_root_hash = contract_trie.root_hash(bonsai_identifier::CONTRACT).unwrap();
```
Mohiiit: let's remove the unwraps?

madara/crates/client/db/src/rocksdb/update_global_trie/mod.rs
```diff
@@ -57,6 +57,18 @@ fn calculate_state_root(contracts_trie_root: Felt, classes_trie_root: Felt) -> F
...
+
+    let class_trie = backend.class_trie();
+    let class_trie_root_hash = class_trie.root_hash(bonsai_identifier::CLASS).unwrap();
```
Mohiiit: same here!

also I am not sure about the placement of this function in this file, or maybe the file name is incorrect 😓

madara/crates/client/gateway/client/src/methods.rs
```diff
@@ -19,93 +19,142 @@ use starknet_core::types::contract::legacy::LegacyContractClass;
...
+                        // Exponential backoff: 100ms * 2^attempt
+                        // attempt 0: 100ms, attempt 1: 200ms, attempt 2: 400ms, attempt 3: 800ms, attempt 4: 1600ms
+                        let delay_ms = 100u64 * 2u64.pow(attempt as u32);
+                        tokio::time::sleep(tokio::time::Duration::from_millis(delay_ms)).await;
+                    }
+                    last_error = Some(e);
+                }
+            }
+        }
+        Err(last_error.unwrap())
```
Mohiiit: can mention that this unwrap is safe!

madara/crates/client/gateway/client/src/methods.rs
```diff
@@ -19,93 +19,142 @@ use starknet_core::types::contract::legacy::LegacyContractClass;
...
+
+        for attempt in 0..MAX_RETRIES {
+            match request_fn().await {
+                Ok(result) => return Ok(result),
+                Err(e) => {
+                    if attempt < MAX_RETRIES - 1 {
+                        tracing::warn!("Failed to get with {:?}, retrying", e);
+                        // Exponential backoff: 100ms * 2^attempt
+                        // attempt 0: 100ms, attempt 1: 200ms, attempt 2: 400ms, attempt 3: 800ms, attempt 4: 1600ms
+                        let delay_ms = 100u64 * 2u64.pow(attempt as u32);
```
Mohiiit: let's move these values to const BASE_TIME, BACKOFF_BASE?

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
 use std::{ops::Range, sync::Arc};
+use tokio::sync::Mutex;
+use crate::pipeline::PipelineStatus;
+use crate::sync_utils::compress_state_diff;
+
+/// Batch size for snap sync state accumulation before flushing to the global trie.
+///
+/// During snap sync, state diffs are accumulated in memory rather than computing
+/// the trie after every block. After this many blocks have been processed, the
+/// accumulated state diffs are compressed and applied to the global trie in a
```
Mohiiit:
```suggestion
/// accumulated state diffs are squashed and applied to the global trie in a
```

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+/// - Larger values: Fewer trie computations (better performance), higher memory usage
+/// - Smaller values: More frequent trie updates (lower performance), less memory usage
+///
+/// The value of 1000 blocks has been empirically tested to provide good performance
+/// on modern hardware without excessive memory consumption. It typically results in
+/// trie updates every few minutes during active syncing.
+///
+/// Not exposed as a CLI option to maintain simplicity (KISS principle). If adjustment
+/// is needed for specific hardware constraints, modify this constant directly.
+const APPLY_STATE_SNAP_BATCH_SIZE: u64 = 1000;
```
Mohiiit, line 31: I think it should be a config for sure, different appchains would have different type of load and block building time and block size

maybe a TODO for future

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
 ) -> ApplyStateSync {
-    PipelineController::new(ApplyStateSteps { importer, disable_tries }, parallelization, batch_size, starting_block_n)
+    PipelineController::new(
+        ApplyStateSteps {
+            importer,
+            backend,
+            disable_tries,
+            snap_sync,
+            state_diff_map: Mutex::new(Default::default()),
+            target_block: Arc::new(std::sync::atomic::AtomicU64::new(u64::MAX)),
```
Mohiiit: if it's always u64::MAX, no need for it be part of the the struct right? could be just a const? or can i change? in case it can be changed, shall it would be taken in the param of the `apply_state_pipeline`

update: got it, the target_block is being updated by the probe directly, although I think the probe should update only it's own state and every other service/thread should read from there (separation of concern between the probe and other services) wdyt?

madara/node/src/cli/l2.rs
```diff
@@ -24,6 +24,13 @@ pub struct L2SyncParams {
     #[clap(env = "MADARA_DISABLE_TRIES", long)]
     pub disable_tries: bool,
 
+    /// Enables snap sync.
+    /// When importing a block, the state root computation is the most expensive operation.
+    /// This feature batches the state root computation, which can significantly speed up syncing.
+    /// While maintaining the global tries.
+    #[clap(env = "MADARA_SNAP_SYNC", long)]
+    pub snap_sync: bool,
```
Mohiiit, line 33: let's add a warning as well, if snap_sync is enabled, then we won't have the storage proofs for each block, something to keep in mind in future!

madara/node/src/service/l2.rs
```diff
@@ -72,6 +72,7 @@ impl Service for SyncService {
         let this = self.start_args.take().expect("Service already started");
         let importer = Arc::new(BlockImporter::new(
             this.db_backend.clone(),
+            // TODO: all_verifications_disabled should be configured from the env
```
Mohiiit: something I picked from the starkware team, I like this idea of having the TODO with POC mentioned, helps with the resolution really quick!

```suggestion
            // TODO(heemank <insert_date>): all_verifications_disabled should be configured from the env
```

madara/crates/client/sync/src/sync.rs
```diff
@@ -193,6 +194,13 @@ impl<P: ForwardPipeline> SyncController<P> {
                 self.set_status(ServiceEvent::Idle);
             }
 
+            if self.config.stop_at_block_n.is_some_and(|stop_at| self.forward_pipeline.next_input_block_n() > stop_at) {
+                if self.get_pending_block.is_some() {
+                    tracing::debug!("Cancelling pending block task as stop_at reached");
+                    self.get_pending_block = None;
+                }
+            }
```
Mohiiit: there is a condition just below it as well, both can be merged right?

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+            let mut state_diff_map = self.state_diff_map.lock().await;
+            *state_diff_map = crate::sync_utils::StateDiffMap::default();
+        }
+        Ok(ApplyOutcome::Success(()))
+    }
+
+    /// Main sync function that decides whether to use snap sync or block-by-block sync
+    /// based on the distance to the target block.
+    /// Main sync function that decides whether to use snap sync or block-by-block sync
+    /// based on the distance to the target block.
```
Mohiiit, line 138:
```suggestion
```

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+    pub async fn sync(
+        self: Arc<Self>,
+        block_range: Range<u64>,
+        input: <ApplyStateSteps as PipelineSteps>::SequentialStepInput
+    ) -> anyhow::Result<ApplyOutcome<()>> {
+        let target_block = self.target_block.load(std::sync::atomic::Ordering::Relaxed);
+        let distance_to_target = target_block.saturating_sub(block_range.end);
```
Mohiiit: shouldn't here we shall subtract from the block_range.start?

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+
+            let mut state_diff_map = self.state_diff_map.lock().await;
+            for single_contract_state_diff in state_diffs {
+                state_diff_map.apply_state_diff(&single_contract_state_diff);
+            }
+        }
+
+        let target_block = self.target_block.load(std::sync::atomic::Ordering::Relaxed);
+
+        if latest_block >= (current_first_block + APPLY_STATE_SNAP_BATCH_SIZE) || latest_block >= target_block {
```
Mohiiit: would there be any case when we have latest_block >= target_block tho? 

target_block could be something sent by the user (in that case, madara shouldn't be syncing blocks more than that)

or in case it's being synced live, could it be the case that target_block is not updated and blocks are synced? 🤔 

if this is the case, maybe we should change the ordering priority maybe?

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+                let has_accumulated_diffs = {
+                    let state_diff_map = self.state_diff_map.lock().await;
+                    let diff_count = state_diff_map.to_raw_state_diff().len();
+                    diff_count > 0
+                };
+
+                if has_accumulated_diffs {
+                    self.clone().flush_accumulated_state_diffs(block_range.start).await?;
+                } else {
+                }
```
Mohiiit:
```suggestion
                let has_accumulated_diffs = {
            let state_diff_map = self.state_diff_map.lock().await;
            !state_diff_map.to_raw_state_diff().is_empty()
        };

        if has_accumulated_diffs {
            self.clone()
                .flush_accumulated_state_diffs(block_range.start)
                .await?;
        }
```

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+            }
+
+            tracing::info!("🔨 Using block-by-block sync");
+            self.sync_each_block(block_range, input).await
+        }
+    }
+
+    /// Flushes accumulated state diffs to the global trie when transitioning from snap sync
+    /// to block-by-block sync mode.
+    async fn flush_accumulated_state_diffs(
```
Mohiiit: this function has duplicate code from the sync_snap function, doing exactly same thing!

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,252 @@ use anyhow::Context;
...
+
+            // Move the trie computation to rayon pool
+            let backend = self.backend.clone();
+
+            self.importer
+                .run_in_rayon_pool_global(move |_| {
+                    // Apply the accumulated state diff to calculate the global state root
+                    let _global_state_root = backend
+                        .write_access()
+                        .apply_to_global_trie(current_first_block, vec![accumulated_state_diff].iter())?;
```
Mohiiit: when calling the apply_to_global_trie, the block_n is used to commit the tries, this commit is used in reorg to revert to desired block as well

earlier, given we were committing per block, block_n was not a range, but it's not the case now, I think we should commit in the trie with latest_block instead of the current_first_block

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+use mc_db::{MadaraBackend, MadaraStorageRead};
+
+/// New function: Compress a raw accumulated state diff by checking against pre_range_block
+/// This should be called ONCE after all accumulation is complete
+pub async fn compress_state_diff(
+    raw_state_diff: StateDiff,
+    pre_range_block: Option<u64>,
+    backend: Arc<MadaraBackend>,
+) -> anyhow::Result<StateDiff> {
+    // tracing::info!("Snap_Sync: Starting compression with pre_range_block={}", pre_range_block);
```
Mohiiit: remove it?
```suggestion
```

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+use std::time::Duration;
+use anyhow::Context;
+use futures::{stream, StreamExt, TryStreamExt};
+use mp_convert::Felt;
+use mp_state_update::{ContractStorageDiffItem, DeclaredClassItem, DeployedContractItem, NonceUpdate, ReplacedClassItem, StateDiff, StorageEntry};
+use mc_db::{MadaraBackend, MadaraStorageRead};
+
+/// New function: Compress a raw accumulated state diff by checking against pre_range_block
```
Mohiiit:
```suggestion
/// Compress a raw accumulated state diff by checking against pre_range_block
```

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    compressed_diff.sort();
+
+    Ok(compressed_diff)
+}
+
+const MAX_CONCURRENT_CONTRACTS_PROCESSING: usize = 400;
+const MAX_CONCURRENT_GET_STORAGE_AT_CALLS: usize = 10000;
+const MAX_GET_STORAGE_AT_CALL_RETRY: u64 = 3;
+const MAX_GET_CLASS_HASH_AT_CALL_RETRY: u64 = 3;
```
Mohiiit: I think these should be just after the imports

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+    nonces: HashMap<Felt, Felt>,
+    replaced_classes: HashMap<Felt, Felt>,
+    touched_contracts: HashSet<Felt>,
+}
+
+impl StateDiffMap {
+
+    /// Apply a single state diff to the existing StateDiffMap
+    pub fn apply_state_diff(&mut self, state_diff: &StateDiff) {
+        // tracing::info!("Snap_Sync: Applying state diff to existing StateDiffMap");
```
Mohiiit:
```suggestion
```

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+        for class_hash in &state_diff.old_declared_contracts {
+            self.deprecated_declared_classes.insert(*class_hash);
+        }
+    }
+
+    /// NEW: Convert to raw StateDiff without any pre_range filtering
+    /// This is MUCH faster as it does no DB lookups
+    pub fn to_raw_state_diff(&self) -> StateDiff {
+
+        // tracing::info!("Snap_Sync: Converting to raw state diff with {} touched contracts", self.touched_contracts.len());
```
Mohiiit:
```suggestion
```

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+                "Failed to get class hash for contract: {} at block {}",
+                contract_address, block_number
+            )
+        })?;
+
+    Ok(class_hash)
+}
+
+
+pub fn retry_sync<F, T, E>(mut func: F, max_retries: u64, delay: Option<Duration>) -> Result<T, E>
```
Mohiiit: this was introduced in the orch code because we were making the calls over network, given as of now it's the call to the disk db

do you think we still need this?

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,409 @@
...
+            declared_classes,
+            old_declared_contracts: deprecated_declared_classes,
+            nonces,
+            replaced_classes,
+        }
+    }
+}
+
+/// NEW: Compress a single contract's storage entries by checking against pre_range_block
+async fn compress_single_contract(
```
Mohiiit, line 190: the need to check if the entry has been updated or not was introduced in the orch because it was affecting the result of the blob but in here, I don't think we need that, this is more of an optimization but I would still test if this is really helping a lot or just making code complicated

reasoning behind this is:

1. in case there was a key which has not been updated (let's say it went from a -> b -> a), worse case is that we will insert the a again at the same key
2. as of now we are checking if the above case has happened and we remove that from the state-diff altogether (but there is some compute happening)

so if time taken in case 2 is more than 1 (which could be the case) then I think we should remove this, wdyt? any numbers on this?

Mohiiit (commented):
gg with the results, thanks for the detailed description as well!

nit: I noticed the new function has NEW: written on the top, which I think should be removed

madara/crates/client/db/src/rocksdb/update_global_trie/mod.rs
heemankv: Done

madara/crates/client/db/src/rocksdb/update_global_trie/mod.rs
heemankv: Yeah, even I was a bit sceptical with the folder name being `update_global_trie` IMO it's majorly because no `get` fn ever existed hence `update_global_tire` was fine to work with,
wdyt on changing it to just `global_trie` ?

madara/crates/client/gateway/client/src/methods.rs
heemankv: You are right, even still to follow better practice I added an expect statement

madara/crates/client/gateway/client/src/methods.rs
heemankv: Done

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, added

madara/crates/client/sync/src/sync_utils.rs
heemankv: Valid, removed

madara/crates/client/sync/src/sync_utils.rs
heemankv: Thanks, removed

madara/crates/client/sync/src/sync_utils.rs
heemankv: Valid, removed

madara/crates/client/sync/src/sync_utils.rs
heemankv: Done

madara/crates/client/sync/src/sync_utils.rs
heemankv: Valid, removed

madara/crates/client/sync/src/sync_utils.rs
heemankv: Valid, removed

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, the common logic is now extracted to apply_accumulated_diffs

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, simplified, this is more idiomatic.

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, applied

madara/crates/client/sync/src/sync.rs
heemankv: Valid, simplified

madara/node/src/service/l2.rs
heemankv: Nice, used

madara/node/src/cli/l2.rs
heemankv: Added !

madara/crates/client/sync/src/apply_state.rs
heemankv: Added a todo for now

madara/crates/client/db/src/rocksdb/update_global_trie/mod.rs
Mohiiit: works!

madara/crates/client/sync/src/sync_utils.rs
heemankv: Makes sense, removed

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, fixed

madara/crates/client/sync/src/apply_state.rs
heemankv: Valid, updated.

madara/crates/client/sync/src/apply_state.rs
heemankv: updated

madara/crates/client/sync/src/apply_state.rs
```diff
@@ -6,21 +6,224 @@ use anyhow::Context;
...
+            pre_range_block_check,
+            self.backend.clone()
+        ).await?;
+
+        // Move the trie computation to rayon pool
+        let backend = self.backend.clone();
+
+        self.importer
+            .run_in_rayon_pool_global(move |_| {
+                // Use latest_block - 1 as the block number for trie commit
```
Mohiiit: if we are committing latest_block, why are we using latest_block - 1 as commit id tho?

madara/crates/client/sync/src/sync_utils.rs
Mohiiit: still using it?

madara/crates/client/sync/src/sync_utils.rs
```diff
@@ -0,0 +1,368 @@
...
+
+/// This function tells if the contract existed at the given block number
+pub async fn check_contract_existed_at_block(
+    backend: Arc<MadaraBackend>,
+    contract_address: Felt,
+    block_number: u64,
+) -> anyhow::Result<bool> {
+    let x = get_class_hash_at(backend.clone(), block_number, contract_address).await?.is_some();
+    Ok(x)
+}
```
Mohiiit, line 336:
```suggestion
pub async fn check_contract_existed_at_block(
    backend: Arc<MadaraBackend>,
    contract_address: Felt,
    block_number: u64,
) -> anyhow::Result<bool> {
    Ok(get_class_hash_at(backend, block_number, contract_address).await?.is_some())
}
```

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -98,6 +98,29 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+            .map_err(StarknetRpcApiError::from)?
+        {
+            if target_block_n < snap_sync_latest_block {
+                return Err(StarknetRpcApiError::ErrUnexpectedError {
+                    error: format!(
+                        "Cannot revert to block {} because snap sync was used up to block {}. Trie data is only available from block {} onwards.",
+                        target_block_n, snap_sync_latest_block, snap_sync_latest_block
+                    ).into()
+                }.into());
+            }
```
Mohiiit, line 121: shouldn't we compare target_block_n > snap_sync_latest_block here?

because errors says "snap sync was used up to the block snap_sync_latest_block"

also, what's the purpose of revert_to? will it revert to the block_n which has it's tries?

Mohiiit (changes requested):
snap_sync with reorg is missing, we should add that before merging!

madara/crates/client/sync/src/sync_utils.rs
heemankv: Valid, applied

Mohiiit (approved):

PR #815, opened 2025-10-26
Add automated Cairo 0 environment setup for builds

heemankv:
## Pull Request type

- [x] Build-related changes
- [x] Documentation content changes

Resolves: #NA

Currently, developers need to manually set up the Cairo 0 environment before building Madara. The build fails with a cryptic error if `cairo-compile` is not available:

```
Failed to verify correct cairo0 package installation. Please make sure you do not have a conflicting installation in your /Users/xxx/.cargo/bin directory.
Original error: Failed to get cairo-compile version: No such file or directory (os error 2)
```
Developers must remember to:
1. Create a Python virtual environment
2. Manually install dependencies from a requirements file
3. Work around NumPy 2.0.2 build issues on some systems (especially Python 3.13)
4. Activate the venv before every build

This creates friction for new contributors and can lead to frustrating build failures.

## What is the new behavior?

Added automated Cairo 0 environment management to the Makefile:

- **`make setup-cairo`** - Automatically sets up the Cairo 0 environment
  - Creates Python virtual environment (`sequencer_venv`) if needed
  - Installs Cairo 0 and dependencies with NumPy compatibility fix (`numpy<2.0` instead of `numpy==2.0.2`)
  - Verifies installation and shows cairo-compile version
  - Idempotent: safe to run multiple times

- **`make build-madara`** - New target for building Madara with automatic Cairo setup
  - Automatically runs `setup-cairo` as a prerequisite
  - Builds in release mode with Cairo environment activated

- **`make pre-push`** - Updated to use Cairo environment
  - Now automatically sets up Cairo 0 before running checks
  - Ensures consistent build environment

- **BUILDING.md** - New documentation file
  - Quick start guide for building Madara
  - Development workflow instructions
  - Troubleshooting common issues

Developers now simply run:
```bash
make build-madara      # Everything is handled automatically
make pre-push    # Cairo environment set up automatically
```

No

This is a developer experience improvement that doesn't affect the runtime behavior or API of Madara. Existing workflows continue to work as before.

## Other information

**Benefits:**
- Reduces onboarding friction for new developers
- Prevents build failures due to missing Cairo 0 environment
- Automatically handles NumPy compatibility issues
- Self-documenting with clear status messages
- Falls back gracefully if sequencer requirements file isn't found

**Testing:**
Verified that `make setup-cairo`, `make build`, and `make pre-push` all work correctly with the new setup.


Makefile
```diff
@@ -193,6 +200,35 @@ artifacts:
...
+	@echo -e "$(DIM)Setting up Cairo 0 environment...$(RESET)"
+	@if [ ! -d "$(VENV)" ]; then \
+		echo -e "$(INFO)Creating Python virtual environment...$(RESET)"; \
+		python3 -m venv $(VENV); \
+	else \
+		echo -e "$(PASS)✅ Virtual environment already exists$(RESET)"; \
+	fi
+	@echo -e "$(INFO)Installing Cairo 0 dependencies...$(RESET)"
+	@if [ ! -f sequencer_requirements.txt ]; then \
+		if [ -f /Volumes/External/rust-builds/git/checkouts/sequencer-be896281adc16bee/e04617e/scripts/requirements.txt ]; then \
```
apoorvsadana: is this path specific your system by any chance? seems specific to a commit too?

apoorvsadana: maybe we can just download it from gh directely?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -231,7 +231,7 @@ impl PipelineSteps for GatewaySyncSteps {
                 .latest_confirmed_block_n();
 
             for block_n in block_range {
-                tracing::info!("📥 Fetching block #{} from gateway", block_n);
+                tracing::debug!("📥 Fetching block #{} from gateway", block_n);
```
Mohiiit, line 235: ins't this useful tho?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -413,7 +413,7 @@ impl PipelineSteps for GatewaySyncSteps {
                         importer.save_transactions(block_n, gateway_block.transactions)?;
                         importer.save_events(block_n, gateway_block.events)?;
 
-                        tracing::info!("✅ Block #{} saved: header, state_diff, transactions, events", block_n);
+                        tracing::debug!("✅ Block #{} saved: header, state_diff, transactions, events", block_n);
```
Mohiiit, line 417: ditto!

Makefile
```diff
@@ -195,9 +203,47 @@ artifacts:
...
+		CARGO_HOME=$${CARGO_HOME:-$$HOME/.cargo}; \
+		GIT_CHECKOUTS=$$(readlink -f "$$CARGO_HOME/git" 2>/dev/null || echo "$$CARGO_HOME/git"); \
+		REQUIREMENTS_PATH=$$(find "$$GIT_CHECKOUTS/checkouts" -type f -path "*/sequencer*/scripts/requirements.txt" 2>/dev/null | head -n 1); \
+		if [ -n "$$REQUIREMENTS_PATH" ]; then \
+			sed 's/numpy==2.0.2/numpy<2.0/' "$$REQUIREMENTS_PATH" > sequencer_requirements.txt; \
+			echo -e "$(INFO)Found requirements.txt at: $$REQUIREMENTS_PATH$(RESET)"; \
+		else \
+			echo -e "$(WARN)⚠️  WARNING: Could not find requirements.txt from sequencer checkout$(RESET)"; \
+			echo -e "$(INFO)Creating basic requirements with cairo-lang...$(RESET)"; \
+			echo "cairo-lang==0.14.0.1" > sequencer_requirements.txt; \
```
Mohiiit, line 227: cairo-lang should be part of the seqeuencer_requirements only (sequencer repo: https://github.com/starkware-libs/sequencer/blob/main-v0.14.1/scripts/requirements.txt#L11)

also, I think the right way to do this would be to dynamically fetch the requirement.txt from the sequencer dependency directly like we are doing in the github workflow:

```bash
# Look for requirements.txt in sequencer checkout directories
        if [ -d "$HOME/.cargo/git/checkouts" ]; then
          echo "Searching for sequencer requirements.txt in cargo checkouts..."
          REQUIREMENTS_FILE=$(find "$HOME/.cargo/git/checkouts" -name "requirements.txt" -path "*/scripts/requirements.txt" 2>/dev/null | grep -i sequencer | head -1)
        fi
```

Makefile
```diff
@@ -195,9 +203,47 @@ artifacts:
...
+			echo "numpy<2.0" >> sequencer_requirements.txt; \
+		fi; \
+	fi
+	@$(VENV_ACTIVATE) && pip install --upgrade pip > /dev/null 2>&1 && pip install -r sequencer_requirements.txt > /dev/null 2>&1
+	@$(VENV_ACTIVATE) && cairo-compile --version > /dev/null 2>&1 && echo -e "$(PASS)✅ Cairo 0 environment ready (cairo-compile $$($(VENV_ACTIVATE) && cairo-compile --version 2>&1))$(RESET)" || (echo -e "$(WARN)❌ Cairo setup failed$(RESET)" && exit 1)
+
+.PHONY: build-madara
+build-madara: setup-cairo
+	@echo -e "$(DIM)Building Madara with Cairo 0 environment...$(RESET)"
+	@$(VENV_ACTIVATE) && cargo build --manifest-path madara/Cargo.toml  --bin madara --release
```
Mohiiit, line 237: don't need it for madara yet tho, just FYI

madara/crates/client/sync/src/gateway/blocks.rs
heemankv: Too much logs in info, becomes very had to track sync progress
image

Makefile
```diff
@@ -195,9 +203,47 @@ artifacts:
...
+		echo -e "$(INFO)Creating Python virtual environment...$(RESET)"; \
+		python3 -m venv $(VENV); \
+	else \
+		echo -e "$(PASS)✅ Virtual environment already exists$(RESET)"; \
+	fi
+	@echo -e "$(INFO)Installing Cairo 0 dependencies...$(RESET)"
+	@if [ ! -f sequencer_requirements.txt ]; then \
+		CARGO_HOME=$${CARGO_HOME:-$$HOME/.cargo}; \
+		GIT_CHECKOUTS=$$(readlink -f "$$CARGO_HOME/git" 2>/dev/null || echo "$$CARGO_HOME/git"); \
+		REQUIREMENTS_PATH=$$(find "$$GIT_CHECKOUTS/checkouts" -type f -path "*/sequencer*/scripts/requirements.txt" 2>/dev/null | head -n 1); \
```
apoorvsadana: i have multiple sequencer folders in my git checkout, probably many have the requirements.txt, wouldn't this command checkout to a random one?

not very necessary to solve, don't want to spend too much time on this but pointing it out in case its a bug and it can be solved in some easy way (Cargo.lock scan?)

Makefile
apoorvsadana: that's happening in the `if` case right? i think this case is when the code fails to find the requirements txt file

apoorvsadana (approved):

Makefile
heemankv: fixed thanks!

Makefile
heemankv: updated

Mohiiit (approved):

PR #821, opened 2025-10-28
add min settlement blocks

byteZorvin:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

## What is the new behavior?

-
-
-


madara/crates/client/settlement_client/src/eth/event.rs
```diff
@@ -85,6 +89,98 @@ impl Stream for EthereumEventStream {
...
+/// A stream wrapper that filters events based on dynamic block confirmation depth
+/// 
+/// This uses a background task to periodically update the latest block number,
+/// making the stream implementation much simpler. It's generic and works with any
+/// settlement layer provider.
+pub struct ConfirmationDepthFilteredStream<S> {
+    inner: S,
+    min_confirmations: u64,
+    latest_block: Arc<AtomicU64>,
+    _update_task: tokio::task::JoinHandle<()>,
```
prkpndy: can we remove it if not used?

madara/node/src/cli/l1.rs
```diff
@@ -122,4 +122,9 @@ pub struct L1SyncParams {
         default_value_t = MadaraSettlementLayer::Eth,
     )]
     pub settlement_layer: MadaraSettlementLayer,
+
+    /// Minimum settlement blocks for L1 sync.
+    /// This is the minimum number of blocks that must be settled on L1 before the node can sync/process the messages.
+    #[clap(env = "MADARA_MIN_SETTLEMENT_BLOCKS", long, default_value = "10")]
```
prkpndy: nit: can we give it a better name if possible? something like `MADARA_MIN_SETTLED_BLOCKS_FOR_MESG_PROCESSING` (this is too long but something along these lines)

madara/crates/client/settlement_client/src/eth/event.rs
```diff
@@ -85,6 +89,98 @@ impl Stream for EthereumEventStream {
     }
 }
 
+/// A stream wrapper that filters events based on dynamic block confirmation depth
+/// 
+/// This uses a background task to periodically update the latest block number,
+/// making the stream implementation much simpler. It's generic and works with any
+/// settlement layer provider.
```
prkpndy: if it can be used with any settlement client, why are we writing it again for starknet client below?

madara/crates/client/settlement_client/src/starknet/event.rs
```diff
@@ -155,6 +158,100 @@ pub fn watch_events(
...
+/// 
+/// This uses a background task to periodically update the latest block number,
+/// making the stream implementation much simpler.
+pub struct ConfirmationDepthFilteredStream<S, P> {
+    inner: S,
+    min_confirmations: u64,
+    latest_block: Arc<AtomicU64>,
+    _update_task: tokio::task::JoinHandle<()>,
+    _phantom: std::marker::PhantomData<fn() -> P>,
+}
```
prkpndy: we also have it in ethereum settlement client? why is it duplicated?

prkpndy: can we create a single struct and a trait and implement that for different clients (if we need to)?

prkpndy (commented):
gg. the logic looks good! have some questions regarding the implementation

madara/crates/client/settlement_client/src/eth/event.rs
```diff
@@ -85,6 +89,98 @@ impl Stream for EthereumEventStream {
...
+                    
+                    // Event doesn't have enough confirmations yet, skip it and continue polling
+                    tracing::debug!(
+                        "Skipping event at block {} (needs {} confirmations, latest L1 block: {})",
+                        event.l1_block_number,
+                        self.min_confirmations,
+                        latest
+                    );
+                    // Continue the loop to get the next event
+                }
```
apoorvsadana: are we sure this works? seeems to me that if the depth isn't met then we will not do anything and that event will never come again?

madara/crates/client/settlement_client/src/eth/event.rs
apoorvsadana: agreed, we should have one fileter and make it work for both

madara/crates/client/settlement_client/src/eth/event.rs
apoorvsadana: it's used. it keeps the latest block updating task in memory so that it's not dropped. we should add a comment @byteZorvin

madara/crates/client/settlement_client/src/eth/event.rs
```diff
@@ -85,6 +89,98 @@ impl Stream for EthereumEventStream {
     }
 }
 
+/// A stream wrapper that filters events based on dynamic block confirmation depth
+/// 
+/// This uses a background task to periodically update the latest block number,
+/// making the stream implementation much simpler. It's generic and works with any
+/// settlement layer provider.
+pub struct ConfirmationDepthFilteredStream<S> {
```
apoorvsadana: how can we write a test case for this?

madara/node/src/cli/l1.rs
```diff
@@ -122,4 +122,9 @@ pub struct L1SyncParams {
         default_value_t = MadaraSettlementLayer::Eth,
     )]
     pub settlement_layer: MadaraSettlementLayer,
+
+    /// Minimum settlement blocks for L1 sync.
+    /// This is the minimum number of blocks that must be settled on L1 before the node can sync/process the messages.
```
apoorvsadana:
```suggestion
    /// Minimum number of block confirmations an event must have before it can be processed from the settlement layer
```

apoorvsadana (commented):
The idea of the filter is good. However, I am unsure if it works as intended.

madara/crates/client/settlement_client/src/messaging/depth_filtered_stream.rs
```diff
@@ -0,0 +1,320 @@
...
+
+        // Now the first event should be released (latest_block = 110, threshold = 100, event at 100 <= 100)
+        let result = filtered_stream.next().await;
+        assert!(result.is_some(), "First event should be released after threshold is met");
+        let released_event = result.unwrap().unwrap();
+        assert_eq!(released_event.l1_block_number, event1_block);
+
+        // Second event should still be buffered (threshold = 100, event at 101 > 100)
+        let result = tokio::time::timeout(Duration::from_millis(50), filtered_stream.next()).await;
+        assert!(result.is_err(), "Second event should still be buffered");
```
apoorvsadana, line 257: what's the err in this case? i was thinking it might be None or something

madara/crates/client/settlement_client/src/messaging/depth_filtered_stream.rs
```diff
@@ -0,0 +1,320 @@
...
+
+        // Now the second event should be released (latest_block = 111, threshold = 101, event at 101 <= 101)
+        let result = filtered_stream.next().await;
+        assert!(result.is_some(), "Second event should be released after threshold is met");
+        let released_event = result.unwrap().unwrap();
+        assert_eq!(released_event.l1_block_number, event2_block);
+    }
+
+    #[tokio::test]
+    async fn test_stream_ends_with_buffered_events() {
```
apoorvsadana, line 273: can this test be combined w/ the upper one by adding 2 lines at the end?

can we also test for the case where stream ends but events are still buffered?

madara/node/src/cli/l1.rs
```diff
@@ -122,4 +122,9 @@ pub struct L1SyncParams {
         default_value_t = MadaraSettlementLayer::Eth,
     )]
     pub settlement_layer: MadaraSettlementLayer,
+
+    /// Minimum number of block confirmations required before an L1 to L2 message can be processed.
+    /// This ensures messages are only processed after they have sufficient confirmations on the settlement layer.
+    #[clap(env = "MADARA_L1_MSG_MIN_CONFIRMATIONS", long, default_value = "10")]
```
apoorvsadana: nitpick: do we use full names like `MESSAGE` in our config normally?

byteZorvin: Makes sense, updated !!

PR #825, opened 2025-10-31
run merge ci once a day or manually

apoorvsadana:
- running merge CI on every merge to main is expensive as we increase commit frequency
- new merge CI runs every day at 12:00 AM UTC ONLY if there's a commit in the last UTC day
- it can also run through manual triggers like before

### DIFF
--- a/.github/workflows/pull-request-merge.yml
+++ b/.github/workflows/pull-request-merge.yml
@@ -97,36 +97,22 @@ jobs:
   # Run coverage tests (Madara)
   test-madara:
     uses: ./.github/workflows/task-do-nothing-test-madara.yml
-    needs: [check-merged-today]
-    if: github.event_name != 'schedule' || needs.check-merged-today.outputs.should_run == 'true'
 
   # Run coverage tests (Orchestrator)
   test-orchestrator:
     uses: ./.github/workflows/task-do-nothing-test-orchestrator.yml
-    needs: [check-merged-today]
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-merge.yml
```diff
@@ -4,10 +4,39 @@ name: Workflow - Pull Request Main (merge)
...
+        with:
+          fetch-depth: 0
+      - name: Check commits on main in previous UTC day
+        id: check
+        shell: bash
+        run: |
+          set -euo pipefail
+          SINCE="$(date -u -d 'yesterday 00:00' +%Y-%m-%dT%H:%M:%SZ)"
+          UNTIL="$(date -u -d 'today 00:00' +%Y-%m-%dT%H:%M:%SZ)"
+          git fetch origin main --prune
```
heemankv, line 32: Nitpick (Can skip) : I think this might not be needed since we are already using `fetch-depth` with '0'.

heemankv (approved):
Overall looks good.

We'll still be creating nightly releases even if e2e fails; IMO after fixing e2e to be compatible with the latest main, we could make nightly dependent on passing e2e, wdyt?

prkpndy (approved):
LGTM

.github/workflows/pull-request-merge.yml
apoorvsadana: Ya, if it's a good to have, will like to skip so we can merge this right now!

PR #828, opened 2025-11-03
feat: update batching logic to check for builtin weights when closing agg batch

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Close the aggregator batch solely on the basis of blob size which is determined by the state update size.
(There are some other checks that are also done, like batch time, etc. but nothing related to the block produced)

Resolves: #NA

## What is the new behavior?

Also check the gas and message_segment_length and if they exceed the limit, close the batch


Mohiiit (commented):
I think we can really optimize the number of calls the code is making to the node for the bouncer weights, ideally it could've only one call per block, although over here the code is making atleast 3 calls

given the intend is to save the bouncer info in the db, maybe we can make a call alongside the state-diff and check it alongside the same conditions as well? wdyt?

orchestrator/src/types/batch.rs
```diff
@@ -66,6 +67,12 @@ pub struct SnosBatchUpdates {
     pub status: Option<SnosBatchStatus>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct AggregatorBatchWeights {
+    pub gas: usize,
```
apoorvsadana:
```suggestion
    pub l1_gas: usize,
```

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -422,12 +429,13 @@ impl BatchingTrigger {
                             )?;
 
                             Ok((
-                                Some(state_update),
+                                Some(squashed_state_update),
```
apoorvsadana, line 436: was this a bug in previous code?

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -422,12 +429,13 @@ impl BatchingTrigger {
...
                                 self.update_aggregator_batch_info(
                                     current_aggregator_batch,
                                     block_number,
                                     Some(new_snos_batch.snos_batch_id),
                                     false,
+                                    combined_weights.unwrap(),
```
apoorvsadana: this is safe because we check if it can close up already? if yes, let's add a comment. 

but i think the code is much more readable if
1. you pass block weights to `should_close`
2. addition happens inside should close and we check if its None or not
3. should close can take a mutable of batch weights and edit it if it needs to change

orchestrator/src/types/batch.rs
```diff
@@ -139,13 +146,42 @@ pub struct AggregatorBatch {
...
     /// All blocks in a batch must have the same Starknet version for prover compatibility
     pub starknet_version: String,
 }
 
+impl AggregatorBatchWeights {
+    pub fn new(gas: usize, message_segment_length: usize) -> Self {
+        Self { gas, message_segment_length }
+    }
+
+    pub fn checked_add(&self, other: &AggregatorBatchWeights) -> Option<AggregatorBatchWeights> {
```
apoorvsadana, line 164: the `checked_add` fn is good design!

orchestrator/src/types/batch.rs
```diff
@@ -139,13 +146,42 @@ pub struct AggregatorBatch {
...
+    pub builtin_weights: AggregatorBatchWeights,
+
     /// Current status of the aggregator batch
     pub status: AggregatorBatchStatus,
     /// Starknet protocol version for all blocks in this batch
     /// All blocks in a batch must have the same Starknet version for prover compatibility
     pub starknet_version: String,
 }
 
+impl AggregatorBatchWeights {
```
apoorvsadana, line 159: should we add some quick unit tests for this?

orchestrator/src/types/batch.rs
prkpndy: sorted

orchestrator/src/types/batch.rs
prkpndy: :)

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: yeah, makes sense. updated!
(was thinking about this, but was thinking around the lines to return the updated value from should_close, which didn't make sense to me. Didn't think about making it mutable)

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: Yes. Missed it earlier

orchestrator/src/types/batch.rs
prkpndy: done

orchestrator/src/types/batch.rs
```diff
@@ -139,13 +146,42 @@ pub struct AggregatorBatch {
...
+    pub fn new(l1_gas: usize, message_segment_length: usize) -> Self {
+        Self { l1_gas, message_segment_length }
+    }
+
+    pub fn checked_add(&self, other: &AggregatorBatchWeights) -> Option<AggregatorBatchWeights> {
+        Some(Self {
+            l1_gas: self.l1_gas.checked_add(other.l1_gas)?,
+            message_segment_length: self.message_segment_length.checked_add(other.message_segment_length)?,
+        })
```
apoorvsadana, line 168: wait, i forgot to point this before but seems like right now it overflows if we cross `usize`? doesn't seem like what we want

orchestrator/src/types/batch.rs
```diff
@@ -139,13 +146,42 @@ pub struct AggregatorBatch {
...
+    }
+
+    pub fn checked_sub(&self, other: &AggregatorBatchWeights) -> Option<AggregatorBatchWeights> {
```
apoorvsadana, line 171: do we use this btw? can remove if not

orchestrator/src/types/batch.rs
```diff
@@ -303,3 +342,188 @@ impl SnosBatch {
...
+            assert_eq!(serialized, expected);
+        }
+
+        #[test]
+        fn test_deserialization() {
+            let json = r#"{"l1_gas":1000,"message_segment_length":500}"#;
+            let weights: AggregatorBatchWeights = serde_json::from_str(json).unwrap();
+            assert_eq!(weights.l1_gas, 1000);
+            assert_eq!(weights.message_segment_length, 500);
+        }
```
apoorvsadana: can remove

prkpndy: done

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -946,12 +980,19 @@ impl BatchingTrigger {
...
         config: &Arc<Config>,
         state_update_len: Option<usize>,
+        current_weights: &AggregatorBatchWeights,
+        combined_weights: &mut AggregatorBatchWeights,
         batch: &AggregatorBatch,
     ) -> bool {
         debug!("checking if we need to close the aggregator batch");
+        *combined_weights = match batch.builtin_weights.checked_add(current_weights) {
+            Some(weights) => weights,
+            None => return true,
```
Mohiiit: let's add a warning over here!

Mohiiit (commented):
LGTM, just have to update with the rest endpoint (currently it's using the admin rpc, the main have moved to the feeder gateway now)

for the bouncer weights

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: resolved

Mohiiit (approved):

apoorvsadana (approved):

PR #832, opened 2025-11-04
Feat/cairo native on 0.14.0

Mohiiit:
# Cairo Native Integration - PR Summary

## Overview

Adds **Cairo Native** support to Madara for faster execution through AOT compilation of Cairo contracts to native machine code.

**Key Innovation**: Asynchronous background compilation with Cairo VM fallback - first execution uses VM (standard speed) while compiling in background, subsequent executions use native code.

## Features

- ✅ **AOT Native Compilation**: Sierra → Native machine code
- ✅ **Async & Blocking Modes**: Async (production) for background compilation with VM fallback, or blocking (testing) to wait for compilation
- ✅ **Runtime Control**: Enable/disable native execution via CLI flag without rebuild (disabled by default for backward compatibility)
- ✅ **Two-tier Caching**: In-memory (DashMap) + persistent disk cache
- ✅ **CLI Configuration**: Full control via CLI flags with environment variable fallbacks
- ✅ **Resource Management**: Configurable limits for compilation concurrency (default: 4), timeouts (5min), memory (1000 classes), disk (10GB)
- ✅ **Observability**: Comprehensive metrics (cache hit rate, compilation time, VM fallback count) with structured logging, timing information, and distinct CAIRO_NATIVE log prefix
- ✅ **Performance Optimizations**: LRU cache eviction, minimized lock contention, timeout-protected disk loading
- ✅ **Smart Logging**: Timeouts logged as warnings (retryable), errors logged for actual failures, structured fields with color-coded output

## Architecture

```
┌─────────────────────────────────────────────────────────────┐
│  Request for Sierra Class Execution                         │
└──────────────────────┬──────────────────────────────────────┘
                       │
                       ▼
         ┌─────────────────────────┐
         │  Check Memory Cache     │
         │  (LRU with access time) │
         └────────┬────────────────┘
                  │
        ┌─────────┴─────────┐
        │ Hit (< 100µs)     │ Miss
        ▼                   ▼
   ┌────────┐      ┌──────────────────┐
   │Execute │      │ Check Disk Cache │
   │Native  │      │ (with 2s timeout)│
   └────────┘      └────────┬─────────┘
                            │
                  ┌─────────┴─────────┐
                  │ Hit (< 10ms)      │ Miss/Timeout
                  ▼                   ▼
          ┌────────────┐      ┌─────────────────┐
          │Load & Cache│      │ Compilation Mode│
          │in Memory   │      │   Check         │
          │Execute     │      └────────┬────────┘
          └────────────┘               │
                             ┌─────────┴──────────┐
                             │                    │
                        ASYNC MODE           BLOCKING MODE
                    (Production/Default)    (Testing/Debug)
                             │                    │
                             ▼                    ▼
                  ┌──────────────────┐   ┌──────────────────┐
                  │Execute Cairo VM  │   │Compile Native    │
                  │+ Spawn Background│   │Synchronously     │
                  │  Compilation     │   │(Wait for result) │
                  │  (with semaphore)│   │                  │
                  └────────┬─────────┘   └────────┬─────────┘
                           │                      │
                           │             ┌────────┴────────┐
                           │             │ Success?        │
                           │             │                 │
                           │             ▼                 ▼
                           │      ┌──────────┐      ┌──────────┐
                           │      │Execute   │      │Fail with │
                           │      │Native    │      │Error     │
                           │      └──────────┘      └──────────┘
                           │
                           ▼
                  ┌─────────────────┐
                  │Background Task: │
                  │- Acquire Permit │
                  │- Compile Native │
                  │- Save to Disk   │
                  │- Cache in Memory│
                  │  (LRU updated)  │
                  │  (Next request  │
                  │   uses native)  │
                  └─────────────────┘
```

### Compilation Modes

**Async Mode (Default - Production)**
- ✅ Immediate execution with Cairo VM (standard speed)
- ✅ Background compilation with semaphore limiting
- ✅ Graceful degradation on compilation failure
- ✅ Subsequent requests benefit from native compilation
- ⚡ Zero latency impact on first execution

**Blocking Mode (Testing/Debugging)**
- ⏸️ Waits for native compilation to complete
- ⏸️ First execution uses native code (if successful)
- ❌ Fails immediately if compilation fails
- 🔍 Better for testing/debugging native compilation issues
- ⏱️ Higher latency on first execution (5-30s depending on contract)

### Caching Strategy

1. **Memory Cache (L1)**: `DashMap<ClassHash, (NativeCompiledClass, Instant)>`
   - Fast access (< 1ms, typically < 100µs)
   - Configurable size limit with **LRU eviction** (based on access time)
   - Thread-safe concurrent access via DashMap
   - Access time updated on every cache hit for accurate LRU tracking

2. **Disk Cache (L2)**: `{cache_dir}/{class_hash}.so`
   - Persistent across restarts
   - **Timeout-protected loading** (2s timeout) to prevent blocking
   - Automatically falls back to VM if disk load times out
   - Configurable via `MADARA_NATIVE_CLASSES_PATH` or `--native-cache-dir`
   - Loaded classes automatically promoted to memory cache

3. **Compilation Queue**:
   - Semaphore-based limiting (default: 4 concurrent)
   - `DashMap` to track in-progress compilations with per-class locks
   - Prevents duplicate compilation of same class
   - Graceful handling of concurrent requests for same class
   - **Timeout handling**: Compilation timeouts logged as warnings (retryable), actual failures logged as errors

## Configuration

Cairo Native is **built-in but disabled by default** for backward compatibility. All configuration is available via CLI flags with environment variable fallbacks:

### CLI Flags

```bash
# Enable/disable native execution (default: false)
--enable-native-execution <BOOL>                    # or MADARA_ENABLE_NATIVE_EXECUTION
                                                    # Set to 'true' to enable native execution

# Cache directory (default: /usr/share/madara/data/classes)
--native-cache-dir <PATH>                           # or MADARA_NATIVE_CACHE_DIR
                                                    # Can also use MADARA_NATIVE_CLASSES_PATH env var

# Memory cache size (default: 1000 classes)
--native-max-memory-cache <SIZE>                    # or MADARA_NATIVE_MAX_MEMORY_CACHE
                                                    # Accepts: positive integer or "unlimited"
                                                    # Example: 2000 or "unlimited"

# Disk cache size in bytes (default: 10 GB)
--native-max-disk-cache-bytes <BYTES>               # or MADARA_NATIVE_MAX_DISK_CACHE_BYTES
                                                    # Accepts: positive integer or "unlimited"
                                                    # Example: 21474836480 (20 GB) or "unlimited"

# Concurrent compilations (default: 4)
--native-max-concurrent-compilations <COUNT>        # or MADARA_NATIVE_MAX_CONCURRENT_COMPILATIONS

# Compilation timeout in seconds (default: 300)
--native-compilation-timeout-secs <SECONDS>         # or MADARA_NATIVE_COMPILATION_TIMEOUT_SECS

# Compilation mode: 'async' or 'blocking' (default: async)
--native-compilation-mode <MODE>                    # or MADARA_NATIVE_COMPILATION_MODE
```

### Compilation Modes

- **async** (default, production): Compile in background, use Cairo VM as fallback
- **blocking** (testing/debugging): Wait for compilation, fail if compilation fails

## Usage

```bash
# Build Madara (Cairo Native is built-in)
cargo build --release

# Run with default settings (native execution disabled - uses Cairo VM)
cargo run --release --bin madara

# Enable native execution with default cache settings
cargo run --release --bin madara \
  --enable-native-execution true

# Run with custom configuration
cargo run --release --bin madara \
  --enable-native-execution true \
  --native-cache-dir /custom/path \
  --native-max-memory-cache 2000 \
  --native-max-disk-cache-bytes 21474836480 \
  --native-max-concurrent-compilations 8

# Use unlimited cache (no eviction)
cargo run --release --bin madara \
  --enable-native-execution true \
  --native-max-memory-cache unlimited \
  --native-max-disk-cache-bytes unlimited

# Use blocking mode for testing/debugging
cargo run --release --bin madara \
  --enable-native-execution true \
  --native-compilation-mode blocking
```

## Logging

Cairo Native logs use a distinct `CAIRO_NATIVE` prefix with structured fields for easy filtering and parsing:

### Log Format

```
[timestamp] CAIRO_NATIVE [WARN|ERROR] <message> class_hash=<hash> [error=<error>] [- <timing>]
```

- **INFO/DEBUG**: No level prefix (clean, fast operations)
- **WARN**: Shows `WARN` prefix (yellow, retryable issues like timeouts)
- **ERROR**: Shows `ERROR` prefix (red, actual failures)

### Examples

```bash
# Memory cache hit (fast) - INFO level
[2025-11-05 19:19:07:022] CAIRO_NATIVE native_from_memory class_hash=0xe2eb8f5672af4e6a4e8a8f1b44989685e668489b0a25437733756c5a34a1d6 - 15.833µs

# Disk cache hit - INFO level
[2025-11-05 19:19:07:022] CAIRO_NATIVE disk_hit class_hash=0x... - 10.234ms

# Compilation success - INFO level
[2025-11-05 19:18:58:287] CAIRO_NATIVE compilation_async_success class_hash=0x... - 1.732s

# VM fallback (compilation in progress) - INFO level
[2025-11-05 19:18:56:555] CAIRO_NATIVE fallback_to_vm_async class_hash=0x...

# Timeout (warning - retryable) - WARN level
[2025-11-05 19:18:56:555] CAIRO_NATIVE WARN compilation_async_timeout class_hash=0x... error=Compilation exceeded timeout

# Error (actual failure) - ERROR level
[2025-11-05 19:18:56:555] CAIRO_NATIVE ERROR compilation_async_failed class_hash=0x... error=...
```

### Log Levels

- **INFO**: Cache hits, compilation success, VM fallback
- **DEBUG**: Cache operations, compilation spawning, detailed flow
- **WARN**: Timeouts (retryable), disk load issues
- **ERROR**: Compilation failures, conversion errors

### Filtering Logs

```bash
# Show only Cairo Native logs at INFO level and above
RUST_LOG=info,madara.cairo_native=info

# Filter out verbose cairo_native library logs (Compiling function...)
RUST_LOG=info,cairo_native=warn

# Show all Cairo Native logs including debug
RUST_LOG=debug,madara.cairo_native=debug
```

All Cairo Native logs use the `madara.cairo_native` target, separate from the external `cairo_native` library logs, allowing independent filtering.


.github/actions/setup-rust/action.yml
```diff
@@ -50,10 +50,46 @@ runs:
...
+        # Note: clang-19 handles both C and C++ compilation
+        echo "CC=clang-19" >> $GITHUB_ENV
+        echo "CXX=clang-19" >> $GITHUB_ENV
+
+        # Configure linker for clang - use lld-19 for Cairo Native runtime compilation
+        # These flags are passed to clang when it invokes the linker
+        echo "LDFLAGS=-fuse-ld=lld-19" >> $GITHUB_ENV
+
+        # Add LLVM 19 binaries to PATH
+        echo "/usr/lib/llvm-19/bin" >> $GITHUB_PATH
```
apoorvsadana: should we make this a make command and reuse across the action file and dockerfile?

build-artifacts/src/lib.rs
```diff
@@ -145,9 +148,27 @@ fn get_artifacts(root: &RootDir, artifacts: &VersionFileArtifacts) -> Result<(),
...
     let mut docker = std::process::Command::new("docker");
-    docker.args(["rm", "-f", &container_name]).status().ok();
+    docker.args([
+        "ps",
+        "-a",
+        "--filter",
+        &format!("name=madara-artifacts-extractor-v{}", version),
+        "--format",
+        "{{.Names}}",
+    ]);
```
apoorvsadana: since containers won't be named without the timestamp after this merge, is this rm is for backward compatibility? also, don't we want to clean the container created w/ the timestamp too?

madara/crates/client/analytics/src/formatter.rs
```diff
@@ -82,6 +82,73 @@ impl RpcCallEventVisitor {
     }
 }
 
+#[derive(Default)]
+struct CairoNativeEventVisitor {
```
apoorvsadana, line 86: this is interesting, what was the intuition behing have a formatter specifically for Cairo Native logs?

madara/crates/client/block_production/Cargo.toml
```diff
@@ -31,6 +31,8 @@ bitvec.workspace = true
 assert_matches.workspace = true
 lazy_static.workspace = true
 serde_json.workspace = true
+tempfile.workspace = true
+criterion = { workspace = true }
```
apoorvsadana: is `tempfile` still needed? couldn't find it being used in the diff.

madara/crates/primitives/class/src/native/cache.rs
```diff
@@ -0,0 +1,420 @@
...
+/// Shorter timeout ensures fast failure and fallback to VM if disk load hangs.
+const DISK_LOAD_TIMEOUT: Duration = Duration::from_secs(2);
+
+/// In-memory cache for compiled native classes.
+///
+/// Uses DashMap for concurrent access without locking.
+/// Stores tuples of `(compiled_class, last_access_time)` to enable LRU eviction.
+/// Access time is updated on every cache hit to track recently used classes.
+pub(crate) static NATIVE_CACHE: std::sync::LazyLock<
+    DashMap<starknet_types_core::felt::Felt, (Arc<NativeCompiledClassV1>, Instant)>,
```
apoorvsadana: should we use `ClassHash` type wherever we've class hash? it's more readable that way.

madara/crates/primitives/class/src/native/cache.rs
```diff
@@ -0,0 +1,420 @@
...
+    config.cache_dir.join(filename)
+}
+
+/// Enforce disk cache size limit by removing oldest files.
+///
+/// Sorts all `.so` files by modification time and removes the oldest ones until
+/// the total size is below `max_size`. If `max_size` is `None`, no limit is enforced.
+///
+/// Called after each successful disk write to prevent unbounded growth.
+pub(crate) fn enforce_disk_cache_limit(cache_dir: &PathBuf, max_size: Option<u64>) -> Result<(), std::io::Error> {
```
apoorvsadana: interesting. is this something that the SW sequencer follows? because, by this same logic, we should also have a cache of CASM files and delete files when they get too big, but over there we assume this is a Madara requirement and we increase disk size.

also, if we need to clear the cache, shouldn't it be based on last accessed time and not modified? in which case will the compiled native binary be modified once its stored.

madara/node/src/cli/cairo_native.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// across long periods. A typical compiled class is 100 KB - 10 MB.
+    #[clap(
+        env = "MADARA_NATIVE_MAX_DISK_CACHE_BYTES",
+        long,
+        default_value_t = DEFAULT_DISK_CACHE_SIZE_BYTES,
+        value_parser = parse_disk_cache_size,
+        value_name = "BYTES"
+    )]
+    #[serde(default = "default_disk_cache_size")]
+    pub native_max_disk_cache_size: u64,
```
apoorvsadana: let's rename to 

`native_max_memory_cache_classes`
and
`native_max_disk_cache_bytes`

currently both are size but one refers to # of classes and the other to size of classes

madara/crates/primitives/class/src/native/cache.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    if let Some(cached_entry) = NATIVE_CACHE.get(class_hash) {
+        let (cached_class, _) = cached_entry.value();
+
+        // Clone before dropping the reference to avoid holding lock during insert
+        let cloned_class = cached_class.clone();
+        drop(cached_entry); // Reference dropped before insert to minimize lock hold time
+
+        // Access time updated for LRU eviction tracking
+        NATIVE_CACHE.insert(*class_hash, (cloned_class.clone(), Instant::now()));
```
apoorvsadana: is there a way to get a mutable here - `NATIVE_CACHE.get(class_hash)` and then we can directly update the time. this will avoid a close of the class when we only need to update the time.

madara/crates/primitives/class/src/native/cache.rs
```diff
@@ -0,0 +1,420 @@
...
+    sierra: &SierraConvertedClass,
+) -> Result<Option<RunnableCompiledClass>, ProgramError> {
+    let start = Instant::now();
+    let path = get_native_cache_path(class_hash, None);
+
+    if !path.exists() {
+        return Ok(None);
+    }
+
+    match try_load_executor_with_timeout(&path, DISK_LOAD_TIMEOUT)? {
```
apoorvsadana: what's the intuition behind a timeout for loading the file? is this something recommended/followed in the sequencer repo too? i am trying to draw an analogy to loading CASM files from the DB in normal execution, we don't have timeouts there, the operation is too primitive to handle edge cases around.

madara/crates/primitives/class/src/native/cache.rs
```diff
@@ -0,0 +1,420 @@
...
+        let config = super::config::get_config();
+        let max_size = config.max_memory_cache_size;
+
+        // Test that eviction threshold is calculated correctly
+        // If Some, should be positive; if None, unlimited cache is valid
+        if let Some(size) = max_size {
+            assert!(size > 0, "Max cache size should be positive when set");
+        }
+        // None is also valid (unlimited cache)
+    }
```
apoorvsadana: I think we can have much better test cases here to see if disk clearance and in memory cache is being cleared correctly.

madara/crates/primitives/class/src/native/compilation.rs
```diff
@@ -0,0 +1,560 @@
...
+pub(crate) static COMPILATION_IN_PROGRESS: std::sync::LazyLock<
+    DashMap<starknet_types_core::felt::Felt, Arc<RwLock<()>>>,
+> = std::sync::LazyLock::new(DashMap::new);
+
+/// Tracks classes that failed compilation in async mode (for retry on next request).
+///
+/// When a compilation fails in async mode, the class hash is added here.
+/// On the next request for the same class, compilation is automatically retried.
+/// Successful compilation removes the entry from this set.
+pub(crate) static FAILED_COMPILATIONS: std::sync::LazyLock<DashMap<starknet_types_core::felt::Felt, ()>> =
```
apoorvsadana: I guess this flag is only for logging purposes? We keep track of failed compilations and if they're present here, we add a specicial log in the next compilation. If yes, maybe we should edit the comment. I thought only classes here are being retried.

OR, it would also make sense to have a CLI flag called `native_should_retry`. If present, we retry compiling previously failed classes, else we skip. So the [`should_retry`](https://github.com/madara-alliance/madara/blob/384487e387d2accb3fbf02dc15be79ff9efc36bb/madara/crates/primitives/class/src/native/execution.rs#L178) here should come from the CLI. Although, I don't know in what case will a second compilation work if the first failed!

madara/crates/primitives/class/src/native/compilation.rs
```diff
@@ -0,0 +1,560 @@
...
+    // Additional check: ensure the hex representation is valid for filename
+    let hash_str = format!("{:#x}", class_hash);
+    if hash_str.is_empty() || hash_str.len() > config::MAX_CLASS_HASH_HEX_LENGTH {
+        return Err(NativeCompilationError::InvalidClassHash(
+            *class_hash,
+            format!("Invalid class hash length: {}", hash_str.len()),
+        ));
+    }
+    Ok(())
+}
```
apoorvsadana:
```suggestion
```

seems like an unnecessary check. if class_hash was invalid then it wouldn't exist in the DB anyways because we check for it during declare txs.

madara/crates/primitives/class/src/native/execution.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+        Entry::Occupied(_) => {
+            // Race condition: another thread started compilation
+            tracing::debug!(
+                target: "madara.cairo_native",
+                class_hash = %format!("{:#x}", class_hash),
+                "compilation_already_in_progress"
+            );
+        }
+    }
```
apoorvsadana: looking at the code and checks inside `spawn_native_compilation`, seems like its possible two class compilations can start in parallel. if yes, isn't it better to have a Mutex or something here, so that only one thread can pick up compilation for one class hash?

apoorvsadana: I think only one file should own `COMPILATION_IN_PROGRESS`. currently, both `spawn_compilation_if_needed` and `spawn_native_compilation` write to it which is confusing.

madara/crates/primitives/class/src/native/compilation.rs
```diff
@@ -0,0 +1,560 @@
...
+
+    // Disk cache limit enforced after successful compilation
+    if let Err(e) = cache::enforce_disk_cache_limit(&config.cache_dir, config.max_disk_cache_size) {
+        tracing::warn!(
+            target: "madara.cairo_native",
+            cache_dir = %config.cache_dir.display(),
+            error = %e,
+            "disk_cache_enforcement_failed"
+        );
+    }
```
apoorvsadana: can use `cache_compiled_native_class`?

madara/crates/primitives/class/src/native/compilation.rs
```diff
@@ -0,0 +1,560 @@
...
+        }
+        Err(_) => {
+            handle_async_compilation_failure(
+                class_hash,
+                "timeout",
+                format!("Compilation exceeded timeout of {:?}", compilation_timeout),
+                &path,
+                timer,
+                true,
+            );
```
apoorvsadana: nitpick, seems like an unnecessary function call. code could probably be here itself.

madara/crates/primitives/class/src/native/config.rs
```diff
@@ -0,0 +1,369 @@
...
+        if !self.cache_dir.is_dir() {
+            return Err(format!("Cache path {:?} is not a directory", self.cache_dir));
+        }
+
+        Ok(())
+    }
+}
+
+/// Global configuration instance
+static CONFIG: std::sync::OnceLock<NativeConfig> = std::sync::OnceLock::new();
```
apoorvsadana: I would avoid global configurations like this. We had this in the orchestrator in the beginning, makes parallel testing very difficult and it's also not coherent with the rest of the Madara code I believe. If the only entrypoint is going to be the blockifier state adapter, then we could store over there and pass down?

madara/crates/primitives/class/src/native/metrics.rs
apoorvsadana, line 1: do we want this on release builds? we could fire OTEL metrics but this seems to be only for local debugging.

madara/crates/tests/src/transaction_flow.rs
```diff
@@ -81,9 +98,29 @@ impl SetupBuilder {
...
+        if self.enable_native_execution {
+            args.push("--native-compilation-mode".into());
+            args.push("blocking".into());
+        }
+
+        args.into_iter().chain(self.block_production_disabled.then_some("--no-block-production".into()))
+    }
+
+    /// Get environment variables for the sequencer, including native cache dir
+    fn sequencer_env(&self) -> Vec<(String, String)> {
```
apoorvsadana: why not pass this via the `native_cache_dir` cli flag itself, like the other args

apoorvsadana (commented):
I think test cases can be more comprehensive. 

In cache, for example, test cases should check if cache is correctly deleting from memory/disk on overflows.

In compilations, we should check non blocking returns the non native version and blocking returns the native versions.

In compilations, we should check if a corrupt native file is used then automatically non native version is returned.

.github/actions/setup-rust/action.yml
Mohiiit: wydm by action files? it's being used by every CI workflow anyway, for making it a part of makefile I think it would require to have support for mac as well, here it's very ubuntu specific

madara/crates/client/analytics/src/formatter.rs
Mohiiit: I was printing cairo_native.<function done> which was kinda annoying, so I went ahead with this

also I noticed how cairo-native would have some time bounded operations as well, having logs helped there too, I was able to create a common format for all the logs types within cairo-native, it shows time taken etc as well

It looked much cleaner to me, attaching a ss

Screenshot 2025-11-06 at 6 06 24 PM

madara/crates/primitives/class/src/native/cache.rs
Mohiiit: SW sequencer also follows something like, they have `ClassManagerConfig` which has info for the cache size etc and yeah I was thinking along the same lines if we should maintain a cache for CASM too

and yeah it should be acessed, will update that

madara/crates/primitives/class/src/native/cache.rs
Mohiiit: for CASM given we have rocksdb, it handles the timeout for it, not the case as of now, once we have moved it behind rocksdb, we won't be doing this

> is this something recommended/followed in the sequencer repo too

they have cache config too yeah

madara/crates/primitives/class/src/native/compilation.rs
Mohiiit: yeah was for logging mostly, can add the native_should_retry flag

build-artifacts/src/lib.rs
Mohiiit: will handle the backward compatibility as well as remove the current created containers too

madara/crates/client/analytics/src/formatter.rs
Mohiiit: on a second thought, I have moved this all to a debug log, would be able to see it with info logs if we use 

`RUST_LOG=info,madara_cairo_native=debug`

.github/actions/setup-rust/action.yml
apoorvsadana: I mean the same code is being used in `[.github/actions/setup-rust/action.yml](https://github.com/madara-alliance/madara/pull/832/files/384487e387d2accb3fbf02dc15be79ff9efc36bb#diff-d83464742fc4e0fa81ee8df43771fd53360c7c7cdae4e5fd7df04d8e9ae99d5c) and in the Docker file. so we can have a ubuntu specific make command maybe and use that in both the place.

madara/crates/primitives/class/src/native/cache.rs
apoorvsadana: just saw the comment here, so every time we read from disk, the cache is updated again?

also, what happens if last access on disk is old because the binary is in memory? looks like we will delete it, we might want to optimise for this use case?

Mohiiit: > so every time we read from disk, the cache is updated again?

Yes, that's correct. When `try_get_from_disk_cache` successfully loads a file from disk, it updates the file's modification time (lines 472-482 in `cache.rs`). This is intentional behavior to track file access for LRU eviction. Reasoning behind updating the modification time is that many filesystems disable `atime` updates for performance. ( I haven't checked upon it yet tho)

> what happens if last access on disk is old because the binary is in memory? looks like we will delete it, we might want to optimise for this use case

if a `class_hash` was evicted from the `memory_cache` (meaning it was LRU `class_hash`), then only it will fall back to `disk_cache`, and in case the `disk_cache` evicted the `class_hash` (because it was not touched upon for a while), it makes sense to just recompile it?

I get the idea behind it tho, maybe there could be a score like system based on the `class_hash's` disk and memory recent usage

also I can foresee the case when we have a smaller `memory_cache` limit, this would be an issue (more frequent re-compilation would happen)

madara/crates/primitives/class/src/native/compilation.rs
Mohiiit: every error has different reasoning tho, hence keeping this function_call

madara/crates/primitives/class/src/native/cache.rs
apoorvsadana: pending?

Mohiiit: I tried different approaches over here, seems this is the only way, in case we get a mutable too, here is how it will go:

```rust
if let Some(mut entry) = NATIVE_CACHE.entry(class_hash).or_insert(...) {
    let (arc, instant) = entry.get_mut(); // Holds lock
    *instant = Instant::now(); // ✅ Update Instant without cloning
    let cached_class = arc.clone(); // Still need to clone Arc
    drop(entry); // Release lock
    // Do expensive operations...
}
```

also the cached/cloned_class in used just below for conversion

madara/crates/primitives/class/src/native/cache.rs
apoorvsadana: Ya, I can imagine for example, the ERC20 fee token will be deleted from disk because its always in memory. But all of this is just for the first txn on a restart maybe, so nothing much to optimise.

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+    #[test]
+    fn test_cache_size_metric() {
+        let _guard = TEST_MUTEX.lock().unwrap();
+        // Clear cache
+        NATIVE_CACHE.clear();
+
+        let config = config::NativeConfig::default();
+        // Cache operations should not panic
+        evict_cache_if_needed(&config);
+    }
```
apoorvsadana: same test as above?

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        let config = config::NativeConfig::default();
+        let max_size = config.max_memory_cache_size;
+
+        // Test that eviction threshold is calculated correctly
+        // If Some, should be positive; if None, unlimited cache is valid
+        if let Some(size) = max_size {
+            assert!(size > 0, "Max cache size should be positive when set");
+        }
+        // None is also valid (unlimited cache)
+    }
```
apoorvsadana: incorrect name? also unsure what this test is testing, default value being set correctly?

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+    use std::sync::Mutex;
+
+    /// Test mutex to serialize test execution for tests that need to clear/modify caches
+    static TEST_MUTEX: Mutex<()> = Mutex::new(());
+
+    /// Counter for generating unique test class hashes
+    static TEST_CLASS_HASH_COUNTER: AtomicU64 = AtomicU64::new(1);
+
+    /// Helper function to create a unique test class hash
+    fn create_unique_test_class_hash() -> ClassHash {
```
apoorvsadana, line 641: nitpcik: test helps can be fixtures

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+
+        // Try to get from memory cache - should succeed
+        let result = try_get_from_memory_cache(&class_hash, &create_simple_test_config(&temp_dir));
+        assert!(result.is_some(), "Memory cache should return Some when class is cached");
+
+        // Verify cache still contains the class (and access time was updated)
+        assert!(NATIVE_CACHE.contains_key(&class_hash), "Class should still be in cache after access");
+        if let Some(entry) = NATIVE_CACHE.get(&class_hash) {
+            let (_, access_time) = entry.value();
+            assert!(*access_time >= initial_time, "Access time should be updated to current or later time");
```
apoorvsadana:
```suggestion
            assert!(*access_time > initial_time, "Access time should be updated to current or later time");
```

`>=` seems like an incorrect check for an update

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        // Insert into cache (simulating what happens after disk load)
+        NATIVE_CACHE.insert(class_hash, (native_class.clone(), Instant::now()));
+
+        // Verify class is now in memory cache
+        assert!(NATIVE_CACHE.contains_key(&class_hash), "Class should be in memory cache");
+
+        // Request the class again - should hit memory cache
+        let result = try_get_from_memory_cache(&class_hash, &create_simple_test_config(&temp_dir));
+        assert!(result.is_some(), "Should hit memory cache after disk load");
+    }
```
apoorvsadana: test seems broken? we're inserting to memory inside the test case and checking if it exits?

shouldn't we call `try_get_from_disk_cache` and then check if it exists in `NATIVE_CACHE`

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+
+    #[test]
+    fn test_memory_cache_hit_updates_access_time() {
```
apoorvsadana: technically this is already being tested in `test_basic_memory_cache_hit`?

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        assert!(
+            updated_access_time > initial_access_time,
+            "Access time should be updated after cache hit. Initial: {:?}, Updated: {:?}",
+            initial_access_time,
+            updated_access_time
+        );
+    }
+
+    #[test]
+    fn test_memory_cache_hit_preserves_class_data() {
```
apoorvsadana: hmm, not sure if this is needed, we already check class hash exists and types ensure it must be a class

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        // Create test Sierra class
+        let sierra = create_test_sierra_class().expect("Failed to create test Sierra class");
+        let temp_dir = TempDir::new().expect("Failed to create temp dir");
+
+        // Create native class
+        let native_class =
+            create_test_native_class(&sierra, &temp_dir, class_hash, None).expect("Failed to create test native class");
+
+        // Insert into cache
+        NATIVE_CACHE.insert(class_hash, (native_class.clone(), Instant::now()));
```
apoorvsadana: looks like this happens in all test cases, can make it a fixture?

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        // Verify class_3 (newest) is still in cache
+        assert!(NATIVE_CACHE.contains_key(&class_hash_3), "Class 3 (newest) should remain in cache");
+
+        // Verify eviction metric was recorded
+        use std::sync::atomic::Ordering;
+        assert_eq!(test_counters::CACHE_EVICTIONS.load(Ordering::Relaxed), 1, "Should have exactly one cache eviction");
+    }
+
+    #[test]
+    fn test_eviction_uses_access_time_correctly() {
```
apoorvsadana: does this test include eveything in `test_lru_eviction_when_limit_reached`

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        use std::sync::atomic::Ordering;
+        assert_eq!(
+            test_counters::CACHE_EVICTIONS.load(Ordering::Relaxed),
+            2,
+            "Should have exactly two cache evictions"
+        );
+    }
+
+    #[test]
+    fn test_eviction_removes_correct_entry() {
```
apoorvsadana: is the only difference in this test that we're checking size of class? if yes, let's just keep this and not others although, as mentioned above, I am unsure if that's really needed

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        use std::sync::atomic::Ordering;
+        assert_eq!(
+            test_counters::CACHE_EVICTIONS.load(Ordering::Relaxed),
+            2,
+            "Should have exactly two cache evictions"
+        );
+    }
+
+    #[test]
+    fn test_limited_memory_cache_config_respected() {
```
apoorvsadana: same, not sure what this test is checking what's not being checked above

madara/crates/client/cairo_native/src/compilation.rs
```diff
@@ -0,0 +1,774 @@
...
+    // Helper to create unique test class hash
+    fn create_unique_test_class_hash() -> ClassHash {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let counter = COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        let base = Felt::from_hex_unchecked("0x3000000000000000000000000000000000000000000000000000000000000000");
+        ClassHash(base + Felt::from(counter))
+    }
+
+    // Helper to create test Sierra class
+    fn create_test_sierra_class() -> Result<SierraConvertedClass, Box<dyn std::error::Error>> {
```
apoorvsadana: duplicate with `cache.rs`?

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1614 @@
...
+        }
+        None => {
+            // Timeout or file locked - treat as cache miss
+            Ok(None)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
apoorvsadana, line 620: feels like we don't need these many test cases. we've a few test cases that encapuslate others.

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1747 @@
...
+
+    // Helper function to create a unique test class hash
+    fn create_unique_test_class_hash() -> ClassHash {
+        let counter = TEST_CLASS_HASH_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        let base = Felt::from_hex_unchecked("0x2000000000000000000000000000000000000000000000000000000000000000");
+        ClassHash(base + Felt::from(counter))
+    }
+
+    // Helper function to create a test SierraConvertedClass (reuse from cache.rs tests)
+    fn create_test_sierra_class() -> Result<SierraConvertedClass, Box<dyn std::error::Error>> {
```
apoorvsadana: duplicate with other files

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1747 @@
...
+
+        // Create config with native disabled
+        let config = Arc::new(
+            config::NativeConfig::default().with_native_execution(false).with_cache_dir(temp_dir.path().to_path_buf()),
+        );
+
+        let _metrics_guard = test_counters::acquire_and_reset();
+
+        // VM class returned immediately when native is disabled
+        let result =
```
apoorvsadana: shouldn't we assert VM type here

apoorvsadana (commented):
review is still pending, need to majorly go through test cases but overview is that there are more test cases then needed (will make it harder to maintain with time)

madara/crates/client/cairo_native/src/compilation.rs
```diff
@@ -0,0 +1,720 @@
...
+            // Spawn compilation
+            spawn_compilation_if_needed(class_hash, sierra.clone(), config.clone());
+        }
+
+        // Wait a bit to allow compilations to start and acquire semaphore permits
+        tokio::time::sleep(Duration::from_millis(500)).await;
+
+        // Verify that at most `max_concurrent` compilations are in progress
+        // (The semaphore should limit how many compilations actually start)
+        let in_progress_count = COMPILATION_IN_PROGRESS.len();
```
apoorvsadana: nitpick
```suggestion
        let in_progress_count = get_current_compilations_count();
```

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+            &sierra_class.info,
+            async_config.clone(),
+        );
+
+        assert!(result.is_ok(), "Should successfully retrieve from memory cache");
+        let runnable = result.unwrap();
+        assert!(std::mem::size_of_val(&runnable) > 0, "Should return valid RunnableCompiledClass");
+
+        // Verify it's still in cache after call
+        assert!(cache::NATIVE_CACHE.contains_key(&class_hash), "Class should remain in memory cache after call");
```
apoorvsadana: hmm, any file being able to directly get access to `NATIVE_CACHE` feels like a bad practice. i think it would be better design to make `NATIVE_CACHE` as private and only accessible via function calls.

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        assert_eq!(
+            test_counters::COMPILATIONS_STARTED.load(Ordering::Relaxed),
+            0,
+            "Should have no compilations (native disabled)"
+        );
+        assert_eq!(
+            test_counters::VM_FALLBACKS.load(Ordering::Relaxed),
+            0,
+            "Should have no VM fallbacks (native disabled)"
+        );
```
apoorvsadana: nitpick: i would create a macro where I can do 

```
assert_counters!(
    CACHE_HITS_MEMORY: 0,
    CACHE_MEMORY_MISS: 0,
    ...
)
```
and it would automatically translate to the code here

apoorvsadana: although this counter approach of testing is pretty cool, allows to test the exact code flow

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+            &sierra_class,
+            &class_hash.to_felt(),
+            &sierra_class.compiled,
+            &sierra_class.info,
+            async_config.clone(),
+        );
+
+        assert!(result.is_ok(), "Should return VM class immediately in async mode");
+        let runnable = result.unwrap();
+        assert!(std::mem::size_of_val(&runnable) > 0, "Should return valid RunnableCompiledClass");
```
apoorvsadana: need to asset its not of Native type

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        let _metrics_guard = test_counters::acquire_and_reset();
+
+        let class_hash = create_unique_test_class_hash();
+
+        // Verify memory cache doesn't have our unique class_hash
+        assert!(!cache::NATIVE_CACHE.contains_key(&class_hash), "Class should not be in memory cache initially");
+
+        // Mark compilation as in progress
+        use tokio::sync::RwLock;
+        compilation::COMPILATION_IN_PROGRESS.insert(class_hash, Arc::new(RwLock::new(())));
```
apoorvsadana: same for these static variables too

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        // Disk cache skipped when compilation is in progress
+        let result = handle_sierra_class(
+            &sierra_class,
+            &class_hash.to_felt(),
+            &sierra_class.compiled,
+            &sierra_class.info,
+            async_config.clone(),
+        );
+
+        assert!(result.is_ok(), "Should return VM class when compilation is in progress");
```
apoorvsadana, line 871: we should assert VM class?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        // Compilation timeout expected in blocking mode with very short timeout
+        let result = handle_sierra_class(
+            &sierra_class,
+            &class_hash.to_felt(),
+            &sierra_class.compiled,
+            &sierra_class.info,
+            config.clone(),
+        );
+
+        assert!(result.is_err(), "Should return an error when compilation fails");
```
apoorvsadana: what happens to this txn if this case happens on production?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+            &sierra_class,
+            &class_hash.to_felt(),
+            &sierra_class.compiled,
+            &sierra_class.info,
+            config.clone(),
+        );
+
+        assert!(result.is_ok(), "Should successfully retrieve from disk cache after memory miss");
+        let runnable = result.unwrap();
+        assert!(std::mem::size_of_val(&runnable) > 0, "Should return valid RunnableCompiledClass");
```
apoorvsadana: should check for VM class?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+    /// `native_memory_cache_timeout_ms` in the config.
+    ///
+    /// Uses an extremely short timeout (1 nanosecond) to reliably trigger the timeout,
+    /// since conversion operations take significantly longer. The test validates:
+    /// - Timeout configuration is respected
+    /// - Fallback to disk cache occurs after memory cache timeout
+    /// - Timeout metric (`CACHE_MEMORY_TIMEOUT`) is recorded
+    /// - Timeout events are logged appropriately
+    #[rstest]
+    fn test_handle_sierra_class_memory_cache_timeout(sierra_class: SierraConvertedClass, temp_dir: TempDir) {
```
apoorvsadana, line 980: nitpick: timeout test cases should ideally live inside `cache`?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+            &class_hash.to_felt(),
+            &sierra_class.compiled,
+            &sierra_class.info,
+            config.clone(),
+        );
+
+        // Async mode returns VM immediately while compilation happens in background
+        assert!(result.is_ok(), "Should return VM class immediately in async mode");
+        let runnable = result.unwrap();
+        assert!(std::mem::size_of_val(&runnable) > 0, "Should return valid RunnableCompiledClass");
```
apoorvsadana: ditto

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+
+    tracing::debug!(
+        target: "madara_cairo_native",
+        class_hash = %format!("{:#x}", class_hash),
+        "returning_vm_class"
+    );
+    Ok(vm_class)
+}
+#[cfg(test)]
+mod tests {
```
apoorvsadana, line 433: pretty elaborate test cases!

madara/crates/client/cairo_native/src/compilation.rs
```diff
@@ -0,0 +1,720 @@
...
+/// Maximum size is limited to prevent unbounded growth (default: 10,000 entries).
+pub(crate) static FAILED_COMPILATIONS: std::sync::LazyLock<DashMap<ClassHash, Instant>> =
+    std::sync::LazyLock::new(DashMap::new);
+
+/// Evict entries from FAILED_COMPILATIONS if it exceeds the size limit.
+///
+/// Uses LRU (Least Recently Used) eviction based on failure timestamp.
+/// Sorts all entries by timestamp (oldest first) and removes entries until
+/// the map size is within the configured limit.
+fn evict_failed_compilations_if_needed(max_failed_compilations: usize) {
```
apoorvsadana, line 164: this seems like an overkill no? at a million classes, this would take ~60mb.

configs/args/config.json
```diff
@@ -108,5 +108,16 @@
...
+  "cairo_native_params": {
+    "enable_native_execution": false,
+    "native_max_memory_cache_size": 1000,
+    "native_max_concurrent_compilations": 4,
+    "native_compilation_timeout_secs": 300,
+    "native_memory_cache_timeout_ms": 100,
+    "native_disk_cache_load_timeout_secs": 2,
+    "native_compilation_mode": "async",
+    "native_enable_retry": true,
+    "native_max_failed_compilations": 10000
```
heemankv, line 125: Any reason to not have max_compilation_retry (per class) ? I would assume if we are tracking native_max_failed_compilations we should also be tracking that?

madara/crates/client/cairo_native/src/compilation.rs
```diff
@@ -0,0 +1,720 @@
...
+/// If accessed before initialization, falls back to default value.
+static COMPILATION_SEMAPHORE: std::sync::OnceLock<Arc<tokio::sync::Semaphore>> = std::sync::OnceLock::new();
+
+/// Initialize the compilation semaphore with the actual config value.
+///
+/// This must be called during config initialization (via `setup_and_log()`).
+/// The semaphore limit controls how many compilations can run concurrently.
+pub fn init_compilation_semaphore(max_concurrent: usize) {
+    COMPILATION_SEMAPHORE.get_or_init(|| Arc::new(tokio::sync::Semaphore::new(max_concurrent)));
+}
```
heemankv, line 226: Should we not use `set` here, we are calling this fn only from `main.rs` (and tests) ? 
IMO `get_or_init` is for lazy initialisation, that would have been apt for cases (as the name suggests) were we try to get value, and if it's not intialised we use init.

If we use `set` here we'll be able to handle the errors as well
```rust
pub fn init_compilation_semaphore(max_concurrent: usize) {
    COMPILATION_SEMAPHORE
        .set(Arc::new(tokio::sync::Semaphore::new(max_concurrent)))
        .expect("Compilation semaphore already initialized");
}
```

`get_or_init` would have made more sense if we were using this fn to get the value as well

madara/crates/client/cairo_native/src/metrics.rs
```diff
@@ -0,0 +1,696 @@
...
+
+/// Get the global metrics instance.
+///
+/// Returns a thread-safe reference to the global metrics singleton.
+/// Safe to call from any thread.
+///
+/// **Note**: For OTEL export to work, `register()` should be called during application
+/// initialization (typically via `Analytics::setup()`). If not called, metrics will
+/// still work but won't be exported.
+pub fn metrics() -> &'static NativeMetrics {
```
heemankv, line 658: nitpick: Let's rename it to `init_metrics()`.

madara/node/src/main.rs
```diff
@@ -221,6 +221,19 @@ async fn main() -> anyhow::Result<()> {
...
+    // Native execution is opt-in and only enabled when the --enable-native-execution CLI flag
+    // is set to true (default: false). When disabled, all contracts will use Cairo VM execution
+    // regardless of cache state.
+    let cairo_native_config = run_cmd.cairo_native_params.to_runtime_config();
+
+    // Setup: validate, initialize semaphore, and log configuration
+    // Note: Validation happens inside setup_and_log() via NativeConfig::validate()
+    mc_cairo_native::config::setup_and_log(&cairo_native_config)
+        .map_err(|e| anyhow::anyhow!("Cairo Native configuration setup failed: {}", e))?;
+
```
heemankv, line 242: Thought process here : 
Currently : 
- CairoNative is a client, we have all the client started and stopped via ServiceMonitor.
- CairoNative is more of a extension to BlockProduction Service right now hence it shall be initiated before and closed after BlockProduction Service, respectively.

Possible Future : 
- Given, CairoNative will be needed in `RPC`, `Full Node` and `Block Production` later on, we can think of it as either an extension to `MadaraBackend` or as a client being managed from `ServiceMonitor`.

Actionables : 
- Should we move to CairoNative as an extension of MadaraBackend or shall we keep it as a client and then use ServiceMonitor to start it.

madara/crates/client/db/src/lib.rs
```diff
@@ -264,6 +265,12 @@ pub struct MadaraBackend<DB = RocksDBStorage> {
     /// Current finalized block_n on L1.
     latest_l1_confirmed: tokio::sync::watch::Sender<Option<u64>>,
 
+    /// Cairo Native execution configuration.
+    ///
+    /// This config is passed through to BlockifierStateAdapter for execution.
+    /// The `enable_native_execution` flag in the config controls whether native execution is used.
+    pub cairo_native_config: Arc<NativeConfig>,
```
heemankv, line 272: Can we not use as chain_config.cairo_native_config ? 
Am I missing something here

madara/node/src/main.rs
```diff
@@ -221,6 +221,19 @@ async fn main() -> anyhow::Result<()> {
     let sys_info = SysInfo::probe();
     sys_info.show();
 
+    // Initialize Cairo Native configuration
+    //
+    // The configuration is validated and passed through parameters (no global state).
+    // Native execution is opt-in and only enabled when the --enable-native-execution CLI flag
+    // is set to true (default: false). When disabled, all contracts will use Cairo VM execution
+    // regardless of cache state.
+    let cairo_native_config = run_cmd.cairo_native_params.to_runtime_config();
```
heemankv, line 236:
```rust
let cairo_native_config = run_cmd.cairo_native_params.to_runtime_config();
```

Any reason to not have this passed through chain_config.rs,
I think we can maintain this there and thus be able to simplify code by not passing it explicitly into MadaraBackend, since we'll already have chain_config present there.

madara/Dockerfile
```diff
@@ -19,7 +19,28 @@ ENV RUSTC_WRAPPER=/bin/sccache
...
+ENV PATH="/usr/lib/llvm-19/bin:${PATH}"
+
+# Set explicit compilers for Cairo Native runtime compilation
+# Note: clang-19 handles both C and C++ compilation
+ENV CC=clang-19
+ENV CXX=clang-19
+
+# Configure linker search paths for Cairo Native runtime compilation
+ENV LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:/usr/lib:/lib/x86_64-linux-gnu:/lib
+ENV LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:/usr/lib:/lib/x86_64-linux-gnu:/lib
```
heemankv, line 43: Nitpick: 
Makefile is very big, if you wanna go bonkers we can actually do this : 
`Makefile` and `Makefile.llvm `
write all the llvm related make commands in `Makefile.llvm` and then import it in Makefile by `include Makefile.llvm`
then you have things working as it is locally and you only need to : 
```bash
COPY Makefile.llvm /tmp/Makefile 
```
here !

madara/crates/client/cairo_native/src/config.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// - Best for testing or when strict native execution is required
+    Blocking,
+}
+
+/// Configuration for Cairo Native execution and compilation.
+///
+/// This struct holds all runtime configuration for native execution.
+/// It is initialized once at startup and cannot be changed afterward.
+#[derive(Debug, Clone)]
+pub struct NativeConfig {
```
heemankv: IMO we should strengthen the type boundations here : 

This way we don't have to care about other params in the Disabled case 
and rather than checking for

```rust
/// Configuration for Cairo Native execution
#[derive(Debug, Clone)]
pub enum NativeConfig {
    /// Native execution is disabled - all contracts use Cairo VM
    Disabled,
    
    /// Native execution is enabled with the following configuration
    Enabled(NativeExecutionConfig),
}

#[derive(Debug, Clone)]
pub struct NativeExecutionConfig {
    /// Directory path for storing compiled native classes
    pub cache_dir: PathBuf,

    /// Maximum number of classes to keep in memory cache (None = unlimited)
    pub max_memory_cache_size: Option<usize>,

    /// Maximum disk cache size in bytes (None = unlimited)
    pub max_disk_cache_size: Option<u64>,

    /// Maximum number of concurrent compilations
    pub max_concurrent_compilations: usize,

    /// Maximum time to wait for a single compilation
    pub compilation_timeout: Duration,

    /// Maximum time to wait for memory cache lookup
    pub memory_cache_timeout: Duration,

    /// Maximum time to wait for disk cache load
    pub disk_cache_load_timeout: Duration,

    /// Compilation mode (async or blocking)
    pub compilation_mode: NativeCompilationMode,

    /// Enable automatic retry of failed compilations.
    ///
    /// When `true`, classes that failed compilation in async mode are automatically retried
    /// on the next request. When `false`, failed compilations are not retried automatically.
    pub enable_retry: bool,

    /// Maximum number of failed compilation entries to keep before evicting oldest.
    ///
    /// When a compilation fails, the class hash is stored in `FAILED_COMPILATIONS` for retry tracking.
    /// This setting limits how many failed entries are kept in memory. When the limit is reached,
    /// the oldest entries are evicted using LRU policy.
    ///
    /// Higher values allow more failed classes to be tracked for retry, but use more memory.
    pub max_failed_compilations: usize,
}```

and then usage will change along:

```rust
let NativeConfig::Enabled(config) = config else {
    return handle_native_disabled(class_hash, compiled, info);
};
```

type checking ensures that all possible cases at compile time itseld, limiting mistakes : 
```rust
match native_config {
    NativeConfig::Disabled => { /* must handle */ },
    NativeConfig::Enabled(config) => { /* must handle */ },
}
````

File: madara/crates/client/cairo_native/src/config.rs [hunk 53]
```diff
@@ -0,0 +1,461 @@
...
+    /// - No VM fallback - failures cause transaction errors
+    /// - Best for testing or when strict native execution is required
+    Blocking,
+}
+
+/// Configuration for Cairo Native execution and compilation.
+///
+/// This struct holds all runtime configuration for native execution.
+/// It is initialized once at startup and cannot be changed afterward.
+#[derive(Debug, Clone)]
```
heemankv, line 87: I see a big opportuinity to cut code by using builder pattern amongst `NativeConfig` & `CairoNativeParams`. 
A self consuming builder is even better! I've used that in `e2e`-> madara-alliance/madara/e2e/src/services/madara/config.rs

madara/crates/client/cairo_native/src/config.rs
```diff
@@ -0,0 +1,461 @@
...
+pub const DEFAULT_COMPILATION_TIMEOUT_SECS: u64 = 300;
+
+/// Default memory cache lookup timeout: 100 milliseconds
+pub const DEFAULT_MEMORY_CACHE_TIMEOUT_MS: u64 = 100;
+
+/// Default disk cache load timeout: 2 seconds
+pub const DEFAULT_DISK_CACHE_LOAD_TIMEOUT_SECS: u64 = 2;
+
+/// Default memory cache size: 1000 classes
+pub const DEFAULT_MEMORY_CACHE_SIZE: usize = 1000;
```
heemankv, line 50: Do we need a capping on the memory usage ?

madara/crates/client/cairo_native/src/metrics.rs
```diff
@@ -0,0 +1,696 @@
...
+    pub fn new() -> Self {
+        metrics().record_compilation_start();
+        Self { start: Instant::now() }
+    }
+
+    pub fn finish(self, success: bool, timeout: bool) {
+        let duration_ms = self.start.elapsed().as_millis() as u64;
+        metrics().record_compilation_end(duration_ms, success, timeout);
+    }
+}
```
heemankv, line 696: Currently, our metrics only track **aggregated** failure counts:
- `cairo_native_compilations_failed_total` - total failures across all classes
- `cairo_native_compilations_timeout_total` - total timeouts across all classes

This makes it impossible to answer critical operational questions:
- **Which specific contract classes are failing to compile?**
- **Which classes timeout most frequently?**
- **Is a specific class causing repeated compilation failures?**

heemankv: please note that just adding the class_hash as an attribute will actually create new time-series with each new class, we might wanna do something about that, it's something to do with cardinality imo

madara/crates/client/db/src/tests/test_open.rs
```diff
@@ -45,20 +47,24 @@ async fn test_open_different_chain_id() {
...
     let chain_config = std::sync::Arc::new(ChainConfig::madara_test());
+    let cairo_native_config = Arc::new(NativeConfig::default());
     assert!(MadaraBackend::open_rocksdb(
         temp_dir.path(),
         chain_config,
         MadaraBackendConfig::default(),
         RocksDBConfig::default(),
+        cairo_native_config,
     )
     .is_err());
```
heemankv, line 69: Not needed once we use cairo_native_config from within chain_config

madara/crates/primitives/class/src/compile.rs
```diff
@@ -134,26 +151,62 @@ impl FlattenedSierraClass {
...
-    pub fn compile_to_native(&self) -> Result<AotContractExecutor, ClassCompilationError> {
+    /// Compile this Sierra class to Cairo Native format.
+    ///
+    /// This function performs ahead-of-time compilation of the Sierra program to native
+    /// code, writing the compiled binary to the specified path. The resulting executor
+    /// can be used for native execution instead of the Cairo VM.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - File system path where the compiled native binary (`.so` file) will be stored.
```
heemankv: Curious : How is the locking handled here ? 
is the file written with some chars that defines lock, does blockifier maintain a lock file ?

madara/crates/client/cairo_native/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+description = "Cairo Native execution and compilation for Madara"
+name = "mc-cairo-native"
```
heemankv: Nitpick : 
I think this should better be `mc-native-exec` to follow the convention that it's an actionable service like `mc-exec`

heemankv: or `mc-cairo-native-exec`

madara/node/src/main.rs
heemankv: After reading the whole codebase now, it makes sense to keep it as a client rn, great work,
I would still push on starting and managing it via service manager

madara/crates/client/cairo_native/Cargo.toml
heemankv: Initially I thought that only the cario-native relevant code must be here based on the description and all,
but seeing the fns `convert_to_vm_class`, `convert_to_vm_class` & `handle_native_disabled` in `execution.rs` broadens the perspective of this client
I am heavily leaning towards `mc-class-exec` or something now, wdyt?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        target: "madara_cairo_native",
+        class_hash = %format!("{:#x}", class_hash),
+        native_enabled = config.enable_native_execution,
+        "handle_sierra_class"
+    );
+
+    // If native execution is disabled, use VM immediately
+    if !config.enable_native_execution {
+        return handle_native_disabled(class_hash, compiled, info);
+    }
```
heemankv: this logic should be isolated from mc-cario-native client (*iif* we keep the name as it is), 
should be handled in `blockifier_state_adapter` to never needing to use mc-cario-native if not needed. 
This might lead to re-arrange `to_blockifier` as well,

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1006 @@
...
+
+/// Get the cache file path for a class hash.
+///
+/// Constructs a path like `{cache_dir}/{class_hash:#x}.so`.
+/// The filename is validated to prevent path traversal attacks.
+pub(crate) fn get_native_cache_path(class_hash: &ClassHash, config: &config::NativeConfig) -> PathBuf {
+    // Convert ClassHash to Felt for formatting (ClassHash implements ToFelt)
+    let felt = class_hash.to_felt();
+    // Ensure filename is safe (no path traversal)
+    let filename = format!("{:#x}.so", felt);
```
heemankv, line 96: Nitpick: We can create a `NativeCompiledClass` struct and add this as a method of it's implementation 
Seems safer while code edits/re-factor.
Being used here + Tests

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -0,0 +1,1006 @@
...
+    let class_hash_for_log = *class_hash;
+    let (tx, rx) = mpsc::channel();
+
+    let thread_handle = thread::spawn(move || {
+        let start = Instant::now();
+
+        if let Some(cached_entry) = NATIVE_CACHE.get(&class_hash_for_log) {
+            // Retrieve the cached class (Arc clone is cheap - just increments reference count, doesn't copy data)
+            let cached_class = cached_entry.value().0.clone();
+            drop(cached_entry); // Release lock quickly to avoid blocking other threads
```
heemankv: Misleading comment, Since we are using Dashmaps here's a better comment:  `// Release shard lock before expensive operations below`

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+            let runnable = RunnableCompiledClass::from(native_class.as_ref().clone());
+            let conversion_elapsed = conversion_start.elapsed();
+            let total_elapsed = start.elapsed();
+            super::metrics::metrics().record_compilation_blocking_complete();
+            super::metrics::metrics()
+                .record_compilation_blocking_conversion_time(conversion_elapsed.as_millis() as u64);
+            tracing::debug!(
+                target: "madara_cairo_native",
+                class_hash = %format!("{:#x}", class_hash.to_felt()),
+                elapsed = ?total_elapsed,
```
heemankv, line 293: Why does it not check for compilation in progress ? 
If two concurrent requests arrive and both skip the disk cache due to COMPILATION_IN_PROGRESS, they will both call compile_native_blocking right ? is that safe ?

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1538 @@
...
+        super::metrics::metrics().record_compilation_in_progress_skip();
+        tracing::debug!(
+            target: "madara_cairo_native",
+            class_hash = %format!("{:#x}", class_hash.to_felt()),
+            elapsed = ?elapsed,
+            elapsed_ms = elapsed.as_millis(),
+            "compilation_in_progress_skipping_disk_cache"
+        );
+        return None;
+    }
```
heemankv: this seems like a guard clause for `try_get_from_disk_cache`, any reason for not having it there?

heemankv (commented):
Looks like a very well implemented PR ngl, gg cursor master !

.github/actions/setup-rust/action.yml
apoorvsadana: still pending?

build-artifacts/src/lib.rs
apoorvsadana: the filter is w/o the timestamp here, so it won't filter our current containers right (which have the timestamp)?

madara/crates/client/cairo_native/src/compilation.rs
apoorvsadana: pending, if you're pushing a new commit, do fix in that, otherwise can skip

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -0,0 +1,1473 @@
...
+            CACHE_HITS_MEMORY: 0,
+            CACHE_MEMORY_MISS: 1,
+            CACHE_HITS_DISK: 0,
+            CACHE_DISK_MISS: 1,
+            VM_FALLBACKS: 1,
+            COMPILATIONS_STARTED: 1,
+        );
+        assert_counters!(
+            COMPILATIONS_SUCCEEDED: 1,
+        );
```
apoorvsadana, line 776: merge to one?

madara/crates/client/cairo_native/src/execution.rs
apoorvsadana: pending?

madara/crates/client/cairo_native/src/execution.rs
apoorvsadana: pending? ig, wherever there's `std::mem::size_of_val`, we can remove it and assert on the class directly.

madara/crates/client/cairo_native/src/execution.rs
apoorvsadana: pending?

Mohiiit: this test makes sense here, all the tests calling `handle_sierra_class` are in execution only!

build-artifacts/src/lib.rs
Mohiiit: done now

madara/crates/client/cairo_native/src/cache.rs
heemankv: Continuing without

madara/crates/client/cairo_native/src/execution.rs
heemankv: We've changed the name to `mc-class-exec` to increase the scope of the client

madara/crates/client/db/src/tests/test_open.rs
heemankv: Not shifting cairo_native_config in chain_config since it's not a chain specific config, rather exec based

madara/crates/client/cairo_native/src/metrics.rs
heemankv: Not continuing with this right now since we will be cognisant about any class failing and it will be looked at priority

heemankv: logs are also enough to track which class fails

heemankv (approved):

apoorvsadana (approved):

PR #835, opened 2025-11-05
replace_chain_tip wait for fsync

heemankv:
Scenarios as such can only happen if graceful shutdown was not done. this code change ensures that we have fsynced with disk before we present with closed block log, ensuring that all criteria related to closing a block have been synced to db and not just the memtable of the db.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/madara/crates/client/db/src/lib.rs
+++ b/madara/crates/client/db/src/lib.rs
@@ -745,11 +745,6 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
     /// In addition, you must have fully imported the block using the low level writing primitives for each of the block
     /// parts.
     pub fn new_confirmed_block(&self, block_number: u64) -> Result<()> {
-        self.inner.db.on_new_confirmed_head(block_number)?; // Update snapshots for storage proofs. (TODO (heemank 06/11/2025): decouple this logic)
-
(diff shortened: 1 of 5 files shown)


apoorvsadana (approved):

madara/crates/client/db/src/lib.rs
```diff
@@ -745,6 +745,11 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
     /// In addition, you must have fully imported the block using the low level writing primitives for each of the block
     /// parts.
     pub fn new_confirmed_block(&self, block_number: u64) -> Result<()> {
+        self.inner.db.on_new_confirmed_head(block_number)?; // Update snapshots for storage proofs. (FIXME: decouple this logic)
```
Mohiiit:
```suggestion
        self.inner.db.on_new_confirmed_head(block_number)?; // Update snapshots for storage proofs. (TODO (heemank 06/11/2025): decouple this logic)
```

Mohiiit (approved):
Just curious, what will happen if after the chain tip has been updated but before flushing madara dies down?

PR #838, opened 2025-11-08
fix: batching logic to handle single block edge case

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

There is an edge case for SNOS batching with a single block.

Resolves: #NA

## What is the new behavior?

All batches are created properly. Update the response structure for bouncer weights.

### DIFF
--- a/orchestrator/src/worker/event_handler/triggers/batching.rs
+++ b/orchestrator/src/worker/event_handler/triggers/batching.rs
@@ -547,7 +547,7 @@ impl BatchingTrigger {
         if self.should_close_snos_batch(config, &current_snos_batch).await? {
             // Close the current batch and start a new batch
             self.update_or_create_snos_batch_in_db(&current_snos_batch, config, SnosBatchStatus::Closed).await?;
-            self.start_snos_batch(current_snos_batch.snos_batch_id + 1, None, block_number)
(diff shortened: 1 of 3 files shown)


Mohiiit (approved):
LGTM!

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -547,7 +547,7 @@ impl BatchingTrigger {
         if self.should_close_snos_batch(config, &current_snos_batch).await? {
             // Close the current batch and start a new batch
             self.update_or_create_snos_batch_in_db(&current_snos_batch, config, SnosBatchStatus::Closed).await?;
-            self.start_snos_batch(current_snos_batch.snos_batch_id + 1, None, block_number)
+            self.start_snos_batch(current_snos_batch.snos_batch_id + 1, None, current_snos_batch.end_block + 1)
```
apoorvsadana, line 550: no change for l2s?

prkpndy: yeah, L2s don't require changes since there the logic is different (should close batch is checked for one block before - so it should be fine. have confirmed this as well with edge cases. will test e2e again after L2s is fixed to handle Fusaka upgrade)

apoorvsadana (approved):

PR #841, opened 2025-11-10
fix: 0.14.0 E2E tests and change settlement to handle fusaka upgrade

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

E2E fails. Settlement fails on Ethereum sepolia

Resolves: #NA

## What is the new behavior?

This PR fixes the issue in the E2E test because of 0.14.0 update in Madara. It also updates the state update job (and ethereum settlement client) to handle the fusaka upgrade on sepolia.
* bumps the SNOS to use 0.14.1 blockifier
* bump c-kzg library to create cell proof
* update trusted_setup.txt file


e2e/config/madara.yaml
```diff
@@ -4,39 +4,39 @@ feeder_gateway_url: "http://localhost:8080/feeder_gateway/"
...
 mempool_max_transactions: 10000
-mempool_tx_limit: 10000
-mempool_declare_tx_limit: 20
-mempool_tx_max_age: null
+mempool_max_declare_transactions: 20
 mempool_ttl: null
 l2_gas_target: 2000000000
-min_l2_gas_price: 0
+min_l2_gas_price: 1000000
 l2_gas_price_max_change_denominator: 48
```
apoorvsadana: we don't need to use the latest chain config w/ version?

orchestrator/crates/da-clients/ethereum/src/lib.rs
```diff
@@ -1,38 +1,48 @@
...
 use async_trait::async_trait;
 use color_eyre::Result;
 use mockall::automock;
 use mockall::predicate::*;
 use orchestrator_da_client_interface::{DaClient, DaVerificationStatus};
-use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use url::Url;
 
+pub type DefaultHttpProvider = alloy::providers::fillers::FillProvider<
```
apoorvsadana, line 15: curious, is this the default way of using alloy providers now, to have this type in code? or we needed to customise something so we did this?

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -126,6 +129,40 @@ pub(crate) async fn prepare_sidecar(
...
+        sidecar_commitments.push(FixedBytes::new(commitment.to_bytes().into_inner()));
+
+        // Compute cells and KZG proofs for this blob
+        // This returns ALL cell proofs (including extension cells)
+        let (_, proofs) = trusted_setup.compute_cells_and_kzg_proofs(&blob)?;
+        // Store all cell proofs for this blob (CELLS_PER_EXT_BLOB proofs, each 48 bytes)
+        let mut blob_cell_proofs = Vec::with_capacity(128);
+        for proof in proofs.iter() {
+            blob_cell_proofs.push(FixedBytes::new(proof.to_bytes().into_inner()));
+        }
```
apoorvsadana: instead of `iter`, you can probably use something like `extend`

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -497,8 +517,67 @@ impl EthereumSettlementClient {
         Ok(tx_signed.into())
     }
 
+    /// Method to create a cell transaction (post-Fusaka, for sepolia/testnet)
+    /// Creates transaction with cell proofs
+    /// Returns the signed transaction variant with EIP7594 sidecar
+    async fn create_cell_transaction(
```
apoorvsadana: if the only difference is a sidecar, we don't need to copy paste the entire code.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -263,36 +258,60 @@ impl SettlementClient for EthereumSettlementClient {
...
-                        mul_factor = self.get_next_mul_factor(mul_factor)?;
-                        continue;
-                    }
-                    SendTransactionError::Other(_) => {
-                        bail!("Failed to send state update transaction: {:?}", e);
-                    }
-                },
+            // Create and send transaction based on whether we're on mainnet or sepolia
+            // Mainnet uses blob proofs (pre-Fusaka), Sepolia uses cell proofs (post-Fusaka)
+            let pending_transaction = if self.is_mainnet {
```
apoorvsadana: a better design to do this would probably be

1. create an enum for BlobTransactionType w/ 2 values - PreEip7594, Eip7594
2. implement two methods for this enum type - create_transaction and send_transaction
2. if possible, make an rpc call on spec version or similar to see of the node on the other end supports Eip7594. if yes, accordingly create the correct enum and call the method. if not available dynamically, then pass it from the flags

is mainnet or not is not the correct flag however, gives very little information to the reader on the relevance of the flag.

e2e/config/madara.yaml
heemankv: +1

bootstrapper/src/helpers/account_actions.rs
```diff
@@ -48,7 +48,7 @@ impl AccountActions for SingleOwnerAccount<&RpcClientProvider, LocalWallet> {
         let calls = vec![Call { to: address, selector: get_selector_from_name(method).unwrap(), calldata }];
         match nonce {
             Some(nonce) => self.execute_v3(calls).nonce(nonce.into()),
-            None => self.execute_v3(calls),
+            None => self.execute_v3(calls).l1_gas(0).l2_gas(0).l1_data_gas(0),
```
heemankv: Comment on why this is needed would be helpful !

e2e/src/services/madara/config.rs
```diff
@@ -89,8 +89,8 @@ impl Default for MadaraConfig {
             charge_fee: false,
             block_time: None,
             l1_endpoint: None,
-            l1_gas_price: 0,
-            blob_gas_price: 0,
+            l1_gas_price: 1,
```
heemankv: Comment explaining this would be helpful

orchestrator/crates/da-clients/ethereum/src/lib.rs
```diff
@@ -1,38 +1,48 @@
...
+                alloy::providers::fillers::JoinFill<
+                    alloy::providers::fillers::NonceFiller,
+                    alloy::providers::fillers::ChainIdFiller,
+                >,
+            >,
+        >,
+    >,
+    alloy::providers::RootProvider<alloy::network::Ethereum>,
+    alloy::network::Ethereum,
+>;
```
heemankv, line 31: The same type is being used in prover-clients as well, can we create it in utils and use it from there ? 
Also comment in why this is the way it is now would be great!

orchestrator/src/cli/settlement/ethereum.rs
```diff
@@ -31,4 +31,9 @@ pub struct EthereumSettlementCliArgs {
 
     #[arg(env = "MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR", long, default_value = "1.5")]
     pub max_gas_price_mul_factor: f64,
+
+    /// Whether settling on Ethereum mainnet (true) or Sepolia testnet (false).
+    /// Mainnet uses blob proofs (pre-Fusaka), Sepolia uses cell proofs (post-Fusaka).
+    #[arg(env = "MADARA_ORCHESTRATOR_ETHEREUM_IS_MAINNET", long, default_value = "false")]
```
heemankv: This is only till December 3 2025 (Fusaka launch date on Mainnet) right ?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: +1

heemankv (commented):
Looks good, 
I think we can make the code more idiomatic tho, 
specially in eth settlement's conversion.rs and lib.rs

e2e/config/madara.yaml
prkpndy: yeah, forgot to push. done now

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: sorted

orchestrator/src/cli/settlement/ethereum.rs
prkpndy: Updated the flag to make it more specific to the usecase as discusses with @apoorvsadana. Not it's called `disable-peerdas`.
Yes, it's till mainnet also have fusaka. After that we can forget about this flag

bootstrapper/src/helpers/account_actions.rs
prkpndy: done

e2e/src/services/madara/config.rs
prkpndy: done

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: updated the code 🫡

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
prkpndy: sorted

orchestrator/crates/da-clients/ethereum/src/lib.rs
prkpndy: yeah, this is the only way i was able to find. `with_recomended_fillers` does not exist now

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -114,16 +131,61 @@ pub(crate) async fn prepare_sidecar(
...
 
-    Ok((sidecar_blobs, sidecar_commitments, sidecar_proofs))
+    Ok(BlobTransactionSidecarVariant::from(BlobTransactionSidecar::new(
+        sidecar_blobs,
+        sidecar_commitments,
+        sidecar_proofs,
+    )))
+}
+
+// Function to prepare sidecar for EIP 7594 transaction
```
apoorvsadana, line 150: why is this type complexity needed here but not in the other function?

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -114,16 +131,61 @@ pub(crate) async fn prepare_sidecar(
...
+
+        // Compute cells and KZG proofs for this blob
+        // This returns ALL cell proofs (including extension cells)
+        let (_, proofs) = trusted_setup.compute_cells_and_kzg_proofs(&blob)?;
+        // Store all cell proofs for this blob (CELLS_PER_EXT_BLOB proofs, each 48 bytes)
+        let blob_cell_proofs: Vec<FixedBytes<48>> =
+            proofs.into_iter().map(|proof| FixedBytes::new(proof.to_bytes().into_inner())).collect();
+        all_cell_proofs.push(blob_cell_proofs);
+    }
+
```
apoorvsadana, line 177: preparing blobs and commitments can still be one function can called from both places?

apoorvsadana (commented):
`prepare_sidecar` functions looks good now!

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
prkpndy: removed

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
prkpndy: in `prepare_sidecar_with_proof` I still need to recreate commitment. don't see a big advantage here

PR #844, opened 2025-11-11
Configurable RocksDB Write Options

heemankv:
# Configurable RocksDB Write Options (WAL and fsync)

## Pull Request type

- [x] Feature

Currently, Madara uses hardcoded RocksDB write options with WAL (Write-Ahead Log) disabled and no fsync (`no-wal-no-fsync`). This configuration is:
- The fastest option but least safe
- Not configurable by users
- Applied inconsistently (trie.rs had local WriteOptions creation)
- May lose data on process crashes or power failures

Resolves: #NA

## What is the new behavior?

Added configurable RocksDB write durability modes via CLI/environment variable:

**Available Modes:**
- `wal-with-fsync` - Safest, slowest (survives crashes and power failures)
- `wal-no-fsync` - **NEW DEFAULT** - Production recommended (survives crashes, faster)
- `no-wal-with-fsync` - Fast with some durability
- `no-wal-no-fsync` - Fastest, least safe (old behavior, testing only)

**Usage:**
```bash
# CLI
madara --db-write-mode wal-no-fsync

# Environment variable
MADARA_DB_WRITE_MODE=wal-with-fsync madara
```

**Changes:**
- Added `DbWriteMode` enum with 4 durability configurations
- Added `--db-write-mode` CLI parameter (default: `wal-no-fsync`)
- Changed default from `no-wal-no-fsync` to `wal-no-fsync` for better production safety
- Fixed inconsistencies: all write operations now use the same configurable settings
- Renamed `writeopts_no_wal` → `writeopts` throughout codebase for clarity

No - the CLI interface is new. However, the **default write behavior has changed** to be more conservative (`wal-no-fsync` instead of `no-wal-no-fsync`), which improves safety with a minor performance trade-off. Users wanting the old behavior can explicitly set `--db-write-mode no-wal-no-fsync`.

## Other information

- Default `wal-no-fsync` is recommended for production sequencers - provides crash safety while maintaining good performance
- Use `wal-with-fsync` for maximum durability in critical environments
- Use `no-wal-no-fsync` only for devnet/testing where speed matters more than data safety
- All write paths (blocks, state, classes, mempool, trie) now consistently use the configured mode

### DIFF
--- a/madara/crates/client/db/src/rocksdb/mod.rs
+++ b/madara/crates/client/db/src/rocksdb/mod.rs
@@ -52,7 +52,7 @@ pub mod update_global_trie;
 type WriteBatchWithTransaction = rocksdb::WriteBatchWithTransaction<false>;
 type DB = DBWithThreadMode<MultiThreaded>;
 
-pub use options::{RocksDBConfig, StatsLevel};
+pub use options::{DbWriteMode, RocksDBConfig, StatsLevel};
 
 const DB_UPDATES_BATCH_SIZE: usize = 1024;
 
@@ -80,7 +80,7 @@ fn deserialize<T: serde::de::DeserializeOwned>(bytes: impl AsRef<[u8]>) -> Resul
 
 struct RocksDBStorageInner {
     db: DB,
-    writeopts_no_wal: WriteOptions,
+    writeopts: WriteOptions,
     config: RocksDBConfig,
 }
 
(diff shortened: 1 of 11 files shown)


heemankv (commented):
Self review done, lgtm

Mohiiit (commented):
given these are two different configurations, doesn't it make sense to have it separately? 

like `--fsync-enable` and `--wal-enable`

simpler to follow as well?

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -80,7 +80,7 @@ fn deserialize<T: serde::de::DeserializeOwned>(bytes: impl AsRef<[u8]>) -> Resul
 
 struct RocksDBStorageInner {
     db: DB,
-    writeopts_no_wal: WriteOptions,
+    writeopts: WriteOptions,
```
prkpndy, line 83: nit: can rename to `writeopts_with_wal` to make it clearer for user ?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -5,14 +5,87 @@ use std::path::PathBuf;
...
+/// RocksDB write durability mode. Controls WAL (Write-Ahead Log) and fsync behavior.
+///
+/// # Performance & Safety Trade-offs
+///
+/// | Mode | WAL | Fsync | Performance | Safety | Use Case |
+/// |------|-----|-------|-------------|--------|----------|
+/// | `WalWithFsync` | ✅ | ✅ | Slowest | Highest | Production critical data |
+/// | `WalNoFsync` | ✅ | ❌ | Medium | High | Production (safe on crash) |
+/// | `NoWalWithFsync` | ❌ | ✅ | Fast | Medium | Testing, can tolerate data loss |
+/// | `NoWalNoFsync` | ❌ | ❌ | Fastest | Lowest | Devnet, testing |
```
prkpndy: Let's format this?

prkpndy (approved):
lgtm

Mohiiit (approved):

PR #845, opened 2025-11-11
Add `--rpc-unsafe` flag to protect dangerous admin RPC methods

heemankv:
# Add `--rpc-unsafe` flag to protect dangerous admin RPC methods

## Pull Request type

- [x] Feature

The admin RPC methods `revertTo` and `setCustomBlockHeader` are exposed without additional safety guards. The `revertTo` method was hidden behind a "no block production" check, which was not the intended protection mechanism.

Resolves: #NA

## What is the new behavior?

- Added a new `--rpc-unsafe` CLI flag (env: `MADARA_RPC_UNSAFE`) that must be explicitly enabled to use dangerous admin methods
- The flag requires `--rpc-admin` to be enabled first
- Both `revertTo` and `setCustomBlockHeader` now check for this flag before execution
- Removed the block production mode check from `revertTo` - it now works in any mode when the unsafe flag is set
- Clear error messages returned when attempting to use these methods without the flag

Usage:
```bash
madara --rpc-admin --rpc-unsafe
```

Yes - existing users calling `revertTo` or `setCustomBlockHeader` will need to add the `--rpc-unsafe` flag to their node configuration.

## Other information

This change improves security by requiring explicit opt-in for methods that can modify blockchain state. The flag name clearly indicates the dangerous nature of these operations.

### DIFF
--- a/madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
+++ b/madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
@@ -99,7 +99,9 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
         }
 
         self.backend.revert_to(&block_hash).map_err(StarknetRpcApiError::from)?;
-        let fresh_chain_tip = self.backend.db.get_chain_tip().unwrap();
+        let fresh_chain_tip = self.backend.db.get_chain_tip()
+            .context("Failed to get chain tip after revert")
+            .map_err(StarknetRpcApiError::from)?;
         let backend_chain_tip = mc_db::ChainTip::from_storage(fresh_chain_tip);
         self.backend.chain_tip.send_replace(backend_chain_tip);
         Ok(())
--- a/madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
+++ b/madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
@@ -11,7 +11,6 @@ use mp_rpc::v0_9_0::{
 };
 use mp_transactions::{L1HandlerTransactionResult, L1HandlerTransactionWithFee};
 use mp_block::header::CustomHeader;
-use mp_utils::service::MadaraServiceId;
 
 
 #[async_trait]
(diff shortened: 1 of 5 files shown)


heemankv (commented):
Self review done, lgtm

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -90,24 +89,22 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
-                error: "This method is only available in full node mode".to_string().into()
-            })?
+        // Check if unsafe RPC methods are enabled
+        if !self.rpc_unsafe_enabled {
+            return Err(StarknetRpcApiError::ErrUnexpectedError {
+                error: "This method requires the --rpc-unsafe flag to be enabled".to_string().into()
+            }.into());
         }
+
+        self.backend.revert_to(&block_hash).map_err(StarknetRpcApiError::from)?;
```
Mohiiit, line 101: so the code won't check if the block production service is on?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -90,24 +89,22 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+        let fresh_chain_tip = self.backend.db.get_chain_tip().unwrap();
```
Mohiiit: let's handle this unwrap gracefully!

apoorvsadana (approved):

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
heemankv: No, we are having to revert seq as well sometimes, hence this PR

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
heemankv: great catch!

Mohiiit (approved):

PR #849, opened 2025-11-12
fix: restart madara 0x2 and bouncer weight issue 

Mohiiit:
# Fix: Re-execute PreconfirmedBlock on Restart to Generate bouncer_weights and state_diff

## Problem

When Madara restarts, if there's a `PreconfirmedBlock` in the database, `close_pending_block_if_exists()` was closing it directly without generating `bouncer_weights` and `state_diff`. These are crucial for:
- **bouncer_weights**: Metrics tracking resource usage during transaction execution, essential for block finalization
- **state_diff**: The aggregated changes to the Starknet state resulting from transactions in a block

Without these, the block would be closed in an incomplete state, potentially causing issues with state consistency and bouncer tracking.

## Solution

Re-execute all transactions in the `PreconfirmedBlock` to recreate the execution context and obtain the `BlockExecutionSummary` containing `bouncer_weights` and `state_diff` before closing the block.

### 1. Re-execution Logic (`lib.rs`)

- **`reexecute_preconfirmed_block()`**: New method that re-executes all transactions in a preconfirmed block to obtain `BlockExecutionSummary`
  - Creates `BlockExecutionContext` from the saved `PreconfirmedHeader` (preserving exact timestamp, gas prices, etc.)
  - Converts `PreconfirmedExecutedTransaction` back to blockifier format
  - Executes transactions using `TransactionExecutor`
  - Calls `finalize()` to get `BlockExecutionSummary` with `bouncer_weights` and `state_diff`

- **`prepare_preconfirmed_tx_for_reexecution()`**: Helper method to convert `PreconfirmedExecutedTransaction` to blockifier format
  - Handles `charge_fee` flag correctly (uses `!self.no_charge_fee` to match original execution)
  - Fetches `declared_class` from state view if missing (for declare transactions)
  - Sets default `paid_fee_on_l1` for L1 handler transactions if missing

- **`close_pending_block_if_exists()`**: Updated to re-execute transactions before closing
  - Checks if preconfirmed block exists and has transactions
  - If empty, closes directly without re-execution
  - If not empty, calls `reexecute_preconfirmed_block()` to get execution summary
  - Extracts consumed L1 nonces from transactions
  - Saves `bouncer_weights` and `state_diff` before closing

### 2. Code Refactoring (`util.rs`)

- **`create_executor_with_block_n_min_10()`**: Extracted common logic for creating `TransactionExecutor` with `block_n-10` state diff handling
  - Handles Starknet protocol requirement: state diff entry on contract `0x1` for block `n-10`
  - Used by both normal execution (`executor/thread.rs`) and re-execution (`lib.rs`)
  - Reduces code duplication

### 3. Comprehensive Test (`lib.rs`)

- **`test_close_pending_block_reexecution_matches_normal_closing`**: New test that verifies re-execution produces identical results
  - Creates a block normally with transactions (invoke, declare)
  - Captures `global_state_root` and `state_diff`
  - Creates a fresh backend (simulating restart) and executes the same transactions
  - Stops before closing to create a preconfirmed block
  - Calls `close_pending_block_if_exists()` to re-execute and close
  - Verifies that `global_state_root` and `state_diff` match exactly
  - Uses two separate `devnet_setup` fixtures for clean state isolation
  - Handles state diff ordering differences by sorting before comparison

### Block Context Preservation

The re-execution preserves the exact `PreconfirmedHeader` values:
- `block_timestamp`: Preserved exactly (converted via `UNIX_EPOCH + Duration::from_secs()`)
- `gas_prices`: Preserved exactly
- `sequencer_address`: Preserved exactly
- `protocol_version`: Preserved exactly

### Block N-10 Handling

Re-execution correctly handles the Starknet protocol requirement for `block_n-10` state diff entry:
- Checks if block `n-10` exists
- If it exists, sets the state diff entry on contract `0x1` with the block hash
- This ensures protocol compliance during re-execution

### Transaction Type Support

The implementation handles all transaction types:
- ✅ Invoke transactions
- ✅ Declare transactions (with `declared_class` fetching)
- ⚠️ L1 handler transactions (with default `paid_fee_on_l1` - note: `paid_fee_on_l1` is not preserved in `PreconfirmedExecutedTransaction`)

## Testing

The test `test_close_pending_block_reexecution_matches_normal_closing` verifies:
1. Normal block creation and closing produces expected state
2. Re-execution on restart produces identical `global_state_root` and `state_diff`
3. Transactions match exactly between normal execution and re-execution

## Notes

- **L1 Handler Transactions**: Currently excluded from the test because `PreconfirmedExecutedTransaction` doesn't store `paid_fee_on_l1`, which affects execution results. During re-execution, we use a high default value (`u128::MAX / 2`) to ensure validation passes. A proper fix would require storing `paid_fee_on_l1` in `PreconfirmedExecutedTransaction` (schema change).

- **charge_fee Configuration**: The implementation uses `!self.no_charge_fee` from `BlockProductionTask` to determine `charge_fee` during re-execution. This assumes that `TransactionValidator` and `BlockProductionTask` use matching configurations. In `DevnetSetup`, both use `disable_fee: false` / `no_charge_fee: false`, ensuring `charge_fee = true` in both original and re-execution. If these don't match, re-execution will produce different results, which is expected behavior.

- **State Diff Ordering**: The test handles state diff ordering differences by sorting both state diffs before comparison, as the order of entries may differ between executions while values remain the same.


configs/args/config.json
```diff
@@ -107,7 +107,6 @@
   },
   "block_production_params": {
     "block_production_disabled": false,
-    "close_preconfirmed_block_upon_restart": false,
```
apoorvsadana, line 110: hmm, why do we remove this? is it always `true` now?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+    ///   including `charge_fee`, `validate`, and `only_query`
+    fn prepare_preconfirmed_tx_for_reexecution(
+        &self,
+        preconfirmed_tx: &PreconfirmedExecutedTransaction,
+        state_view: &MadaraStateView,
+    ) -> anyhow::Result<blockifier::transaction::transaction_execution::Transaction> {
+        // Convert PreconfirmedExecutedTransaction to ValidatedTransaction
+        // Use the actual charge_fee value from configuration (charge_fee = !no_charge_fee)
+        let mut validated_tx = preconfirmed_tx.to_validated();
+        validated_tx.charge_fee = !self.no_charge_fee;
```
apoorvsadana, line 346: charge_fee should also be stored as a part of the `PreconfirmedExecutedTransaction` right? otherwise restart can cause failures if the flag was changes.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+        let Some(block_n_min_10) = block_n.checked_sub(10) else {
+            return Ok(None);
+        };
+
+        if let Some(view) = backend.block_view_on_confirmed(block_n_min_10) {
+            let block_hash = view.get_block_info().context("Getting block hash of block_n - 10")?.block_hash;
+            Ok(Some((block_n_min_10, block_hash)))
+        } else {
+            // Block doesn't exist yet - this is fine, we'll skip it
+            Ok(None)
```
apoorvsadana: is this fine? seems unreachable to me.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+        // Convert PreconfirmedExecutedTransaction to ValidatedTransaction
+        // Use the actual charge_fee value from configuration (charge_fee = !no_charge_fee)
+        let mut validated_tx = preconfirmed_tx.to_validated();
+        validated_tx.charge_fee = !self.no_charge_fee;
+
+        // If declared_class is missing and transaction is Declare, fetch it from state_view
+        if validated_tx.declared_class.is_none() {
+            if let Some(declare_tx) = validated_tx.transaction.as_declare() {
+                validated_tx.declared_class = Some(
+                    state_view
```
apoorvsadana, line 356: `reexecute_preconfirmed_block` sends the parent view here. if the declare txn is in the preconfirmed block, then fetching the class on the parent view will technically fail right?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+            block_number: header.block_number,
+            sequencer_address: header.sequencer_address,
+            block_timestamp: UNIX_EPOCH + Duration::from_secs(header.block_timestamp.0),
+            protocol_version: header.protocol_version,
+            gas_prices: header.gas_prices.clone(),
+            l1_da_mode: header.l1_da_mode,
+        };
+
+        // Create LayeredStateAdapter
+        let state_adaptor =
```
apoorvsadana:
```suggestion
        let state_adapter =
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+
+        // Execute all transactions
+        let execution_results = executor.execute_txs(&blockifier_txs, /* execution_deadline */ None);
+
+        // Check for execution errors (though we don't need to process results, we should check for panics)
+        for result in &execution_results {
+            if let Err(err) = result {
+                tracing::warn!("Transaction execution error during re-execution: {err:?}");
+            }
+        }
```
apoorvsadana: do we want to assert receipts to the receipts stored?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+    ///
+    /// - The re-execution uses the exact header values from the preconfirmed block (timestamp, gas_prices, etc.)
+    /// - The `charge_fee` flag is determined by `self.no_charge_fee` configuration
+    /// - L1 handler transactions preserve their `paid_fee_on_l1` value (stored during `append_batch`)
+    ///
+    /// # TODO (mohit, 13/11/2025)
+    ///
+    /// - Handle cases where Starknet version has been updated between shutdown and restart
+    /// - Handle cases where version constants or bouncer weights have changed
+    async fn close_pending_block_if_exists(&mut self) -> anyhow::Result<()> {
```
apoorvsadana:
```suggestion
    async fn close_preconfirmed_block_if_exists(&mut self) -> anyhow::Result<()> {
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+
+            // Convert state_diff and close block
+            let state_diff: mp_state_update::StateDiff = block_exec_summary.state_diff.into();
+            backend
+                .write_access()
+                .close_preconfirmed(/* pre_v0_13_2_hash_override */ true, Some(state_diff))
+                .context("Closing preconfirmed block on startup")?;
+
+            anyhow::Ok(())
+        })
```
apoorvsadana: let's make the code inside `EndBlock` in `process_reply` a function and then call that function from here?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -293,40 +304,243 @@ impl BlockProductionTask {
...
+    ///    - `bouncer_weights`: Required for block finalization
+    ///    - `state_diff`: Required for block closing
+    /// 3. Extracts consumed L1 handler nonces from transactions
+    /// 4. Removes consumed L1 to L2 message nonces from the database
+    /// 5. Saves bouncer weights to the database
+    /// 6. Closes the preconfirmed block with the regenerated state_diff
+    ///
+    /// # Why Re-execution is Necessary
+    ///
+    /// When Madara shuts down during block production, the preconfirmed block is saved with executed transactions
```
apoorvsadana, line 562: I am thinking, shouldn't we handle the case during graceful shutdowns that the block gets closed automatically? This saves time on re execution and avoids this edge case until something unexpected, like an OOM, happens.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -563,8 +772,7 @@ pub(crate) mod tests {
                 self.mempool.clone(),
                 self.metrics.clone(),
                 Arc::new(self.l1_client.clone()),
-                true,
-                true, // close_preconfirmed_block_upon_restart - default to true for tests
+                false, /* no_charge_fee = false */
```
apoorvsadana, line 847: if tests are passing, this is prob not an issue but `no_charge_fee` was true here before.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -873,94 +1081,310 @@ pub(crate) mod tests {
...
     }
 
-    /// This test makes sure that if a pending block is already present in db
-    /// at startup, then it is closed and stored in db.
-    ///
-    /// This happens if a full node is shutdown (gracefully or not) midway
-    /// during block production.
+    //
+    // This test verifies that when Madara restarts with a preconfirmed block, the re-execution process
+    // Test that `close_pending_block_if_exists` correctly re-executes transactions and produces matching results.
```
apoorvsadana:
```suggestion
```

madara/crates/client/block_production/src/lib.rs
```diff
@@ -873,94 +1081,310 @@ pub(crate) mod tests {
...
+        )
+        .await;
+
+        // 5. Another invoke transaction
+        sign_and_add_invoke_tx(
+            &restart_devnet_setup.contracts.0[1],
+            &restart_devnet_setup.contracts.0[3],
+            &restart_devnet_setup.backend,
+            &restart_devnet_setup.tx_validator,
+            Felt::ZERO, // Different account, so nonce starts at ZERO
```
apoorvsadana: this part can be a functon inside the test itself

madara/crates/client/block_production/src/lib.rs
```diff
@@ -873,94 +1081,310 @@ pub(crate) mod tests {
...
+        let reexecuted_block_info =
+            restart_devnet_setup.backend.block_view_on_confirmed(block_number).unwrap().get_block_info().unwrap();
+
+        // Verify the header fields match the pre-execution pre-confirmed block's header
+        assert_eq!(restart_preconfirmed_block.header.block_timestamp, reexecuted_block_info.header.block_timestamp);
+        assert_eq!(restart_preconfirmed_block.header.protocol_version, reexecuted_block_info.header.protocol_version);
+        assert_eq!(restart_preconfirmed_block.header.l1_da_mode, reexecuted_block_info.header.l1_da_mode);
+        assert_eq!(restart_preconfirmed_block.header.gas_prices, reexecuted_block_info.header.gas_prices);
+        assert_eq!(restart_preconfirmed_block.header.sequencer_address, reexecuted_block_info.header.sequencer_address);
+        assert_eq!(restart_preconfirmed_block.header.block_number, reexecuted_block_info.header.block_number);
```
apoorvsadana, line 1372: should we match block hash too?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -873,94 +1081,310 @@ pub(crate) mod tests {
...
+    #[timeout(Duration::from_secs(100))]
     #[tokio::test]
-    #[allow(clippy::too_many_arguments)]
-    async fn block_prod_pending_close_on_startup_pass(
-        #[future] devnet_setup: DevnetSetup,
-        #[with(Felt::ONE)] tx_invoke_v0: TxFixtureInfo,
-        #[from(converted_class_legacy)]
-        #[with(Felt::ZERO)]
-        converted_class_legacy_0: mp_class::ConvertedClass,
+    async fn test_close_pending_block_reexecution_matches_normal_closing(
```
apoorvsadana: nice test!

madara/crates/client/block_production/src/lib.rs
Mohiiit: block hash won't be there for pre-confirmed block right?

madara/crates/client/block_production/src/lib.rs
Mohiiit: the issue was how the tests are written, there is a function called `to_validated` and it uses disable_fee param to add charge_fee as true or false, it was using the default values, and hence that was causing the issue, so I just made sure that our test is using the same value as of the default values 

but I can also ensure that the disable_fee uses the value we give via the config (might be a slightly bigger change)

madara/crates/client/block_production/src/lib.rs
Mohiiit: and keep this as backup?

madara/crates/client/block_production/src/lib.rs
Mohiiit: yes there are a bunch of params apart from charge_fee, that can change the result, I was thinking should we handle that case for once and all? 

or do you think handling charge_fee for now will do?

configs/args/config.json
Mohiiit: yup, this would be always true, in case of replay, the node won't save the preconfirmed block

Mohiiit: issue with continuing is same, the executor won't have the context for 0x2 and bouncer weights

madara/crates/client/block_production/src/lib.rs
apoorvsadana: will move to next PR

madara/crates/client/block_production/src/lib.rs
apoorvsadana: let's make it `unreachable!` here

madara/crates/client/block_production/src/lib.rs
apoorvsadana: we should panic here

madara/crates/client/block_production/src/lib.rs
apoorvsadana: discussed on call, will move preconfirmed block fetch before drop

madara/crates/client/db/src/preconfirmed.rs
```diff
@@ -21,7 +25,7 @@ impl PreconfirmedExecutedTransaction {
     pub fn to_validated(&self) -> ValidatedTransaction {
         ValidatedTransaction {
             transaction: self.transaction.transaction.clone(),
-            paid_fee_on_l1: None,
+            paid_fee_on_l1: self.paid_fee_on_l1,
```
prkpndy, line 28: curious, how it used to work till now? was the logic broken or we expected this?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -532,6 +532,7 @@ pub fn gateway_preconfirmed_block_sync(
                         state_diff,
                         declared_class: None, // It seems we can't get the declared classes from the preconfirmed block :/
                         arrived_at,
+                        paid_fee_on_l1: None, // Gateway blocks don't contain paid_fee_on_l1, set to None
```
prkpndy: > Gateway blocks don't contain paid_fee_on_l1

why?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -272,7 +283,7 @@ impl BlockProductionTask {
             executor_commands_recv: Some(recv),
             l1_client,
             bypass_tx_input: Some(bypass_tx_input),
-            close_preconfirmed_block_upon_restart,
+            no_charge_fee,
```
prkpndy, line 293: to confirm, this is to ensure that we re-execute the txn with same `no_charge_fee` flag, even tho it might have been changed in the madara flag?

madara/crates/client/block_production/src/lib.rs
prkpndy: +1

prkpndy (approved):
lgtm

madara/crates/client/block_production/src/lib.rs
Mohiiit: it might change for now, there would be another PR that ensures that on re-execution we are using the same configs

apoorvsadana (approved):

prkpndy (approved):

PR #852, opened 2025-11-12
feat: add support for versioned constants in Orchestrator

byteZorvin:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

- Orchestrator does not have support for custom version constants

Resolves: #NA

## What is the new behavior?
- Orchestrator has custom version constant support

-
-
-


Mohiiit (changes requested):
instead of taking the path and then converting the yaml/json file to the versionconstants, we can do something like this:

```rust
fn parse_constants(path: &str) -> Result<VersionedConstants, String> {
    std::fs::read_to_string(path)
        .map_err(|e| e.to_string())
        .and_then(|content| {
            serde_json::from_str::<VersionedConstants>(&content)
                .map_err(|e| e.to_string())
        })
}

#[derive(Parser, Debug)]
pub struct Args {
    #[arg(
        env = "MADARA_ORCHESTRATOR_VERSIONED_CONSTANTS_PATH",
        long,
        value_parser = parse_constants
    )]
    pub versioned_constants: Option<VersionedConstants>,
}
```

this way we won't have to add `versioned_constants` in the `ConfigParam` and it would be part of the `snos_config` only and can be used directly then

orchestrator/src/core/config.rs
```diff
@@ -120,6 +121,8 @@ pub struct ConfigParam {
     /// * Aggregator Proof
     pub store_audit_artifacts: bool,
     pub bouncer_weights_limit: BouncerWeights,
+    /// Optional versioned constants loaded from file. If None, defaults from blockifier will be used.
+    pub versioned_constants: Option<VersionedConstants>,
```
Mohiiit: do we still need this in the `ConfigParams` tho? I think having `versioned_constants` within the `snos_config` is enough?

Mohiiit (changes requested):

orchestrator/src/core/config.rs
byteZorvin: Makes sense, removed !!

orchestrator/src/worker/event_handler/jobs/snos.rs
```diff
@@ -75,8 +75,8 @@ trait OsHintsConfigurationFromLayer {
 impl OsHintsConfigurationFromLayer for OsHintsConfiguration {
     fn with_layer(layer: Layer) -> OsHintsConfiguration {
         match layer {
-            Layer::L2 => OsHintsConfiguration { debug_mode: true, full_output: true, use_kzg_da: false },
-            Layer::L3 => OsHintsConfiguration { debug_mode: true, full_output: false, use_kzg_da: false },
+            Layer::L2 => OsHintsConfiguration { debug_mode: true, full_output: true, use_kzg_da: true },
```
Mohiiit: should it be true for L2 as well 👀

byteZorvin: i think its already true?

byteZorvin: updated the flag

Mohiiit (approved):

apoorvsadana (approved):

PR #857, opened 2025-11-15
Fix: Reset block production executor state after chain reorg

heemankv:
# Fix: Reset block production executor state after chain reorg

Pull Request Type
- [x] Bugfix

## Problem Statement

When the `revertTo` RPC method was called while Madara was actively in block production state, the system encountered errors. The root cause was that the block production executor maintained stale state after the database was reverted:

- The executor's state adapter still pointed to the pre-revert chain state
- Pending transactions in the execution queue belonged to blocks that had just been reverted
- Block state flags (gas consumed, empty block status) were inconsistent with the new chain tip
- Attempting to continue block production with this stale state caused execution errors and inconsistencies

## Solution

Implemented a state reset mechanism for the block production executor that triggers after chain reorganizations:

### Core Changes

1. **Added Reset Command**: Introduced a new `ResetState` executor command that:
   - Reinitializes the state adapter from the current database state
   - Clears all pending transactions from the execution queue
   - Resets block state tracking flags (gas consumed, block empty status)
   - Transitions the executor to a clean `NewBlock` state
   - Uses `continue` to naturally re-enter the wait loop for new transactions

2. **RPC Integration**: The `revertTo` RPC method now calls `reset_state()` on the block production handle after reverting the chain, ensuring the executor state stays synchronized with the database state.

3. **State Synchronization**: Added `StateReset` message to notify the block production task when the executor has been reset, allowing proper state machine transitions.

### Additional Improvements

- **Refactored custom header access**: Moved `get_custom_header()` methods to `impl<D: MadaraStorageRead>` block to make them available for all storage trait bounds, eliminating manual mutex lock patterns throughout the codebase
- **Cleaned up logging**: Removed verbose debug tracing statements while retaining essential operational logs
- **Code simplification**: Eliminated redundant intermediate variables and manual lock handling

No

## Other Information

The fix ensures that block production can safely resume after any chain reorganization triggered by the `revertTo` RPC, with all executor state properly synchronized to the new chain tip.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -14,3 +14,8 @@ orchestrator/build/
 orchestrator/node_modules/
 lib/
 s3/
+
+*.rs
+*.toml
+Cargo.toml
+Cargo.lock
--- a/madara/crates/client/block_production/src/executor/thread.rs
+++ b/madara/crates/client/block_production/src/executor/thread.rs
@@ -283,45 +283,6 @@ impl ExecutorThread {
                             let _ = callback.send(Ok(()));
                             Default::default()
                         }
-                        super::ExecutorCommand::ResetState(callback) => {
-                            tracing::info!("🔄 Resetting executor state after reorg");
-                            // Reinitialize the state adapter from the current database state
-                            let new_state_adaptor = match mc_exec::LayeredStateAdapter::new(self.backend.clone()) {
-                                Ok(adaptor) => adaptor,
-                                Err(e) => {
-                                    tracing::error!("Failed to reinitialize state adapter after reorg: {:#}", e);
-                                    let _ = callback.send(Err(super::ExecutorCommandError::ChannelClosed));
-                                    return Err(e.into());
-                                }
-                            };
-
-                            let latest_block_n = new_state_adaptor.previous_block_n();
-                            tracing::info!("✅ State adapter reinitialized to block_n={}", new_state_adaptor.block_n());
-
-                            // Reset the executor state to NewBlock with the fresh state adapter
-                            state = ExecutorThreadState::NewBlock(ExecutorStateNewBlock {
-                                state_adaptor: new_state_adaptor,
-                                consumed_l1_to_l2_nonces: HashSet::new(),
-                            });
-
-                            // Clear any pending transactions by clearing both fields
-                            to_exec.txs.clear();
-                            to_exec.additional_info.clear();
-
-                            // Reset block state flags
-                            block_empty = true;
-                            force_close = false;
-                            l2_gas_consumed_block = 0;
-
-                            // Notify the main task that state was reset
-                            if self.replies_sender.blocking_send(super::ExecutorMessage::StateReset { latest_block_n }).is_err() {
-                                tracing::error!("Failed to send StateReset message to block production task");
-                            }
-
-                            let _ = callback.send(Ok(()));
-                            // Continue the loop to wait for new transactions - don't return a batch
-                            continue;
-                        }
                     },
                     // Channel closed. Exit gracefully.
                     // Before exiting, check if we have an executing block that needs to be closed.
--- a/madara/crates/client/block_production/src/executor/thread.rs
+++ b/madara/crates/client/block_production/src/executor/thread.rs
@@ -283,6 +283,45 @@ impl ExecutorThread {
                             let _ = callback.send(Ok(()));
                             Default::default()
                         }
+                        super::ExecutorCommand::ResetState(callback) => {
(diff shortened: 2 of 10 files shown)


madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -281,6 +281,45 @@ impl ExecutorThread {
...
+                                }
+                            };
+
+                            let latest_block_n = new_state_adaptor.previous_block_n();
+                            tracing::info!("✅ State adapter reinitialized to block_n={}", new_state_adaptor.block_n());
+
+                            // Reset the executor state to NewBlock with the fresh state adapter
+                            state = ExecutorThreadState::NewBlock(ExecutorStateNewBlock {
+                                state_adaptor: new_state_adaptor,
+                                consumed_l1_to_l2_nonces: HashSet::new(),
```
Mohiiit: what would happen to the nonces that were in the hasmap earlier? will those nonces get saved in the db?

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -281,6 +281,45 @@ impl ExecutorThread {
...
+                                Ok(adaptor) => adaptor,
+                                Err(e) => {
+                                    tracing::error!("Failed to reinitialize state adapter after reorg: {:#}", e);
+                                    let _ = callback.send(Err(super::ExecutorCommandError::ChannelClosed));
+                                    return Err(e.into());
+                                }
+                            };
+
+                            let latest_block_n = new_state_adaptor.previous_block_n();
+                            tracing::info!("✅ State adapter reinitialized to block_n={}", new_state_adaptor.block_n());
```
Mohiiit: latest_block_n and block_n() are different blocks?

.prettierignore
```diff
@@ -14,3 +14,8 @@ orchestrator/build/
 orchestrator/node_modules/
 lib/
 s3/
+
+*.rs
+*.toml
+Cargo.toml
+Cargo.lock
```
Mohiiit, line 21: so prettier is for .js, .md files only?

Mohiiit (commented):
what would happen to the l1->l2 deposits? will it get reverted in the bridge as well? 

what about withdrawals? (Ik we will have a window of some time before that could happen)

is this for replay-service only? 

I think revertTO in sequencer mode should specify the assumption it's making

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -107,6 +107,21 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+              tracing::debug!("revertTo: resetting block production state");
+              block_prod_handle
+                  .reset_state()
+                  .await
+                  .context("Resetting block production executor state after reorg")
+                  .map_err(StarknetRpcApiError::from)?;
+              tracing::debug!("revertTo: block production state reset complete");
+          }
+        }
+
```
apoorvsadana: this is business logic, should not live in the rpc crate

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -107,6 +107,21 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
             .map_err(StarknetRpcApiError::from)?;
         let backend_chain_tip = mc_db::ChainTip::from_storage(fresh_chain_tip);
         self.backend.chain_tip.send_replace(backend_chain_tip);
+
+        // Reset block production executor state if block production is enabled
+        let bp_status = self.ctx.service_status(MadaraServiceId::BlockProduction);
+        if matches!(bp_status, mp_utils::service::MadaraServiceStatus::On) {
```
apoorvsadana: why do we need to check if service is on?

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -281,6 +281,45 @@ impl ExecutorThread {
                             let _ = callback.send(Ok(()));
                             Default::default()
                         }
+                        super::ExecutorCommand::ResetState(callback) => {
+                            tracing::info!("🔄 Resetting executor state after reorg");
+                            // Reinitialize the state adapter from the current database state
+                            let new_state_adaptor = match mc_exec::LayeredStateAdapter::new(self.backend.clone()) {
```
apoorvsadana: we can just do `state = initial_state()`, send the message `StateReset` and return from callback. all resetting of values isn't needed imo.

PR #859, opened 2025-11-19
Feat/close preconfirmed on graceful shutdown

Mohiiit:
# Graceful Shutdown with Preconfirmed Block Closing

## Summary

Implements graceful shutdown for the block production service that properly closes any open preconfirmed block during shutdown, using the executor's existing state without re-execution.

## Problem

Previously, when graceful shutdown was triggered, the block production service would exit immediately, leaving any open preconfirmed block unclosed. This required re-execution of transactions on restart, which is inefficient and can cause inconsistencies.

## Solution

The implementation adds a graceful shutdown flow that:

1. **Detects cancellation** before the batcher completes
2. **Sends `CloseBlock` command** to the executor thread (which is still running)
3. **Waits for `EndBlock` message** from the executor with the block's execution summary
4. **Closes the block** using the executor's existing state (no re-execution needed)
5. **Handles edge cases** with timeout protection and graceful degradation

### Shutdown Flow

The shutdown process follows 5 concurrent paths in a `tokio::select!` loop:

1. **Cancellation Detection**: Detects shutdown request and attempts to close block
2. **Batcher Completion**: Handles batcher task completion and fallback block closing
3. **EndBlock Processing**: Receives and processes `EndBlock` message from executor
4. **Timeout Protection**: Prevents infinite waiting with 5-second timeout
5. **Executor Shutdown**: Handles executor thread completion

### State Management

- `ShutdownState`: Encapsulates shutdown state (`shutting_down`, `batcher_completed`, `end_block_deadline`)
- `try_close_block_on_shutdown()`: Attempts to close block and returns `ControlFlow` for loop control

## Note

If graceful shutdown fails for any reason, the block remains preconfirmed and is handled on restart via the existing `close_preconfirmed_block_if_exists()` mechanism.


madara/crates/client/block_production/src/lib.rs
```diff
@@ -1631,4 +1893,74 @@ pub(crate) mod tests {
...
+
+    // This test verifies that graceful shutdown properly closes any open preconfirmed block
+    // without requiring re-execution. When shutdown is triggered, the block production service
+    // should close the preconfirmed block using the executor's existing state.
+    #[rstest::rstest]
+    #[timeout(Duration::from_secs(30))]
+    #[tokio::test]
+    async fn test_graceful_shutdown_closes_preconfirmed_block(
+        #[future]
+        #[with(Duration::from_secs(3000000000), false)]
```
prkpndy: this seems to be too big? should we not reduce it?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1631,4 +1893,74 @@ pub(crate) mod tests {
...
+        assert_eq!(preconfirmed_view.num_executed_transactions(), 1);
+
+        // Step 3: Trigger graceful shutdown by cancelling ServiceContext
+        ctx_clone.cancel_global();
+
+        // Step 4: Wait for shutdown to complete
+        task.await.unwrap();
+
+        // Give a small delay to ensure database writes complete
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
prkpndy, line 1822: why is this needed tho? aren't all service (including DB) is closed once we shutdown?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +841,207 @@ impl BlockProductionTask {
...
         );
 
-        // Graceful shutdown: when the service is asked to stop, the `batcher_task` will stop,
-        //  which will close the `send_batch` channel (by dropping it). The executor thread then will see that the channel
-        //  is closed next time it tries to receive from it. The executor thread shuts down, dropping the `executor.stop` channel,
-        //  therefore closing it as well.
+        // Graceful shutdown flow:
+        //
+        // ┌─────────────────────────────────────────────────────────────────┐
+        // │                    Graceful Shutdown Flow                        │
```
prkpndy: can format this

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1631,4 +1893,74 @@ pub(crate) mod tests {
...
         );
     }
+
+    // This test verifies that graceful shutdown properly closes any open preconfirmed block
+    // without requiring re-execution. When shutdown is triggered, the block production service
+    // should close the preconfirmed block using the executor's existing state.
+    #[rstest::rstest]
+    #[timeout(Duration::from_secs(30))]
+    #[tokio::test]
+    async fn test_graceful_shutdown_closes_preconfirmed_block(
```
prkpndy, line 1724: Is it possible to add some more test cases for edge cases?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +841,207 @@ impl BlockProductionTask {
...
+                // This prevents infinite waiting if:
+                //   - Executor crashes before sending EndBlock
+                //   - EndBlock gets lost in transit
+                //   - Any other unexpected failure
+                _ = async {
+                    if let Some(deadline) = shutdown_state.end_block_deadline.as_ref() {
+                        tokio::time::sleep_until(*deadline).await
+                    } else {
+                        std::future::pending().await
+                    }
```
prkpndy: I think we can remove the if-else statement and directly sleep until deadline

prkpndy (approved):

madara/crates/client/block_production/src/lib.rs
```diff
@@ -622,6 +661,53 @@ impl BlockProductionTask {
...
+                    // Successfully sent CloseBlock command to executor
+                    // The executor will process this and send EndBlock message back
+                    tracing::debug!("CloseBlock command sent successfully, waiting for EndBlock");
+                    shutdown_state.set_waiting_for_end_block();
+                    Ok(ControlFlow::Continue(()))
+                }
+                Err(e) => {
+                    // Executor channel is closed - executor has already shut down
+                    // This can happen if executor detected channel closure before we sent CloseBlock
+                    // In this case, the block will remain preconfirmed and be handled on restart
```
apoorvsadana: is this possible? if yes, we should fix it as a part of the PR. we shouldn't have a known race condition. if not, this comment probably needs to change.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +841,207 @@ impl BlockProductionTask {
...
+        //    └─> Executor detects send_batch channel closure → exits gracefully
+        //        └─> Signals via executor.stop channel
+        //
+        // Key Safety Features:
+        // - Timeout protection: Won't wait indefinitely for EndBlock
+        // - State validation: Only sends CloseBlock when block exists
+        // - Graceful degradation: If executor already shut down, block handled on restart
+        // - No re-execution: Uses executor's existing state (no transaction re-execution needed)
+        //
+        // Note: The executor thread shuts down, dropping the `executor.stop` channel, therefore closing it as well.
```
apoorvsadana: i think these comments are more detailed than they need to be. i am worried they will get outdated very easily because someone will update the executor thread and forget to change this comment because its in a completely different place. maybe this detail makes sense in the PR desc

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +841,207 @@ impl BlockProductionTask {
...
+                // ============================================================
+                // Safety mechanism: if we're waiting for EndBlock but it doesn't arrive within
+                // the timeout period, we exit gracefully. The block will remain preconfirmed
+                // and be handled on restart (via close_preconfirmed_block_if_exists).
+                //
+                // This prevents infinite waiting if:
+                //   - Executor crashes before sending EndBlock
+                //   - EndBlock gets lost in transit
+                //   - Any other unexpected failure
+                _ = async {
```
apoorvsadana: doesn't the services layer already have a deadline after which it forcefully kills the service?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +841,207 @@ impl BlockProductionTask {
...
+                // signals the executor to shut down.
+                //
+                // Flow:
+                //   1. Batcher completes → closes send_batch channel
+                //   2. If we already sent CloseBlock → continue waiting for EndBlock
+                //   3. If we haven't sent CloseBlock yet → try to close block now
+                //   4. If CloseBlock succeeds → wait for EndBlock
+                //   5. If CloseBlock fails or no block → shutdown complete
+                res = &mut batcher_task, if !shutdown_state.batcher_completed => {
+                    res.context("In batcher task")?;
```
apoorvsadana: if the batcher task closes because of an error, then `?` will technically return before we gracefully shutdown?

apoorvsadana (commented):
As discussed, let's try an approach where

1. we pass cancellation ctx to executor
2. block production service waits for batcher, executor and for block to be closed

madara/crates/client/block_production/src/lib.rs
apoorvsadana: comment still applies?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +755,132 @@ impl BlockProductionTask {
...
+                // During graceful shutdown, the executor automatically sends EndBlock when
+                // it detects the send_batch channel closure (if a block exists).
                 Some(reply) = executor.replies.recv() => {
+                    let is_end_block = matches!(reply, ExecutorMessage::EndBlock(_));
+
+                    // Process the reply (this will close the block if it's EndBlock)
                     self.process_reply(reply).await.context("Processing reply from executor thread")?;
+
+                    // If we're shutting down and just processed EndBlock, shutdown is complete
+                    if shutting_down && is_end_block {
```
apoorvsadana: it's possible batcher hasn't closed yet right? image a scenario where batcher didn't detect signal yet, `shutting_down` has been made true and executor closed block because block time was reached.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +755,132 @@ impl BlockProductionTask {
...
+                        // If there was a block, executor should have sent EndBlock before exiting
+                        // If we're here, either no block existed or executor panicked
+                        // In case of panic, it will propagate naturally
+                        tracing::debug!("Executor shut down during graceful shutdown");
+                        return Ok(());
+                    }
+                    // Normal executor completion (not during shutdown)
+                    // If executor panicked, recv() will resume the panic (handled by StopErrorReceiver)
+                    res.context("In executor thread")?;
+                    return Ok(());
```
apoorvsadana: can't we just check if `res` is an error, if yes we add context, otherwise we don't? or maybe we add context all the time. unsure what this if condition is solving here.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1631,4 +1732,269 @@ pub(crate) mod tests {
...
+        // No preconfirmed block should exist (still)
+        assert!(!devnet_setup.backend.has_preconfirmed_block());
+    }
+
+    // This test verifies that graceful shutdown properly closes a preconfirmed block
+    // containing multiple transactions, ensuring all transactions are preserved correctly.
+    #[rstest::rstest]
+    #[timeout(Duration::from_secs(30))]
+    #[tokio::test]
+    async fn test_graceful_shutdown_closes_preconfirmed_block_with_multiple_transactions(
```
apoorvsadana: let's combine this test and the 1 transaction test?
i think having multiple cases like these make more sene when we've smaller unit tests, not in these ones (or the ones we had in cairo native)

madara/crates/client/block_production/src/lib.rs
```diff
@@ -1631,4 +1732,269 @@ pub(crate) mod tests {
...
+    // the send_batch channel closure (WaitTxBatchOutcome::Exit). This test ensures shutdown
+    // completes gracefully regardless of timing.
+    //
+    // The scenario: Cancellation → batcher completes → closes send_batch → executor detects Exit
+    // → executor closes block automatically → sends EndBlock → main loop closes block → shutdown complete
+    #[rstest::rstest]
+    #[timeout(Duration::from_secs(30))]
+    #[tokio::test]
+    async fn test_graceful_shutdown_closeblock_fails_executor_channel_closed(
+        #[future]
```
apoorvsadana: what's the edge case here? not sure how is this different from the previous test case?

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -283,7 +283,61 @@ impl ExecutorThread {
...
+                                            .is_err()
+                                        {
+                                            // Receiver closed - main loop already shut down
+                                            // Block will remain preconfirmed and be handled on restart
+                                            tracing::warn!(
+                                                "Could not send EndFinalBlock during shutdown, block will remain preconfirmed"
+                                            );
+                                        }
+                                    }
+                                    Err(e) => {
```
apoorvsadana, line 315: still send the `EndFinalBlock` command?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +807,63 @@ impl BlockProductionTask {
...
+                        Ok(()) => tracing::debug!("Batcher task completed normally"),
+                        Err(e) => {
+                            let error = e.context("In batcher task");
+                            tracing::warn!("Batcher task errored: {error:?}");
+                            if self.backend.has_preconfirmed_block() {
+                                batcher_error = Some(error);
+                                tracing::warn!("Batcher errored with preconfirmed block, attempting graceful shutdown");
+                            } else {
+                                batcher_error = Some(error);
+                            }
```
apoorvsadana, line 829:
```suggestion
                            batcher_error = Some(error);
                            if self.backend.has_preconfirmed_block() {
                                tracing::warn!("Batcher errored with preconfirmed block, attempting graceful shutdown");
                            }
```

apoorvsadana (approved):

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +807,61 @@ impl BlockProductionTask {
...
+                        end_final_block_received = true;
+                        tracing::debug!("EndFinalBlock processed, executor completed");
+                    }
                 }
 
-                // Bubble up errors from the executor thread, or graceful shutdown.
-                // We do this after processing all the replies to ensure we don't lose some of the state by accident.
-                res = executor.stop.recv() => return res.context("In executor thread"),
+                // Path 3: Executor thread stopped (normal completion or panic)
+                // This fires when executor exits. EndFinalBlock should have already been processed
```
byteZorvin: I guess `EndFinalBlock` is not fired when the Executor panics but only in normal exits, the comment confused me a bit.

byteZorvin: Also I dont think the EndFinalBlock is guranteed to be already processed in this branch execution. High chances this could be processed in the next loop iteration.

madara/crates/client/block_production/src/lib.rs
```diff
@@ -753,33 +807,61 @@ impl BlockProductionTask {
...
+                // (executor always sends EndFinalBlock during shutdown - Some(summary) if block exists, None if no block).
+                res = executor.stop.recv() => {
+                    res.context("In executor thread")?;
```
byteZorvin, line 851: We can mention that we are explicity bubbling up the error incase of panic, since during no EndOfFinalBlock can be fired. And the last resort remains reexecution of transactions

byteZorvin: Ohkay, saw the comment on top of file, nvm

byteZorvin (commented):
GG work, really creative on the new message EndFinalBlock

byteZorvin (approved):

madara/crates/client/block_production/src/lib.rs
```diff
@@ -58,17 +58,54 @@
...
+//! ### Executor Panic
+//!
+//! If the executor thread panics:
+//! - The panic is caught and propagated via the `stop` channel
+//! - The main loop resumes the panic, causing the block to remain preconfirmed
+//! - The preconfirmed block will be handled on restart
+//!
+//! The loop exits when:
+//! - Both batcher and executor have completed → returns `Ok(())` or the saved batcher error
+//! - Batcher errored without a preconfirmed block → returns the error immediately
```
byteZorvin: I could not match the logic with the comment...`Batcher errored without a preconfirmed block → returns the error immediately`

Mohiiit: fair, it's changed now, updating the doc!

madara/crates/client/block_production/src/lib.rs
Mohiiit: we will get the one-shot stop signal only after the executor has sent EndFinalBlock message

byteZorvin (approved):
GG !!

PR #860, opened 2025-11-20
feat: save configs for re-execution on re-starts

Mohiiit:
# Save Runtime Execution Config for Restarts

## Summary

Implements persistence of runtime execution configuration to ensure pre-confirmed blocks are re-executed with the same config values used during original execution, even after node restarts with different configuration.

## Problem

When a node restarts with a pre-confirmed block, re-execution must use the exact same configuration (ChainConfig, VersionedConstants, and `no_charge_fee`) that was used during original execution. Without this, transaction receipts and state diffs would differ between original execution and re-execution, causing validation failures.

## Solution

1. **New Module**: Created `runtime_exec_config.rs` in `primitives/chain_config` to handle serialization of runtime execution config with version-aware ChainConfig handling

2. **Database Persistence**: Added database operations to save/load `RuntimeExecutionConfig` on restart

3. **Re-execution Logic**: Modified `close_preconfirmed_block_if_exists` to:
   - Load saved config from database
   - Use saved `ChainConfig`, `VersionedConstants`, and `no_charge_fee` for re-execution
   - Update config with current values after re-execution completes

4. **Version Handling**: Implemented round-trip YAML serialization that automatically handles ChainConfig version changes

## Key Changes

- **New file**: `madara/crates/primitives/chain_config/src/runtime_exec_config.rs` - RuntimeExecutionConfig serialization
- **Modified**: `block_production/src/lib.rs` - Save/load config on restart, use saved values for re-execution
- **Modified**: `db/src/rocksdb/meta.rs` - Database operations for runtime exec config
- **Modified**: `db/src/storage.rs`, `db/src/lib.rs`, `db/src/rocksdb/mod.rs` - Storage trait and implementations

## Testing

- Added `test_reexecution_uses_saved_no_charge_fee_value` to verify saved config is used for re-execution
- Added serialization tests using mainnet.yaml config

## Breaking Changes

None - backward compatible.


madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -38,7 +38,7 @@ mod events_bloom_filter;
 mod iter_pinned;
 mod l1_to_l2_messages;
 mod mempool;
-mod meta;
+pub mod meta;
```
apoorvsadana: why does this specifically need to be public? seems like a leak in business logic.

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+//!
+//! ## What Gets Saved
+//!
+//! - `chain_config` fields (via YAML serialization)
+//! - `protocol_version` (string, e.g., "0.13.2")
+//! - `no_charge_fee` flag
+//!
+//! ## What Does NOT Get Saved
+//!
+//! - `versioned_constants` - Reconstructed from current node's built-in constants using saved `protocol_version`
```
apoorvsadana: this comment is wrong? we're storing the `versioned_constants` right (and these might not be the built in ones because its possible the previous run gave some other constants in the chain config)

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+//! - `private_key` - Set to `None` (re-execution doesn't need it; signing uses current backend's key)
+//!
+//! ## Assumption
+//!
+//! For a given `protocol_version`, the versioned constants (gas costs, limits) are **immutable**
```
apoorvsadana: this comment too might be inaccurate if we assume the user can give custom constants across runs.

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub no_charge_fee: bool,
+}
+
+/// Serializable form for database storage.
+#[derive(Serialize, Deserialize)]
+pub struct RuntimeExecutionConfigSerializable {
+    /// Config version for validation
+    config_version: u32,
+    /// YAML-serialized chain config (without private_key, without versioned_constants)
+    chain_config_json: String,
```
apoorvsadana:
```suggestion
    chain_config_yaml: String,
```

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+        Ok(Self { chain_config: config_for_saving, exec_constants, no_charge_fee })
+    }
+
+    /// Prepares a chain config for saving by creating a clean copy.
+    ///
+    /// This uses YAML round-trip serialization to:
+    /// 1. Exclude private_key (has #[serde(skip)])
+    /// 2. Exclude versioned_constants (has #[serde(skip)])
+    /// 3. Handle version migrations automatically
+    fn prepare_chain_config_for_saving(config: &ChainConfig) -> Result<ChainConfig> {
```
apoorvsadana: why do this round trip serialization? why not just take a `mut` variable and update values inside it directly.

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+        if chain_config.chain_id != current_backend_config.chain_id {
+            anyhow::bail!(
+                "Chain ID mismatch: saved config has chain_id={}, but current node has chain_id={}",
+                chain_config.chain_id,
+                current_backend_config.chain_id
+            );
+        }
+
+        // Step 4: Reconstruct versioned_constants from current node
+        chain_config.versioned_constants = Self::copy_versioned_constants_from_current(current_backend_config);
```
apoorvsadana: why not take the versioned constants we stored?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,247 @@
...
+        current_backend_config: &ChainConfig,
+    ) -> Result<Self> {
+        // Step 1: Validate saved config against current node
+        Self::validate_saved_config(&saved, current_backend_config)?;
+
+        // Step 2: Deserialize saved chain config
+        let mut chain_config = Self::deserialize_saved_chain_config(&saved.chain_config_json)?;
+
+        // Step 3: Validate chain ID matches
+        if chain_config.chain_id != current_backend_config.chain_id {
```
apoorvsadana: this check should be in `validate_saved_config`

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: versioned_constants are not serializable, hence couldn't do that, one way to do this would be to create versionedConstants on our side 😓

madara/crates/client/db/src/rocksdb/mod.rs
Mohiiit: reverted, earlier the code of config was here, moved that to primitives now and not needed now!

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: as discussed, not storing the vc as of now

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: `ChainConfig` doesn't implement `Clone`, so we cannot simply clone and mutate it directly. The round-trip serialization serves as a way to create a copy of the config.

madara/node/src/cli/mod.rs
```diff
@@ -301,8 +300,8 @@ impl RunCmd {
         };
 
         chain_config.private_key = match self.private_key.take() {
-            Some(s) => s.try_into().context("Failed to parse private key")?,
-            None => ZeroingPrivateKey::default(),
+            Some(s) => Some(s.try_into().context("Failed to parse private key")?),
+            None => None,
```
byteZorvin: Why not use the default?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,262 @@
...
+//! ## What Gets Saved
+//!
+//! - `chain_config` fields (via YAML serialization)
+//! - `protocol_version` (string, e.g., "0.13.2") - used to look up versioned constants on restart
+//! - `no_charge_fee` flag
+//! - `config_version` - for compatibility validation
+//!
+//! ## What Does NOT Get Saved (and why)
+//!
+//! - **`versioned_constants`**: The `VersionedConstants` type from blockifier does not implement
```
byteZorvin, line 14: We can add a small comment on the versioned_constant versions(v1, v2..) implementation could be pursued in future with wrapper type serialization, or gh issue link may be

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,262 @@
...
+//! - **`private_key`**: Set to `None` in saved configs. Re-execution doesn't require signing;
+//!   the current node's private key (from CLI/chain-config) is used for any signing operations.
+//!
+//! ## Important Assumption
+//!
+//! **Versioned constants are immutable for a given protocol version.** This means:
+//! - For protocol version "0.14.0", the gas costs, limits, and other constants are the same
+//!   across all node versions and restarts.
+//! - If you use custom `versioned_constants` (via `versioned_constants_path` in chain config),
+//!   you must ensure they remain consistent across restarts.
```
byteZorvin, line 33: According to discussion, we can add that if we want to update custom versioned constants then protocol versioned must also be bumped

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
byteZorvin: Why dont we implement Clone from ChainConfig, my guess that its intensive operation, but saw that VersionedConstants do implement Clone then may be can?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -58,12 +58,27 @@
...
+//! is executed (`BatchExecuted` message). This persistence allows us to recover the pre-confirmed
+//! block upon restart.
+//!
+//! ### Restart Recovery
+//!
+//! When Madara starts, it checks for an existing pre-confirmed block. If found:
+//! 1. It loads the saved **runtime execution configuration** (ChainConfig, VersionedConstants, etc.)
+//!    to ensure re-execution uses the exact same parameters as the original execution.
+//! 2. It **re-executes** all transactions in the pre-confirmed block. This is necessary because
+//!    intermediate execution artifacts (like bouncer weights and state diffs) are not fully persisted.
```
byteZorvin, line 80: Why keep bouncer weights in the category of state diff, they will be persisted in the db right?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,262 @@
...
+        // Convert to canonical ChainConfig
+        let mut chain_config = ChainConfig::try_from(versioned).context("Failed to convert to ChainConfig")?;
+
+        // Copy versioned_constants back (needed for in-memory operations, not persisted)
+        chain_config.versioned_constants = {
+            let mut vc = ChainVersionedConstants::default();
+            for (version, constants) in &config.versioned_constants.0 {
+                vc.add(*version, constants.clone());
+            }
+            vc
```
byteZorvin: Why do we do explicity set the default values?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,262 @@
...
+            anyhow::bail!(
+                "Config version mismatch: saved config has version {}, but current node expects version {}. \
+                 You may need to migrate your database or use a compatible node version.",
+                saved.config_version,
+                current.config_version
+            );
+        }
+
+        // Check chain ID - must match exactly
+        if saved_chain_config.chain_id != current.chain_id {
```
byteZorvin: This is only for reexecuting the preconfirmed block right and we can change chain id later on?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -586,15 +621,36 @@ impl BlockProductionTask {
...
             block_number
         );
 
+        // Load saved runtime execution config
+        let saved_config = self.backend.get_runtime_exec_config().context("Getting runtime execution config")?;
+
+        // Extract saved values for re-execution without modifying self
+        let (saved_chain_config, saved_no_charge_fee) = if let Some(config) = saved_config {
+            // Log warning if saved config differs from current config (for debugging)
+            if config.chain_config.chain_id != self.backend.chain_config().chain_id {
```
byteZorvin: Saw this check is also present [here](https://github.com/madara-alliance/madara/blob/9f6e13fe11e3f1473b67ba814c0dbbfdabfec541/madara/crates/client/db/src/lib.rs#L329) is this redundant then?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
byteZorvin: I guess not, as there is a check on MadaraBackend init

byteZorvin (commented):
Great one. Mostly few minor questions !!

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: even if `VersionedConstants` has clone, it doesn't have serialize, no point of cloning it then?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: will add a TODO about that

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: I think the next task would be picked up immediately after this, but will add this for now!

madara/crates/client/block_production/src/lib.rs
Mohiiit: these are the bouncer_weights associated with each block (when we call finalize on the executor)

these bouncer_weights are then used in snos to calculate the batch size

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
Mohiiit: shouldn't be the case, a lot of changes required, config_hash also changes, not recommended to change the chain_id

byteZorvin (approved):
Lgtm !!

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
```diff
@@ -0,0 +1,208 @@
...
+//!
+//! ## Important Assumption
+//!
+//! **Versioned constants are immutable for a given protocol version.** This means:
+//! - For protocol version "0.14.0", the gas costs, limits, and other constants are the same
+//!   across all node versions and restarts.
+//! - If you use custom `versioned_constants` (via `versioned_constants_path` in chain config),
+//!   you must ensure they remain consistent across restarts.
+//! - **Important:** If you need to update custom versioned constants, you MUST also bump the
+//!   protocol version to maintain the immutability guarantee.
```
apoorvsadana: this comment is more appropriate for when we support this, we don't support this at the moment right?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
apoorvsadana: seems like a heavy and hacky way to create a clone however. what's the problem with implementing Clone on ChainConfig?

madara/crates/primitives/chain_config/src/runtime_exec_config.rs
apoorvsadana: as discussed, let's also add closing the pre confirmed block point

apoorvsadana (approved):

apoorvsadana (approved):

PR #863, opened 2025-11-22
fix: onchain data calculation is dead code

byteZorvin:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

- Onchain data hash/size calculation is dead code and is not required now
- Also, when and l3 chain is run with `use_kzg_da`: true, this breaks with not able to find `fact_root`

## What is the new behavior?
- Removes onchain data hash and size calculation.

- No

### DIFF
--- a/orchestrator/src/worker/event_handler/jobs/snos.rs
+++ b/orchestrator/src/worker/event_handler/jobs/snos.rs
@@ -10,7 +10,7 @@ use crate::types::jobs::metadata::{JobMetadata, JobSpecificMetadata, SnosMetadat
 use crate::types::jobs::status::JobVerificationStatus;
 use crate::types::jobs::types::{JobStatus, JobType};
 use crate::worker::event_handler::jobs::JobHandlerTrait;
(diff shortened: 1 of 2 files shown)


orchestrator/src/worker/event_handler/jobs/snos.rs
```diff
@@ -180,11 +180,7 @@ impl JobHandlerTrait for SnosJobHandler {
         }
 
         debug!("Storing SNOS outputs");
-        if config.layer() == &Layer::L3 {
-            // Store the on-chain data path
-            self.store_l2(internal_id.clone(), config.storage(), &snos_metadata, cairo_pie, os_output, program_output)
-                .await?;
-        } else if config.layer() == &Layer::L2 {
+        if config.layer() == &Layer::L2 {
```
Mohiiit: this if condition can be removed now

byteZorvin: Makes sense updated

apoorvsadana (approved):

prkpndy (approved):

PR #864, opened 2025-11-25
Fix/blob-settlement

byteZorvin:
Fix L3 settlement when KZG DA flag is enabled

## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

When running an L3 chain with the KZG DA flag enabled (`use_kzg_da = 1`), settlement breaks because the code attempts to use blob-based settlement (`update_state_with_blobs`) which is not supported by L3 core contracts. The L3 settlement logic previously only handled the case where `use_kzg_da = 0`, and when the flag was set to `1`, it would incorrectly try to use blob settlement, causing the settlement process to fail.

## What is the new behavior?

- L3 chains now always use calldata-based settlement regardless of the KZG DA flag value (0 or 1)
- The condition for L3 state updates now accepts both `Felt::ZERO` and `Felt::ONE` for the `use_kzg_da` flag
- Removed the blob-based settlement path for L3 chains that was causing failures
- Simplified the calldata update by removing unused onchain data fetching logic
- Added documentation comments explaining that L3 core contracts do not support blobs and always require calldata for state updates

No

## Other information

This fix ensures that L3 chains can successfully perform state updates to the settlement layer even when KZG DA is enabled. The core contract for L3 chains does not support blob transactions, so all state updates must use calldata. When KZG DA is enabled on L3, this configuration effectively provides private DA functionality, as the state diff is not included in the snos_output while still maintaining the ability to update state on the settlement layer.

### DIFF
--- a/orchestrator/src/worker/event_handler/jobs/state_update.rs
+++ b/orchestrator/src/worker/event_handler/jobs/state_update.rs
@@ -3,7 +3,7 @@ use crate::error::job::state_update::StateUpdateError;
 use crate::error::job::JobError;
 use crate::error::other::OtherError;
 use crate::types::batch::{AggregatorBatchStatus, SnosBatchStatus};
-use crate::types::constant::{ON_CHAIN_DATA_FILE_NAME, PROOF_FILE_NAME, PROOF_PART2_FILE_NAME};
+use crate::types::constant::{PROOF_FILE_NAME, PROOF_PART2_FILE_NAME};
 use crate::types::jobs::job_item::JobItem;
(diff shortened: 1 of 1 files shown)


Mohiiit (commented):
shouldn't we make this a config? instead of hardcoding this? or this is the way we are going forward for other future L3s as well?

apoorvsadana (approved):

Mohiiit (approved):

PR #865, opened 2025-11-27
feat(orchestrator): Add default proving gas for empty blocks

Mohiiit:
#### Problem

When a block has no transactions, the bouncer weights API returns `proving_gas = 0`. However, the Starknet OS still performs work for every block (even empty ones).

This results in approximately **~4,775 steps (~477,500 proving gas)** per empty block that wasn't being accounted for in batch capacity calculations.

#### Solution

Introduced a configurable `default_empty_block_proving_gas` parameter that's applied when a block returns zero proving gas from the bouncer weights API.

**Default value: 1,500,000 gas (~3x safety margin over measured ~477,500 gas)**

---

#### Empirical Data

| Empty Blocks in Batch | Total Steps | Per-Block (Marginal) |
|----------------------|-------------|----------------------|
| 1 | 4,775 | 4,775 |
| 2 | 8,880 | 4,105 |
| 3 | 12,985 | 4,105 |
| 5 | 21,195 | 4,105 |
| 10 | 41,720 | 4,105 |
| 25 | 134,002 | ~6,152 |
| 50 | 294,404 | ~6,416 |
| 100 | 624,478 | ~6,601 |

**Key Observations:**
- **First block**: ~4,775 steps (includes one-time setup)
- **Additional blocks (small batch)**: ~4,105 steps each
- **At scale**: Per-block cost grows to ~6,500+ steps


apoorvsadana (approved):

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -1044,18 +1044,29 @@ impl BatchingTrigger {
         }
 
         let mut current_builtins = Some(BouncerWeights::empty());
+        let mut last_processed_block = batch.start_block.saturating_sub(1);
         for block_number in batch.start_block..=(batch.end_block + 1) {
+            // Check if the previous iteration caused an overflow
+            if current_builtins.is_none() {
```
byteZorvin, line 1056: Nit: Can add a comment that none checks overflow due to checked add returning option in previous iteration

byteZorvin (approved):
Looks good !!

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -1044,18 +1044,29 @@ impl BatchingTrigger {
...
+
             // Get the bouncer weights for this block
             let bouncer_weights = self
                 .get_block_builtin_weights(config, block_number)
                 .await
                 .map_err(|e| JobError::ProviderError(e.to_string()))?;
+
             if let Some(cb) = &mut current_builtins {
                 current_builtins = cb.checked_add(bouncer_weights);
-            } else {
```
byteZorvin, line 1055: Somehow this was more readable to me

Mohiiit: if it's a hard requirement, will redo it, ideally we should get rid of the code here tbh! this file needs thorough cleaning

PR #866, opened 2025-11-29
feat: L1→L2 Message Finality Support

Mohiiit:
### Description

Adds configurable finality checking for L1→L2 messages to ensure messages are only processed after reaching sufficient block confirmations.

#### Changes

**Core Logic (`messaging.rs`)**
- Messages are now queued in a `VecDeque` and processed only after reaching the required finality threshold
- Uses `tokio::select!` to poll the stream while periodically checking finality on queued events
- Refactored into helper functions: `get_start_block()`, `process_finalized_events()`

**Configuration (`chain_config.rs`)**
- Added `l1_messages_finality_blocks` parameter (default: 10 blocks)
- Configurable per-chain via preset files

**Tests**
- Added `test_finality_blocks_delays_processing` - verifies messages wait for finality
- Added `test_finality_blocks_processes_after_threshold` - verifies messages process after reaching finality

#### Behavior

```
L1 Event (block 100) → Queue → Wait for finality
                              ↓
              Latest L1 block reaches 110 (10 confirmations)
                              ↓
                    Process & store message
```

#### Why

Prevents processing L1 messages from blocks that could be reorged, ensuring message integrity.

---

### DIFF
--- a/configs/presets/devnet.yaml
+++ b/configs/presets/devnet.yaml
@@ -32,7 +32,7 @@ bouncer_config:
       mul_mod: 604
       range_check96: 56
 sequencer_address: "0x123"
-eth_core_contract_address: "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512"
+eth_core_contract_address: "0x003f5ea0ec5fad0b020839fec467a872a620e6be6093801a5ac92b1e58d0d33a"
 eth_gps_statement_verifier: "0xf294781D719D2F4169cE54469C28908E6FA752C1"
 mempool_max_transactions: 10000
 mempool_max_declare_transactions: 20
--- a/configs/presets/devnet.yaml
+++ b/configs/presets/devnet.yaml
@@ -32,7 +32,7 @@ bouncer_config:
       mul_mod: 604
       range_check96: 56
 sequencer_address: "0x123"
-eth_core_contract_address: "0x003f5ea0ec5fad0b020839fec467a872a620e6be6093801a5ac92b1e58d0d33a"
+eth_core_contract_address: "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512"
 eth_gps_statement_verifier: "0xf294781D719D2F4169cE54469C28908E6FA752C1"
 mempool_max_transactions: 10000
 mempool_max_declare_transactions: 20
--- a/madara/crates/client/settlement_client/src/messaging.rs
+++ b/madara/crates/client/settlement_client/src/messaging.rs
@@ -1,13 +1,14 @@
 use crate::client::{ClientType, SettlementLayerProvider};
 use crate::error::SettlementClientError;
 use alloy::primitives::{B256, U256};
-use anyhow::Context;
-use futures::{StreamExt, TryStreamExt};
+use futures::StreamExt;
 use mc_db::MadaraBackend;
 use mp_transactions::L1HandlerTransactionWithFee;
 use mp_utils::service::ServiceContext;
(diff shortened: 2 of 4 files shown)


madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -96,73 +97,188 @@ async fn sync_inner(
...
+                "Event at L1 block {} not finalized yet ({} < {} blocks required)",
+                event.l1_block_number,
+                blocks_since_event,
+                finality_blocks
+            );
+            break; // Events are ordered, so all remaining events are also not finalized
+        }
+
+        // Event is finalized - pop and process it
+        let event = pending_events.pop_front().unwrap();
```
apoorvsadana: let's use the `event` directly? or add a comment why unwrap isn safe

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -96,73 +97,188 @@ async fn sync_inner(
...
+        .await?;
+
+        if processed_count > 0 {
+            tracing::debug!("Processed {} finalized events", processed_count);
+        }
+    }
+}
+
+/// Determines the L1 block to start syncing from.
+async fn determine_start_block(
```
apoorvsadana:
```suggestion
async fn get_start_block(
```

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -96,73 +97,188 @@ async fn sync_inner(
...
+
+    // Buffer for events waiting for finality
+    let mut pending_events: VecDeque<MessageToL2WithMetadata> = VecDeque::new();
+    let mut stream_exhausted = false;
+
+    loop {
+        // Step 1: Receive new events from stream OR wait for finality check interval
+        if !stream_exhausted {
+            // Try to receive from stream with a short timeout
+            let timeout = tokio::time::sleep(Duration::from_millis(100));
```
apoorvsadana: we didn't have this before, do we need it now?

configs/presets/devnet.yaml
```diff
@@ -32,7 +32,7 @@ bouncer_config:
       mul_mod: 604
       range_check96: 56
 sequencer_address: "0x123"
-eth_core_contract_address: "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512"
+eth_core_contract_address: "0x003f5ea0ec5fad0b020839fec467a872a620e6be6093801a5ac92b1e58d0d33a"
```
byteZorvin: can change this back

madara/crates/client/settlement_client/src/messaging.rs
byteZorvin: I think we need to have a delay between 2 consecutive block numbers

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -96,73 +97,188 @@ async fn sync_inner(
...
+                            stream_exhausted = true;
+                        }
+                    }
+                }
+
+                _ = timeout => {
+                    // Timeout - proceed to check finality on pending events
+                }
+            }
+        } else {
```
byteZorvin: We can remove this as discussed

byteZorvin (commented):
Looks good, much more simplified. For passing the tests need to add the env var in e2e config json too

apoorvsadana (approved):

byteZorvin (approved):
Looks good !!

PR #867, opened 2025-11-29
Feat: Add Database Migration System

Mohiiit:
### Problem

Currently, Madara has no migration strategy - if the database schema changes, users must delete and resync from scratch. This is problematic for nodes with large databases.

### Solution

Implement a migration framework inspired by [Pathfinder](https://github.com/eqlabs/pathfinder/blob/main/crates/storage/src/schema.rs) that:

1. **Detects version mismatch** on startup by comparing `.db-version` file with binary's expected version
2. **Runs migrations sequentially** from current → target version
3. **Creates backup** before migrating (RocksDB checkpoint)
4. **Supports crash recovery** via state checkpointing - interrupted migrations resume on restart
5. **Reports progress** for long-running migrations

### Behavior

| Database State | Action |
|----------------|--------|
| Fresh | Create at current version |
| Same version | Open normally |
| Older (migratable) | Run migrations |
| Too old | Error with resync instruction |
| Newer than binary | Error with upgrade instruction |

### Adding Future Migrations

When a schema change is needed:
1. Create a migration file
2. Register it in the migration registry  
3. Bump `current_version` in `.db-versions.yml`

Documentation included in `MIGRATION.md`.


madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+                Err(e) => {
+                    // Can't determine age, treat as stale
+                    tracing::warn!("Cannot read lock metadata: {}, treating as stale", e);
+                    true
+                }
+            };
+
+            if is_stale {
+                tracing::warn!("Removing stale migration lock");
+                fs::remove_file(&lock_path)?;
```
prkpndy: possibility of a race condition here I think

madara/crates/client/db/src/lib.rs
```diff
@@ -442,20 +446,98 @@ impl MadaraBackend<RocksDBStorage> {
...
+                    db_version,
+                    binary_version
+                );
+            }
         }
+
+        // Now open with the proper RocksDBStorage wrapper
         let db_path = base_path.join("db");
-        let db = RocksDBStorage::open(&db_path, rocksdb_config).context("Opening rocksdb storage")?;
+        let db = RocksDBStorage::open(&db_path, rocksdb_config).context("Opening RocksDB storage")?;
```
prkpndy: why are we opening again here? can't we reuse the one from above?

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+    abort_flag: Arc<AtomicBool>,
+    skip_backup: bool,
+}
+
+impl MigrationRunner {
+    pub fn new(base_path: &Path, required_version: u32, base_version: u32) -> Self {
+        Self {
+            base_path: base_path.to_path_buf(),
+            required_version,
+            base_version,
```
prkpndy, line 105: can we add an assert here that required_version = base_version + 1

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// Creates a backup using RocksDB checkpoint (uses hard links, very fast).
+    fn create_backup(&self, db: &DB) -> Result<(), MigrationError> {
+        let backup_path = self.base_path.join(BACKUP_DIR_NAME);
+
+        if backup_path.exists() {
+            fs::remove_dir_all(&backup_path)?;
+        }
+
+        // Checkpoint creates hard links to SST files - fast and space-efficient
+        let checkpoint =
```
prkpndy: can flush DB before creating backup to ensure all data from memory is on disk (not sure of rocksdb does this automaticlly). also don't think this would ever happen since we'll be running migration in the start only, but still a good practice imo.

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+        Ok(())
+    }
+
+    fn acquire_lock(&self) -> Result<MigrationLock, MigrationError> {
+        let lock_path = self.base_path.join(DB_MIGRATION_LOCK);
+
+        if lock_path.exists() {
+            let is_stale = match fs::metadata(&lock_path).and_then(|m| m.modified()) {
+                Ok(modified) => match modified.elapsed() {
+                    Ok(age) => age > std::time::Duration::from_secs(24 * 60 * 60),
```
prkpndy: Let's move the time to a constant?

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+            MigrationStatus::DatabaseTooOld { current_version, base_version } => {
+                Err(MigrationError::DatabaseTooOld { current_version, base_version })
+            }
+            MigrationStatus::MigrationRequired { current_version, target_version, .. } => {
+                self.execute_migrations(db, current_version, target_version)
+            }
+        }
+    }
+
+    fn execute_migrations(&self, db: &DB, from_version: u32, to_version: u32) -> Result<(), MigrationError> {
```
prkpndy, line 189: should we also add a disk space check while creating backup and doing the migration? can give better errors to the dev! just a suggestion tho

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,457 @@
...
+                Ok(())
+            }
+            MigrationStatus::DatabaseNewer { db_version, binary_version } => {
+                Err(MigrationError::DatabaseNewerThanBinary { db_version, binary_version })
+            }
+            MigrationStatus::DatabaseTooOld { current_version, base_version } => {
+                Err(MigrationError::DatabaseTooOld { current_version, base_version })
+            }
+            MigrationStatus::MigrationRequired { current_version, target_version, .. } => {
+                self.execute_migrations(db, current_version, target_version)
```
prkpndy, line 184: nit: can we add a verification step as well after migration is complete?

madara/crates/client/db/src/migration/error.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[error("Serialization error: {0}")]
+    Serialization(String),
+
+    #[error("Migration aborted by user")]
+    Aborted,
+}
+
+impl From<rocksdb::Error> for MigrationError {
+    fn from(e: rocksdb::Error) -> Self {
+        MigrationError::RocksDb(e.to_string())
```
prkpndy: when improving logs for orch, I noticed that we lose the actual error context when using `.to_string` method. can we confirm this?

prkpndy (commented):
Should we add a list of supported DB versions?

madara/crates/client/db/src/migration/mod.rs
Mohiiit: there would be tests post migration which would validate the migration, that would right? or are you asking something else?

madara/crates/client/db/src/migration/mod.rs
Mohiiit: required_version could be more than base_version + 1, so there would be range of versions we will support, but there would always be a base version (the least version we support)

.github/workflows/task-test-migration.yml
```diff
@@ -0,0 +1,121 @@
...
+        uses: ./.github/actions/setup-rust
+        with:
+          rust-version: ${{ env.BUILD_RUST_VERSION }}
+          cache-key: madara-${{ runner.os }}
+
+      - name: Run migration tests
+        working-directory: madara
+        run: cargo test -p mc-e2e-tests migration_validation -- --ignored --nocapture
+        env:
+          MIGRATION_TEST_RPC_URL: "http://localhost:9944"
```
heemankv, line 113: I was wondering, we could potentially remove the burden of doing a rust setup here just to run the migration tests by utilising the Base vs Current version detection of job `Read versions from .db-versions.yml`
and using that as a trigger in the `test-madara` CI that will dynamically add this crate to the testing scope only when a version difference is detected, leading to saving time consumed by `Rust Setup` ~1.5 min per run

heemankv: wdyt?

madara/crates/client/db/build.rs
```diff
@@ -83,25 +86,30 @@ fn get_db_version() -> Result<(), BuildError> {
...
 
-fn parse_version(content: &str) -> Result<u32, BuildError> {
+fn parse_version(content: &str, key: &str) -> Result<u32, BuildError> {
+    let prefix = format!("{}:", key);
     content
         .lines()
-        .find(|line| line.starts_with("current_version:"))
-        .ok_or(BuildError::Parse(Cow::Borrowed("Could not find current_version")))?
+        .find(|line| line.starts_with(&prefix))
+        .ok_or_else(|| BuildError::Parse(Cow::Owned(format!("Could not find {}", key))))?
```
heemankv, line 106: I'm confused, what is the necessity to use Owned here ?

madara/crates/client/db/src/migration/mod.rs
```diff
@@ -0,0 +1,465 @@
...
+const BACKUP_DIR_NAME: &str = "backup_pre_migration";
+const STALE_LOCK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(24 * 60 * 60);
+
+#[derive(Debug)]
+pub enum MigrationStatus {
+    FreshDatabase,
+    NoMigrationNeeded,
+    MigrationRequired { current_version: u32, target_version: u32, migration_count: usize },
+    DatabaseTooOld { current_version: u32, base_version: u32 },
+    DatabaseNewer { db_version: u32, binary_version: u32 },
```
heemankv, line 79: Can these two `DatabaseTooOld` & `DatabaseNewer` be adopted from `MigrationError` ? 
seems like we are re-writing these two

.github/workflows/task-test-migration.yml
```diff
@@ -0,0 +1,121 @@
...
+          IMAGE="ghcr.io/madara-alliance/db-fixture:v${VERSION}"
+
+          echo "📦 Pulling base DB v${VERSION} from ${IMAGE}..."
+          mkdir -p /tmp/migration-test
+
+          if ! docker pull "${IMAGE}"; then
+            echo "❌ Base DB fixture not found!"
+            echo "Please run: ./scripts/create-base-db.sh ${VERSION}"
+            exit 1
+          fi
```
heemankv, line 51: I think we can automate this as well using the strategy I've mentioned for `Read versions from .db-versions.yml`

heemankv: Like, in the file .db_version.yml if the db version is updated then we run this before proceeding.

heemankv: Like, having to ensure that we just need to write the code and CI will handle everything else will be superb

madara/crates/client/db/src/migration/MIGRATION.md
```diff
@@ -0,0 +1,268 @@
+# Database Migration System
```
heemankv, line 1: Should this file not be in root/madara-root, unsure of the convention here

.github/workflows/task-test-migration.yml
Mohiiit: we will run the CI for each PR, because there could be cases where type changes could lead to madara not supporting certain lower db versions

madara/crates/client/db/build.rs
Mohiiit: `BuildError::Parse` takes `Cow<'static, str>` - this pattern allows both static strings (`Cow::Borrowed("static error")`) without allocation AND dynamic strings (`Cow::Owned(format!(...))`) when we need to include runtime values like `key`. Since we're using `format!()` to include the key name in the error message, we need `Owned`. Could simplify to just `String` but `Cow` is more flexible if we ever add static error messages.

madara/crates/client/db/src/migration/mod.rs
Mohiiit: Good observation! These are intentionally separate:

- `MigrationStatus` (enum) describes **states** - it's informational, returned by `check_status()` to tell you what state the DB is in
- `MigrationError` (error enum) describes **failures** - actionable errors when operations fail

The flow is: `check_status()` returns a status → `run_migrations()` converts certain statuses to errors. Keeping them separate maintains the distinction between "what state is the DB in?" vs "what went wrong?". Could reference errors from status but the explicit separation makes the intent clearer IMO.

madara/crates/client/db/src/migration/MIGRATION.md
Mohiiit: The current location keeps documentation close to the migration code it documents - this is a common pattern (co-locating module-specific docs with the module). Moving to root would increase visibility but separate docs from implementation. Happy to move if there's a preference for root-level docs, but the co-location with code is intentional.

madara/crates/client/db/src/migration/mod.rs
prkpndy: yeah, fair

madara/crates/client/db/src/lib.rs
Mohiiit: The DB is only opened once and reused after migration. See lines 514-521 in lib.rs - we open it, run migrations, then return the same `db` instance. The comment "Reuse the same DB instance instead of reopening" confirms this intent.

madara/crates/client/db/src/migration/mod.rs
Mohiiit: The lock uses `OpenOptions::create_new(true)` which maps to `O_CREAT | O_EXCL` flags - this is atomic at the OS level. If two processes try to create the lock simultaneously, only one will succeed, the other will get `AlreadyExists` error. No race condition possible with this approach.

madara/crates/client/db/src/migration/mod.rs
Mohiiit: Already implemented! `db.flush_wal(true)` is called before creating the checkpoint (line 324). This ensures all WAL data is flushed to SST files for a consistent backup.

madara/crates/client/db/src/migration/mod.rs
Mohiiit: Already a constant! See line 35: `const STALE_LOCK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(24 * 60 * 60);`

madara/crates/client/db/src/migration/mod.rs
Mohiiit: Good suggestion! Implemented this - now checking disk space before backup using `libc::statvfs`. If available space is less than DB size + 10% buffer, we return a clear `InsufficientDiskSpace` error with the required vs available bytes.

madara/crates/client/db/src/migration/error.rs
Mohiiit: Good catch! Fixed this - changed `BackupFailed(String)` to `BackupFailed { message: &'static str, #[source] source: rocksdb::Error }` so the error chain is preserved via the `#[source]` attribute. Now calling `.source()` on the error will return the underlying RocksDB error.

madara/crates/client/db/src/migration/mod.rs
prkpndy: I was thinking that after the db is migrated, we can verify that it has been migrated correctly before deleting the backup. is this already present?

prkpndy (approved):
asked for a clarification. looks good otherwise. gg!

heemankv (approved):

PR #868, opened 2025-11-29
fix(feeder_gateway): fix preconfirmed endpoint

byteZorvin:
## Pull Request type

- [x] Bugfix

The `handle_get_preconfirmed_block` endpoint in the feeder gateway uses `block_view_on_preconfirmed()` which returns `None` when there's no preconfirmed block available. This differs from how the RPC implementation handles the same scenario, which uses `block_view_on_preconfirmed_or_fake()` that always returns a block (even a fake one). This inconsistency can lead to confusing behavior and less informative error messages.

Resolves: #NA

## What is the new behavior?

- Uses `block_view_on_preconfirmed_or_fake()` instead of `block_view_on_preconfirmed()` to align with the RPC implementation behavior
- Adds `tracing::info!` logging for incoming pre-confirmed block requests to improve observability
- Provides a more informative error message when the requested block number doesn't match, now including the current pre-confirmed block number in the error response

No

The API contract remains the same. The endpoint still returns the pre-confirmed block when available and an error when not found. The only difference is:
1. More consistent internal implementation with RPC
2. Improved error messages with additional context

## Other information

The changes are localized to `crates/client/gateway/server/src/handler.rs` in the `handle_get_preconfirmed_block` function.


madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -50,13 +50,21 @@ pub async fn handle_get_preconfirmed_block(
...
         .map_err(|e: std::num::ParseIntError| StarknetError::new(StarknetErrorCode::MalformedRequest, e.to_string()))?;
 
-    let mut block =
-        backend.block_view_on_preconfirmed().filter(|block| block.block_number() == block_number).ok_or_else(|| {
-            StarknetError::new(
-                StarknetErrorCode::BlockNotFound,
-                format!("Pre-confirmed block with number {block_number} was not found."),
-            )
-        })?;
+    tracing::info!(target: "feeder_gateway", "Requesting pre-confirmed block with number: {block_number}");
```
Mohiiit: do we need this? don't have in other functions tho

madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -50,13 +50,21 @@ pub async fn handle_get_preconfirmed_block(
...
+
+    // Use block_view_on_preconfirmed_or_fake() - this always returns a block
+    let mut block = backend
```
Mohiiit, line 55: do we need this to be mutable tho? below we are creating new block I think, not sure myself

Mohiiit (approved):

heemankv (approved):

heemankv (approved):
Please ack @Mohiiit 's comments, I've same, 
rest looks good.

PR #870, opened 2025-12-01
fix: send fee_token_address while creating Atlantic bucket

byteZorvin:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

When creating a bucket with the Atlantic prover service, the `fee_token_address` is not being sent in the `aggregator_params`. This is required by the Atlantic API for proper bucket creation.

Resolves: MDR-655

## What is the new behavior?

- Added `fee_token_address` field to `AtlanticAggregatorParams` struct
- Updated `create_bucket` function to accept and pass `fee_token_address` parameter
- Updated `AtlanticProverService` to store and pass `fee_token_address` when creating buckets
- The `strk_fee_token_address` from SNOS config is now passed when building the prover service

The bucket creation request now includes `fee_token_address` in the `aggregator_params`:
```json
{
  "aggregator_params": {
    "use_kzg_da": true,
    "full_output": false,
    "chain_id_hex": "0x...",
    "fee_token_address": "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
  }
}
```

No

## Other information

Files modified:
- `orchestrator/crates/prover-clients/atlantic-service/src/types.rs`
- `orchestrator/crates/prover-clients/atlantic-service/src/client.rs`
- `orchestrator/crates/prover-clients/atlantic-service/src/lib.rs`
- `orchestrator/src/core/config.rs`
- `orchestrator/src/tests/config.rs`
- `orchestrator/crates/prover-clients/atlantic-service/tests/lib.rs`

### DIFF
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -188,6 +188,7 @@ impl AtlanticClient {
         atlantic_api_key: impl AsRef<str>,
         mock_proof: bool,
         chain_id_hex: Option<String>,
+        fee_token_address: Option<String>,
     ) -> Result<AtlanticBucketResponse, AtlanticError> {
         // TODO(prakhar,19/11/2025): Use the aggregator version calculated from Madara Version being passed through ENV
         let response = self
@@ -206,6 +207,7 @@ impl AtlanticClient {
                     use_kzg_da: AGGREGATOR_USE_KZG_DA,
                     full_output: AGGREGATOR_FULL_OUTPUT,
                     chain_id_hex,
+                    fee_token_address,
                 },
                 mock_proof,
             })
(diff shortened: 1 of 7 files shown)


orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -188,6 +188,7 @@ impl AtlanticClient {
         atlantic_api_key: impl AsRef<str>,
         mock_proof: bool,
         chain_id_hex: Option<String>,
+        fee_token_address: Option<String>,
```
Mohiiit: ideally, we should have a aggregatorInput struct here, which can be taken from: https://github.com/starkware-libs/sequencer/blob/main-v0.14.1/crates/starknet_os/src/hint_processor/aggregator_hint_processor.rs#L42-L52

can be a todo for now, given for 0.14.1 we would need to send the public_keys as well

Mohiiit (approved):
LGTM

heemankv (commented):
Overall looks good,
Can you possibly see if we can use Felt for this address, 
if we can use clap's parsing abilities for that, it would be great!

heemankv (approved):

PR #873, opened 2025-12-01
fix: max concurrent flag not being applied from CLI args

byteZorvin:
## Pull Request type

- [x] Bugfix

CLI arguments `--max-concurrent-snos-jobs` and `--max-concurrent-proving-jobs` were not being applied to the event worker queue control, causing the flags to have no effect.

Resolves: #NA

## What is the new behavior?

- Override queue control `max_message_count` with service config values when CLI/env args are provided
- Added logging to show when overrides are applied

No

## Other information

N/A

### DIFF
--- a/orchestrator/src/worker/controller/event_worker.rs
+++ b/orchestrator/src/worker/controller/event_worker.rs
@@ -51,27 +51,19 @@ impl EventWorker {
         // Override with service config values if provided via CLI/env args
         // This ensures CLI arguments like --max-concurrent-snos-jobs actually take effect
         let service_config = config.service_config();
-        match queue_type {
-            QueueType::SnosJobProcessing => {
-                if let Some(max_concurrent) = service_config.max_concurrent_snos_jobs {
-                    info!(
-                        "Overriding SnosJobProcessing max_message_count from {} to {} (from service config)",
-                        queue_control.max_message_count, max_concurrent
-                    );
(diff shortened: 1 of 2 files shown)


orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -46,7 +46,33 @@ impl EventWorker {
...
+                if let Some(max_concurrent) = service_config.max_concurrent_proving_jobs {
+                    info!(
+                        "Overriding ProvingJobProcessing max_message_count from {} to {} (from service config)",
+                        queue_control.max_message_count, max_concurrent
+                    );
+                    queue_control.max_message_count = max_concurrent;
+                }
+            }
+            _ => {}
+        }
```
Mohiiit:
```suggestion
let override_val = match queue_type {
    QueueType::SnosJobProcessing => service_config.max_concurrent_snos_jobs,
    QueueType::ProvingJobProcessing => service_config.max_concurrent_proving_jobs,
    _ => None,
};

if let Some(max_concurrent) = override_val {
    info!(
        "Overriding {:?} max_message_count from {} to {} (from service config)",
        queue_type, queue_control.max_message_count, max_concurrent
    );
    queue_control.max_message_count = max_concurrent;
}
```

Mohiiit (approved):
LGTM!

heemankv (changes requested):
Let's discuss this offline

PR #875, opened 2025-12-01
feat:  Migrate from starknet to starknet-rust crate

Mohiiit:
### Summary
Updates the codebase to use the renamed `starknet-rust` crate (formerly `starknet`). The crate was moved to a new package name by the maintainers.

### Changes
Replaced `felt!` and `selector!` macro usages with runtime equivalents.

### Why macro changes were needed

The proc macros (`felt!`, `selector!`) use `$crate` internally, which resolves to the actual package name (`starknet_rust`). Since we alias the import:

```toml
starknet = { package = "starknet-rust", version = "0.17.0" }
```

The macros fail to resolve with:
```
error: could not find `starknet_rust` in the list of imported crates
```

Importing with both names (`starknet` and `starknet_rust`) is not allowed:
```
error: the crate depends on crate `starknet-rust` multiple times with different names
```

### Solution
Replace macros with runtime equivalents:
- `felt!("0x123")` → `Felt::from_hex("0x123").unwrap()`
- `selector!("fn_name")` → `get_selector_from_name("fn_name").unwrap()`

---


bootstrapper-v2/src/setup/madara/mod.rs
```diff
@@ -138,7 +138,8 @@ impl MadaraSetup {
         let universal_deployer_class = self.require_class_hash(&MadaraClass::UniversalDeployer)?;
 
         let calldata = Vec::from([*universal_deployer_class, Felt::ZERO, Felt::ONE, Felt::ZERO]);
-        let calls = vec![Call { to: account_address, selector: selector!("deploy_contract"), calldata }];
+        let calls =
+            vec![Call { to: account_address, selector: get_selector_from_name("deploy_contract").unwrap(), calldata }];
```
prkpndy: can add a comment specifying that unwrap is safe!

prkpndy: or maybe a better design would be to create `get_selector_from_name_or_panic` function and return Felt instead of Result. wdyt?

prkpndy (approved):
some suggestions. otherwise looks good. gg

bootstrapper-v2/src/setup/madara/mod.rs
Mohiiit: resolved

heemankv (approved):
LGTM

PR #880, opened 2025-12-03
Remove unused `aggregator_index` parameter and enable debug symbols in release builds

heemankv:
Pull Request type

- [x] Build-related changes

1. The `get_aggregator_task_id` method in the `ProverClient` trait accepts an `aggregator_index` parameter that is unused by all implementations (Atlantic and Sharp). Both implementations ignore this parameter and use their own lookup mechanisms instead.

2. Release builds do not include debug symbols, making it difficult to diagnose issues in production - stack traces show `<unknown source file>:<unknown line>` instead of actual file paths and line numbers.

Resolves: #NA

## What is the new behavior?

- Removed the unused `aggregator_index` parameter from `get_aggregator_task_id` across:
  - `ProverClient` trait definition (`prover-client-interface/src/lib.rs`)
  - Atlantic implementation (`atlantic-service/src/lib.rs`)
  - Sharp implementation (`sharp-service/src/lib.rs`)
  - Caller in `aggregator.rs`
  - Test mocks

- Added `[profile.release]` with `debug = true` to orchestrator's `Cargo.toml` for better stack traces and profiling support in release builds.

No. The `aggregator_index` parameter was unused by all implementations, so removing it has no functional impact.

## Other information

The Atlantic implementation finds the aggregator job by its step type (`FactHashRegistration`) rather than by index, as noted in the code comment referencing `docs/bugs/aggregator_task_id_race_condition.md`. This approach is more reliable than using an index that depends on `num_snos_batches`.


Mohiiit (approved):

orchestrator/crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -179,11 +179,11 @@ pub fn prepare_sidecar_with_cells(
     // Format: all proofs for blob 0, then all proofs for blob 1, etc.
     let flat_cell_proofs: Vec<FixedBytes<48>> = all_cell_proofs.into_iter().flatten().collect();
 
-    // Create the sidecar with cell proofs
-    Ok(BlobTransactionSidecarVariant::from(BlobTransactionSidecar::new(
+    // Create the EIP-7594 sidecar with cell proofs (required for Fusaka/PeerDAS)
+    Ok(BlobTransactionSidecarVariant::Eip7594(BlobTransactionSidecarEip7594::new(
```
prkpndy, line 183: nit: we can keep this as it is imo. more readable and similar to `prepare_sidecar_with_proof` function above.

prkpndy (approved):
lgtm

orchestrator/Dockerfile
```diff
@@ -140,9 +140,9 @@ WORKDIR /app
 COPY --from=builder-rust /app/target/release/orchestrator /bin
 
 # Copy trusted setup files
-COPY --from=builder-rust /app/orchestrator/crates/da-clients/ethereum/trusted_setup.txt \
+COPY --from=builder-rust /app/orchestrator/crates/settlement-clients/ethereum/src/trusted_setup.txt \
```
byteZorvin, line 143: Curious how is not failing on main right now

orchestrator.dockerfile
byteZorvin, line 1: Is this not being used?

.markdownlintignore
```diff
@@ -2,3 +2,4 @@
 node_modules/**/*.md
 **/target/**/*.md
 docs/**/*.md
+AI_docs/**/*.md
```
byteZorvin, line 5: nit: I dont think everybody has. May be you can use .git/info/exclude to achieve the same thing

byteZorvin (approved):
Lgtm

PR #881, opened 2025-12-03
feat: add max_blocks_per_snos_batch check to batch closing logic

byteZorvin:
## Pull Request type

- [x] Feature

The `should_close_snos_batch` function only checked `fixed_blocks_per_snos_batch` (used for testing) but did not check `max_blocks_per_snos_batch` as a separate condition. When `max_blocks_per_snos_batch` was set, it was not being used to determine if a batch should be closed.

Resolves: #NA

## What is the new behavior?

- Added separate check for `max_blocks_per_snos_batch` in `should_close_snos_batch` function
- If `max_blocks_per_snos_batch` is set and batch has reached that limit, returns `true` to close the batch
- If condition is not met, continues with other batch closing checks (time-based, bouncer weights)
- Added `fixed_blocks_per_snos_batch` CLI argument and config parameter (for testing purposes)

No

## Other information

- `fixed_blocks_per_snos_batch` takes precedence over `max_blocks_per_snos_batch` when both are set
- `max_blocks_per_snos_batch` is now properly checked as a separate condition before other closing criteria

### DIFF
--- a/orchestrator/src/worker/event_handler/triggers/batching.rs
+++ b/orchestrator/src/worker/event_handler/triggers/batching.rs
@@ -102,7 +102,6 @@ impl JobTrigger for BatchingTrigger {
             Layer::L3 => self.get_range_for_assigning_batches_l3(&config).await?,
         };
 
-        tracing::info!("in batching code");
         // Invoking method to assign batches to all the blocks in the range
         if start_block < end_block {
             match config.layer() {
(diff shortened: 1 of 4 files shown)


Mohiiit (approved):
instead of adding another number to the config, do you think it would be beneficial to have just a bool flag: --fixed-snos-blocks-per-batch boolean

by default it could be false and in case of true, max_blocks_per_snos_batch would be used to get the exact number?

PS: not a blocking comment, can be ignored for now 🙇

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -102,6 +102,7 @@ impl JobTrigger for BatchingTrigger {
             Layer::L3 => self.get_range_for_assigning_batches_l3(&config).await?,
         };
 
+        tracing::info!("in batching code");
```
prkpndy: let's remove this

orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -1036,11 +1037,17 @@ impl BatchingTrigger {
     /// 2. Update it incrementally when adding each block
     /// 3. Here, only fetch the next block's weights (end_block + 1) and check overflow
     async fn should_close_snos_batch(&self, config: &Arc<Config>, batch: &SnosBatch) -> Result<bool, JobError> {
-        if let Some(max_blocks_per_snos_batch) = config.params.batching_config.max_blocks_per_snos_batch {
+        if let Some(fixed_blocks_per_snos_batch) = config.params.batching_config.fixed_blocks_per_snos_batch {
             // If the MADARA_ORCHESTRATOR_MAX_BLOCKS_PER_SNOS_BATCH env is set, we use that value
```
prkpndy: let's update this comment

prkpndy (changes requested):
minor changes

prkpndy (approved):

PR #885, opened 2025-12-04
fix: <0.14.1 will sync now

Mohiiit:
Patch for networks with <0.14.1 but full-node needs to sync 

PS: won't work for >=0.14.1


prkpndy (approved):
lgtm

heemankv (approved):

PR #887, opened 2025-12-05
feat(orchestrator): fetch chain details at startup with retry

heemankv:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, the orchestrator fetches chain configuration (chain_id, fee token addresses) during SNOS job processing. This means:
1. Every SNOS job makes an RPC call to fetch `chain_id`
2. Fee token addresses (`strk_fee_token_address`, `eth_fee_token_address`) must be manually configured via environment variables
3. If the RPC is temporarily unavailable during job processing, the job fails immediately without retry

Resolves: #NA

## What is the new behavior?

- **Chain details fetched once at startup**: A new `ChainDetails` struct is fetched at orchestrator startup with retry logic (5s interval, 5min timeout)
  - `chain_id` fetched from RPC (`starknet_chainId`)
  - Fee token addresses fetched from feeder gateway (`/feeder_gateway/get_contract_addresses`)

- **No more env vars for fee tokens**: Fee token addresses are now dynamically fetched from the node, removing the need for `MADARA_ORCHESTRATOR_STRK_NATIVE_FEE_TOKEN_ADDRESS` and `MADARA_ORCHESTRATOR_ETH_NATIVE_FEE_TOKEN_ADDRESS` environment variables

- **Fail-fast at startup**: If the node is unreachable, the orchestrator fails at startup rather than mid-job processing

**Yes** - The following environment variables are no longer used and can be removed:
- `MADARA_ORCHESTRATOR_STRK_NATIVE_FEE_TOKEN_ADDRESS`
- `MADARA_ORCHESTRATOR_ETH_NATIVE_FEE_TOKEN_ADDRESS`

Migration: Simply remove these environment variables from your configuration. The orchestrator will now fetch these values automatically from the node's feeder gateway.

### Architecture

```
Startup Flow:
┌─────────────────┐     ┌──────────────┐     ┌─────────────────────┐
│  Orchestrator   │────▶│ ChainDetails │────▶│  RPC: chain_id      │
│    Startup      │     │   ::fetch()  │     │  FG: fee tokens     │
└─────────────────┘     └──────────────┘     └─────────────────────┘
                              │ retry (5s interval, 5min timeout)
                              ▼
                        ┌──────────────┐
                        │   Config     │
                        │ (cached)     │
                        └──────────────┘

SNOS Job Flow:
┌─────────────────┐     ┌──────────────────────────────┐
│   SNOS Job      │────▶│ config.chain_details()       │ (instant, no RPC)
│  process_job()  │     │   .to_chain_config()         │
└─────────────────┘     └──────────────────────────────┘
```

image


heemankv (commented):
Self Review done, lgtm

heemankv (commented):
Re-reviewed, cleaned some code, lgtm ++ now

orchestrator/crates/utils/src/chain_details.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// Attempt to fetch chain details (single attempt, no retry).
+    async fn try_fetch(rpc_url: &Url, feeder_gateway_url: &Url) -> Result<Self> {
+        let chain_id = Self::fetch_chain_id(rpc_url).await?;
+        let addresses = Self::fetch_contract_addresses(feeder_gateway_url).await?;
+
+        Ok(Self {
+            chain_id,
+            strk_fee_token_address: addresses.strk_l2_token_address,
+            eth_fee_token_address: addresses.eth_l2_token_address,
+            is_l3: false, // Set by caller after fetch
```
byteZorvin: Nit: Would prefer the design if we pass `is_l3` as arg, as if user forgets to set we wont be able to catch. Or if we can have testcase that ensures if there is l3 flag passed from env and the user doesnt set the flag this init fails(not sure how that test would look like).

byteZorvin (changes requested):
Looks good

orchestrator/crates/utils/src/chain_details.rs
heemankv: makes sense, I've refactored to ensure we can't have two places setting the value.

byteZorvin (approved):

orchestrator/crates/utils/src/chain_details.rs
```diff
@@ -0,0 +1,225 @@
...
+                        error!("Failed to fetch chain details after timeout");
+                        return Err(eyre!(
+                            "Failed to fetch chain details after {} attempts over {} seconds: {}",
+                            attempt,
+                            elapsed.as_secs(),
+                            e
+                        ));
+                    }
+
+                    info!("Failed to fetch chain details, retrying");
```
Mohiiit, line 136: maybe a warning?

orchestrator/crates/utils/src/chain_details.rs
```diff
@@ -0,0 +1,225 @@
...
+            let addresses = Self::fetch_contract_addresses(feeder_gateway_url).await;
+
+            match (chain_id, addresses) {
+                (Ok(chain_id), Ok(addresses)) => {
+                    debug!(
+                        chain_id = %chain_id,
+                        strk_fee_token = %addresses.strk_l2_token_address,
+                        eth_fee_token = %addresses.eth_l2_token_address,
+                        "Chain details fetched"
+                    );
```
Mohiiit, line 115: nit: maybe the fetching chain details should be debug and chain details fetched should be info?

orchestrator/src/core/config.rs
```diff
@@ -243,25 +249,29 @@ impl Config {
...
         }
 
+        // Fetch chain details from node with retry logic
+        let chain_details =
+            ChainDetails::fetch(&params.madara_rpc_url, &params.madara_feeder_gateway_url, layer.is_l3())
+                .await
+                .map_err(|e| {
+                    OrchestratorError::ConfigError(format!("Failed to fetch chain details from node: {}", e))
+                })?;
+        info!(chain_id = %chain_details.chain_id, is_l3 = %chain_details.is_l3, "Chain details fetched successfully");
```
Mohiiit, line 259: nit: duplicate log! same log exists in the chain_details.rs

Mohiiit (approved):
some nits related to logs, else LGTM!

PR #888, opened 2025-12-05
feat: add support for priority queue

prkpndy:
# Add Priority Queue Mechanism for Job Processing

  ## Pull Request type

  - [ ] Bugfix
  - [x] Feature
  - [ ] Build-related changes
  - [ ] Testing

  All jobs are processed inorder through their respective queues. There's no mechanism to prioritize urgent jobs.

  Resolves: #NA

  ## What is the new behavior?

Implements a priority queue mechanism that allows urgent jobs to "jump the line" while respecting per-job-type concurrency limits.

  ### Key Changes:

  - **New API Parameter**: Jobs can be submitted with `?priority=true` query parameter
    ```bash
    GET /jobs/{id}/process?priority=true
    GET /jobs/{id}/verify?priority=true

  - Priority Injection Architecture:
    - Added shared PriorityJobQueue
    - All 12 job workers check priority queue before their normal queues
    - Priority messages contain {job_id, job_type, action} for routing
    - Workers filter and process only matching messages
  - Concurrency Limits Respected: Priority jobs still respect existing limits (SNOS max 5, others max 10)

  Files Modified:

  - New: orchestrator/src/worker/parser/priority_queue_message.rs
  - Modified: 8 files (queue types, event worker, job service, API routes)

  Does this introduce a breaking change?

  Yes - Internal API signature changes (backward compatible for external users)


Mohiiit (commented):
## Code Review Summary

This PR introduces a priority queue feature. I've identified several issues that need attention:

🚨 **Critical:** Worker starvation bug that must be fixed
⚠️ **Major:** Data consistency concerns
🔍 **Code Quality:** Several DRY violations and error handling improvements needed
🐛 **Additional Issues:** Documentation, metrics, and type safety improvements

See inline comments for details.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -397,13 +434,38 @@ impl EventWorker {
...
-                // Handle shutdown signal
+                // 2. Handle shutdown signal
                 _ = self.cancellation_token.cancelled() => {
                     info!("Shutdown signal received, breaking from main loop");
                     break;
                 }
 
-                // Process new messages (with backpressure)
+                // 3. NEW: Check priority queue BEFORE normal queue
+                priority_result = self.get_priority_message(),
```
Mohiiit: ## 🚨 Critical: Worker Starvation Bug

The `tokio::select!` loop with `biased` mode creates a **critical starvation bug** that prevents normal queue processing:

1. `get_priority_message()` returns immediately (non-blocking) with `Ok(None)` when the priority queue is empty
2. `get_message()` is a blocking call that waits for messages
3. In a `biased` select, if the first branch (`priority_result`) is ready (even with `None`), it is selected
4. The `get_message()` future is dropped/cancelled
5. The loop repeats, `get_priority_message()` returns `None` instantly again
6. **Result:** The worker busy-loops, consuming 100% CPU, and **never** waits for the normal queue

### Solution

Move the priority check **outside** the `tokio::select!` loop to prevent starvation:

```rust
loop {
    // Check if shutdown was requested
    if self.is_shutdown_requested() {
        info!("Shutdown requested, stopping message processing");
        break;
    }

    // 1. Check priority queue first (non-blocking)
    if tasks.len() < max_concurrent_tasks && self.should_check_priority_queue() {
        match self.get_priority_message().await {
            Ok(Some((delivery, priority_msg))) => {
                // Convert priority message to JobQueueMessage for processing
                let job_queue_msg = JobQueueMessage { id: priority_msg.id };
                let parsed_message = ParsedMessage::JobQueue(Box::new(job_queue_msg));

                let worker = self.clone();
                tasks.spawn(async move {
                    worker.process_message(delivery, parsed_message).await
                });
                debug!("Spawned PRIORITY task, active: {}", tasks.len());
                continue; // Process next iteration
            }
            Ok(None) => {
                // No priority messages, fall through to normal queue
            }
            Err(e) => {
                error!("Error processing priority queue: {:?}", e);
                sleep(Duration::from_millis(100)).await;
                continue;
            }
        }
    }

    // 2. Block on normal queue or shutdown (only when no priority messages)
    tokio::select! {
        biased;
        Some(result) = tasks.join_next(), if !tasks.is_empty() => {
            Self::handle_task_result(result);
            // ... rest of handling
        }
        _ = self.cancellation_token.cancelled() => {
            info!("Shutdown signal received, breaking from main loop");
            break;
        }
        message_result = self.get_message(), if tasks.len() < max_concurrent_tasks => {
            // ... process normal queue
        }
    }
}
```

This ensures the normal queue is always checked when priority queue is empty, preventing the busy-loop.

orchestrator/src/worker/parser/priority_queue_message.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::error::ConsumptionError;
+use crate::types::jobs::types::JobType;
+use crate::types::queue::JobAction;
+use crate::worker::traits::message::MessageParser;
+use omniqueue::Delivery;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct PriorityQueueMessage {
```
Mohiiit: ## ⚠️ Data Consistency Issue

The `PriorityQueueMessage` struct duplicates `job_type`:

```rust
pub struct PriorityQueueMessage {
    pub id: Uuid,
    pub job_type: JobType, // <--- Redundant?
    pub action: JobAction,
}
```

However, the worker later fetches the job from the DB (`JobService::get_job`) which contains the **authoritative** `job_type`. This creates a potential inconsistency:

- **Risk:** If a job's state changes in the DB between enqueue and processing, the worker might filter based on stale message data but process based on fresh DB data
- The message's `job_type` is used for routing (in `message_matches_worker`), but the actual processing uses the DB's `job_type`

### Solution Options

**Option 1:** Simplify the message to just `{ id, action }` (if action is needed for routing):
```rust
pub struct PriorityQueueMessage {
    pub id: Uuid,
    pub action: JobAction, // Keep action for routing
}
```
Then in `message_matches_worker`, fetch the job from DB first to check if it matches.

**Option 2:** Keep current structure but document that `job_type` is for routing optimization only, and the DB is the source of truth. Add validation in `message_matches_worker` to verify consistency.

**Recommendation:** Option 1 is cleaner, but Option 2 is acceptable if you want to avoid DB hits for wrong workers (performance trade-off).

orchestrator/src/worker/traits/message.rs
```diff
@@ -1,12 +1,14 @@
...
 use crate::worker::parser::job_queue_message::JobQueueMessage;
+use crate::worker::parser::priority_queue_message::PriorityQueueMessage;
 use crate::worker::parser::worker_trigger_message::WorkerTriggerMessage;
 use omniqueue::Delivery;
 
 #[derive(Debug, Clone)]
 pub enum ParsedMessage {
     WorkerTrigger(Box<WorkerTriggerMessage>),
     JobQueue(Box<JobQueueMessage>),
+    PriorityJobQueue(Box<PriorityQueueMessage>),
```
Mohiiit: ## 🔍 Unnecessary Enum Variant

`ParsedMessage::PriorityJobQueue` is defined but effectively unused. The code immediately converts it to `JobQueueMessage` inside the worker loop:

```rust
// In get_priority_message():
let parsed_msg = PriorityQueueMessage::parse_message(&delivery)?;

// Then immediately converted:
let job_queue_msg = JobQueueMessage { id: priority_msg.id };
let parsed_message = ParsedMessage::JobQueue(Box::new(job_queue_msg));
```

The `PriorityJobQueue` variant only appears in error handling spans, which is unnecessary complexity.

### Solution

Remove the `PriorityJobQueue` variant entirely. Parse the priority message, extract the ID, and directly construct a `JobQueueMessage`:

```rust
// In get_priority_message(), return:
Ok(Some((delivery, priority_msg.id))) // Just return the ID

// In the run loop:
match self.get_priority_message().await {
    Ok(Some((delivery, job_id))) => {
        let job_queue_msg = JobQueueMessage { id: job_id };
        let parsed_message = ParsedMessage::JobQueue(Box::new(job_queue_msg));
        // ... rest of processing
    }
    // ...
}
```

This simplifies the code and removes dead code.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -434,6 +496,120 @@ impl EventWorker {
...
+            Err(e) => {
+                // Log error but don't fail - just skip priority check this iteration
+                debug!("Priority queue check failed (will retry): {:?}", e);
+                Ok(None)
+            }
+        }
+    }
+
+    /// Checks if a priority message matches this worker's responsibility
+    fn message_matches_worker(&self, msg: &PriorityQueueMessage) -> bool {
```
Mohiiit: ## 🔍 DRY Violation: Repetitive Matching Logic

The `message_matches_worker` function is a long, repetitive `match` statement (60+ lines) that repeats the same pattern:

```rust
QueueType::SnosJobProcessing => msg.job_type == JobType::SnosRun && msg.action == JobAction::Process,
QueueType::SnosJobVerification => msg.job_type == JobType::SnosRun && msg.action == JobAction::Verify,
// ... repeated 12 times
```

This violates DRY (Don't Repeat Yourself) and makes the code harder to maintain.

### Solution

Implement helper methods on `QueueType` to extract the target job type and action:

```rust
// In types/queue.rs
impl QueueType {
    /// Returns the job type this queue processes, if applicable
    pub fn target_job_type(&self) -> Option<JobType> {
        match self {
            Self::SnosJobProcessing | Self::SnosJobVerification => Some(JobType::SnosRun),
            Self::ProvingJobProcessing | Self::ProvingJobVerification => Some(JobType::ProofCreation),
            Self::ProofRegistrationJobProcessing | Self::ProofRegistrationJobVerification => {
                Some(JobType::ProofRegistration)
            }
            Self::DataSubmissionJobProcessing | Self::DataSubmissionJobVerification => {
                Some(JobType::DataSubmission)
            }
            Self::UpdateStateJobProcessing | Self::UpdateStateJobVerification => {
                Some(JobType::StateTransition)
            }
            Self::AggregatorJobProcessing | Self::AggregatorJobVerification => Some(JobType::Aggregator),
            _ => None,
        }
    }

    /// Returns the action this queue performs, if applicable
    pub fn target_action(&self) -> Option<JobAction> {
        match self {
            Self::SnosJobProcessing
            | Self::ProvingJobProcessing
            | Self::ProofRegistrationJobProcessing
            | Self::DataSubmissionJobProcessing
            | Self::UpdateStateJobProcessing
            | Self::AggregatorJobProcessing => Some(JobAction::Process),
            Self::SnosJobVerification
            | Self::ProvingJobVerification
            | Self::ProofRegistrationJobVerification
            | Self::DataSubmissionJobVerification
            | Self::UpdateStateJobVerification
            | Self::AggregatorJobVerification => Some(JobAction::Verify),
            _ => None,
        }
    }
}

// In worker
fn message_matches_worker(&self, msg: &PriorityQueueMessage) -> bool {
    let Some(target_type) = self.queue_type.target_job_type() else { return false; };
    let Some(target_action) = self.queue_type.target_action() else { return false; };
    
    msg.job_type == target_type && msg.action == target_action
}
```

This reduces the function from 60+ lines to 4 lines and makes it easier to maintain.

orchestrator/src/server/route/jobs.rs
```diff
@@ -38,25 +39,49 @@ use crate::worker::service::JobService;
...
-            Err(JobRouteError::ProcessingError(e.to_string()))
+            ORCHESTRATOR_METRICS.failed_job_operations.add(
+                1.0,
+                &[
+                    KeyValue::new("operation_type", "queue_process"),
+                    KeyValue::new("priority", if query.priority { "true" } else { "false" }),
+                ],
+            );
+            // Map PriorityQueueFull to QueueCapacityExceeded for proper HTTP status code
+            match e {
```
Mohiiit: ## 🔍 Error Handling: Manual Error Mapping

The manual mapping of `JobError::PriorityQueueFull` to `JobRouteError` inside the route handler is repetitive and appears in both `handle_process_job_request` and `handle_verify_job_request`:

```rust
match e {
    crate::error::job::JobError::PriorityQueueFull { current_size, max_size } => {
        Err(JobRouteError::QueueCapacityExceeded(format!(
            "Priority queue is full (current: {}, max: {}). Please try again later or use normal queue.",
            current_size, max_size
        )))
    }
    _ => Err(JobRouteError::ProcessingError(e.to_string())),
}
```

### Solution

Implement `From<JobError> for JobRouteError` to centralize this conversion:

```rust
// In orchestrator/src/server/error.rs
impl From<JobError> for JobRouteError {
    fn from(err: JobError) -> Self {
        match err {
            JobError::PriorityQueueFull { current_size, max_size } => {
                JobRouteError::QueueCapacityExceeded(format!(
                    "Priority queue is full (current: {}, max: {}). Please try again later or use normal queue.",
                    current_size, max_size
                ))
            }
            _ => JobRouteError::ProcessingError(err.to_string()),
        }
    }
}

// Then in route handlers, simplify to:
match JobService::queue_job_for_processing(job_id, config.clone(), query.priority).await {
    Ok(_) => { /* success */ }
    Err(e) => Err(e.into()), // Automatic conversion
}
```

This follows Rust idioms and reduces duplication.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -434,6 +496,120 @@ impl EventWorker {
...
+                                warn!("Failed to nack non-matching priority message: {:?}", e);
+                            }
+
+                            Ok(None)
+                        }
+                    }
+                    Err(e) => {
+                        error!("Failed to parse priority message: {:?}", e);
+                        // Nack the malformed message
+                        let _ = delivery.nack().await;
```
Mohiiit: ## 🐛 Silent Failure in Message Parsing Error Handling

When a priority message fails to parse, the code nacks it but silently ignores the nack error:

```rust
Err(e) => {
    error!("Failed to parse priority message: {:?}", e);
    // Nack the malformed message
    let _ = delivery.nack().await;  // ← Silent failure
    Ok(None)
}
```

### Impact

If nack fails, the malformed message stays in the queue and will be consumed again, potentially causing an infinite retry loop on corrupt messages. This wastes resources and could cause the worker to get stuck processing the same bad message repeatedly.

### Solution

Log nack failures and consider moving the message to a DLQ if available:

```rust
Err(e) => {
    error!("Failed to parse priority message: {:?}", e);
    // Nack the malformed message
    if let Err(nack_err) = delivery.nack().await {
        error!(
            "CRITICAL: Failed to nack malformed priority message: {:?}. Message may be stuck in queue.",
            nack_err
        );
        // Consider: Return error to escalate, or implement retry counter
    }
    Ok(None)
}
```

Alternatively, track parse failures per message and move to DLQ after N failures.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -434,6 +496,120 @@ impl EventWorker {
...
+                            Ok(Some((delivery, *parsed_msg)))
+                        } else {
+                            // Message doesn't match - nack it to release for other workers
+                            debug!(
+                                "Priority message doesn't match, releasing: job_type={:?}, action={:?}",
+                                parsed_msg.job_type, parsed_msg.action
+                            );
+
+                            // Nack the message to make it available for other workers
+                            if let Err(e) = delivery.nack().await {
```
Mohiiit: ## 🐛 Nack Error Handling Inconsistency

There are two different patterns for handling nack errors in the same function:

1. **Non-matching message** (line ~535): `if let Err(e) = delivery.nack().await { warn!(...) }`
2. **Parse error** (line ~545): `let _ = delivery.nack().await;` (silent)

### Impact

Inconsistent error handling makes it harder to debug issues and could hide critical failures. If nacking fails, it's a serious issue that should be logged consistently.

### Solution

Use consistent error handling throughout:

```rust
// For non-matching messages:
if let Err(e) = delivery.nack().await {
    warn!("Failed to nack non-matching priority message: {:?}", e);
}

// For parse errors:
if let Err(e) = delivery.nack().await {
    error!("CRITICAL: Failed to nack malformed priority message: {:?}", e);
}
```

Or better yet, extract to a helper function:

```rust
async fn nack_with_logging(delivery: &Delivery, context: &str) -> Result<(), ConsumptionError> {
    delivery.nack().await.map_err(|e| {
        error!("Failed to nack priority message ({}): {:?}", context, e);
        ConsumptionError::FailedToAcknowledgeMessage(e.0.to_string())
    })
}
```

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -397,13 +434,38 @@ impl EventWorker {
...
+                            tasks.spawn(async move {
+                                worker.process_message(delivery, parsed_message).await
+                            });
+                            debug!("Spawned PRIORITY task, active: {}", tasks.len());
+                        }
+                        Ok(None) => {
+                            // No matching priority messages, will check normal queue
+                        }
+                        Err(e) => {
+                            error!("Error processing priority queue: {:?}", e);
```
Mohiiit: ## 🐛 Priority Queue Check Error Swallows Real Issues

When `get_priority_message()` returns an error (not `NoData`), the code logs an error message and sleeps 100ms, but the error is swallowed:

```rust
Err(e) => {
    error!("Error processing priority queue: {:?}", e);
    sleep(Duration::from_millis(100)).await;
}
```

### Impact

- **Transient errors** (network hiccups) will retry, which is fine
- **Persistent errors** (permissions, queue misconfiguration, queue doesn't exist) will cause constant error logging without surfacing as a critical failure
- No way to distinguish between transient and persistent failures
- Could mask configuration issues in production

### Solution

Track consecutive errors and escalate after a threshold:

```rust
let mut consecutive_priority_errors = 0;
const MAX_CONSECUTIVE_ERRORS: usize = 10;

// In the loop:
Err(e) => {
    consecutive_priority_errors += 1;
    error!(
        "Error processing priority queue (consecutive: {}): {:?}",
        consecutive_priority_errors, e
    );
    
    if consecutive_priority_errors >= MAX_CONSECUTIVE_ERRORS {
        error!(
            "CRITICAL: Priority queue has failed {} consecutive times. This may indicate a configuration issue.",
            MAX_CONSECUTIVE_ERRORS
        );
        // Consider: Return error to stop worker, or send alert
    }
    
    sleep(Duration::from_millis(100)).await;
}

// Reset counter on success:
Ok(Some(_)) | Ok(None) => {
    consecutive_priority_errors = 0;
}
```

This helps distinguish between transient issues and persistent configuration problems.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -434,6 +496,120 @@ impl EventWorker {
         Ok(())
     }
 
+    /// Determines if this worker should check the priority queue
+    /// Only job processing and verification workers check priority queue
+    /// WorkerTrigger and JobHandleFailure workers do not
+    fn should_check_priority_queue(&self) -> bool {
```
Mohiiit: ## 🐛 Missing Documentation

The function excludes `PriorityJobQueue` itself from checking the priority queue, but the reasoning isn't documented:

```rust
fn should_check_priority_queue(&self) -> bool {
    !matches!(self.queue_type, QueueType::WorkerTrigger | QueueType::JobHandleFailure | QueueType::PriorityJobQueue)
}
```

This makes it unclear if it's preventing infinite recursion or if there's another reason.

### Solution

Add a doc comment explaining the reasoning:

```rust
/// Determines if this worker should check the priority queue.
///
/// Only job processing and verification workers check the priority queue.
/// WorkerTrigger and JobHandleFailure workers do not, as they handle special
/// system-level operations.
///
/// PriorityJobQueue workers are excluded to prevent circular consumption:
/// a PriorityJobQueue worker checking the PriorityJobQueue would create
/// an infinite loop where priority messages are consumed by the wrong worker type.
fn should_check_priority_queue(&self) -> bool {
    !matches!(self.queue_type, QueueType::WorkerTrigger | QueueType::JobHandleFailure | QueueType::PriorityJobQueue)
}
```

orchestrator/src/worker/service.rs
```diff
@@ -70,22 +72,69 @@ impl JobService {
...
 
+    /// Adds a job to the priority queue
+    async fn add_job_to_priority_queue(
+        id: Uuid,
+        job_type: JobType,
+        action: JobAction,
+        config: Arc<Config>,
+    ) -> Result<(), JobError> {
+        // Check priority queue depth before adding
+        let queue_depth = config.queue().get_queue_depth(QueueType::PriorityJobQueue).await?;
```
Mohiiit: ## 🐛 Inconsistent Error Propagation

If `get_queue_depth` fails, it propagates as a generic error, making it hard to distinguish between:
- "Queue depth exceeded" (expected, user-facing error)
- "Failed to check queue depth" (system error, should be logged differently)

### Solution

Add explicit error handling to distinguish between these cases:

```rust
let queue_depth = match config.queue().get_queue_depth(QueueType::PriorityJobQueue).await {
    Ok(depth) => depth,
    Err(e) => {
        error!(
            "Failed to check priority queue depth: {:?}. Allowing job addition but this should be investigated.",
            e
        );
        // Option 1: Fail fast
        return Err(JobError::Other(format!("Failed to check queue depth: {}", e)));
        // Option 2: Allow but log (if queue depth check is best-effort)
        // 0 // Assume empty if check fails
    }
};

if queue_depth >= max_size {
    return Err(JobError::PriorityQueueFull { current_size: queue_depth, max_size });
}
```

This provides better observability and distinguishes between user-facing and system errors.

orchestrator/src/core/client/queue/mod.rs
```diff
@@ -26,4 +26,11 @@ pub trait QueueClient: Send + Sync {
     /// * `Ok(())` - If the queue service is healthy and accessible
     /// * `Err(QueueError)` - If the health check fails
     async fn health_check(&self) -> Result<(), QueueError>;
+
+    /// Get the approximate number of messages in a queue
+    ///
+    /// # Returns
+    /// * `Ok(i32)` - Approximate number of messages in the queue
+    /// * `Err(QueueError)` - If the operation fails
+    async fn get_queue_depth(&self, queue: QueueType) -> Result<i32, QueueError>;
```
Mohiiit: ## 🐛 Type Safety: i32 for Queue Depth

Queue depth is returned as `i32`:

```rust
async fn get_queue_depth(&self, queue: QueueType) -> Result<i32, QueueError>;
```

**Issues:**
- SQS queue depths can exceed `i32::MAX` (2.1 billion) in extreme cases
- `i32` allows negative values, which don't make sense for queue depth
- SQS `ApproximateNumberOfMessages` returns a string that's parsed to `i32`, but could theoretically be larger

### Impact

Potential overflow/underflow issues, though unlikely for a priority queue with max size 20. However, for consistency and future-proofing, this is a type safety concern.

### Solution

Use `u64` or `usize` for queue depth:

```rust
// In queue/mod.rs
async fn get_queue_depth(&self, queue: QueueType) -> Result<u64, QueueError>;

// In queue_control.rs
pub static MAX_PRIORITY_QUEUE_SIZE: LazyLock<u64> = LazyLock::new(|| {
    get_env_var_or_default("MADARA_ORCHESTRATOR_MAX_PRIORITY_QUEUE_SIZE", "20")
        .parse()
        .expect("MADARA_ORCHESTRATOR_MAX_PRIORITY_QUEUE_SIZE must be a valid integer")
});

// In sqs.rs
let count = attributes
    .attributes()
    .and_then(|attrs| attrs.get(&QueueAttributeName::ApproximateNumberOfMessages))
    .and_then(|value| value.parse::<u64>().ok()) // Changed to u64
    .unwrap_or(0);
```

This provides better type safety and prevents negative values.

orchestrator/src/types/queue_control.rs
```diff
@@ -173,5 +181,14 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
             supported_layers: vec![Layer::L2],
         },
     );
+    map.insert(
+        QueueType::PriorityJobQueue,
```
Mohiiit: ## 🐛 Visibility Timeout Risk with Message Bouncing

The PriorityJobQueue has a `visibility_timeout: 30` seconds. When a worker nacks a non-matching message:

1. Message becomes visible again immediately (visibility timeout is 0 on nack)
2. If no worker of the correct type is available, the message will bounce between wrong workers
3. If a message type has **NO active workers** (e.g., all DataSubmission workers are down), the message could bounce indefinitely consuming resources

### Impact

- **Resource waste:** Messages bouncing between workers consume CPU and SQS API calls
- **Potential message loss:** If visibility timeout handling is inconsistent, messages could get stuck
- **No circuit breaker:** No mechanism to detect when a message type has no active workers

### Solution

**Option 1:** Document this behavior and add monitoring:
```rust
// Add comment in queue_control.rs
// PriorityJobQueue visibility timeout is set to 30 seconds.
// Note: When messages are nacked due to non-matching worker types,
// they become visible immediately. If no matching worker is available,
// messages will bounce until a matching worker comes online or visibility
// timeout expires. Consider implementing a max-bounce counter or DLQ
// routing for messages that bounce too many times.
```

**Option 2:** Implement max-bounce counter:
```rust
// Add to PriorityQueueMessage:
pub struct PriorityQueueMessage {
    pub id: Uuid,
    pub job_type: JobType,
    pub action: JobAction,
    pub bounce_count: Option<u32>, // Track bounces
}

// In get_priority_message, when nacking:
if bounce_count.unwrap_or(0) >= MAX_BOUNCES {
    // Move to DLQ or log critical error
    error!("Message {} has bounced {} times, no matching worker available", id, bounce_count);
}
```

**Option 3:** Add alerting for messages that exceed visibility timeout multiple times.

Recommendation: Start with Option 1 (documentation + monitoring), then add Option 2 if bouncing becomes a problem.

prkpndy (commented):
this type is used in process_message, post_processing, etc. for matching the message. I don't think that we can remove the type. Or am I missing something?

orchestrator/src/worker/traits/message.rs
prkpndy: this enum variant is being used in process_message, post_processing, etc. I don't think it's redundant. unless I am misssing something 🤔 ?

orchestrator/src/worker/parser/priority_queue_message.rs
prkpndy: don't think this is an issue. fetching the job type from db would add additional time

orchestrator/src/worker/controller/event_worker.rs
prkpndy: not sure if this is relevant. we are following the same pattern in other queues as well!

orchestrator/src/worker/controller/event_worker.rs
prkpndy: i think it's self explanatory that we don't want to move dlq and worker trigger messages to priority queues

orchestrator/src/worker/service.rs
prkpndy: not sure what this will achieve practically?

orchestrator/src/types/queue_control.rs
prkpndy: increased it to 300 seconds. this is the max time we have given for any job, so ideally we should be able to complete the job in that time

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -434,6 +502,78 @@ impl EventWorker {
...
+                        let matches = self.message_matches_worker(&parsed_msg);
+
+                        if matches {
+                            debug!(
+                                "Priority message matched: job_type={:?}, action={:?}",
+                                parsed_msg.job_type, parsed_msg.action
+                            );
+                            Ok(Some((delivery, *parsed_msg)))
+                        } else {
+                            // Message doesn't match - nack it to release for other workers
```
byteZorvin: When a worker nacks a non-matching message, it becomes immediately visible to all workers again. With multiple worker types polling simultaneously:

1. SNOS worker gets ProofCreation msg → nacks
2. Proving worker gets SnosRun msg → nacks  
3. Repeat until lucky match

This wastes API calls and adds unpredictable latency.

orchestrator/src/types/queue_control.rs
```diff
@@ -4,6 +4,14 @@ use orchestrator_utils::env_utils::get_env_var_or_default;
 use std::collections::HashMap;
 use std::sync::LazyLock;
 
+/// Maximum number of messages allowed in the priority queue
+/// Can be configured via MADARA_ORCHESTRATOR_MAX_PRIORITY_QUEUE_SIZE environment variable
+pub static MAX_PRIORITY_QUEUE_SIZE: LazyLock<usize> = LazyLock::new(|| {
+    get_env_var_or_default("MADARA_ORCHESTRATOR_MAX_PRIORITY_QUEUE_SIZE", "20")
```
byteZorvin: This design does not take values from clap, we already have similar tech debt for queue initializaiton.

orchestrator/src/types/queue.rs
```diff
@@ -38,6 +45,8 @@ pub enum QueueType {
     JobHandleFailure,
     #[strum(serialize = "worker_trigger")]
     WorkerTrigger,
+    #[strum(serialize = "priority_job_queue")]
+    PriorityJobQueue,
```
byteZorvin: This should be added to get_l2_queues and get_l3_queues

byteZorvin (commented):
We can have a test for the setup resources test that priority queues are also built

byteZorvin (commented):
We can have a testcase that priority jobs are being picked first than other jobs

orchestrator/src/types/queue_control.rs
```diff
@@ -173,5 +181,14 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
             supported_layers: vec![Layer::L2],
         },
     );
+    map.insert(
+        QueueType::PriorityJobQueue,
+        QueueConfig {
+            visibility_timeout: 300,
+            dlq_config: None, // No DLQ - failed priority jobs go through normal error handling
```
byteZorvin: With `dlq_config: None`, a malformed message that fails parsing will be nacked and redelivered indefinitely, consuming resources.

Consider adding: `dlq_config: Some(DlqConfig { max_receive_count: 5 })`

byteZorvin (changes requested):
GG work !! Easy read too.
Mostly a few tests and we are good.

orchestrator/src/server/service/admin.rs
```diff
@@ -0,0 +1,129 @@
...
+        job_types: Vec<JobType>,
+        config: Arc<Config>,
+    ) -> Result<BulkJobResult, JobError> {
+        Self::process_jobs_by_status(
+            JobStatus::Failed,
+            job_types,
+            config.clone(),
+            "retry failed",
+            // Use normal queue (non-priority) for bulk retry operations
+            |id, _, cfg| async move { JobHandlerService::retry_job(id, cfg, false).await },
```
byteZorvin, line 86: Nit: We should retry jobs with priority only, but since orch does not pick up jobs anyways if there are failed jobs in db, it gives the same effect

byteZorvin: If in future we process jobs while having failed jobs in db, we would need to update this

orchestrator/src/types/queue_control.rs
```diff
@@ -12,6 +12,31 @@ pub static MAX_PRIORITY_QUEUE_SIZE: LazyLock<usize> = LazyLock::new(|| {
         .expect("MADARA_ORCHESTRATOR_MAX_PRIORITY_QUEUE_SIZE must be a valid integer")
 });
 
+/// Maximum time (in seconds) to wait for the priority slot to become empty.
+/// Can be configured via MADARA_ORCHESTRATOR_PRIORITY_SLOT_WAIT_TIMEOUT environment variable
+pub static PRIORITY_SLOT_WAIT_TIMEOUT_SECS: LazyLock<u64> = LazyLock::new(|| {
+    get_env_var_or_default("MADARA_ORCHESTRATOR_PRIORITY_SLOT_WAIT_TIMEOUT", "300")
```
byteZorvin: Can we not move this to clap?

orchestrator/src/types/queue_control.rs
```diff
@@ -12,6 +12,31 @@ pub static MAX_PRIORITY_QUEUE_SIZE: LazyLock<usize> = LazyLock::new(|| {
...
+pub static PRIORITY_SLOT_STALENESS_TIMEOUT_SECS: LazyLock<u64> = LazyLock::new(|| {
+    get_env_var_or_default("MADARA_ORCHESTRATOR_PRIORITY_SLOT_STALENESS_TIMEOUT", "300")
+        .parse()
+        .expect("MADARA_ORCHESTRATOR_PRIORITY_SLOT_STALENESS_TIMEOUT must be a valid integer")
+});
+
+/// Interval (in milliseconds) between priority slot availability checks.
+/// Can be configured via MADARA_ORCHESTRATOR_PRIORITY_SLOT_CHECK_INTERVAL_MS environment variable
+pub static PRIORITY_SLOT_CHECK_INTERVAL_MS: LazyLock<u64> = LazyLock::new(|| {
+    get_env_var_or_default("MADARA_ORCHESTRATOR_PRIORITY_SLOT_CHECK_INTERVAL_MS", "1000")
```
byteZorvin: this too

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -0,0 +1,288 @@
...
+    async fn wait_for_slot_empty(&self) -> bool {
+        let timeout = Duration::from_secs(*PRIORITY_SLOT_WAIT_TIMEOUT_SECS);
+        let staleness_secs = *PRIORITY_SLOT_STALENESS_TIMEOUT_SECS;
+        let check_interval = Duration::from_millis(*PRIORITY_SLOT_CHECK_INTERVAL_MS);
+        let start = Instant::now();
+
+        loop {
+            if self.is_shutdown_requested() {
+                return false;
+            }
```
byteZorvin, line 161: I guess this is not needed, as the run function already has it. Although I am fine to be present here too

orchestrator/src/types/priority_slot.rs
```diff
@@ -0,0 +1,152 @@
+//! Priority Job Slot Module
```
byteZorvin, line 1: there is bit of duplication here, we can improve the logic. One option is an enum
Say
```rust
//! Priority Job Slot Module
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum SlotType {
    Processing,
    Verification,
}

impl SlotType {
    fn get_slot(&self) -> &'static tokio::sync::Mutex<Option<PriorityJobSlot>> {
        match self {
            SlotType::Processing => &*PRIORITY_PROCESSING_SLOT,
            SlotType::Verification => &*PRIORITY_VERIFICATION_SLOT,
        }
    }
}

pub async fn take_if_matches(slot_type: SlotType, job_type: &JobType) -> Option<PriorityJobSlot> { ... }
pub async fn place(slot_type: SlotType, job: PriorityJobSlot) -> bool { ... }
pub async fn is_empty(slot_type: SlotType) -> bool { ... }
pub async fn clear_stale(slot_type: SlotType, max_age_secs: u64) -> Option<Delivery> { ... }
```

byteZorvin (approved):
Looks good, few nits. Can be ignored

orchestrator/src/server/service/admin.rs
prkpndy: didn't get your point properly but, priority queue have a max size. bulk can fill that size very quickly. so we should only call individual jobs for priority processing

orchestrator/src/types/priority_slot.rs
prkpndy: can do but it'll complicate the logic
also I have tested the current implementation so don't want to change the core logic 😅 
would prefer to keep it as it is if you are okay

Mohiiit (commented):
Review comments

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -1266,6 +1270,56 @@ impl DatabaseClient for MongoDbClient {
...
+        job_statuses: Vec<JobStatus>,
+    ) -> Result<Option<JobItem>, DatabaseError> {
+        let start = Instant::now();
+        let pipeline = vec![
+            doc! {
+                "$match": {
+                    "job_type": bson::to_bson(&job_type)?
+                }
+            },
+            doc! {
```
Mohiiit, line 1285: 🚨 **Critical MongoDB Query Bug**

The `$match` stage is missing the `"$match"` key. The status filter should be wrapped in a `$match` stage. Currently it's:
```rust
doc! {
      "status": {
    "$nin": job_statuses.iter()...
},
    },
```
Should be:
```rust
doc! {
    "$match": {
        "status": {
            "$nin": job_statuses.iter().map(|status| bson::to_bson(status).unwrap_or(Bson::Null)).collect::<Vec<Bson>>()
        }
    }
},
```
This will cause a runtime error when the query executes. Consider adding a test to verify this function works correctly with various job statuses.

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -0,0 +1,263 @@
...
+        // Create slot entry and place in slot
+        let slot_job = PriorityJobSlot::new(parsed_msg.id, job.job_type, delivery);
+
+        if self.place_in_slot(slot_job).await {
+            debug!("PQ Worker ({}): Placed job {} in slot", self.action, parsed_msg.id);
+        } else {
+            // This shouldn't happen since we waited for slot to be empty,
+            // but if there's a race condition, log it. The delivery is now
+            // consumed by the failed placement attempt.
+            error!("PQ Worker ({}): Race condition - failed to place job {} in slot", self.action, parsed_msg.id);
```
Mohiiit, line 273: **Race Condition in Slot Placement - Clarification Needed**

We noticed the error logging for a potential race condition when `place_in_slot()` returns `false` after `wait_for_slot_empty()` returned `true`. Our understanding is that there's only one PriorityQueueWorker per action type (Processing/Verification) per orchestrator instance, and slot operations are protected by `tokio::sync::Mutex`, so this race condition shouldn't occur between PriorityQueueWorkers. Could you clarify:

1. Is our understanding correct that there's only one PriorityQueueWorker per action type per instance?
2. Under what scenario would `place_in_slot()` return `false` after `wait_for_slot_empty()` returned `true`?
3. Is the error logging at line 258 defensive programming for an edge case, or is there a specific scenario we should be aware of?

orchestrator/src/worker/event_handler/triggers/mod.rs
```diff
@@ -39,7 +39,7 @@ pub trait JobTrigger: Send + Sync {
     async fn is_worker_enabled(&self, config: Arc<Config>) -> color_eyre::Result<bool> {
         let failed_jobs = config
             .database()
-            .get_jobs_by_types_and_statuses(vec![], vec![JobStatus::Failed, JobStatus::VerificationTimeout], Some(1))
+            .get_jobs_by_types_and_statuses(vec![], vec![JobStatus::Failed, JobStatus::VerificationTimeout], None)
```
Mohiiit, line 42: **Empty Job Types Vector Behavior**

The call to `get_jobs_by_types_and_statuses(vec![], vec![JobStatus::Failed, JobStatus::VerificationTimeout], None)` passes an empty vector for job types. Based on the implementation, an empty vector means "all job types" (no filter is added), so this will return ALL failed/verification timeout jobs across all types, not 0. Is this the intended behavior? If so, consider adding a comment to clarify that an empty vector means "all types".

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -51,8 +51,53 @@ impl JobTrigger for SnosJobTrigger {
             error!(error = %e, "Failed to heal orphaned SNOS jobs, continuing with normal processing");
         }
 
+        // Get the target buffer size from config (configurable via env, defaults to 50)
+        let snos_job_buffer_size = config.service_config().snos_job_buffer_size;
+
+        // Calculate the maximum number of snos jobs to create
```
Mohiiit: **SNOS Job Calculation Logic Refactoring**

The nested if-let logic for calculating `max_jobs_to_create` (lines 57-93) is complex and could benefit from extraction into a separate function. This would:
- Make the code easier to follow
- Allow for unit testing of the calculation logic
- Follow more idiomatic Rust patterns

Consider extracting this into a function like `calculate_max_snos_jobs_to_create()` with appropriate tests.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -409,6 +457,36 @@ impl EventWorker {
...
 
+    /// Check if there's a priority job in the global slot that matches this worker.
+    ///
+    /// Only job processing and verification workers check the priority slot.
+    /// System queues (WorkerTrigger, JobHandleFailure, PriorityQueues) do not.
+    ///
+    /// # Returns
+    /// * `Some(PriorityJobSlot)` if a matching priority job was found and taken
+    /// * `None` if no matching job or this worker shouldn't check the slot
+    async fn check_priority_slot(&self) -> Option<PriorityJobSlot> {
```
Mohiiit, line 469: **Priority Slot Checking Logic Clarification**

The `check_priority_slot()` function excludes `PriorityProcessingQueue` and `PriorityVerificationQueue` from checking the slot, which makes sense since those are the queues that place messages INTO the slot. The regular job processing/verification workers (like `SnosJobProcessing`, `ProvingJobProcessing`, etc.) check and take messages FROM the slot. Could you confirm our understanding is correct? Also, are there tests covering the `check_priority_slot()` logic, especially the filtering logic for which workers should check the slot?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -356,19 +371,49 @@ impl EventWorker {
...
                 _ = self.cancellation_token.cancelled() => {
                     info!("Shutdown signal received, breaking from main loop");
                     break;
                 }
 
-                // Process new messages (with backpressure)
+                // 3. Check priority slot - pattern only matches if there's a message
+                // If check_priority_slot() returns None, this branch is skipped
+                // and select proceeds to normal queue. This prevents starvation.
+                Some(priority_slot) = self.check_priority_slot(), if tasks.len() < max_concurrent_tasks => {
```
Mohiiit, line 384: **Priority Slot Handling Code Extraction**

The priority slot handling code (lines 383-407) could be extracted into a separate function like `handle_priority_slot_message()` to improve readability and make the main `run()` loop cleaner. This would also make it easier to test the priority slot handling logic independently.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -384,10 +429,13 @@ impl EventWorker {
...
-                                tasks.spawn(async move {
-                                    worker.process_message(message, parsed_message).await
-                                });
+                                        let worker = self.clone();
+                                        tasks.spawn(async move {
+                                            worker.process_message(message, parsed_message).await
+                                        });
+                                    }
+                                },
+                                None => sleep(Duration::from_secs(1)).await,
```
Mohiiit: **Hardcoded Sleep Duration**

The `sleep(Duration::from_secs(1))` should be extracted into a named constant (e.g., `const NO_MESSAGE_SLEEP_DURATION: Duration = Duration::from_secs(1);`) for better maintainability and clarity.

orchestrator/src/worker/controller/worker_controller.rs
```diff
@@ -55,6 +56,37 @@ impl WorkerController {
             Layer::L3 => Self::get_l3_queues(),
         };
         let mut worker_set = tokio::task::JoinSet::new();
+
+        // Spawn the dedicated Priority Queue Workers (one for processing, one for verification)
```
Mohiiit, line 60: **DRY Violation in Worker Spawning**

The code for spawning PriorityQueueWorkers (Processing and Verification) follows the same pattern and could be refactored to follow DRY principles. Consider extracting a helper function like `spawn_priority_queue_worker(action: PriorityAction)` to reduce duplication.

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -0,0 +1,263 @@
...
+                // Timeout or shutdown, continue to check shutdown and retry
+                continue;
+            }
+
+            // Try to get a message from the queue
+            match self.get_message().await {
+                Ok(Some(delivery)) => {
+                    if let Err(e) = self.process_message(delivery).await {
+                        error!("PQ Worker ({}): Error processing message: {:?}", self.action, e);
+                        sleep(Duration::from_millis(100)).await;
```
Mohiiit: **Hardcoded Sleep Durations**

Multiple hardcoded sleep durations (`Duration::from_millis(100)` and `Duration::from_millis(500)`) should be extracted into named constants for better maintainability. Consider:
```rust
const PQ_WORKER_ERROR_SLEEP_MS: u64 = 100;
const PQ_WORKER_NO_MESSAGE_SLEEP_MS: u64 = 100;
const PQ_WORKER_QUEUE_ERROR_SLEEP_MS: u64 = 500;
```

orchestrator/src/server/route/jobs.rs
```diff
@@ -38,25 +39,40 @@ use crate::worker::service::JobService;
...
-            ORCHESTRATOR_METRICS.failed_job_operations.add(1.0, &[KeyValue::new("operation_type", "queue_process")]);
-            Err(JobRouteError::ProcessingError(e.to_string()))
+            ORCHESTRATOR_METRICS.failed_job_operations.add(
+                1.0,
+                &[
+                    KeyValue::new("operation_type", "queue_process"),
+                    KeyValue::new("priority", if query.priority { "true" } else { "false" }),
+                ],
+            );
+            Err(e.into())
```
Mohiiit, line 75: **Error Conversion Improvement**

nit: error conversion is changed here and it's different than other functions, is this one better than the rest?

orchestrator/src/server/route/jobs.rs
```diff
@@ -82,23 +99,40 @@ async fn handle_process_job_request(
...
-            ORCHESTRATOR_METRICS.failed_job_operations.add(1.0, &[KeyValue::new("operation_type", "queue_verify")]);
-            Err(JobRouteError::ProcessingError(e.to_string()))
+            ORCHESTRATOR_METRICS.failed_job_operations.add(
+                1.0,
+                &[
+                    KeyValue::new("operation_type", "queue_verify"),
+                    KeyValue::new("priority", if query.priority { "true" } else { "false" }),
+                ],
+            );
+            Err(e.into())
```
Mohiiit, line 135: **Error Conversion Improvement**

ditto as above

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -443,4 +443,26 @@ impl QueueClient for SQS {
...
+        let attributes = self
+            .inner
+            .client()
+            .get_queue_attributes()
+            .queue_url(&queue_url)
+            .attribute_names(QueueAttributeName::ApproximateNumberOfMessages)
+            .send()
+            .await?;
+
+        let count = attributes
```
Mohiiit, line 460: **Queue Depth Return Value Clarification**

The `unwrap_or(0)` will return `0` in multiple scenarios:
- Queue is actually empty (expected)
- Attribute doesn't exist (error case)
- Parsing fails (error case)

Could you clarify: Should we distinguish between "queue is empty" vs "error occurred"? Or is it acceptable to treat all cases as "empty" for the capacity check? If errors should be handled differently, consider returning a `Result` or logging a warning when parsing fails.

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -0,0 +1,263 @@
...
+                    error!("PQ Worker ({}): Failed to NACK stale delivery: {:?}", self.action, e);
+                }
+            }
+
+            // Check if slot is empty
+            if self.is_slot_empty().await {
+                return true;
+            }
+
+            sleep(check_interval).await;
```
Mohiiit: **Staleness Check Interval Clarification**

The staleness check happens inside the wait loop, which means stale messages are only detected during the wait period. Our understanding is that if a message becomes stale while a worker is processing it, it will be detected in the next wait cycle (which happens soon). Could you confirm if our understanding is correct? If a message is in the slot but no worker is waiting (e.g., all workers are busy processing), will it be detected as stale when the next wait cycle starts?

orchestrator/src/tests/server/job_routes.rs
```diff
@@ -44,10 +46,12 @@ async fn setup_trigger() -> (SocketAddr, Arc<Config>) {
...
 }
 
-#[tokio::test]
 #[rstest]
+#[case(false)]
+#[case(true)]
+#[tokio::test]
 #[allow(clippy::await_holding_lock)]
-async fn test_trigger_process_job(#[future] setup_trigger: (SocketAddr, Arc<Config>)) {
+async fn test_trigger_process_job(#[case] is_priority: bool, #[future] setup_trigger: (SocketAddr, Arc<Config>)) {
```
Mohiiit, line 54: **Test Coverage**

Tests cover basic priority queue functionality, but consider adding more comprehensive tests for: queue capacity limits (rejecting when full), slot staleness handling, concurrent priority queue submissions, error scenarios (DB failures, queue full, etc.), and the `check_priority_slot()` filtering logic.

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -0,0 +1,263 @@
+//! Priority Queue Worker Module
```
Mohiiit, line 1: **Documentation**

Consider adding documentation about: priority queue behavior under load, how staleness timeout affects message delivery, the relationship between priority queues and normal queues, and operational guidance for monitoring priority queue depth and slot staleness.

orchestrator/src/cli/service.rs
```diff
@@ -25,4 +25,26 @@ pub struct ServiceCliArgs {
     /// Timeout in seconds for jobs stuck in LockedForProcessing status before self-healing recovery.
     #[arg(env = "MADARA_ORCHESTRATOR_JOB_PROCESSING_TIMEOUT_SECONDS", long, default_value = "1800")]
     pub job_processing_timeout_seconds: u64,
+
```
Mohiiit, line 36: **Configuration Validation**

Consider adding validation to ensure `max_priority_queue_size > 0`, `priority_slot_wait_timeout_secs > 0`, `priority_slot_staleness_timeout_secs > 0`, etc. Invalid configurations could cause runtime issues. Could you add a TODO to implement configuration validation to avoid human errors? Ideally, we should validate these values at startup.

orchestrator/src/worker/controller/worker_controller.rs
Mohiiit, line 60: **Priority Queue Worker Shutdown**

The `shutdown()` method only signals EventWorkers, but not the PriorityQueueWorkers. They're spawned in `run()` but not tracked in the `workers` vector. However, they do respond to the cancellation token (`is_shutdown_requested()` checks `cancellation_token.is_cancelled()`), so they should shut down when the parent token is cancelled. Consider documenting this behavior or ensuring the shutdown flow is clear.

orchestrator/src/worker/controller/priority_queue_worker.rs
prkpndy: 1. Yes
2. It should not happen in any case
3. It should not happen under any scenario. But since `place_in_slot` isn't aware that we have checked for slot's emptiness before, it has to return a bool

orchestrator/src/worker/event_handler/triggers/mod.rs
prkpndy: done

prkpndy: this is intended. also, this code isn't changed in this PR. I just updated the call to return all job (updated the limit from Some(1) to None) so that the warn log is correct

orchestrator/src/worker/controller/event_worker.rs
prkpndy: Understanding is correct. Will add tests!

orchestrator/src/worker/controller/event_worker.rs
prkpndy: done

orchestrator/src/worker/controller/worker_controller.rs
prkpndy: done

orchestrator/src/worker/controller/priority_queue_worker.rs
prkpndy: done

orchestrator/src/server/route/jobs.rs
prkpndy: didn't get you 😅

orchestrator/src/core/client/queue/sqs.rs
prkpndy: this is an aws defined attribute, so should always exist and parsing should also not be an issue. also, if it fails in rare case it's not a disaster tbh. it's anyways an approximate comparision :)

orchestrator/src/worker/controller/priority_queue_worker.rs
prkpndy: priority queue worker processors are not that heavy. they just put the message in the slot and we have already checked that the slot is empty (hence no wait). so this should be quick.
So the flow would be - we wait for the slot to be empty (with staleness check). as soon as it gets empty, we get the message from the queue (if available) and process it (which is just putting the message in the slot).
Then we again go to the wait cycle. So the wait cycle will be started very quickly after we put the message in the slot (i.e., after process_message is called)

orchestrator/src/cli/service.rs
prkpndy: some of these are not needed to be configurable tbh. I have removed those. only max size of pq is now in config. i have also added a check for that while loading the config!

orchestrator/src/worker/controller/priority_queue_worker.rs
prkpndy: added in priority queue worker file :)

orchestrator/src/server/route/jobs.rs
prkpndy: sorted

orchestrator/src/worker/controller/worker_controller.rs
prkpndy: discussed offline

orchestrator/src/core/client/database/mongodb.rs
prkpndy: sorted!

orchestrator/src/worker/event_handler/triggers/snos.rs
prkpndy: done

orchestrator/src/worker/controller/event_worker.rs
prkpndy: done

orchestrator/src/worker/controller/event_worker.rs
prkpndy: hmm.. all the logic for handling the messages is present in run function itself - both for priority queue and normal queue. would prefer to keep it there only if it's okay with you

orchestrator/src/tests/server/job_routes.rs
prkpndy: have added the test for DB method introduced in this PR. Adding TODOs for other tests that we'll add in the future :)

Mohiiit (approved):

PR #889, opened 2025-12-05
fix(estimate_fee): should use charge_fee from config

byteZorvin:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

The estimate_fee RPC method always sets charge_fee to false when creating execution flags, ignoring the node's charge_fee configuration.

Resolves: #NA

## What is the new behavior?

The estimate_fee method now respects the charge_fee configuration from the node's validator parameters. The charge_fee value is passed through the RPC service and used when creating execution flags for fee estimation.

No


Mohiiit (approved):

byteZorvin (commented):
The bug was identified and it wasnt due to madara/blockifier but the custom account from starkpay where they handle txn versions differently. 
In the above context, closing this PR

PR #892, opened 2025-12-09
fix: orchestrator minor fixes for batching and logging

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

**Issue 1: Lock not released on error in batching worker**
- The batching worker acquires a lock at the start of execution
- If an error occurs during batch processing, the function returns early with `?`
- The lock release code is never reached, causing the lock to remain held until timeout
- This can block subsequent batching worker executions

**Issue 2: Unnecessary log noise**
- Madara logs a message about preconfirmed blocks not being saved in DB, which is expected behavior and creates unnecessary log noise

**Issue 3: No visibility into custom versioned constants usage**
- When custom versioned constants are loaded from a file, there's no indication in the logs
- Makes it difficult to debug issues related to versioned constants configuration

Resolves: #NA

## What is the new behavior?

**Fix 1: Guaranteed lock release (batching.rs)**
- Wrapped the main batching work in an async block to capture the result
- Lock is now **always** released after work completes, regardless of success or failure
- Proper error handling ensures original errors are preserved while lock release failures are logged
- Only scenario where lock won't release is a panic (which still relies on timeout)

**Fix 2: Remove preconfirmed block log (madara)**
- Removed the unnecessary log message about preconfirmed blocks not being saved
- Reduces log noise for expected behavior

**Fix 3: Add versioned constants logging (snos.rs)**
- Added debug logging when custom versioned constants are loaded from a file
- Shows the file path being used for better observability
- Helps operators understand which constants configuration is active

No

This PR contains only bug fixes and logging improvements. No API changes, no schema changes, no breaking changes to existing functionality.

### Testing
- Batching lock release has been verified to work correctly with error paths
- Logging changes have been verified in local testing

### Related Files Changed
- `orchestrator/src/worker/event_handler/triggers/batching.rs` - Lock release fix
- `madara/crates/client/db/src/lib.rs` - Remove unnecessary log
- `madara/crates/primitives/chain_config/src/chain_config.rs` - Add versioned constants logging
- `orchestrator/src/cli/snos.rs` - Add versioned constants logging

### Impact
- **Low risk**: Changes are isolated to error handling and logging
- **Positive impact**: Prevents lock starvation in batching worker and improves observability

### DIFF
--- a/orchestrator/src/worker/event_handler/triggers/batching.rs
+++ b/orchestrator/src/worker/event_handler/triggers/batching.rs
@@ -96,24 +96,37 @@ impl JobTrigger for BatchingTrigger {
             }
         }
 
-        // Getting the block range to process
-        let (start_block, end_block) = match config.layer() {
-            Layer::L2 => self.get_range_for_assigning_batches_l2(&config).await?,
-            Layer::L3 => self.get_range_for_assigning_batches_l3(&config).await?,
-        };
+        // Execute the main work and capture the result
+        let result = async {
+            // Getting the block range to process
+            let (start_block, end_block) = match config.layer() {
+                Layer::L2 => self.get_range_for_assigning_batches_l2(&config).await?,
+                Layer::L3 => self.get_range_for_assigning_batches_l3(&config).await?,
+            };
 
-        // Invoking method to assign batches to all the blocks in the range
-        if start_block < end_block {
-            match config.layer() {
-                Layer::L2 => self.assign_batch_to_blocks_l2(start_block, end_block, &config).await?,
-                Layer::L3 => self.assign_batch_to_blocks_l3(start_block, end_block, &config).await?,
+            // Invoking method to assign batches to all the blocks in the range
+            if start_block < end_block {
+                match config.layer() {
+                    Layer::L2 => self.assign_batch_to_blocks_l2(start_block, end_block, &config).await?,
+                    Layer::L3 => self.assign_batch_to_blocks_l3(start_block, end_block, &config).await?,
+                }
             }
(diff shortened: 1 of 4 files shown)


orchestrator/src/worker/event_handler/triggers/batching.rs
```diff
@@ -96,24 +96,37 @@ impl JobTrigger for BatchingTrigger {
             }
         }
 
-        // Getting the block range to process
-        let (start_block, end_block) = match config.layer() {
-            Layer::L2 => self.get_range_for_assigning_batches_l2(&config).await?,
-            Layer::L3 => self.get_range_for_assigning_batches_l3(&config).await?,
-        };
+        // Execute the main work and capture the result
```
byteZorvin, line 99: This was interesting, good one !!

byteZorvin (approved):
Looks good. Thanks for this  !

orchestrator/src/worker/event_handler/triggers/batching.rs
Mohiiit, line 1: can we have a test case about this? I think that would help!

Mohiiit (commented):
not a blocker, but I think our batcher lacks some edge cases like these, would be a good edition indeed

orchestrator/src/worker/event_handler/triggers/batching.rs
prkpndy: sure. will add in priority queue PR :)

Mohiiit (approved):

PR #897, opened 2025-12-10
fix(orchestrator): requeue SNOS jobs when RPC is unavailable

byteZorvin:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

When the SNOS RPC endpoint (Pathfinder) is down, all SnosRun jobs fail one by one, sending alerts for each failure. This causes:
1. Multiple unnecessary job failures when external infrastructure is temporarily unavailable
2. Alert fatigue from continuous failure notifications
3. Jobs being marked as failed when they should simply wait for the dependency to recover

Resolves: #NA

## What is the new behavior?

- **Health check before processing**: Added `check_ready_to_process()` method to `JobHandlerTrait` that allows handlers to verify dependencies are available before job processing begins
- **SNOS health check**: `SnosJobHandler` implements this method to check if SNOS RPC is reachable using `chain_id()` RPC call
- **Graceful requeue**: When SNOS is unavailable, jobs are requeued with a 60-second delay instead of failing
- **Single alert per downtime**: Uses `AtomicBool` to send only one alert when SNOS goes down, preventing alert fatigue
- **Auto-recovery**: Alert flag resets when SNOS recovers, so the next downtime will trigger a new alert

### Flow:
```
process_job() called
    │
    ├─► check_ready_to_process()
    │       │
    │       ├─► SNOS healthy → Ok() → continue processing
    │       │
    │       └─► SNOS down → Err(60s delay)
    │               │
    │               ├─► Send alert (once per downtime)
    │               └─► Requeue job with delay
    │
    └─► Job not marked as failed, status remains "Created"
```

No - This is an additive change. The `check_ready_to_process()` method has a default implementation returning `Ok(())`, so existing job handlers continue to work without modification.

### Files changed:
- `orchestrator/src/worker/event_handler/jobs/mod.rs` - Added `check_ready_to_process` to trait
- `orchestrator/src/worker/event_handler/jobs/snos.rs` - Implemented health check for SNOS
- `orchestrator/src/worker/event_handler/service.rs` - Call check before processing
- `orchestrator/src/worker/service.rs` - Made `add_job_to_queue` pub(crate)
- `orchestrator/src/tests/jobs/snos_job/mod.rs` - Added health check tests
- `orchestrator/src/tests/jobs/mod.rs` - Added requeue behavior test

### DIFF
--- a/orchestrator/src/tests/jobs/mod.rs
+++ b/orchestrator/src/tests/jobs/mod.rs
@@ -417,7 +417,6 @@ async fn process_job_with_job_exists_in_db_with_invalid_job_processing_status_er
 /// 4. No failure is recorded
 #[rstest]
 #[tokio::test]
-#[allow(clippy::await_holding_lock)]
 async fn process_job_requeues_when_check_ready_to_process_fails() {
     // Acquire test lock to serialize this test with others that use mocks
     let _test_lock = acquire_test_lock();
--- a/orchestrator/src/tests/jobs/mod.rs
+++ b/orchestrator/src/tests/jobs/mod.rs
@@ -278,6 +278,8 @@ async fn process_job_with_job_exists_in_db_and_valid_job_processing_status_works
     database_client.create_job(job_item.clone()).await.unwrap();
 
     let mut job_handler = MockJobHandlerTrait::new();
+    // Expecting check_ready_to_process to return Ok (dependencies are ready)
+    job_handler.expect_check_ready_to_process().times(1).returning(|_| Ok(()));
     // Expecting process job function in job processor to return the external ID.
     job_handler.expect_process_job().times(1).returning(move |_, _| Ok("0xbeef".to_string()));
     job_handler.expect_verification_polling_delay_seconds().return_const(1u64);
@@ -345,6 +347,8 @@ async fn process_job_handles_panic() {
     database_client.create_job(job_item.clone()).await.unwrap();
 
     let mut job_handler = MockJobHandlerTrait::new();
+    // Expecting check_ready_to_process to return Ok (dependencies are ready)
+    job_handler.expect_check_ready_to_process().times(1).returning(|_| Ok(()));
     // Setting up mock to panic when process_job is called
     job_handler
         .expect_process_job()
@@ -511,6 +515,9 @@ async fn process_job_two_workers_process_same_job_works() {
     let _test_lock = acquire_test_lock();
 
     let mut job_handler = MockJobHandlerTrait::new();
+    // Expecting check_ready_to_process to return Ok (dependencies are ready)
+    // Both workers will call check_ready_to_process, so expect 2 calls
+    job_handler.expect_check_ready_to_process().times(2).returning(|_| Ok(()));
     // Expecting process job function in job processor to return the external ID.
     job_handler.expect_process_job().times(1).returning(move |_, _| Ok("0xbeef".to_string()));
     job_handler.expect_verification_polling_delay_seconds().return_const(1u64);
@@ -567,6 +574,8 @@ async fn process_job_job_handler_returns_error_works() {
     let _test_lock = acquire_test_lock();
 
     let mut job_handler = MockJobHandlerTrait::new();
+    // Expecting check_ready_to_process to return Ok (dependencies are ready)
+    job_handler.expect_check_ready_to_process().times(1).returning(|_| Ok(()));
     // Expecting process job function in job processor to return the external ID.
     let failure_reason = "Failed to process job";
     job_handler
--- a/orchestrator/src/tests/jobs/mod.rs
+++ b/orchestrator/src/tests/jobs/mod.rs
@@ -409,6 +409,68 @@ async fn process_job_with_job_exists_in_db_with_invalid_job_processing_status_er
     assert_matches!(queue_error, QueueError::ErrorFromQueueError(_));
 }
 
+/// Tests `process_job` function when `check_ready_to_process` returns Err (dependencies not ready).
+/// This test verifies that:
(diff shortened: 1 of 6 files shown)


orchestrator/src/worker/event_handler/jobs/snos.rs
```diff
@@ -208,6 +228,30 @@ impl JobHandlerTrait for SnosJobHandler {
     fn verification_polling_delay_seconds(&self) -> u64 {
         1
     }
+
+    async fn check_ready_to_process(&self, config: Arc<Config>) -> Result<(), Duration> {
+        let snos_url = config.snos_config().rpc_for_snos.to_string();
```
heemankv: `rpc_for_snos` is already a Url, any reason to re-parse it ?

orchestrator/src/worker/event_handler/jobs/snos.rs
```diff
@@ -208,6 +228,30 @@ impl JobHandlerTrait for SnosJobHandler {
...
+            warn!(snos_url = %snos_url, "SNOS RPC is unavailable, job will be requeued");
+
+            // Send alert only once per downtime period
+            if !SNOS_UNAVAILABLE_ALERT_SENT.swap(true, Ordering::SeqCst) {
+                let alert_msg =
+                    format!("SNOS RPC {} is unavailable. SnosRun jobs will be requeued until it recovers.", snos_url);
+                if let Err(e) = config.alerts().send_message(alert_msg).await {
+                    error!(error = ?e, "Failed to send SNOS unavailability alert");
+                }
+            }
```
heemankv: If config.alerts().send_message(alert_msg) fails, the flag is already set to true, so the alert will never be attempted again (even if SNOS recovers and goes down again later) !

heemankv (changes requested):

orchestrator/src/worker/event_handler/jobs/snos.rs
heemankv: Also this will only solve alert aggregation properly only when we have 1 pod of orchestrator running. 
Over on multiple pods the aggregation reduce inversely with number of pods

heemankv (changes requested):

orchestrator/src/worker/event_handler/jobs/snos.rs
heemankv: Should we just handle aggregation outside of orchestrator ?

byteZorvin (commented):
[removed comment]

byteZorvin (commented):
[removed comment]

byteZorvin (commented):
[removed comment]

byteZorvin (commented):
[removed comment]

orchestrator/src/worker/event_handler/jobs/snos.rs
byteZorvin: Good catch on comment 1! Updated `check_snos_health` to accept &Url directly instead of re-parsing the string.

orchestrator/src/worker/event_handler/jobs/snos.rs
byteZorvin: you're correct that if alert sending fails, the flag stays set and prevents future retries during the same downtime period. I've updated the code to reset the flag on failure, so the next job attempt will retry the alert send.

byteZorvin: Acknowledged. For single-pod deployments this works well. For multi-pod, we'd get at most N alerts (one per pod) which is bounded and significantly better than the current unbounded alerts. For true single-alert behavior across pods, we'd to store that in mongodb. 
We could go for Sns deduplication also when needed but that is only supported for Fifo sns, and we dont use that.
What external approach are you suggesting specifically ?

orchestrator/src/tests/jobs/mod.rs
```diff
@@ -409,6 +409,68 @@ async fn process_job_with_job_exists_in_db_with_invalid_job_processing_status_er
...
 
+/// Tests `process_job` function when `check_ready_to_process` returns Err (dependencies not ready).
+/// This test verifies that:
+/// 1. The job is NOT processed (process_job on handler should not be called)
+/// 2. The job is requeued to the processing queue with a delay
+/// 3. The job status remains unchanged (still Created)
+/// 4. No failure is recorded
+#[rstest]
+#[tokio::test]
+#[allow(clippy::await_holding_lock)]
```
prkpndy: can add a comment explaining the need for this

orchestrator/src/worker/service.rs
```diff
@@ -40,7 +40,7 @@ impl JobService {
     ///
     /// # Returns
     /// * `Result<(), JobError>` - Success or an error
-    async fn add_job_to_queue(
+    pub(crate) async fn add_job_to_queue(
```
prkpndy, line 43: why do we need to make it pub?

prkpndy (approved):
lgtm

orchestrator/src/tests/jobs/mod.rs
byteZorvin: Actually this is not needed, somehow this got pushed

orchestrator/src/worker/service.rs
byteZorvin: We need this as we call it outside of this module. Also the public function [here](https://github.com/madara-alliance/madara/blob/f1437a689ac62063f48d5352c6fe92bfd12ee175/orchestrator/src/worker/service.rs#L57-L58) does not have delay param available.

heemankv (approved):

PR #899, opened 2025-12-12
Prevent duplicate Atlantic job submissions

byteZorvin:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

Atlantic jobs are always submitted even when an identical external_id already exists, and bucket metadata is not validated before reuse. There was no client helper to search queries by external_id.

Resolves: #NA

## What is the new behavior?

- Add an Atlantic query search helper (with response types/errors) and pass external_id through job creation payloads
- Skip resubmission when an existing non-failed job matches the external_id while asserting bucket_id/bucket_job_index consistency; resubmit only when the existing job failed
- Add regression test `atlantic_client_does_not_resubmit_when_job_exists` to ensure duplicates are not re-sent

No


orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -204,6 +206,98 @@ impl AtlanticClient {
...
+    /// Returns a list of Atlantic queries matching the search criteria and total count
+    pub async fn search_atlantic_queries(
+        &self,
+        atlantic_api_key: impl AsRef<str>,
+        search_string: impl AsRef<str>,
+        limit: Option<u32>,
+        offset: Option<u32>,
+        network: Option<&str>,
+        status: Option<&str>,
+        result: Option<&str>,
```
prkpndy: can remove these params if not used (we are only using this method at one place i think and not using these params). Or can put these into a struct

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -66,7 +66,54 @@ impl ProverClient for AtlanticProverService {
...
+                        AtlanticQueryStatus::Failed => {
+                            tracing::warn!(
+                                atlantic_query_id = %existing_job.id,
+                                external_id = %external_id,
+                                status = ?existing_job.status,
+                                "Existing Atlantic job found in failed state. Resubmitting."
+                            );
+                        }
+                        _ => {
+                            Self::ensure_bucket_details_match(&existing_job, &bucket_id, bucket_job_index)?;
```
prkpndy, line 102: hmm.. why are we doing confirmation after we got the result from the API?

prkpndy: also, why only in the not failed case?

orchestrator/crates/prover-clients/atlantic-service/tests/lib.rs
```diff
@@ -53,13 +53,121 @@ async fn atlantic_client_submit_task_when_mock_works() {
...
+
+    let search_mock = mock_server.mock(|when, then| {
+        when.method("GET")
+            .path("/atlantic-queries")
+            .query_param("search", external_id.as_str())
+            .query_param("limit", "1")
+            .query_param("network", "TESTNET");
+
+        then.status(200).header("content-type", "application/json").json_body(serde_json::json!({
+            "atlanticQueries": [{
```
prkpndy: can use the AtlanticQuery struct here ig instead of writing all things

prkpndy (changes requested):
minor refactoring suggestions.
btw, any reason why we didn't go with checking the response message code approach? (like, send the request, check if the response has that specific error code which atlantic sends when job already exists with the given bucket id and job index and then check the status of the job)

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
byteZorvin: As discussed f2f, atlantic does not apply unique constraint on the external id, so this the only approach that i could think of

prkpndy (approved):
approved

heemankv (approved):

Issue #824, opened 2025-10-29
Clean Batching Code


prkpndy:
The batching code is very cluttered and difficult to understand for a new person. Organise the code properly using traits if needed

heemankv:
Status update from codebase review on 2026-05-05.

Classification: Still relevant.

Current state:
Still relevant. Current batching/compression/applicative-recursion code works, but it is still sprawling and a good cleanup target.

Keeping this issue open based on the current tree.

PR #901, opened 2025-12-15
refactor: batching logic

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

SNOS and Aggregator batching used to happen simultaneously in a single worker. This was making the code very complicated.

Resolves:
* https://github.com/madara-alliance/madara/issues/824

## What is the new behaviour?

Now we have separated SNOS and Aggregator batching logic

No


orchestrator/src/cli/batching.rs
```diff
@@ -25,7 +25,7 @@ pub struct BatchingCliArgs {
     /// Keep this None if you don't want to specify a hard limit.
     /// This can be used to test with RPC other than Madara.
     #[arg(env = "MADARA_ORCHESTRATOR_MAX_BLOCKS_PER_SNOS_BATCH", long)]
-    pub max_blocks_per_snos_batch: Option<u64>,
+    pub max_blocks_per_snos_batch: u64,
```
byteZorvin, line 28: Why are we moving away from option? Feature wise optionality seems better.

orchestrator/src/core/config.rs
```diff
@@ -558,13 +588,13 @@ impl Config {
...
 
         BouncerWeights {
-            l1_gas: 1_000_000,                 // 1M L1 gas
-            message_segment_length: 100_000,   // 100K message segment length
-            n_events: 5_000,                   // 5K events
-            state_diff_size: 100_000,          // 100K state diff size
-            sierra_gas: GasAmount(10_000_000), // 10M sierra gas
-            n_txs: 1_000,                      // 1K transactions
-            proving_gas: GasAmount(5_000_000), // 5M proving gas
+            l1_gas: 10_000_000,                 // 1M L1 gas
```
byteZorvin: Any specific reason to change this?

byteZorvin: The comments can be updated besides the numbers though

orchestrator/src/core/client/lock/mongodb.rs
```diff
@@ -88,10 +88,11 @@ impl MongoLockClient {
                 )
                 .build(),
             // Unique index on the key for atomic operations
-            IndexModel::builder()
-                .keys(doc! { "key": 1 })
-                .options(IndexOptions::builder().unique(true).name("key_unique_index".to_string()).build())
-                .build(),
+            // key is stored as _id which already has unique index
+            // IndexModel::builder()
```
byteZorvin: This can be removed?

orchestrator/src/core/config.rs
Mohiiit: The inline comments are out of sync with the actual values (e.g., comment says "1M L1 gas" but value is 10M). Could you either update the comments to match the new values, or add a TODO/link to the official spec (ideally with a commit hash) explaining where these limits come from?

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,142 @@
...
+        debug!(first_block_to_assign_batch = %first_block_to_assign_batch, "Calculated first block number to batch.");
+
+        let max_blocks_to_process_at_once = self.max_blocks_to_process_at_once();
+        let end_block =
+            min(last_block_to_assign_batch, first_block_to_assign_batch + max_blocks_to_process_at_once - 1);
+        Ok((first_block_to_assign_batch, end_block))
+    }
+
+    fn max_blocks_to_process_at_once(&self) -> u64 {
+        10
```
Mohiiit: This is hardcoded to 10 (the old `BatchingTrigger` used 25). Should this be made configurable via CLI/env? Maybe add a TODO for future if not needed now.

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
```diff
@@ -0,0 +1,297 @@
...
+        block_aggregator_batch_index: Option<u64>,
+        batch_limits: &SnosBatchLimits,
+    ) -> Result<Option<Self>, JobError> {
+        // If a fixed size is set, use only that to decide if we should close the SNOS batch
+        if let Some(fixed_blocks_per_snos_batch) = batch_limits.fixed_batch_size {
+            // If the MADARA_ORCHESTRATOR_FIXED_BLOCKS_PER_SNOS_BATCH env is set, we use that value
+            // Mostly, it'll be used for testing purposes
+            debug!("Using MADARA_ORCHESTRATOR_FIXED_BLOCKS_PER_SNOS_BATCH env variable to close snos batch with max blocks = {}", fixed_blocks_per_snos_batch);
+            return Ok(None);
+        }
```
Mohiiit, line 261: When `fixed_blocks_per_snos_batch` is set, this returns `Ok(None)` immediately without checking:
1. Whether `self.batch.num_blocks >= fixed_blocks_per_snos_batch` (to know if we should close)
2. Whether the bouncer weights would overflow

Even with a fixed batch size for testing, shouldn't we still validate bouncer weights to prevent breaking limits? Otherwise tests could pass but production would fail due to weight violations.

orchestrator/src/tests/workers/batching/mod.rs
```diff
@@ -246,65 +265,22 @@ async fn test_batching_worker_with_multiple_blocks() -> Result<(), Box<dyn Error
...
     // Mock starknet_getStateUpdate - separate mock for each block
-    for block_num in 4..=7 {
+    for block_num in start_block..=end_block {
+        let pattern = format!(r#".*"block_number"\s*:\s*{}[,\}}].*"#, block_num);
+        server.mock(move |when, then| {
+            when.path("/").body_includes("starknet_getBlockWithTxHashes").body_matches(pattern.as_str());
+            then.status(200).body(
+                serde_json::to_vec(&json!({
+                    "jsonrpc":"2.0",
+                    "result": get_dummy_block_with_tx_hashes(block_num, "0.13.2"),
```
Mohiiit, line 321: earlier we were testing with the 0.13.3 version as well, which ensured that the batches are closing across different starknet version? shall we maintain that or it's being tested somewhere else as well?

orchestrator/src/types/batch.rs
```diff
@@ -39,18 +41,20 @@ pub enum AggregatorBatchStatus {
     StateUpdateFailed,
 }
 
+impl AggregatorBatchStatus {
+    pub fn is_closed(&self) -> bool {
```
Mohiiit, line 43: batch status would be closed if it's not open? what about other status?

orchestrator/src/types/batch.rs
```diff
@@ -197,26 +194,22 @@ impl AggregatorBatch {
...
         start_block: u64,
-        squashed_state_updates_path: String,
-        blob_path: String,
         bucket_id: String,
+        blob_len: usize,
         builtin_weights: AggregatorBatchWeights,
-        starknet_version: String,
+        starknet_version: StarknetVersion,
     ) -> Self {
+        // TODO: move get file path methods to this struct
```
Mohiiit: let's follow the TODO(author date): (todo)

orchestrator/src/types/batch.rs
```diff
@@ -244,6 +267,12 @@ pub enum SnosBatchStatus {
     Completed,
 }
 
+impl SnosBatchStatus {
```
Mohiiit, line 267: shall we remove other status code then?

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -0,0 +1,429 @@
...
+                    biguint_vec_to_u8_vec(blob.as_slice()).into(),
+                    &AggregatorBatch::get_blob_file_path(state.batch.index, i as u64 + 1),
+                )
+                .await?;
+        }
+
+        // Update batch status in the database
+        self.config
+            .database()
+            .update_or_create_aggregator_batch(&state.batch, &AggregatorBatchUpdates::default())
```
Mohiiit: do we need batch and batchUpdates both here? batch would have all the info right?

orchestrator/src/worker/event_handler/triggers/batching/mod.rs
```diff
@@ -0,0 +1,15 @@
+pub mod aggregator;
+pub mod snos;
+pub mod utils;
+
+pub enum BlockProcessingResult<S, NES> {
```
Mohiiit, line 5: not sure if mod is the right place for this

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct NonEmptySnosState {
+    pub batch: SnosBatch,
+}
+
+pub struct SnosStateHandler {
+    config: Arc<Config>,
+}
+
+impl SnosStateHandler {
+    #[allow(dead_code)]
```
Mohiiit: let's remove the dead code then

orchestrator/src/core/config.rs
prkpndy: The old values were too low actually. Haven't picked these up from anywhere. Just multiplied them by 10 so that we atleast get more than one block in a batch on sepolia

prkpndy: I'll update the comments and add a TODO

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: From my experience 10 is a decent enough limit. It takes ~2 mins for sepolia for batching 10 blocks (on average). Can make snos one larger but won't have much effect since it's upper capped by aggregator batches

orchestrator/src/worker/event_handler/triggers/aggregator.rs
```diff
@@ -108,14 +104,33 @@ impl AggregatorJobTrigger {
...
+            let last = snos_batches.last().unwrap();
+
+            let mut start = first.start_block;
+            for batch in snos_batches.iter() {
+                if !batch.status.is_closed() || batch.start_block != start {
+                    return Ok(false);
+                } else {
+                    start = batch.end_block + 1;
+                }
+            }
```
Mohiiit: This check validates that SNOS batches are contiguous (no gaps between them), but it doesn't validate that:
1. The first SNOS batch's `start_block` matches the aggregator batch's `start_block`
2. The last SNOS batch's `end_block` matches the aggregator batch's `end_block`

If there's a gap at the beginning or end, this wouldn't catch it. For example:
- Aggregator batch: blocks 100-200
- SNOS batches: blocks 110-200 (missing 100-109)

The check would pass because the SNOS batches are internally contiguous. Should we add validation against the aggregator batch's block range?

Mohiiit: not sure about this ^

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
prkpndy: we set `MADARA_ORCHESTRATOR_FIXED_BLOCKS_PER_SNOS_BATCH` env only when bouncer weights are inaccessible. It's meant only for testing with non-madara RPC

prkpndy: I should have an if check there though. Have fixed that 🙏

orchestrator/src/types/batch.rs
prkpndy: It'll be closed if the status is not open since all other status can come only after Closed. The flow is Open -> Close -> Others... There was one status `BatchCreationFailed` which isn't and can't be used. Have removed that.

orchestrator/src/types/batch.rs
prkpndy: hmm.. we are using all except `SnosJobCreated`. Will add a TODO to use that as well. or did you mean something else?

orchestrator/src/worker/event_handler/triggers/aggregator.rs
prkpndy: Good catch. Fixed it. Thanks!

orchestrator/src/types/batch.rs
prkpndy: this is resolve

orchestrator/src/tests/workers/batching/mod.rs
prkpndy: I have added new tests which covers that

orchestrator/src/types/batch.rs
prkpndy: done

byteZorvin (commented):
looks good, few questions only.

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,142 @@
...
+            config.clone(),
+            AggregatorBatchLimits::from_config(&config.params),
+            config.params.batching_config.default_empty_block_proving_gas,
+        );
+
+        let state_handler = AggregatorStateHandler::from_config(&config);
+
+        // Execute the main work and capture the result
+        let result = async {
+            let (start_block, end_block) = self.calculate_range(&config).await?;
```
byteZorvin, line 51: If the worker panic here, we wont be able to release lock?

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,142 @@
...
+
+            // If there are no blocks to process, return early
+            if start_block > end_block {
+                debug!("No Aggregator blocks to process: start_block ({}) > end_block ({})", start_block, end_block);
+                return Ok(());
+            }
+
+            info!("Processing Aggregator batches for blocks {} to {}", start_block, end_block);
+
+            let mut state = state_handler.load_batch_state().await?;
```
byteZorvin, line 61: or here

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
byteZorvin: What if there are many empty blocks say for starkpay, or running in locally?

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,142 @@
...
+        let batching_handler = AggregatorHandler::new(
+            config.clone(),
+            AggregatorBatchLimits::from_config(&config.params),
+            config.params.batching_config.default_empty_block_proving_gas,
+        );
+
+        let state_handler = AggregatorStateHandler::from_config(&config);
+
+        // Execute the main work and capture the result
+        let result = async {
```
byteZorvin, line 50: Oh never mind, this should work

orchestrator/src/core/config.rs
prkpndy: done

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: Yes. We have all none values in batchUpdates

orchestrator/src/cli/batching.rs
prkpndy: it's good to always specify the max limit for SNOS batch. The code unnecessarily gets complicated because of None. Starkware also reported that they have a limit for aggregator batches

orchestrator/src/core/config.rs
prkpndy: Comments updated. The earlier values were too low and all batches were being created with 1 block

orchestrator/src/core/client/lock/mongodb.rs
prkpndy: done

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: added env for this

orchestrator/src/cli/batching.rs
apoorvsadana: nitpick: we could do a default value!

orchestrator/src/cli/batching.rs
```diff
@@ -40,4 +40,8 @@ pub struct BatchingCliArgs {
     /// This value is used when proving_gas is zero.
     #[arg(env = "MADARA_ORCHESTRATOR_DEFAULT_EMPTY_BLOCK_PROVING_GAS", long, default_value = "1500000")]
     pub default_empty_block_proving_gas: u64,
+
+    /// Maximum number of blocks to batch in one go in aggregator and snos batching worker
+    #[arg(env = "MADARA_ORCHESTRATOR_MAX_BLOCKS_TO_BATCH_AT_ONCE", long, default_value = "10")]
+    pub max_blocks_to_batch_at_once: u64,
```
apoorvsadana: does this sound better?
```suggestion
    pub max_batch_processing_size: u64,
```

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -78,6 +79,72 @@ impl MongoDbClient {
         Ok(Self { client, database })
     }
 
+    pub async fn ensure_indexes(&self) -> Result<(), DatabaseError> {
```
apoorvsadana, line 82: just to confirm, we call create everytime but if it already exists, nothing happens?

apoorvsadana: mongodb has a pretty good js library or handling migrations already. it manages a changelog table inside mongodb and ensures we're up to date. we were already using it here (madara-alliance/madara/orchestrator/migrate-mongo-config.js). in the dockerfile, we used to check for migrations on every start. i think that got lost in some docker upgrade.

for now, can keep it like this, i believe this is how it was already happening. however, can note this down for a future upgrade.

orchestrator/src/setup/mod.rs
```diff
@@ -31,7 +31,7 @@ pub async fn setup(setup_cmd: &SetupCmd) -> OrchestratorResult<()> {
     debug!("Alert Params: {:?}", alert_params);
     debug!("Cron Params: {:?}", cron_params);
 
-    lock_client.initialize().await.map_err(|e| OrchestratorError::SetupError(e.to_string()))?;
+    lock_client.ensure_indexes().await.map_err(|e| OrchestratorError::SetupError(e.to_string()))?;
```
apoorvsadana: shouldn't we also do this for the normal client?

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -0,0 +1,880 @@
...
+    }
+
+    /// Save the given state in DB and Storage
+    ///
+    /// 1. Update or Create doc in DB
+    /// 2. Update or Create the blob and other assets in Storage
+    ///
+    /// IMPORTANT:
+    /// 1. Assuming all the details are already updated in state
+    /// 2. Not making database and storage updates atomically. It might happen that one fail and other pass
```
apoorvsadana:
```suggestion
    /// 2. Not making database and storage updates atomically. It might happen that one fails and the other passes
```

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,138 @@
...
+                // Returning safely
+                debug!("AggregatorBatchingWorker failed to acquire lock, returning safely: {}", err);
+                return Ok(());
+            }
+        }
+
+        let batching_handler = AggregatorHandler::new(
+            config.clone(),
+            AggregatorBatchLimits::from_config(&config.params),
+            config.params.batching_config.default_empty_block_proving_gas,
```
apoorvsadana: nitpick: `default_empty_block_proving_gas` can be a part of `AggregatorBatchLimits` itself (maybe rename to aggregator batch config).

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+impl AggregatorBatchingTrigger {
+    async fn calculate_range(&self, config: &Arc<Config>) -> Result<(u64, u64), JobError> {
+        // Get the latest aggregator batch from the DB
+        let latest_batch = config.database().get_latest_aggregator_batch().await?;
+
+        // Getting the latest block numbers for aggregator and snos batches from DB
+        let last_block_in_batches = latest_batch.map_or(-1, |batch| batch.end_block as i64);
```
apoorvsadana:
```suggestion
        let last_created_batch = latest_batch.map_or(-1, |batch| batch.end_block as i64);
```

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,138 @@
...
+        // Calculating the last block number that needs to be assigned to a batch
+        let last_block_to_assign_batch = config
+            .service_config()
+            .max_block_to_process
+            .map_or(last_block_in_provider, |max_block| min(max_block, last_block_in_provider));
+
+        debug!(latest_block_number = %last_block_to_assign_batch, "Calculated last block number to batch.");
+
+        // Calculating the first block number to for which a batch needs to be assigned
+        let first_block_to_assign_batch =
```
apoorvsadana:
```suggestion
        let start_block =
```

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,138 @@
...
+        // Getting the latest block numbers for aggregator and snos batches from DB
+        let last_block_in_batches = latest_batch.map_or(-1, |batch| batch.end_block as i64);
+
+        // Getting the latest block number from the sequencer
+        let provider = config.madara_rpc_client();
+        let last_block_in_provider =
+            provider.block_number().await.map_err(|e| JobError::ProviderError(e.to_string()))?;
+
+        // Calculating the last block number that needs to be assigned to a batch
+        let last_block_to_assign_batch = config
```
apoorvsadana:
```suggestion
        let end_block = config
```

this shouldn't conflict w/ the next `let end_block=` because rust allows multiple declarations
and might be more readable that we're overriding the end_block based on `max_blocks_to_process_at_once`

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -0,0 +1,138 @@
...
+                    state_handler.save_batch_state(&state).await?;
+                }
+            }
+
+            Ok(())
+        }
+        .await;
+
+        // Always release the lock, regardless of whether work succeeded or failed
+        if let Err(e) = config.lock().release_lock("AggregatorBatchingWorker", None).await {
```
apoorvsadana: `AggregatorBatchingWorker` should be a constant (can be used in tests too)

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -0,0 +1,880 @@
...
+        }
+
+        Ok(Some(NonEmptyAggregatorState {
+            batch: self.batch.update(block_num, blob_len, combined_weights, None),
+            blob: squashed_state_update,
+        }))
+    }
+
+    pub fn close(&self) -> NonEmptyAggregatorState {
+        let mut batch = self.batch.clone();
```
apoorvsadana: you can pass &mut self and skip the cloning

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -0,0 +1,880 @@
...
+            )))
+        })?;
+        debug!(index = %index, bucket_id = %bucket_id, "Created new bucket successfully");
+
+        // Getting the builtin weights for the start_block and adding it in the DB
+        let weights = AggregatorBatchWeights::from(
+            &get_block_builtin_weights(
+                start_block,
+                self.config.madara_feeder_gateway_client(),
+                self.empty_block_proving_gas,
```
apoorvsadana: do we need to store `empty_block_proving_gas` in the DB? sounds like this is an orch config instead of a batch config

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
```diff
@@ -0,0 +1,895 @@
...
+                // Get the starknet version of the current block
+                let block_version = get_block_version(block_num, self.config.madara_rpc_client()).await?;
+                let block_weights = get_block_builtin_weights(
+                    block_num,
+                    self.config.madara_feeder_gateway_client(),
+                    self.empty_block_proving_gas,
+                )
+                .await?;
+                let new_state = self
+                    .start_snos_batch(
```
apoorvsadana: `start_snos_batch` seems like a wrong name here. its just creating the the object. `build` might be correct but even better would be to just copy that line here.

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
```diff
@@ -0,0 +1,895 @@
...
+            return SnosBatchCheckResult::TimeLimitExceeded;
+        }
+
+        // Check aggregator batch index (for L2)
+        if let Some(snos_batch_aggregator_batch_index) = self.batch.aggregator_batch_index {
+            match block_aggregator_batch_index {
+                None => {
+                    return SnosBatchCheckResult::MissingAggregatorBatch;
+                }
+                Some(block_agg_index) => {
```
apoorvsadana:
```suggestion
                Some(block_aggregator_batch_index) => {
```

apoorvsadana: the naming is very confusing w/ `snos_batch_aggregator_batch_index` and `block_agg_index`. in which case does this check get triggered?

orchestrator/src/worker/event_handler/triggers/aggregator.rs
```diff
@@ -108,14 +104,38 @@ impl AggregatorJobTrigger {
...
+        } else {
+            // unwraps are safe here
+            let first = snos_batches.first().unwrap();
+            let last = snos_batches.last().unwrap();
+
+            if first.start_block != aggregator_batch.start_block || last.end_block != aggregator_batch.end_block {
+                return Ok(false);
+            }
+
+            let mut start = first.start_block;
```
apoorvsadana: `first_block_aggregator_batch`

reading both `first` and `start` in 5 lines is confusing!

orchestrator/src/worker/event_handler/triggers/aggregator.rs
```diff
@@ -108,14 +104,38 @@ impl AggregatorJobTrigger {
...
+            for batch in snos_batches.iter() {
+                if !batch.status.is_closed() || batch.start_block != start {
```
apoorvsadana: why do we've this check `batch.start_block != start`

apoorvsadana (commented):
the code is much cleaner than before! two places that can be improved imo

1. variable naming. added comments in places where i felt it was relevant.
2. more comments in code, it will make it easier to review and AI does a great job :)

orchestrator/src/cli/batching.rs
prkpndy: done

orchestrator/src/core/client/database/mongodb.rs
prkpndy: will create an issue for this and handle in another PR 🙏

orchestrator/src/setup/mod.rs
prkpndy: have removed mongo setup during setup phase

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: 🙏

orchestrator/src/worker/event_handler/triggers/aggregator.rs
prkpndy: this is for confirming that there are no gaps in the batches

orchestrator/src/worker/event_handler/triggers/aggregator.rs
prkpndy: updated the variable names

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
prkpndy: updated the names

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
prkpndy: removed the method

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: empty block proving gas is used when the block's weights are zero. we only store the accumulated weights of the batch, not the default proving gas

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: done

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: made it `last_processed_block`

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: refactored

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: gotcha

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
prkpndy: done

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: done

byteZorvin (approved):

apoorvsadana (approved):

PR #910, opened 2025-12-18
Turning Vanilla

heemankv:
Removes Emojis and Color from Orchestrator


orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -408,11 +367,27 @@ impl EventWorker {
                     match message_result {
                         Ok(message) => {
                             if let Ok(parsed_message) = self.parse_message(&message) {
+                                // Log received message with queue and payload
+                                match &parsed_message {
+                                    ParsedMessage::WorkerTrigger(msg) => {
+                                        info!(
```
byteZorvin: Nit: why adding more log here

byteZorvin (approved):
Looks good

orchestrator/src/cli/mod.rs
```diff
@@ -179,7 +179,7 @@ pub struct RunCmd {
 
     /// Graceful shutdown timeout in seconds.
     /// Provides sufficient time for workers to complete their current tasks before forcing termination.
-    #[arg(env = "MADARA_ORCHESTRATOR_GRACEFUL_SHUTDOWN_TIMEOUT", long, default_value = "120")]
+    #[arg(env = "MADARA_ORCHESTRATOR_GRACEFUL_SHUTDOWN_TIMEOUT", long, default_value = "30")]
```
Mohiiit: 30 second is enough for the graceful shutdown?

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -241,9 +241,7 @@ impl QueueClient for SQS {
     /// The consumer is used to receive messages from the queue.
     async fn get_consumer(&self, queue: QueueType) -> Result<SqsConsumer, QueueError> {
         let queue_name = self.get_queue_name(&queue)?;
-        tracing::debug!("Getting queue url for queue name {}", queue_name);
         let queue_url = self.inner.get_queue_url_from_client(queue_name.as_str()).await?;
-        tracing::debug!("Found queue url {}", queue_url);
```
Mohiiit, line 246: just debug logs tho?

orchestrator/Cargo.toml
```diff
@@ -115,6 +115,7 @@ tracing-subscriber = { workspace = true, features = [
 ] }
 
 anyhow = { workspace = true }
+color-spantrace = "0.2.1"
```
Mohiiit: is this being used?

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -339,25 +318,7 @@ impl EventWorker {
         let span = self.create_job_span(&parsed_message);
```
Mohiiit, line 318: span was removed in the above functions but not here? I think we should keep it consistent or maybe it's deliberate? in that case what's special about this span?

orchestrator/src/worker/controller/event_worker.rs
Mohiiit: +1 would be lot of logs because of this?

orchestrator/src/cli/mod.rs
heemankv: reverted, using envs to govern this

orchestrator/src/core/client/queue/sqs.rs
heemankv: these were logging the queue url with additional aws details on to the logs, hence removed

orchestrator/Cargo.toml
heemankv: removed, good catch

orchestrator/src/worker/controller/event_worker.rs
heemankv: this nested span leads to improper span trace, removing these fixes that. otherwise we always get OtherError in span trace

orchestrator/src/worker/controller/event_worker.rs
heemankv: 1. Need these to ensure the orch is able to get messages properly from queues, these help in debugging from time to time.

2. Only a lot of logs if the queue had a lot of messages, on a legitimate run these are not a lot.

3. Need them for debugging right now, will remove then if left mundane

PR #914, opened 2025-12-19
Update Atlantic client to 0.14.1

prkpndy:
## Pull Request type

- [x] Feature

Atlantic client uses aggregator version `SnosAggregator0_13_3`.

Resolves: #NA

## What is the new behavior?

- Update Atlantic aggregator version to `SnosAggregator0_14_1`
- Add `da_public_keys` parameter support for bucket creation
- Add `MADARA_ORCHESTRATOR_DA_PUBLIC_KEYS` CLI argument

Yes - requires `MADARA_ORCHESTRATOR_DA_PUBLIC_KEYS` environment variable to be set.

## Other information

N/A

### DIFF
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -435,6 +435,7 @@ impl AtlanticClient {
         mock_proof: bool,
         chain_id_hex: Option<String>,
         fee_token_address: Option<Felt252>,
+        da_public_keys: Option<Vec<String>>,
     ) -> Result<AtlanticBucketResponse, AtlanticError> {
         let context = format!("mock_proof: {}, chain_id_hex: {:?}", mock_proof, chain_id_hex);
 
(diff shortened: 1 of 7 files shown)


orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -316,6 +318,7 @@ impl ProverClient for AtlanticProverService {
 }
 
 impl AtlanticProverService {
+    #[allow(clippy::too_many_arguments)]
```
byteZorvin, line 321: Nit: I would say its the right to time to move to a struct 😅

byteZorvin (approved):
Looks good, lets go 14.1 🚀

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -454,12 +455,13 @@ impl AtlanticClient {
                 let bucket_request = AtlanticCreateBucketRequest {
                     external_id: None,
                     node_width: None,
-                    aggregator_version: AtlanticAggregatorVersion::SnosAggregator0_13_3,
+                    aggregator_version: AtlanticAggregatorVersion::SnosAggregator0_14_1,
```
heemankv, line 458: This should be dynamic right ?
for 0.14.0 and 0.14.1 support via the block information ? 
ideally we should have the version maintained in the agg that it gets from the blocks and then accordingly set the value here, that will be a better design

heemankv (changes requested):

PR #920, opened 2025-12-23
fix: sdd support for atlantic-provided DA segment for L2 state updates

Mohiiit:
## Summary

Aggregator 0.14.1+ encrypts the state diff internally, causing a KZG commitment mismatch when using locally-generated blob data. This PR adds support for using the DA segment provided by the aggregator's trace generation instead.

## Problem
The orchestrator was generating blob data locally from state updates, but the aggregator now encrypts the state diff with a symmetric key before creating KZG commitments. This caused y_0 value mismatches during state update verification.

## Solution

Instead of encrypting locally (which would require access to the symmetric key and random data), we now:

1.  Receive the pre-FFT DA segment as an artifact from the aggregator's trace generation
2. Parse and transform it using the new `da_segment_to_blobs` function
3. Use this for KZG proof verification in L2 state updates


orchestrator/src/worker/utils/encrypted_blob.rs
```diff
@@ -0,0 +1,128 @@
...
+/// 1. Pads to BLOB_LEN (4096) boundary
+/// 2. Applies FFT transformation
+/// 3. Converts to blob bytes
+///
+/// # Arguments
+/// * `da_segment` - The DA segment as a vector of Felt252
+///
+/// # Returns
+/// * `Vec<Vec<u8>>` - A vector of blobs, each blob is 131072 bytes (4096 * 32)
+pub fn da_segment_to_blobs(da_segment: Vec<Felt252>) -> Result<Vec<Vec<u8>>, DaSegmentError> {
```
prkpndy, line 50: I think we already have this method - convert_felt_vec_to_blob_data - in `orchestrator/src/compression/blob/mod.rs`

Mohiiit: hmm I see there would be some dead code, I am planning to remove this part in a dedicated effort 

also I have added the test for now, which ensures that code is not broken

orchestrator/src/tests/artifacts/da_blob_index_1.json
```diff
@@ -0,0 +1,188 @@
+[
```
prkpndy: we can put the artifacts in https://github.com/madara-alliance/madara-test-artifacts repo like we are doing for compression tests? this will unnecessary bloat the main repo

Mohiiit: interesting, how are we using this repo tho? can you share the related link?

orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -107,8 +107,11 @@ pub struct AggregatorMetadata {
     /// downloaded. If `Some(value)`, the proof will be downloaded and stored to the specified path
     /// in the provided storage.
     pub download_proof: Option<String>,
-    /// Path of blob data
+    /// Path of blob data (legacy, used for L3)
     pub blob_data_path: String,
+    /// Path to the DA segment file from the prover (used for L2 with encryption)
+    /// TODO (mohit 23/12/2025): verify the actual file structure with atlantic team
```
prkpndy: can we remove this TODO now?

orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -107,8 +107,11 @@ pub struct AggregatorMetadata {
...
+    pub da_segment_path: String,
```
prkpndy, line 113: I guess, `encrypted_blob_data_path` would be a better name?

prkpndy: also, this would be the path of the directory containing many blobs, right?

orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -198,8 +201,13 @@ pub struct StateUpdateMetadata {
...
     /// Paths to program output files for each block/batch
     pub program_output_paths: Vec<String>,
-    /// Paths to blob data files for each block/batch
+    /// Paths to blob data files for each block/batch (legacy, used for L3)
     pub blob_data_paths: Vec<String>,
+    /// Paths to DA segment files for each batch (used for L2 with encryption)
+    /// Note: One DA segment per aggregator job, one state update per aggregator job
+    /// TODO (mohit 23/12/2025): verify file structure with atlantic team
+    #[serde(default)]
+    pub da_segment_paths: Vec<String>,
```
prkpndy, line 209:
```suggestion
    pub encrypted_blob_data_paths: Vec<String>,
```

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -40,13 +40,22 @@ impl JobHandlerTrait for StateUpdateJobHandler {
...
+        // L3: requires program_output_paths + blob_data_paths + snos_output_paths
+        if state_metadata.program_output_paths.is_empty() {
+            error!("program_output_paths is required for all state updates");
+            return Err(JobError::Other(OtherError(eyre!("Missing required program_output_paths in metadata"))));
+        }
+
+        let is_l2_config = !state_metadata.da_segment_paths.is_empty();
+        let is_l3_config = !state_metadata.blob_data_paths.is_empty() && !state_metadata.snos_output_paths.is_empty();
+
+        if !is_l2_config && !is_l3_config {
```
byteZorvin, line 54: Should we assert that only either L2 case or L3 case is present, could lead to undefined behaviour

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -119,6 +120,18 @@ impl JobHandlerTrait for AggregatorJobHandler {
                 )
                 .await?;
 
+                // Fetch DA segment from prover and store it in storage
+                // The DA segment contains the encrypted/compressed state diff for KZG proof
+                if !metadata.da_segment_path.is_empty() {
```
byteZorvin: this should be always true right, since in aggregator worker we populate the path

byteZorvin (approved):
Looks good, gg

orchestrator/src/types/jobs/metadata/mod.rs
Mohiiit: it's not necessary that the data would be encrypted tho!

orchestrator/src/types/jobs/metadata/mod.rs
Mohiiit: this would be one file only, we have to split it ourselves, afaik

orchestrator/src/worker/event_handler/jobs/state_update.rs
Mohiiit: this whole code piece requires some cleaning, I agree, I think the current code will remain the same, although while creating the job itself, the worker should ensure, based on the layer, which path to fill!

orchestrator/src/types/jobs/metadata/mod.rs
prkpndy: Got it

orchestrator/src/types/jobs/metadata/mod.rs
prkpndy: Got it. I thought we were getting blobs directly from Atlantic. I guess they are giving us a vector of felts, and we have to convert it to a blob ourselves? But in case we provide the public keys, the vector of felts would be encrypted, right? (I guess they are doing squashing -> stateful comp -> stateless comp -> encryption. Then we'll convert to blob data and do the state update. Right?)

prkpndy (approved):
LFG 🚀

PR #926, opened 2025-12-27
fix: cairo-native path + runtime issue + madara version update

Mohiiit:
1. cairo-native classes will use the same base path as madara-db now
2. cairo-native compilation requires tokio runtime to compile, it was not present in certain cases, that's been resolved now
3. madara version was still 0.8.0, moved that to 0.10.0
4. cairo-native version moved from 0.7.1 to 0.7.3, as 0.7.2 has some known bugs
5. madara's blockifier moved to the same tag as orch blockifier (same as snos one)


prkpndy (approved):
LGTM

madara/crates/client/cairo_native/src/compilation.rs
```diff
@@ -595,16 +621,31 @@ pub(crate) fn spawn_compilation_if_needed(
...
         Err(_) => {
-            tracing::debug!(
-                target: "madara_cairo_native",
-                class_hash = %format!("{:#x}", class_hash.to_felt()),
-                "compilation_async_no_runtime_context"
-            );
-            remove_compilation_in_progress(&class_hash);
-            return;
+            // No runtime in current thread - try stored handle (for non-Tokio thread contexts)
+            match get_stored_runtime_handle() {
```
byteZorvin, line 629: Nit: would have preferred `get_or_init`

byteZorvin (approved):

madara/crates/client/cairo_native/src/compilation.rs
Mohiiit: it is deliberately done, to ensure we are calling the exact functionality!

PR #927, opened 2025-12-29
chore: add message retention period (TTL) for worker trigger queue

prkpndy:
## Summary
- Adds configurable message retention period (TTL) support for SQS queues
- Sets 60 second TTL for WorkerTrigger queue to auto-expire stale worker trigger messages
- Worker triggers are time-sensitive and stale triggers should not be processed

## Changes
- Added `message_retention_period: Option<u32>` field to `QueueConfig`
- Updated `create_queue` to accept and set `MessageRetentionPeriod` SQS attribute
- Updated SQS setup to pass message retention period when creating queues
- Added test verification for message retention period

## Test plan
- [x] Existing SQS setup tests pass
- [x] New test verification confirms 60s TTL is set for WorkerTrigger queue
- [x] `cargo test -p orchestrator --lib tests::setup::sqs` passes

### DIFF
--- a/orchestrator/src/setup/aws/sqs.rs
+++ b/orchestrator/src/setup/aws/sqs.rs
@@ -60,7 +60,9 @@ impl Resource for InnerSQS {
                     let queue_name = InnerSQS::get_queue_name_from_type(name, queue_type);
 
                     // Creating the queue
-                    let queue_url = self.create_queue(queue_name.clone(), queue.visibility_timeout).await?;
+                    let queue_url = self
+                        .create_queue(queue_name.clone(), queue.visibility_timeout, queue.message_retention_period)
+                        .await?;
 
                     tracing::info!("Queue created for type {}", queue_type);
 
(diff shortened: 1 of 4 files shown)


orchestrator/src/setup/aws/sqs.rs
```diff
@@ -78,19 +80,18 @@ impl Resource for InnerSQS {
...
-                                        .get(&dlq_config.dlq_name)
-                                        .ok_or_else(|| {
-                                            OrchestratorError::SetupError(format!(
-                                                "Failed to get DLQ {} in QUEUES",
-                                                &dlq_config.dlq_name
-                                            ))
-                                        })?
-                                        .visibility_timeout,
+                                    dlq_queue_config.visibility_timeout,
+                                    dlq_queue_config.message_retention_period,
```
heemankv, line 94: Just to confirm, we need to add message retention period to dlq as well ? 
Hoq does this work ?

prkpndy: No. we don't have retention period for DLQs. Only for worker trigger. It's just getting the field from QUEUES. The value there is None

heemankv (approved):

byteZorvin (approved):
Looks good, thanks for this @prkpndy 🙌

PR #929, opened 2025-12-30
feat: RocksDB write stall prevention and monitoring

prkpndy:
## Summary

- Add write stall prevention settings to RocksDB configuration
- Add monitoring metrics for detecting write stalls via OpenTelemetry
- Make new settings configurable via CLI/ENV variables
- Tune defaults for ~20 GiB volumes

### Write Stall Prevention Settings
Added configurable thresholds to prevent RocksDB from stalling writes:

| Setting | Default | Description |
|---------|---------|-------------|
| `max_write_buffer_number` | 5 | Max memtables before stall |
| `level_zero_slowdown_writes_trigger` | 20 | L0 files to start throttling |
| `level_zero_stop_writes_trigger` | 36 | L0 files to stop writes |
| `soft_pending_compaction_bytes_limit` | 6 GiB | Soft limit for pending compaction |
| `hard_pending_compaction_bytes_limit` | 12 GiB | Hard limit for pending compaction |

### New CLI Flags
```
--db-max-write-buffer-number
--db-l0-slowdown-trigger
--db-l0-stop-trigger
--db-soft-pending-compaction-gib
--db-hard-pending-compaction-gib
```

### Monitoring Metrics
New OpenTelemetry metrics for write stall detection:

| Metric | Description |
|--------|-------------|
| `db_is_write_stopped` | 1 if writes blocked, 0 otherwise |
| `db_pending_compaction_bytes` | Bytes waiting for compaction |
| `db_l0_files_count` | Number of L0 files |
| `db_num_immutable_memtables` | Memtables waiting to flush |
| `db_memtable_size_bytes` | Total memtable size |

### Alerting Thresholds

| Metric | Warning | Critical |
|--------|---------|----------|
| `db_is_write_stopped` | - | = 1 |
| `db_pending_compaction_bytes` | > 4 GiB | > 6 GiB |
| `db_l0_files_count` | >= 15 | >= 20 |
| `db_num_immutable_memtables` | >= 3 | >= 4 |

## Test plan

- [ ] Verify node starts with default settings
- [ ] Verify CLI flags override defaults
- [ ] Verify metrics are exported via Prometheus/OTLP endpoints
- [ ] Monitor metrics during sync to validate thresholds


madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
             backup_dir: None,
             restore_from_latest_backup: false,
             write_mode: DbWriteMode::default(),
+            // Write stall prevention defaults (tuned for ~20 GiB volumes)
+            max_write_buffer_number: 5,
+            level_zero_slowdown_writes_trigger: 20,
+            level_zero_stop_writes_trigger: 36,
+            soft_pending_compaction_bytes_limit: 6 * GiB,  // ~30% of 20 GiB
```
byteZorvin, line 263: I would say we should have this 65%, and the hard limit be 85. 30 seems too small to bump the disk volume to avoid slow writes.
Or may be I am missing the dynamics/nuances completely here

madara/crates/client/db/src/rocksdb/metrics.rs
```diff
@@ -1,18 +1,53 @@
...
+//!
+//! These metrics are exported via OpenTelemetry and can be scraped by Prometheus
+//! or pushed to an OTLP endpoint.
+//!
+//! ## Write Stall Detection
+//!
+//! Key metrics for detecting write stalls:
+//! - `db_is_write_stopped`: 1 if writes are completely blocked
+//! - `db_pending_compaction_bytes`: Backlog of data waiting to be compacted
+//! - `db_l0_files_count`: Number of L0 files (slowdown at 20, stop at 36)
```
Mohiiit: is this just an example? values doesn't match with the values present below!

madara/crates/client/db/src/rocksdb/metrics.rs
```diff
@@ -1,18 +1,53 @@
...
     pub cache_total: Gauge<u64>,
+
+    // Write stall detection metrics
+    pub is_write_stopped: Gauge<u64>,
+    pub pending_compaction_bytes: Gauge<u64>,
+    pub l0_files_count: Gauge<u64>,
+
+    // MemTable monitoring
+    pub num_immutable_memtables: Gauge<u64>,
+    pub memtable_size_bytes: Gauge<u64>,
```
Mohiiit: nit: memory monitoring and memory metrics are from memtable right? maybe we should bring it under one topic? more readable

madara/crates/client/db/src/rocksdb/metrics.rs
```diff
@@ -21,58 +56,125 @@ impl DbMetrics {
...
             "".to_string(),
         );
 
         let column_sizes = register_gauge_metric_instrument(
             &meter,
             "column_sizes".to_string(),
-            "Sizes of RocksDB columns".to_string(),
+            "Size of each RocksDB column family in bytes".to_string(),
             "".to_string(),
         );
```
Mohiiit, line 77: will bytes be changed to something more readable like MB/GB?

madara/crates/client/db/src/rocksdb/metrics.rs
```diff
@@ -21,58 +56,125 @@ impl DbMetrics {
...
+            "db_pending_compaction_bytes".to_string(),
+            "Estimated bytes pending compaction (slowdown at 64GiB, stop at 256GiB)".to_string(),
             "".to_string(),
         );
 
-        Ok(Self { db_size, column_sizes, mem_table_total, mem_table_unflushed, mem_table_readers_total, cache_total })
+        let l0_files_count = register_gauge_metric_instrument(
+            &meter,
+            "db_l0_files_count".to_string(),
+            "Number of files at Level 0 (slowdown at 20, stop at 36)".to_string(),
```
Mohiiit: I think we should use const values or cli args maybe (with default value for now), easy to miss when we update the level 0 files config!

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -1,6 +1,100 @@
...
+//! ┌──────────────────────────────────────────────────────────────────────────────┐
+//! │                                  MEMORY                                      │
+//! ├──────────────────────────────────────────────────────────────────────────────┤
+//! │                                                                              │
+//! │   ┌─────────────────┐    ┌──────────────────┐    ┌──────────────────┐        │
+//! │   │ Active MemTable │    │Immutable MemTable│    │Immutable MemTable│        │
+//! │   │ (writes here)   │    │ (waiting flush)  │    │ (waiting flush)  │        │
+//! │   └─────────────────┘    └──────────────────┘    └──────────────────┘        │
+//! │          │                        │                      │                   │
+//! │          │         max_write_buffer_number = 5           │                   │
```
Mohiiit, line 22: clarification: 5 is the default value?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -1,6 +1,100 @@
...
+//!
+//! ### 3. Compaction Limits
+//! - `soft_pending_compaction_bytes_limit`: Soft limit for pending compaction
+//! - `hard_pending_compaction_bytes_limit`: Hard limit (writes stop)
+//!
+//! ### 4. I/O Smoothing
+//! - `bytes_per_sync`: Incremental sync during SST writes (prevents I/O spikes)
+//! - `enable_pipelined_write`: Parallel WAL + memtable writes
+//!
+//! ## References
```
Mohiiit, line 91: nice!!

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -112,6 +214,34 @@ pub struct RocksDBConfig {
...
+
+    /// Number of L0 files that triggers write slowdown.
+    /// When L0 file count reaches this, writes are throttled.
+    /// Default: 20
+    pub level_zero_slowdown_writes_trigger: i32,
+
+    /// Number of L0 files that triggers complete write stop.
+    /// When L0 file count reaches this, writes are blocked until compaction catches up.
+    /// Default: 36
+    pub level_zero_stop_writes_trigger: i32,
```
Mohiiit, line 233: any reason for these to be i32 tho?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
             backup_dir: None,
             restore_from_latest_backup: false,
             write_mode: DbWriteMode::default(),
+            // Write stall prevention defaults (tuned for ~20 GiB volumes)
+            max_write_buffer_number: 5,
+            level_zero_slowdown_writes_trigger: 20,
+            level_zero_stop_writes_trigger: 36,
+            soft_pending_compaction_bytes_limit: 6 * GiB,  // ~30% of 20 GiB
+            hard_pending_compaction_bytes_limit: 12 * GiB, // ~60% of 20 GiB
```
Mohiiit, line 264: clarification: so if we have a system running on 500GB, setting these values to hardcoded 6 and 12 won't make sense right? 

idea is to update the arguments for specific system? or shall we have a % here instead of numbers?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+    // while the immutable one is still flushing, writes stall immediately.
+    //
+    // Setting to 5 allows 4 immutable memtables to queue while flushing,
+    // providing much more headroom during write bursts.
+    options.set_max_write_buffer_number(config.max_write_buffer_number);
+
+    // Merge 2 memtables before flushing to L0.
+    // This reduces L0 file count (fewer files = less read amplification)
+    // and removes duplicate keys, reducing overall data size.
+    options.set_min_write_buffer_number_to_merge(2);
```
Mohiiit, line 386: clarification: what's the default value? and how changing(increasing the value) changes things in compaction?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+    // ═══════════════════════════════════════════════════════════════════════════
+    // WRITE STALL PREVENTION - PENDING COMPACTION LIMITS
+    // ═══════════════════════════════════════════════════════════════════════════
+    // When compaction falls behind, uncompacted data accumulates.
+    // These limits prevent unbounded growth of pending compaction work.
+    //
+    // Soft limit: Start slowing writes when pending compaction exceeds the limit.
+    // This provides early warning before hitting the hard limit.
+    // Default: 6 GiB (tuned for ~20 GiB volumes, ~30% of capacity)
+    options.set_soft_pending_compaction_bytes_limit(config.soft_pending_compaction_bytes_limit);
```
Mohiiit: in the top we are calculating the number of cores we have in the system, similarity maybe we can get the pvc size as well and instead of a number have a x% and use that here? 

ex: 

soft_pending_compaction_limit = 30% 
so upon setting this value on y size, it would be x*0.3 and we can always tweak this later as per the performance

for me, it reduces the human error

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+    // Without incremental sync, RocksDB writes entire SST files (potentially
+    // hundreds of MB) before syncing, causing I/O spikes that can stall writes.
+    //
+    // bytes_per_sync: Sync every 1MB during SST file writes.
+    // This spreads I/O evenly, preventing spikes.
+    options.set_bytes_per_sync(1 * MiB as u64);
+
+    // wal_bytes_per_sync: Sync WAL every 512KB.
+    // WAL writes are smaller but more frequent; smaller sync interval is appropriate.
+    options.set_wal_bytes_per_sync(512 * KiB as u64);
```
Mohiiit, line 427: this should be part of the config as well?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+
+    // ═══════════════════════════════════════════════════════════════════════════
+    // I/O SMOOTHING
+    // ═══════════════════════════════════════════════════════════════════════════
+    // Without incremental sync, RocksDB writes entire SST files (potentially
+    // hundreds of MB) before syncing, causing I/O spikes that can stall writes.
+    //
+    // bytes_per_sync: Sync every 1MB during SST file writes.
+    // This spreads I/O evenly, preventing spikes.
+    options.set_bytes_per_sync(1 * MiB as u64);
```
Mohiiit, line 423: same as wal_bytes_per_sync, should be part of config?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+    //
+    // Soft limit: Start slowing writes when pending compaction exceeds the limit.
+    // This provides early warning before hitting the hard limit.
+    // Default: 6 GiB (tuned for ~20 GiB volumes, ~30% of capacity)
+    options.set_soft_pending_compaction_bytes_limit(config.soft_pending_compaction_bytes_limit);
+
+    // Hard limit: Stop writes completely when pending compaction is too high.
+    // This is a safety valve to prevent disk exhaustion.
+    // Default: 12 GiB (tuned for ~20 GiB volumes, ~60% of capacity)
+    options.set_hard_pending_compaction_bytes_limit(config.hard_pending_compaction_bytes_limit);
```
Mohiiit, line 413: should be calculated dynamically!

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+
+    // ═══════════════════════════════════════════════════════════════════════════
+    // PIPELINED WRITES
+    // ═══════════════════════════════════════════════════════════════════════════
+    // By default, WAL write and memtable insert are sequential.
+    // Pipelining allows the next write to start its WAL write while the
+    // previous write is still doing memtable insert.
+    //
+    // This improves throughput for concurrent writers.
+    options.set_enable_pipelined_write(true);
```
Mohiiit: config for this as well?

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +261,281 @@ impl Default for RocksDBConfig {
...
+
+    // ═══════════════════════════════════════════════════════════════════════════
+    // LOGGING & FILE MANAGEMENT
+    // ═══════════════════════════════════════════════════════════════════════════
     options.set_max_log_file_size(10 * MiB);
     options.set_max_open_files(2048);
     options.set_keep_log_file_num(3);
     options.set_log_level(rocksdb::LogLevel::Warn);
```
Mohiiit, line 438: maybe we can add these to the config as well, with these as the default values?

Mohiiit (changes requested):
mostly clarification! a few changes!

madara/crates/client/db/src/rocksdb/metrics.rs
prkpndy: Metrics are exported in bytes as a convention for Prometheus/OTLP - this is the standard practice. Unit conversion to MB/GB should be done in the visualization layer (Grafana). Grafana has built-in unit formatters

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: Yes, 5 is the default value for max_write_buffer_number. RocksDB default is 2, but we increased it to 5 to provide more headroom during write bursts.

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: :)

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: These are i32 because the RocksDB Rust bindings expect i32 for these methods

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: Correct. These are configurable via CLI. Defaults are for 20Gi which is what we use as default in our K8s

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: Default for level_zero_slowdown_writes_trigger is 20. When L0 files reach this count, RocksDB starts throttling writes (slowing them down). Increasing this value delays the slowdown, allowing faster writes but risking more L0 files accumulating (which hurts read performance since L0 files overlap).

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: Yeah. I thought about it earlier, but getting the mounted volume's size, especially in a container is not very straighforward in rust - and also not very reliable (acc to Claude). Have we done this anywhere in our code? (maybe in migration code?)

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: Checked about it - seems like these are fairly standard values and rarely needs tuning

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: same

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: These are RocksDB internal log settings (not application logs). They control the LOG file in the RocksDB data directory. I don't think we would ever need to configure these

madara/crates/client/db/src/rocksdb/metrics.rs
prkpndy: fixed the comments

madara/crates/client/db/src/rocksdb/metrics.rs
prkpndy: good catch. updated 🙏

madara/crates/client/db/src/rocksdb/metrics.rs
prkpndy: Yeah, missed this. I'll remove the values from description. Not really needed

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: bytes_per_sync controls how frequently RocksDB syncs data to disk during SST file writes to prevent I/O spikes. It's not related to disk size - it's about smoothing I/O patterns.

Without it: RocksDB writes a 256MB SST file, then does one massive sync → I/O spike.
With bytes_per_sync=1MB: RocksDB syncs every 1MB during the write → smooth I/O, no spikes.

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: enable_pipelined_write allows concurrent WAL and memtable writes, improving throughput. It's enabled by default and there's rarely a reason to disable it. I don't think disabling it would offer any benifit (cost would be decreased throughput)

prkpndy: enable_pipelined_write allows concurrent WAL and memtable writes, improving throughput. It's enabled by default and there's rarely a reason to disable it. I don't think disabling it would offer any benifit (cost would be decreased throughput)

prkpndy (commented):
resolved comments

madara/crates/client/db/src/rocksdb/options.rs
Mohiiit: we have a code which checks the disk space yes! in migration before making the copy of the db!

madara/crates/client/db/src/rocksdb/options.rs
Mohiiit: decreasing this means the compaction will slow down the writes? but given it's just 2 files, it would be faster to compact?

madara/crates/client/db/src/rocksdb/options.rs
Mohiiit: the comment is about something else I believe

madara/crates/client/db/src/rocksdb/options.rs
Mohiiit: is it enabled by default?

prkpndy: yups

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: The above clarification was about some other setting.
Regarding this, this specifies that we merge two memtables before putting them in L0 files. This reduces L0 file count (help in preventing/delaying write stall because of that) and also remove duplicate keys (reduce this load later on when L0 files are merged) and also increases read speed

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: yes, my bad. so this is about the same thing, right? (calculating this based on the volume size and giving percentages in cli)

Mohiiit (approved):

byteZorvin (approved):

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -131,56 +257,280 @@ impl Default for RocksDBConfig {
...
+
+    // ═══════════════════════════════════════════════════════════════════════════
+    // PIPELINED WRITES
+    // ═══════════════════════════════════════════════════════════════════════════
+    // By default, WAL write and memtable insert are sequential.
+    // Pipelining allows the next write to start its WAL write while the
+    // previous write is still doing memtable insert.
+    //
+    // This improves throughput for concurrent writers.
+    options.set_enable_pipelined_write(true);
```
byteZorvin: This was giving conflicts to me, with wal afaik i remember

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: these limits shouldn't depend on disk size actually. these are dependent on rocksdb's performance, consistency and incoming traffic mainly

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: fixed!

PR #930, opened 2025-12-30
Orchestrator: Rework(db): internal_id -> string to u64

heemankv:
Pull Request type

- [x] Build-related changes

`internal_id` field in jobs is stored as `String` type, which causes incorrect sorting (lexicographic instead of numeric) and inefficient queries requiring `$toInt`/`$toLong` conversions.

Resolves: #NA

## What is the new behavior?

- Changed `internal_id` from `String` to `u64` across the codebase
- Added custom serde module for MongoDB compatibility (BSON only supports signed i64)
- Simplified MongoDB queries by removing `$toInt`/`$toLong` conversions
- Added migration script to convert existing data from String to NumberLong
- Moved index creation from JS migrations to Rust `ensure_indexes()` (called during `orchestrator setup`)
- Added `--mongodb` as required argument for `orchestrator setup` command
- Enabled migration locking via `migrate-mongo` (`lockTtl: 60`) for safe concurrent deployments
- Docker now runs migrations automatically on startup

Yes

**Migration steps:**

1. **Run `orchestrator setup` first** (creates indexes):
   ```bash
   cargo run --package orchestrator -- setup \
     --mongodb \
     --mongodb-connection-url "mongodb://<host>:<port>" \
     --mongodb-database-name "<database>" \
     --aws \
     # ... other required args
   ```

2. **Run data migrations** (converts internal_id String → NumberLong):
   ```bash
   cd orchestrator
   npm install
   MADARA_ORCHESTRATOR_MONGODB_CONNECTION_URL="mongodb://<host>:<port>" \
   MADARA_ORCHESTRATOR_DATABASE_NAME="<database>" \
   npm run migrate
   ```

3. **Docker (automatic):** Migrations run automatically on container startup:
   ```bash
   docker run \
     -e MADARA_ORCHESTRATOR_MONGODB_CONNECTION_URL="mongodb://<host>:<port>" \
     -e MADARA_ORCHESTRATOR_DATABASE_NAME="<database>" \
     ghcr.io/madara-alliance/orchestrator:<tag>
   ```

## Other information

**Migration script (`00000000000001-internal-id-to-long.js`):**
- Converts all `internal_id` values from String to NumberLong
- Includes rollback support via `down()` function
- Tracked in `changelog` collection (idempotent - safe to run multiple times)
- Migration locking prevents race conditions when multiple orchestrators start simultaneously

**Index management:**
- Indexes are now created by Rust code in `MongoDbClient::ensure_indexes()`
- Called during `orchestrator setup` (required before first run)
- Also called during `orchestrator run` to ensure indexes exist
- Deleted `00000000000000-init.js` migration (indexes no longer managed via JS)
### First run on the database : 
image

### Second run on the database : 
Screenshot 2025-12-30 at 1 10 39 PM

### Here's how the db collection look like :
Screenshot 2025-12-30 at 1 11 30 PM

### Here's how the migration status looks like :
image


heemankv (commented):
Self review done, lgtm

orchestrator/src/types/jobs/job_item.rs
```diff
@@ -7,13 +7,41 @@ use mongodb::bson::serde_helpers::{chrono_datetime_as_bson_datetime, uuid_1_as_b
...
+mod u64_as_bson_int64 {
+    use serde::{self, Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S>(value: &u64, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // Safe to cast since block/batch numbers won't exceed i64::MAX
+        serializer.serialize_i64(*value as i64)
+    }
```
byteZorvin, line 25:
```suggestion
pub fn serialize<S>(value: &u64, serializer: S) -> Result<S::Ok, S::Error>                                                                                                                                                                                                                                                              
  where                                                                                                                                                                                                                                                                                                                                   
      S: Serializer,                                                                                                                                                                                                                                                                                                                      
  {                                                                                                                                                                                                                                                                                                                                       
      if *value > i64::MAX as u64 {                                                                                                                                                                                                                                                                                                       
          return Err(serde::ser::Error::custom(                                                                                                                                                                                                                                                                                           
              format!("internal_id {} exceeds i64::MAX", value)                                                                                                                                                                                                                                                                           
          ));                                                                                                                                                                                                                                                                                                                             
      }                                                                                                                                                                                                                                                                                                                                   
      serializer.serialize_i64(*value as i64)                                                                                                                                                                                                                                                                                             
  } 
```

.gitignore
```diff
@@ -113,3 +113,4 @@ sequencer_venv/
 sequencer_requirements.txt
 # AI-generated documentation
 AI_docs/
+orchestrator/package-lock.json
```
byteZorvin, line 116: Why not have a package lock if we are adding the js file to git

orchestrator/Dockerfile
```diff
@@ -67,25 +67,25 @@ FROM base-rust AS planner
...
-    fi && \
-    if [ -n "$REQUIREMENTS_FILE" ] && [ -f "$REQUIREMENTS_FILE" ]; then \
-        echo "✓ Found requirements.txt at: $REQUIREMENTS_FILE"; \
-        cp "$REQUIREMENTS_FILE" /app/sequencer-requirements.txt; \
-        cat /app/sequencer-requirements.txt; \
-    else \
-        echo "ERROR: Could not find requirements.txt in cargo checkouts"; \
-        exit 1; \
-    fi
+  cargo fetch && \
```
byteZorvin, line 75: is this a lint, or some local config?

orchestrator/migrations/00000000000000-init.js
```diff
@@ -1,14 +1,13 @@
...
-      // primarily for get_jobs_without_successor
-      { key: { job_type: 1, status: 1 } },
-    ]);
+    const jobs = db.collection("jobs");
+    await jobs.createIndex({ id: 1 });
+    await jobs.createIndex({ job_type: 1, internal_id: -1 }, { unique: true });
+    await jobs.createIndex({ job_type: 1, status: 1, internal_id: -1 });
+    await jobs.createIndex({ status: 1 });
+    // primarily for get_jobs_without_successor
+    await jobs.createIndex({ job_type: 1, status: 1 });
```
byteZorvin: This looks redudant, as being a prefix of  `await jobs.createIndex({ job_type: 1, status: 1, internal_id: -1 });`

.gitignore
heemankv: as far as I know, pushing lock file is not recommended till you have brittle dependencies that work with specific versions only, here we don't have that case hence no point in pushing the lock file

orchestrator/Dockerfile
heemankv: lint

.gitignore
byteZorvin: I would still prefer having it anyways; keeps things safe and consistent everywhere, any downside?

byteZorvin (approved):

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -133,14 +133,13 @@ impl JobHandlerTrait for StateUpdateJobHandler {
...
             // Get the artifacts for the block/batch
-            let snos_output =
-                match fetch_snos_for_block(internal_id.clone(), i, config.clone(), &snos_output_paths).await {
-                    Ok(snos_output) => Some(snos_output),
-                    Err(err) => {
-                        debug!("failed to fetch snos output, proceeding without it: {}", err);
-                        None
-                    }
-                };
+            let snos_output = match fetch_snos_for_block(*internal_id, i, config.clone(), &snos_output_paths).await {
```
prkpndy: why not use the internal id directly instead of copying a reference of it in the start and dereferencing it here?

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -238,7 +237,7 @@ impl StateUpdateJobHandler {
         config: &Arc<Config>,
         nums_settled: &[u64],
         job_id: &Uuid,
-        internal_id: &str,
+        internal_id: &u64,
```
prkpndy: why take a reference of u64? copying it would make the code cleaner imo

orchestrator/src/worker/event_handler/jobs/snos.rs
```diff
@@ -175,7 +175,7 @@ impl JobHandlerTrait for SnosJobHandler {
 
         debug!("Storing SNOS outputs");
         // Store the Cairo Pie path
-        self.store(internal_id.clone(), config.storage(), &snos_metadata, cairo_pie, os_output, program_output).await?;
+        self.store(*internal_id, config.storage(), &snos_metadata, cairo_pie, os_output, program_output).await?;
```
prkpndy: same

orchestrator/src/worker/event_handler/jobs/proof_registration.rs
```diff
@@ -56,7 +56,7 @@ impl JobHandlerTrait for RegisterProofJobHandler {
         // Format proof for submission
         let formatted_proof = format!("{{\n\t\"proof\": {}\n}}", proof);
 
-        let task_id = job.internal_id.clone();
+        let task_id = job.internal_id.to_string();
```
prkpndy: hmm. we can actually send the u64 as the task id, no?

prkpndy: also, a nit but some places, we are using both internal id copied form job.internal_id in the beginning and then also using job.internal_id directly in the same function (like in this function). can be made cleaner

orchestrator/migrations/00000000000000-init.js
```diff
@@ -1,14 +1,12 @@
...
     // Create indexes for the 'jobs' collection
-    await db.collection("jobs").createIndexes([
-      { key: { id: 1 } },
-      { key: { job_type: 1, internal_id: -1 }, unique: true },
-      { key: { job_type: 1, status: 1, internal_id: -1 } },
-      { key: { status: 1 } },
-      // primarily for get_jobs_without_successor
-      { key: { job_type: 1, status: 1 } },
-    ]);
+    const jobs = db.collection("jobs");
```
prkpndy: I think we should add indexes in the ensure_indexes function in config. That's where we are making sure that other indexes for DB are also present (this is what we discussed in standup as well afair)

orchestrator/Dockerfile
```diff
@@ -113,42 +113,65 @@ COPY build-artifacts build-artifacts
...
+RUN npm install --omit=dev
+
+# Create entrypoint script that runs migrations before starting orchestrator
+RUN echo '#!/bin/bash\n\
+  set -e\n\
+  \n\
+  # Run database migrations if MongoDB connection is configured\n\
+  if [ -n "$MADARA_ORCHESTRATOR_MONGODB_CONNECTION_URL" ] && [ -n "$MADARA_ORCHESTRATOR_DATABASE_NAME" ]; then\n\
+  echo "Running database migrations..."\n\
+  cd /app && npx migrate-mongo up\n\
```
prkpndy, line 169: Few questions:
1. Will we keep a track of migrations in the DB anywhere? Or will the library handle this automatically for us? If yes, afaik, ideally we should create new migration files and not edit the already existing ones. But you can check to confirm this
2. why use npx when we have installed the dependencies already above? I think we can run `npm run migrate` directly

orchestrator/migrations/00000000000000-init.js
prkpndy: also afaik, it's not a good practice to edit existing migration files. we should create new ones which will be run sequentially by the library (not sure tho. you can confirm the best practices)

prkpndy (changes requested):
Looks good overall. A few suggestions/clarifications

orchestrator/migrate-mongo-config.js
```diff
@@ -23,6 +23,13 @@ const config = {
   // The mongodb collection where the applied changes are stored. Only edit this when really necessary.
   changelogCollectionName: "changelog",
 
+  // The mongodb collection where the lock will be created.
+  lockCollectionName: "changelog_lock",
+
+  // The value in seconds for the TTL index that will be used for the lock.
+  // Value of 0 will disable the feature.
+  lockTtl: 60,
```
byteZorvin, line 31: isnt this too less ?
cc: @prkpndy

orchestrator/src/error/mod.rs
```diff
@@ -40,6 +41,9 @@ pub enum OrchestratorError {
     #[error("Database error: {0}")]
     DatabaseCoreError(#[from] DatabaseError),
 
+    #[error("Lock error: {0}")]
+    LockCoreError(#[from] LockError),
```
prkpndy: what's this for?

prkpndy: not sure if it's being used anywhere

orchestrator/src/setup/mod.rs
```diff
@@ -18,6 +21,9 @@ pub async fn setup(setup_cmd: &SetupCmd) -> OrchestratorResult<()> {
 
     info!("Setting up resources for Orchestrator...");
 
+    // Setup database indexes
+    setup_database(setup_cmd).await?;
```
prkpndy: nit: do we need to add indexes during setup tho? it's just adding an extra dependency during setup - and we'll always run the same piece of code during run as well

byteZorvin (approved):

orchestrator/src/types/params/database.rs
```diff
@@ -24,3 +24,20 @@ impl TryFrom<RunCmd> for DatabaseArgs {
         })
     }
 }
+
```
prkpndy: extra dependency/code for ensuring indexes during setup :)

orchestrator/src/cli/mod.rs
```diff
@@ -235,6 +235,12 @@ pub struct RunCmd {
             .multiple(false)
             .requires("provider")
     ),
+    group(
```
prkpndy: extra dependency :)

orchestrator/migrate-mongo-config.js
prkpndy: should be enough i think, but can be made higher to be on the safer side? how long did it take to migrate paradex's mock db @heemankv?

heemankv: This is lock for running the migration which usually completes in 5 seconds

heemankv: We can increase this over time based on the new migrations

Makefile
```diff
@@ -475,24 +475,24 @@ git-hook:
 	@git config core.hooksPath .githooks
 
 .PHONY: setup-l2
-setup-l2:
+setup-l2: setup-cairo
 	@echo -e "$(DIM)Setting up orchestrator with L2 layer...$(RESET)"
-	@cargo run --package orchestrator -- setup --layer l2 --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type schedule
+	@cargo run --package orchestrator --release -- setup --layer l2  --mongodb --aws --aws-s3 --aws-sqs --aws-sns --aws-event-bridge --event-bridge-type schedule
```
prkpndy: is `--release` needed for local development?

prkpndy (commented):
few nits. mostly looks good. gg!!

orchestrator/migrate-mongo-config.js
prkpndy: cool

orchestrator/src/error/mod.rs
heemankv:
```
 let lock_client = MongoLockClient::new(&db_args).await?;
    lock_client.ensure_indexes().await?;
 ```
 LockCoreError is used to map from LockError of mongodb

Makefile
heemankv: good catch, reverted

orchestrator/src/error/mod.rs
prkpndy: makes sense. extra dependency tho :)

orchestrator/src/cli/mod.rs
heemankv: Made optional

orchestrator/src/types/params/database.rs
heemankv: Made optional

orchestrator/src/setup/mod.rs
heemankv: IMO we should, setup should setup all the dependencies be it AWS / Mongodb anything.

Made optional

prkpndy (approved):
LGTM 🚀

PR #931, opened 2025-12-30
chore: optimize SQS client with queue URL caching

prkpndy:
## Summary

- Add queue URL caching to SQS client to avoid redundant `GetQueueUrl` API calls
- For ARN-based identifiers: URLs are computed directly without any API call
- For Name-based identifiers: URLs are fetched once via `GetQueueUrl` and cached per queue type
- Remove unnecessary clone in event worker

## Changes

**`orchestrator/src/core/client/queue/sqs.rs`**
- Add `queue_url_cache` field using `Arc<RwLock<HashMap<QueueType, Arc<OnceCell<String>>>>>`
- Add `get_or_resolve_queue_url()` helper method with fast path for cached URLs
- Update all `QueueClient` trait methods to use cached URLs

**`orchestrator/src/worker/controller/event_worker.rs`**
- Remove unnecessary `Arc` clone when calling methods on config

## Test plan

- [x] `cargo check` passes
- [x] `cargo clippy` passes
- [x] `cargo build --bin orchestrator` succeeds

### DIFF
--- a/orchestrator/src/core/client/queue/sqs.rs
+++ b/orchestrator/src/core/client/queue/sqs.rs
@@ -19,7 +19,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{OnceCell, RwLock};
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct InnerSQS(Client);
 
 impl InnerSQS {
@@ -158,16 +158,6 @@ pub struct SQS {
     queue_url_cache: Arc<RwLock<HashMap<QueueType, Arc<OnceCell<String>>>>>,
 }
 
-impl std::fmt::Debug for SQS {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("SQS")
-            .field("inner", &self.inner)
-            .field("queue_template_identifier", &self.queue_template_identifier)
-            .field("queue_url_cache", &"<cache>")
-            .finish()
-    }
-}
-
 impl SQS {
     /// new - Create a new SQS client, with both client and option for the client;
     /// we've needed to pass the aws_config and args to the constructor.
@@ -243,6 +233,13 @@ impl SQS {
 
     pub fn client(&self) -> &Client {
         self.inner.client()
+    }
+
+    /// Returns the number of queue URLs currently cached.
+    /// This is primarily useful for testing to verify caching behavior.
+    #[cfg(test)]
+    pub async fn cached_url_count(&self) -> usize {
+        self.queue_url_cache.read().await.len()
     }
 
     /// get_queue_name - Get the queue name
--- a/orchestrator/src/core/client/queue/sqs.rs
+++ b/orchestrator/src/core/client/queue/sqs.rs
@@ -206,6 +206,31 @@ impl SQS {
 
 #[async_trait]
 impl QueueClient for SQS {
+    /// TODO: if possible try to reuse the same producer which got created in the previous run
+    /// get_producer - Get the producer for the given queue
+    /// This function returns the producer for the given queue.
+    /// The producer is used to send messages to the queue.
+    async fn get_producer(&self, queue: QueueType) -> Result<SqsProducer, QueueError> {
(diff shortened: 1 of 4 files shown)


orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -142,10 +144,28 @@ impl InnerSQS {
...
+/// Queue URLs are cached per QueueType to avoid redundant GetQueueUrl API calls.
+/// - For ARN-based identifiers: URLs are computed directly without any API call
+/// - For Name-based identifiers: URLs are fetched once and cached
+#[derive(Clone)]
 pub struct SQS {
     pub inner: InnerSQS,
     queue_template_identifier: AWSResourceIdentifier,
+    /// Cache for queue URLs, keyed by QueueType.
+    /// Each entry is lazily initialized on first use.
+    queue_url_cache: Arc<RwLock<HashMap<QueueType, Arc<OnceCell<String>>>>>,
```
heemankv, line 158: Arc<OnceCell<String>>
any reason we can't make it
Arc<OnceCell<Url>> ?

orchestrator/src/core/client/queue/sqs.rs
```diff
@@ -142,10 +144,28 @@ impl InnerSQS {
...
+}
+
+impl std::fmt::Debug for SQS {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SQS")
+            .field("inner", &self.inner)
+            .field("queue_template_identifier", &self.queue_template_identifier)
+            .field("queue_url_cache", &"<cache>")
+            .finish()
+    }
```
heemankv: When is this being used ?

orchestrator/src/core/client/queue/sqs.rs
prkpndy: you mean why can't we use URL instead of string?
because we need to pass String and not URL to the aws sdk. Using URL would increase dependency and unnecessary parsing to and from URL to string

orchestrator/src/core/client/queue/sqs.rs
prkpndy: the struct had Debug derived so we needed to maintain that capability. with custom impl we ensure that the logs are not cluttered with the whole cache hash map

heemankv: then I would question where are we using SQS struct in debug, we ultimately do not want to leak the queue urls since it has account details over on to the logs

orchestrator/src/core/client/queue/sqs.rs
heemankv: Noted, thanks

orchestrator/src/core/client/queue/sqs.rs
heemankv: and queue_url_cache will leak the url to the logs

prkpndy: that's one more reason to not print it. and we are not printing it, no?

orchestrator/src/core/client/queue/sqs.rs
byteZorvin, line 158: Why do we need locks ?

byteZorvin (approved):

orchestrator/src/core/client/queue/sqs.rs
prkpndy: because it can be accessed (written to/read from) concurrently by the workers

heemankv (approved):

orchestrator/src/core/client/queue/sqs.rs
byteZorvin: Hmm, if we make it write once(during init) then locks wont be needed, that loooks simpler to me, wdyt

prkpndy: The SQS client is a generic client that does not have context of what all queues are available or are being used in the code. That's why I kept it this way. It has it's own caching layer to which it'll keep on storing stuff whenever it's asked to do something with a QueueType

PR #932, opened 2025-12-30
fix: rpc 10 state-update spec

Mohiiit:
1. removed unwanted bash scripts 
2. added tests to ensure migrated_classes is present


byteZorvin (approved):
super clean, lets goo !

madara/crates/tests/src/rpc/migration.rs
```diff
@@ -131,36 +131,32 @@ async fn test_migration_validation_state_update_intact() {
...
-    //     }
-    // }
+    /// Verify that state updates include the migrated_compiled_classes field.
+    ///
+    /// The field should be present in all state updates (empty or with values).
+    /// This test queries a block that is expected to have no migrations.
+    #[rstest]
+    #[tokio::test]
+    #[ignore = "Requires external madara instance on port 9944"]
+    async fn test_migration_validation_v9_migrated_classes_field() {
```
prkpndy, line 166: should it be v10?

madara/crates/tests/src/rpc/migration.rs
```diff
@@ -131,36 +131,32 @@ async fn test_migration_validation_state_update_intact() {
 
 /// Tests specific to v8 → v9 migration (SNIP-34: CASM hash migration)
```
prkpndy, line 152: v9 -> v10? not sure tbh

prkpndy (approved):
asked for a small clarification regarding test naming. otherwise looks good!

madara/crates/tests/src/rpc/migration.rs
Mohiiit: it's db version and not rpc version

madara/crates/tests/src/rpc/migration.rs
Mohiiit: same as above!

PR #934, opened 2026-01-05
feat: add versioning to orchestrator

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/orchestrator/src/core/client/database/mongodb.rs
+++ b/orchestrator/src/core/client/database/mongodb.rs
@@ -659,6 +659,7 @@ impl DatabaseClient for MongoDbClient {
         job_type: Vec<JobType>,
         job_status: Vec<JobStatus>,
         limit: Option<i64>,
+        orchestrator_version: Option<String>,
     ) -> Result<Vec<JobItem>, DatabaseError> {
         let start = Instant::now();
 
@@ -674,6 +675,10 @@ impl DatabaseClient for MongoDbClient {
         if !job_status.is_empty() {
             let serialized_statuses: Result<Vec<Bson>, _> = job_status.iter().map(bson::to_bson).collect();
             filter.insert("status", doc! { "$in": serialized_statuses? });
+        }
+
+        if let Some(version) = &orchestrator_version {
+            filter.insert("metadata.common.orchestrator_version", version.as_str());
         }
 
         let find_options = limit.map(|val| FindOptions::builder().limit(Some(val)).build());
@@ -1371,7 +1376,12 @@ impl DatabaseClient for MongoDbClient {
         Ok(results)
     }
 
-    async fn get_orphaned_jobs(&self, job_type: &JobType, timeout_seconds: u64) -> Result<Vec<JobItem>, DatabaseError> {
+    async fn get_orphaned_jobs(
+        &self,
+        job_type: &JobType,
+        timeout_seconds: u64,
+        orchestrator_version: Option<String>,
+    ) -> Result<Vec<JobItem>, DatabaseError> {
         let start = Instant::now();
 
         // Calculate the cutoff time (current time - timeout)
@@ -1379,13 +1389,17 @@ impl DatabaseClient for MongoDbClient {
 
         // Query for jobs of the specific type in LockedForProcessing status with
         // process_started_at older than cutoff
-        let filter = doc! {
+        let mut filter = doc! {
             "job_type": bson::to_bson(job_type)?,
             "status": bson::to_bson(&JobStatus::LockedForProcessing)?,
             "metadata.common.process_started_at": {
                 "$lt": cutoff_time.timestamp()
             }
         };
+
+        if let Some(version) = &orchestrator_version {
+            filter.insert("metadata.common.orchestrator_version", version.as_str());
+        }
 
         let jobs: Vec<JobItem> = self.get_job_collection().find(filter, None).await?.try_collect().await?;
 
--- a/orchestrator/src/core/client/database/mongodb.rs
+++ b/orchestrator/src/core/client/database/mongodb.rs
@@ -554,7 +554,7 @@ impl DatabaseClient for MongoDbClient {
         job_a_type: JobType,
         job_a_status: JobStatus,
         job_b_type: JobType,
-        orchestrator_version: Option<&str>,
+        orchestrator_version: Option<String>,
     ) -> Result<Vec<JobItem>, DatabaseError> {
         let start = Instant::now();
         // Convert enums to Bson strings
@@ -567,8 +567,8 @@ impl DatabaseClient for MongoDbClient {
             "job_type": job_a_type_bson,
             "status": job_a_status_bson,
         };
-        if let Some(version) = orchestrator_version {
-            match_filter.insert("metadata.common.orchestrator_version", version);
+        if let Some(version) = &orchestrator_version {
+            match_filter.insert("metadata.common.orchestrator_version", version.as_str());
         }
 
         // Construct the aggregation pipeline
@@ -622,7 +622,7 @@ impl DatabaseClient for MongoDbClient {
         job_type: JobType,
         job_status: JobStatus,
         internal_id: u64,
-        orchestrator_version: Option<&str>,
+        orchestrator_version: Option<String>,
     ) -> Result<Vec<JobItem>, DatabaseError> {
         let start = Instant::now();
         // Cast u64 to i64 because BSON only supports signed 64-bit integers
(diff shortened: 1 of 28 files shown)


orchestrator/src/types/constant.rs
Mohiiit, line 46: does it take the version 0.10.0? the release version?

Mohiiit (commented):
looks good, some tests would be nice 🙏

orchestrator/src/types/constant.rs
prkpndy: It'll take the version from version in `Cargo.toml` in `orchestrator` directory. i.e. `0.1.0`

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -725,6 +748,31 @@ impl DatabaseClient for MongoDbClient {
         Ok(batch)
     }
 
+    async fn get_oldest_aggregator_batch(
+        &self,
+        version: Option<String>,
```
Mohiiit: Nit: Let's keep naming as `orchestrator_version` for consistency and clarity across the codebase.

orchestrator/src/core/client/database/mongodb.rs
```diff
@@ -501,13 +501,25 @@ impl DatabaseClient for MongoDbClient {
         }
     }
 
-    async fn get_latest_job_by_type(&self, job_type: JobType) -> Result<Option<JobItem>, DatabaseError> {
+    async fn get_latest_job_by_type(
+        &self,
+        job_type: JobType,
+        orchestrator_version: Option<String>,
```
Mohiiit, line 507: Thought: Given `orchestrator_version` is from `Cargo.toml` and hardcoded per binary, we could potentially add the version filter by default to all queries. The explicit `Option` approach here is more flexible though, allowing queries across versions when needed (e.g., for migrations or debugging).

orchestrator/src/tests/utils.rs
```diff
@@ -128,6 +145,50 @@ pub fn build_snos_batch(index: u64, aggregator_batch_index: Option<u64>, start_b
...
+        start_block,
+        default_test_bouncer_weights(),
+        StarknetVersion::V0_13_2,
+    );
+    batch.orchestrator_version = orchestrator_version;
+    batch
+}
+
+/// Build an aggregator batch with a custom orchestrator version for testing version filtering
+pub fn build_batch_with_version(
```
Mohiiit: Nit: Consider renaming to `build_aggregator_batch_with_version` for consistency with `build_snos_batch_with_version` naming pattern.

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
```diff
@@ -140,30 +141,40 @@ impl SnosBatchingTrigger {
...
-    /// - Start: max(last block SNOS batching processed, min_block_to_process)
-    /// - End: min(last block processed by aggregator batching, max_block_to_process, start + max_blocks_to_process_at_once - 1)
+    /// - Start: max(last block SNOS batching processed, min_block_to_process, oldest aggregator batch of current version's start_block)
+    /// - End: min(last block processed by aggregator batching of current version, max_block_to_process, start + max_blocks_to_process_at_once - 1)
     async fn calculate_range_l2(&self, config: &Arc<Config>) -> Result<(u64, u64), JobError> {
         // Get the latest SNOS batch from the DB
         let latest_snos_batch = config.database().get_latest_snos_batch().await?;
 
         // Get the latest aggregator batch from the DB
         let latest_aggregator_batch = config.database().get_latest_aggregator_batch().await?;
```
Mohiiit, line 151: Consider: `get_latest_aggregator_batch()` is called without version filter, but `get_oldest_aggregator_batch()` uses version filter. This asymmetry could cause edge cases where `last_block_in_aggregator_batches` is from a different orchestrator version than `first_block_in_aggregator_batches`. Should both queries use the same version filter?

orchestrator/src/worker/event_handler/triggers/snos.rs
```diff
@@ -49,7 +49,7 @@ fn calculate_max_snos_jobs_to_create(
 ) -> u64 {
     match (latest_job_internal_id, oldest_incomplete_internal_id) {
         (Some(latest), Some(oldest_incomplete)) => {
-            let num_jobs_in_buffer = latest - oldest_incomplete + 1;
+            let num_jobs_in_buffer = latest.saturating_sub(oldest_incomplete) + 1;
```
Mohiiit, line 52: 👍 Good use of `saturating_sub` - prevents underflow in edge cases.

orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -37,7 +37,8 @@ pub struct CommonMetadata {
     /// Orchestrator version that created this job.
     /// Used to ensure only compatible orchestrator versions process jobs they can handle.
     /// This prevents version conflicts in multi-version deployments.
-    pub orchestrator_version: Option<String>,
+    #[serde(default)]
+    pub orchestrator_version: String,
```
Mohiiit, line 41: Nit: With this change, legacy jobs will deserialize with empty string `""` rather than `None`. Might be worth documenting that empty string should be treated as "legacy/no version" for backward compatibility.

Mohiiit (commented):
Good implementation of orchestrator versioning for safe multi-version deployments. The database-level filtering approach is cleaner than application-level filtering. Left a few suggestions for consideration.

orchestrator/src/core/client/database/mongodb.rs
prkpndy: Yeah, I did think about it. But for some methods, we require both the cases - Some and None. For eg with `get_latest_job_by_type`, where state update job needs to get the latest job of any version since state updates need to happen sequentially no matter the version - though even if we don't do like this, the case where a state update of previous version is happening while the orch is ready to pick state update of new version would be rare, practically speaking 🤔

orchestrator/src/types/jobs/metadata/mod.rs
prkpndy: I was thinking of adding a migration to move all legacy jobs - i.e. without any version to have version `0.1.0`. Will add that 🙏

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
prkpndy: It won't matter tbh. We don't want to gett he latest snos batch of current version only. To calculate the first block to batch, we are capping that by the start block of oldest aggregator of current version.
In other places also, it doesn't need to be of the current version only

orchestrator/src/core/client/database/mongodb.rs
prkpndy: done

orchestrator/src/tests/utils.rs
prkpndy: done

orchestrator/src/types/jobs/metadata/mod.rs
prkpndy: done

prkpndy (commented):
Resolved comments. Also realised that there are some redundant db methods, so removed them as well. Added version filtering in 2 more db methods

Mohiiit (approved):

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
```diff
@@ -140,30 +141,40 @@ impl SnosBatchingTrigger {
...
+                _ => {
+                    // No aggregator batch exists, so we can't process any SNOS batches
+                    debug!("No aggregator batch exists, cannot process SNOS batches for L2");
+                    return Ok((1, 0)); // Return invalid range to skip processing
+                }
+            };
 
         // Calculate the first block to assign to SNOS batch
-        let first_block = max(config.service_config().min_block_to_process, (last_block_in_snos_batches + 1) as u64);
+        let first_block = max(
```
byteZorvin, line 174: Nit: This will prevent orch to process batches of block B if there is any batch with block > B. Although this should never be the case in production but could happen in testing. For someone running the orch, this behaviour could be unintuitive.

let me know if there is some reasoning to it i dont know of

orchestrator/src/worker/event_handler/triggers/batching/snos.rs
```diff
@@ -116,10 +117,22 @@ impl SnosHandler {
         state: SnosState,
     ) -> Result<BlockProcessingResult<SnosState, NonEmptySnosState>, JobError> {
         info!("Including block {} in snos batch", block_num);
+
+        // Get the starknet version of the current block (applies to all states)
+        let block_version = get_block_version(block_num, self.config.madara_rpc_client()).await?;
+
+        // Check if block's Starknet version is supported (applies to all states)
+        if !block_version.is_supported() {
```
byteZorvin, line 125: This is great !!

byteZorvin (approved):
Solid implementation of orchestrator versioning. Just one nit.

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
prkpndy: the current logic is that it'll take the minimum of the followings:
1. min block to process (given by the user) - needed as it's a user config
2. latest snos block (created by the orch) - needed for the orch to move forward
3. oldest block aggregated by the current version of orchestrator - needed to make sure we don't create batches for agg batches of older versions
not sure how this will be unintutive for the user 😅? for the above case you described, afaiu, we should never want to create batches for blocks before the latest batch in DB. or am i missing something?

PR #936, opened 2026-01-06
Simplify StateTransition job architecture

heemankv:
# Simplify StateTransition job architecture

### Problem
1. StateTransition job processed multiple blocks/batches in a loop, maintaining arrays for paths and tx_hashes - but this added complexity without benefit since jobs were already being created one at a time.

2. `verification_attempt_no` was incremented only after the timeout check, causing it to not increase properly when verification was pending.

### Solution

**Single block/batch per job:**
- Reworked StateTransition to process exactly one block/batch per job call
- Changed `StateUpdateMetadata` fields from arrays to optional single values:
  - `tx_hashes: Vec<String>` → `tx_hash: Option<String>`
  - `snos_output_paths` → `snos_output_path: Option<String>`
  - `program_output_paths` → `program_output_path: Option<String>`
  - `blob_data_paths` → `blob_data_path: Option<String>`
  - `da_segment_paths` → `da_segment_path: Option<String>`
- Set `job.external_id` to the tx_hash for easier job tracking

**Verification fix:**
- Moved `verification_attempt_no` increment to the start of verification, ensuring it's always counted

**Already-settled handling:**
- Only set tx_hash to `0x0...64` if not already set, preserving existing values on retries

**Code cleanup:**
- Removed `max_items_to_process_in_single_job()` and `find_successive_items_in_vector()`
- Removed unused `fetch_blob_data_for_batch()`
- Simplified fetch utility functions to take `Option<String>` directly
- Added migration script to convert existing jobs to new schema

image
Screenshot 2026-01-08 at 2 01 05 PM


orchestrator/src/types/jobs/metadata/mod.rs
```diff
@@ -192,7 +192,7 @@ pub struct SnosMetadata {
 ///
 /// # Field Management
 /// - Worker-initialized fields: blocks and paths configurations
-/// - Job-populated fields: last_failed_block_no and tx_hashes (during processing)
+/// - Job-populated fields: last_failed_block_no and tx_hash (during processing)
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
 pub struct StateUpdateMetadata {
```
prkpndy, line 200: If we are assuming that one state update job will always send one txn, let's also move the path fields from Vec<String> to String?

heemankv: Done

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -41,20 +39,20 @@ impl JobHandlerTrait for StateUpdateJobHandler {
...
 
         // Validate required paths based on layer configuration
-        // L2: requires program_output_paths + da_segment_paths
-        // L3: requires program_output_paths + blob_data_paths + snos_output_paths
-        if state_metadata.program_output_paths.is_empty() {
-            error!("program_output_paths is required for all state updates");
-            return Err(JobError::Other(OtherError(eyre!("Missing required program_output_paths in metadata"))));
+        // L2: requires program_output_path + da_segment_path
+        // L3: requires program_output_path + blob_data_path + snos_output_path
+        if state_metadata.program_output_path.is_none() {
```
prkpndy, line 47: should we make program output non-optional since it's required by both?

prkpndy: nit: actually a better design imo would be to make an enum for paths with two variants - Blob/Da and have non-optional required fields in both. no chance of making mistakes and code would be cleaner as well. wdyt?

orchestrator/src/worker/event_handler/triggers/update_state.rs
```diff
@@ -177,114 +177,63 @@ impl UpdateStateJobTrigger {
     async fn collect_paths_l3(
```
prkpndy, line 165: with new design, these can be methods on the enum!

prkpndy (commented):
some suggestions :)

orchestrator/src/worker/event_handler/triggers/update_state.rs
heemankv: Adding a TODO

prkpndy (approved):
will move to a better design in a different PR as discussed with @heemankv 
rest all looks good!

orchestrator/src/tests/workers/update_state/mod.rs
```diff
@@ -228,30 +211,13 @@ async fn update_state_worker_next_block_missing() {
...
-            format!("{}/{}", 4, BLOB_DATA_FILE_NAME),
-        ],
-        da_segment_paths: vec![],
-        tx_hashes: Vec::new(),
+        snos_output_path: Some(format!("{}/{}", 0, SNOS_OUTPUT_FILE_NAME)),
+        program_output_path: Some(format!("{}/{}", 0, PROGRAM_OUTPUT_FILE_NAME)),
+        blob_data_path: Some(format!("{}/{}", 0, BLOB_DATA_FILE_NAME)),
+        da_segment_path: None,
+        tx_hash: None,
         context: SettlementContext::Block(SettlementContextData { to_settle: vec![0, 1, 2, 3, 4], last_failed: None }),
```
byteZorvin: Lets update `SettlementContext` to have one block instead of vec?

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -111,41 +109,40 @@ impl JobHandlerTrait for StateUpdateJobHandler {
...
-                sent_tx_hashes.push(format!("0x{:0>64}", ""));
-                state_metadata.tx_hashes = sent_tx_hashes.clone();
-                job.metadata.specific = JobSpecificMetadata::StateUpdate(state_metadata.clone());
-                continue;
+        if !self.should_send_state_update_txn(&config, to_settle_num).await? {
+            // State update not needed (already settled on-chain)
+            // Setting tx_hash to an empty string, to denote that no transaction was sent in current run
+            // Only if tx_hash is not already set
+            if state_metadata.tx_hash.is_none() {
+                state_metadata.tx_hash = Some(format!("0x{:0>64}", ""));
```
byteZorvin: Should we do `const EMPTY_TX_HASH: &str = "0x0000000000000000000000000000000000000000000000000000000000000000";` and use that instead of `format!()`

orchestrator/src/worker/event_handler/jobs/state_update.rs
```diff
@@ -111,41 +109,40 @@ impl JobHandlerTrait for StateUpdateJobHandler {
...
-        let da_segment_paths = state_metadata.da_segment_paths.clone();
-
-        let mut nonce = config.settlement_client().get_nonce().await.map_err(|e| JobError::Other(OtherError(e)))?;
+        let nonce = config.settlement_client().get_nonce().await.map_err(|e| JobError::Other(OtherError(e)))?;
 
-        let mut sent_tx_hashes: Vec<String> = Vec::with_capacity(filtered_indices.len());
+        // Process the batch/block (filtered_indices should contain exactly one index)
+        let i = *filtered_indices
+            .first()
+            .ok_or_else(|| JobError::Other(OtherError(eyre!("No blocks/batches to process after filtering"))))?;
```
byteZorvin: Filtering confuses without context
```suggestion
            .ok_or_else(|| JobError::Other(OtherError(eyre!("No blocks/batches to settle"))))?;
```

byteZorvin (commented):
Solid changes.

byteZorvin (approved):

PR #937, opened 2026-01-08
Fix/heal orphan

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

## What is the new behavior?

Moved the logic to heal orphan jobs (jobs stuck in LFP stage) from worker trigger to job processing handler

### DIFF
--- a/orchestrator/src/types/queue_control.rs
+++ b/orchestrator/src/types/queue_control.rs
@@ -1,3 +1,4 @@
+use crate::cli::service::DEFAULT_TIMEOUT_SECONDS;
 use crate::types::queue::QueueType;
 use crate::types::Layer;
 use orchestrator_utils::env_utils::get_env_var_or_default;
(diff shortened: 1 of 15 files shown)


orchestrator/src/types/queue_control.rs
```diff
@@ -204,8 +239,11 @@ pub static QUEUES: LazyLock<HashMap<QueueType, QueueConfig>> = LazyLock::new(||
     map.insert(
         QueueType::PriorityProcessingQueue,
         QueueConfig {
-            visibility_timeout: 600, // 2x expected max processing time for slot wait + processing
-            dlq_config: Some(DlqConfig { max_receive_count: 5, dlq_name: QueueType::JobHandleFailure }),
+            visibility_timeout: 2 * QUEUE_VISIBILITY_TIMEOUT_SECONDS, // 2x expected max processing time for slot wait + processing
```
apoorvsadana, line 242: what's slot wait?

apoorvsadana (approved):

orchestrator/src/types/queue_control.rs
prkpndy: This is for priority queue. We have a slot (a mutex) for storing the job from queue. It's 2x for the case where:
1. One message (from priority queue) comes and sits on slot
2. This is picked up by the worker so the slot is empty
3. Now another message is picked from the queue and kept here
4. For this message, the VT should cover the time to process the first job (while it's waiting in slot) + time to process this job

apoorvsadana: makes sense!

heemankv (approved):

PR #940, opened 2026-01-10
update: nack message instead of marking failed when job processing fail

prkpndy:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Job is marked as failed if the job handler returns an error/panics just after one try.

Resolves: #NA

## What is the new behavior?

NACK the message if the job fails so we can re-process it when the message comes back.
If the job fails repeatedly, we mark the job as failed through DLQ route.

No

### DIFF
--- a/orchestrator/src/worker/event_handler/service.rs
+++ b/orchestrator/src/worker/event_handler/service.rs
@@ -11,6 +11,7 @@ use crate::core::config::Config;
 use crate::error::job::JobError;
 use crate::error::other::OtherError;
 use crate::types::jobs::external_id::ExternalId;
+use crate::types::jobs::job_item::JobItem;
 use crate::types::jobs::job_updates::JobItemUpdates;
 use crate::types::jobs::metadata::JobMetadata;
 use crate::types::jobs::status::JobVerificationStatus;
@@ -366,7 +367,6 @@ impl JobHandlerService {
                 external_id
             }
             Ok(Err(e)) => {
-                // Reset job state so it can be retried when the message comes back
                 error!(
                     job_id = ?id,
                     job_type = ?job.job_type,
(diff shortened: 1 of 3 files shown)


heemankv (approved):

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -360,18 +366,30 @@ impl JobHandlerService {
...
                     job_id = ?id,
                     job_type = ?job.job_type,
                     internal_id = %job.internal_id,
                     status = ?job.status,
                     error = ?e,
-                    "Failed to process job"
+                    "Failed to process job, resetting state for retry"
                 );
-                return JobService::move_job_to_failed(&job, config.clone(), format!("Processing failed: {}", e)).await;
+                job.metadata.common.process_started_at = None;
```
Mohiiit: Consider extracting the job reset logic into a helper function to apply DRY - this same pattern is repeated in the panic branch below (lines 405-418).

A helper like `reset_job_for_retry()` could:
1. Clear `process_started_at`
2. Update job to `original_status`
3. **Properly propagate the DB error** instead of just logging it

Currently if `update_job` fails, the message is NACKed but the job remains in `LockedForProcessing` state, potentially causing inconsistency. The helper could return the DB error, letting the caller decide whether to return the original error or the DB error.

Mohiiit (commented):
Good approach to improve resilience by using NACK + retry instead of immediate failure. The change makes sense and tests are properly updated. Left one suggestion about extracting the duplicated reset logic into a helper.

Mohiiit (approved):

orchestrator/src/worker/event_handler/service.rs
prkpndy: done

PR #942, opened 2026-01-13
fix: clean up orphaned .lock files on timeout and startup + docker issue resolved

Mohiiit:
## Summary
  - Fix orphaned `.lock` file cleanup when cairo-native compilation times out
  - Add startup cleanup for `.lock` files left behind from ungraceful shutdowns

  ## Problem

  When cairo-native compilation times out:
  1. The `.lock` file created by `compile_to_native()` was NOT being cleaned up (only `.so` was removed)
  2. The `spawn_blocking` task continues running after timeout, but its result is discarded
  3. On ungraceful shutdown, orphaned `.lock` files remain on disk
  4. Future compilation attempts see the stale `.lock` file and fail with "Failed to extract sierra program"

  ## Solution

  1. **DRY cleanup helper** - Added `cleanup_compilation_artifacts()` that removes both `.so` and `.lock` files
  2. **Fixed 3 timeout locations** - Updated all timeout handlers to use the new cleanup helper
  3. **Startup cleanup** - Added `cleanup_orphaned_lock_files()` called during `setup_and_log()` to clean up any `.lock` files left from previous crashes

  ## Test plan
  - [x] Added unit test for `cleanup_compilation_artifacts()`
  - [x] Added unit tests for `cleanup_orphaned_lock_files()` (normal case + empty directory)
  - [x] All existing tests pass


madara/Dockerfile
```diff
@@ -118,11 +118,30 @@ RUN --mount=type=cache,target=$SCCACHE_DIR,sharing=locked \
...
+COPY Makefile /tmp/Makefile
+
+# Install LLVM 19 for Cairo Native runtime compilation (compiles Sierra to native on-the-fly)
 RUN apt-get -y update && \
-    apt-get install -y openssl ca-certificates tini curl &&\
-    apt-get autoremove -y; \
-    apt-get clean; \
-    rm -rf /var/lib/apt/lists/*
+    apt-get install -y openssl ca-certificates tini curl wget gnupg software-properties-common make && \
+    cd /tmp && make -f Makefile install-llvm19 CODENAME=bookworm && \
```
byteZorvin, line 127: This was a difficult one, lets go !!

byteZorvin (approved):
Cairo native go brr !!

heemankv (approved):

PR #945, opened 2026-01-14
Orchestrator: Migrate Atlantic idempotency from externalId to dedupId

heemankv:
# PR: Migrate Atlantic idempotency from externalId to dedupId

## Pull Request type

The Atlantic service uses `search_atlantic_queries` with `externalId` for idempotency checks. This approach:
- Makes a search API call to `/atlantic-queries` with the external ID
- Returns a list of all matching queries
- Filters client-side to find an exact match
- Results in O(n) complexity where n is the number of matching results

## What is the new behavior?

- Uses Atlantic's native `dedupId` field instead of `externalId` for deduplication
- Adds new `get_query_by_dedup_id` method that calls `/atlantic-query-by-dedup-id` endpoint
- Direct O(1) lookup - returns the exact query or 404 if not found
- Removes the now-unused `search_atlantic_queries` function
- Removes unused `AtlanticQueriesListResponse` type
- Updates tests to use the new endpoint

### Changes:
1. **client.rs**: Added `get_query_by_dedup_id()`, changed form field from `externalId` to `dedupId` in `add_job()`, removed `search_atlantic_queries()`
2. **lib.rs**: Updated `submit_task` to use `get_query_by_dedup_id` for idempotency check
3. **types.rs**: Removed unused `AtlanticQueriesListResponse` struct
4. **tests/lib.rs**: Updated mocks to use `/atlantic-query-by-dedup-id` endpoint

No. The change is internal to how idempotency is checked. The `ProverClient` trait interface remains unchanged.

## Other information

This change was enabled by Herodotus updating their API to allow `/atlantic-query-by-dedup-id` lookups using just the API key (deriving project from the key server-side), removing the need for `projectId` parameter.

### DIFF
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -21,7 +21,8 @@ use crate::transport::ApiKeyAuth;
 use crate::types::{
     AtlanticAddJobResponse, AtlanticAggregatorParams, AtlanticAggregatorVersion, AtlanticBucketResponse,
     AtlanticCairoVersion, AtlanticCairoVm, AtlanticCreateBucketRequest, AtlanticGetBucketResponse,
-    AtlanticGetStatusResponse, AtlanticQuery, AtlanticQueryStep, AtlanticSharpProver,
+    AtlanticGetStatusResponse, AtlanticQueryByDedupIdResponse, AtlanticQuerySimple, AtlanticQueryStep,
+    AtlanticSharpProver,
 };
 use crate::AtlanticValidatedArgs;
 
@@ -341,7 +342,7 @@ impl AtlanticClient {
         &self,
         dedup_id: &str,
         api_key: &str,
-    ) -> Result<Option<AtlanticQuery>, AtlanticError> {
+    ) -> Result<Option<AtlanticQuerySimple>, AtlanticError> {
         let context = format!("dedup_id: {}", dedup_id);
         let dedup_id = dedup_id.to_string();
         let api_key = api_key.to_string();
@@ -387,7 +388,7 @@ impl AtlanticClient {
                 }
 
                 if status.is_success() {
-                    let response: AtlanticGetStatusResponse = response
+                    let response: AtlanticQueryByDedupIdResponse = response
                         .json()
                         .await
                         .map_err(|e| AtlanticError::parse_error("get_query_by_dedup_id", e.to_string()))?;
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -380,10 +380,8 @@ impl AtlanticClient {
                         dedup_id = %dedup_id,
                         "Query not found (first submission)"
                     );
-                    return Ok(None);
-                }
-
-                if status.is_success() {
+                    Ok(None)
+                } else if status.is_success() {
                     let response: AtlanticQueryByDedupIdResponse = response
                         .json()
                         .await
@@ -396,19 +394,19 @@ impl AtlanticClient {
                         status = ?response.atlantic_query.status,
                         "Query found"
                     );
-                    return Ok(Some(response.atlantic_query));
+                    Ok(Some(response.atlantic_query))
+                } else {
+                    // Other error
+                    let (error_text, status) = extract_http_error_text(response, "get query by dedup id").await;
+                    debug!(
+                        operation = "get_query_by_dedup_id",
+                        dedup_id = %dedup_id,
+                        status = %status,
+                        error = %error_text,
+                        "Failed to get query by dedup ID"
+                    );
+                    Err(AtlanticError::from_http_error_response("get_query_by_dedup_id", status, error_text))
                 }
-
-                // Other error
-                let (error_text, status) = extract_http_error_text(response, "get query by dedup id").await;
-                debug!(
-                    operation = "get_query_by_dedup_id",
-                    dedup_id = %dedup_id,
-                    status = %status,
-                    error = %error_text,
-                    "Failed to get query by dedup ID"
-                );
-                Err(AtlanticError::from_http_error_response("get_query_by_dedup_id", status, error_text))
             }
         })
         .await
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -340,6 +340,13 @@ impl AtlanticClient {
         api_key: &str,
     ) -> Result<Option<AtlanticQuery>, AtlanticError> {
         let context = format!("dedup_id: {}", dedup_id);
+
+        debug!(
+            operation = "get_query_by_dedup_id",
+            dedup_id = %dedup_id,
+            "Looking up query by dedup ID"
+        );
+
         let dedup_id = dedup_id.to_string();
         let api_key = api_key.to_string();
 
@@ -348,12 +355,6 @@ impl AtlanticClient {
             let api_key = api_key.clone();
 
             async move {
-                debug!(
-                    operation = "get_query_by_dedup_id",
-                    dedup_id = %dedup_id,
-                    "Looking up query by dedup ID"
-                );
-
                 let response = self
                     .client
                     .request()
--- a/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/client.rs
@@ -21,7 +21,7 @@ use crate::transport::ApiKeyAuth;
 use crate::types::{
     AtlanticAddJobResponse, AtlanticAggregatorParams, AtlanticAggregatorVersion, AtlanticBucketResponse,
     AtlanticCairoVersion, AtlanticCairoVm, AtlanticCreateBucketRequest, AtlanticGetBucketResponse,
-    AtlanticGetStatusResponse, AtlanticQueriesListResponse, AtlanticQuery, AtlanticQueryStep, AtlanticSharpProver,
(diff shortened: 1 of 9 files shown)


Mohiiit (commented):
## Review: Atlantic idempotency migration to dedupId

### Summary
Good performance-oriented refactor (O(n) → O(1) lookup) with clean code removal. However, there's a **critical discrepancy** between implementation and tests regarding response types.

---

### 🔴 Critical Issue: Response Type Mismatch

**Location**: `client.rs:361-372` and `tests/lib.rs:107-115`

**Problem**: Implementation expects `AtlanticGetStatusResponse` but tests mock bare `AtlanticQuery`:

```rust
// Implementation (client.rs)
let response: AtlanticGetStatusResponse = response.json().await.map_err(/*...*/)?;
return Ok(Some(response.atlantic_query));

// Test mock (tests/lib.rs) - INCONSISTENT
let response = AtlanticQuery { /* ... */ };  // Should be wrapped!
then.status(200).json_body_obj(&response);
```

**Action Required**: 
1. Verify `/atlantic-query-by-dedup-id` actual endpoint response structure against [Atlantic API docs](https://atlantic.api.herodotus.cloud/docs/)
2. Fix either implementation or tests to match reality

**Suggested fix** (if endpoint returns wrapped response):
```rust
// tests/lib.rs
let response = AtlanticGetStatusResponse {
    atlantic_query: AtlanticQuery { /* ... */ },
    metadata_urls: vec![],
};
then.status(200).json_body_obj(&response);
```

---

### ✅ Strengths

1. **Performance**: O(n) search → O(1) direct lookup
2. **Code quality**: Removes ~100 lines of unused code
3. **Semantics**: Uses native `dedupId` field instead of repurposing `externalId`
4. **Safety**: No unsafe code, proper error handling
5. **API**: Public interface unchanged (internal refactor only)
6. **Idempotency**: Correctly handles failed job resubmission

---

### 📝 Minor Suggestions

**1. Naming consistency** (optional, low priority):
The variable is still called `external_id` internally but form field is now `dedupId`. Consider renaming variable to `dedup_id` in future PR for consistency.

**2. Add clarifying comment** (client.rs:361):
```rust
// Check 404 first - query doesn't exist (first submission)
if status == StatusCode::NOT_FOUND {
```

---

### 🔍 Testing Notes

- Tests correctly updated to use new endpoint
- Mock structure needs fixing (see critical issue above)
- Consider adding test case for 404 response path explicitly

---

### Verdict

**LGTM with critical fix required**. Please resolve the response type discrepancy and I'll approve. The refactoring itself is well-structured and achieves the stated performance goals.

cc @heemankv

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -324,104 +324,97 @@ impl AtlanticClient {
...
+
+                let response = self
+                    .client
+                    .request()
+                    .method(Method::GET)
+                    .header(
+                        ApiKeyAuth::header_name(),
+                        HeaderValue::from_str(&api_key).map_err(|e| AtlanticError::Other {
+                            operation: "get_query_by_dedup_id".to_string(),
+                            message: format!("Invalid API key header value: {}", e),
```
Mohiiit: **Response type verification needed**

The implementation expects `AtlanticGetStatusResponse` which wraps the query in a nested structure:

```rust
let response: AtlanticGetStatusResponse = response
    .json()
    .await
    .map_err(|e| AtlanticError::parse_error("get_query_by_dedup_id", e.to_string()))?;

return Ok(Some(response.atlantic_query));
```

Please verify against the [Atlantic API Swagger](https://atlantic.api.herodotus.cloud/docs/) that `/atlantic-query-by-dedup-id` returns the same structure as `/atlantic-query/{id}` (which does return `AtlanticGetStatusResponse`).

If it returns a bare `AtlanticQuery`, change the type annotation to:
```rust
let response: AtlanticQuery = response.json().await...?;
return Ok(Some(response));
```

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -79,27 +79,16 @@ impl ProverClient for AtlanticProverService {
...
-                    .atlantic_queries
-                    .into_iter()
-                    .find(|query| {
-                        query.external_id.as_deref() == Some(external_id.as_str()) || query.id == external_id
-                    });
+                // Direct lookup by dedup ID - O(1), no list filtering needed
+                let existing_job =
+                    self.atlantic_client.get_query_by_dedup_id(&external_id, &self.atlantic_api_key).await?;
 
                 if let Some(existing_job) = existing_job {
```
Mohiiit: **Nice optimization! 🎯**

This is the key improvement - direct O(1) lookup instead of O(n) search with client-side filtering. The old code would:
1. Fetch a list (potentially large)
2. Filter client-side with `.find(|query| query.external_id.as_deref() == Some(external_id.as_str()) || query.id == external_id)`

The new approach is much cleaner and more efficient.

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -324,104 +324,92 @@ impl AtlanticClient {
...
+        let context = format!("dedup_id: {}", dedup_id);
+        let dedup_id = dedup_id.to_string();
+        let api_key = api_key.to_string();
+
+        self.retry_request("get_query_by_dedup_id", &context, || {
+            let dedup_id = dedup_id.clone();
+            let api_key = api_key.clone();
+
+            async move {
+                debug!(
```
Mohiiit: **Question**: In the old `search_atlantic_queries`, the `debug!` statement was placed *before* the `retry_request` call (fired once). Here it's *inside* the async block (fires on each retry attempt). Was this change intentional? It's arguably more useful for debugging retries, but wanted to confirm this was deliberate.

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -79,27 +79,16 @@ impl ProverClient for AtlanticProverService {
...
                     match existing_job.status {
                         AtlanticQueryStatus::Failed => {
                             tracing::warn!(
                                 atlantic_query_id = %existing_job.id,
-                                external_id = %external_id,
+                                dedup_id = %external_id,
```
Mohiiit: **Naming consistency**: The log message uses `dedup_id` but the variable is `external_id` (from `CreateJobInfo`). Would you prefer to:
1. Rename the field in `CreateJobInfo` to `dedup_id` for full consistency, or
2. Keep the current naming and just update logs to say `external_id` (since that's the variable name)?

Either works, just want to ensure consistency across the codebase.

orchestrator/crates/prover-clients/atlantic-service/tests/lib.rs
```diff
@@ -187,10 +177,98 @@ async fn atlantic_client_does_not_resubmit_when_job_exists() {
...
+#[tokio::test]
+async fn atlantic_client_get_query_by_dedup_id_found() {
+    let _ = env_logger::try_init();
+    dotenvy::from_filename_override("../.env.test").expect("Failed to load the .env file");
+    let atlantic_params = AtlanticValidatedArgs {
+        atlantic_api_key: get_env_var_or_panic("MADARA_ORCHESTRATOR_ATLANTIC_API_KEY"),
+        atlantic_service_url: Url::parse(&get_env_var_or_panic("MADARA_ORCHESTRATOR_ATLANTIC_SERVICE_URL")).unwrap(),
+        atlantic_rpc_node_url: Url::parse(&get_env_var_or_panic("MADARA_ORCHESTRATOR_ATLANTIC_RPC_NODE_URL")).unwrap(),
+        atlantic_mock_fact_hash: get_env_var_or_panic("MADARA_ORCHESTRATOR_ATLANTIC_MOCK_FACT_HASH"),
+        atlantic_prover_type: get_env_var_or_panic("MADARA_ORCHESTRATOR_ATLANTIC_PROVER_TYPE"),
```
Mohiiit, line 198: **Suggestion**: This test submits a real job to Atlantic just to verify the lookup. Consider simplifying by using a known/existing dedup_id (perhaps from a test fixture or a dedicated test query that already exists) rather than creating a new job each time. This would make the test faster and avoid creating unnecessary state on Atlantic's side.

Mohiiit (commented):
Thanks for this clean refactoring! The move from O(n) search to O(1) direct lookup is a nice improvement. I have a few questions and suggestions below.

**Overall**: The code changes look correct and well-structured. The 404-as-`Ok(None)` pattern is idiomatic Rust.

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
heemankv: Moved

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
heemankv: Fixed

orchestrator/crates/prover-clients/atlantic-service/tests/lib.rs
heemankv: Fixed

byteZorvin (approved):
Solid changes for dedup, tests were really helpful too.

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -709,7 +698,7 @@ impl AtlanticClient {
                             .form_text("cairoVersion", &AtlanticCairoVersion::Cairo0.as_str())
                             .form_text("cairoVm", &cairo_vm.as_str())
                             .form_text("sharpProver", &sharp_prover.as_str())
-                            .form_text("externalId", &external_id)
+                            .form_text("dedupId", &external_id)
```
byteZorvin: This is slightly confusing though, as there is external id too in atlantic still and we are passing the external id as dedup id. Would prefer if we rename externalId to dedup id everywhere.

Mohiiit (approved):

PR #946, opened 2026-01-14
Orchestrator: Automatic Storage Cleanup for Settled Artifacts

heemankv:
# PR: Automatic Storage Cleanup for Settled Artifacts

## Pull Request type

- [x] Feature

S3 storage grows continuously with no cleanup mechanism. Artifacts from completed jobs (cairo_pie.zip, snos_output.json, program_output.txt, blob data, etc.) remain indefinitely after StateTransition jobs complete.

- PC Devnet reached **1.4 TB** storage consumption
- Growth rate: **>100 GB/day**
- No TTL or lifecycle policies exist
- Files remain indefinitely, causing unnecessary storage costs

Resolves: #NA

## What is the new behavior?

Implements automatic storage cleanup using S3 Lifecycle Rules with object tagging:

- **S3 Lifecycle Rule**: On setup, configures lifecycle rule to auto-delete objects tagged with `expire-after-settlement=true` after 14 days from creation
- **StorageCleanupTrigger**: New worker that runs via EventBridge (every 10 mins), tags artifacts from completed StateTransition jobs
- **Tracking**: `storage_artifacts_tagged_at` field in StateUpdateMetadata prevents re-processing
- **Batch Limiter**: Processes max 200 jobs per run to avoid overwhelming S3 API
- **Distributed Lock**: Uses MongoDB locks collection to prevent concurrent runs
- **Fast Cleanup**: Old artifacts (>14 days) get deleted within 24h of tagging

## ⚠️ Required IAM Permissions

This feature requires the following IAM permissions to be added to the orchestrator's IAM role/policy:

```json
{
    "Effect": "Allow",
    "Action": [
        "s3:PutObjectTagging",
        "s3:PutLifecycleConfiguration",
        "s3:GeLifecycleConfiguration"
    ],
    "Resource": [
        "arn:aws:s3:::<bucket-name>",
        "arn:aws:s3:::<bucket-name>/*"
    ]
}
```

**Permission breakdown:**
- **`s3:PutObjectTagging`** (Required): Allows the StorageCleanup worker to tag artifacts for expiration. Without this, the worker will fail with a "service error" when tagging objects.
- **`s3:PutLifecycleConfiguration`** (Required): Allows the setup phase to configure the lifecycle rule that auto-deletes tagged objects after 14 days. Without this, the `setup` command will fail.
- **`s3:GetLifecycleConfiguration`** (Recommended): Best practice for checking existing lifecycle rules before applying new ones to avoid overwriting other configurations.

## Storage Paths Covered

The worker discovers and tags **ALL files** in these directories using `list_files_in_dir()`:

| Path Pattern | Description | Example Files |
|--------------|-------------|---------------|
| `artifacts/batch/{job_id}/` | Aggregator artifacts (new format) | `cairo_pie.zip`, `program_output.txt`, `snos_output.json`, `da_segment.json`, `proof.json` |
| `blob/batch/{job_id}/` | Blob data files | `1.txt`, `2.txt`, etc. |
| `{job_id}/` | SNOS artifacts (old format, root level) | `cairo_pie.zip`, `program_output.txt`, `snos_output.json`, `blob_data.txt`, `onchain_data.json`, `proof.json`, `proof_part2.json` |
| `state_update/batch/{job_id}.json` | State update JSON file | Single JSON file per job |

**Key point**: The worker uses `list_files_in_dir()` to discover files dynamically, so ALL files in each directory are tagged - not just hardcoded filenames. This ensures complete cleanup even if new file types are added.

No

The new field `storage_artifacts_tagged_at` in `StateUpdateMetadata` uses `#[serde(default)]`, so existing jobs will deserialize with `None` (untagged). The cleanup worker will process them on subsequent runs.

## Deployment Steps

1. **Add IAM permissions** (see Required IAM Permissions section above) to orchestrator's role
2. Deploy the code changes
3. Run setup to apply lifecycle rule to existing bucket (idempotent)
4. EventBridge rule is auto-created for StorageCleanup trigger

## How It Works

```
StateTransition completes
        ↓
StorageCleanupTrigger runs (every 10 min via EventBridge)
        ↓
Acquires distributed lock (MongoDB)
        ↓
Query: completed StateTransition jobs WHERE storage_artifacts_tagged_at IS NULL LIMIT 200
        ↓
For each job:
  - List all files in artifacts/batch/{id}/, blob/batch/{id}/, {id}/
  - Tag all discovered files with "expire-after-settlement=true"
  - Tag state_update/batch/{id}.json
  - Update metadata: storage_artifacts_tagged_at = now()
        ↓
Release lock
        ↓
AWS S3 Lifecycle (runs daily)
        ↓
Delete objects where tag exists AND creation_date > 14 days ago
```

## Design Decisions

- **Provider-agnostic naming**: `StorageCleanup` not `S3Cleanup`
- **Tag-based expiration**: AWS handles deletion, minimal code
- **14-day retention**: Buffer for debugging before deletion
- **Creation date basis**: Old artifacts deleted within 24h of tagging
- **Dynamic file discovery**: Uses `list_files_in_dir()` instead of hardcoded paths
- **Efficient DB query**: New `get_jobs_without_storage_artifacts_tagged()` method queries directly for untagged jobs
- **Graceful error handling**: Non-existent files (NoSuchKey) are skipped without failing the job

### DIFF
--- a/orchestrator/src/core/client/storage/mod.rs
+++ b/orchestrator/src/core/client/storage/mod.rs
@@ -29,4 +29,18 @@ pub trait StorageClient: Send + Sync {
     /// * `Ok(())` - If the storage service is healthy and accessible
     /// * `Err(StorageError)` - If the health check fails
     async fn health_check(&self) -> Result<(), StorageError>;
+
+    /// Tag an object with key-value pairs
+    ///
+    /// This method applies tags to an existing object in storage.
+    /// Tags can be used for lifecycle management, cost allocation, or organization.
+    ///
+    /// # Arguments
+    /// * `key` - The key of the object to tag
+    /// * `tags` - A vector of (key, value) pairs representing the tags
+    ///
+    /// # Returns
+    /// * `Ok(())` - If tagging was successful
+    /// * `Err(StorageError)` - If tagging fails
+    async fn tag_object(&self, key: &str, tags: Vec<(String, String)>) -> Result<(), StorageError>;
 }
--- a/orchestrator/src/core/client/storage/mod.rs
+++ b/orchestrator/src/core/client/storage/mod.rs
@@ -37,10 +37,10 @@ pub trait StorageClient: Send + Sync {
     ///
     /// # Arguments
     /// * `key` - The key of the object to tag
-    /// * `tags` - A slice of (key, value) pairs representing the tags (passed by reference to avoid cloning)
+    /// * `tags` - A slice of (key, value) string slice pairs representing the tags
     ///
     /// # Returns
     /// * `Ok(())` - If tagging was successful
     /// * `Err(StorageError)` - If tagging fails
-    async fn tag_object(&self, key: &str, tags: &[(String, String)]) -> Result<(), StorageError>;
+    async fn tag_object<'a>(&self, key: &str, tags: &'a [(&'a str, &'a str)]) -> Result<(), StorageError>;
 }
--- a/orchestrator/src/core/client/storage/mod.rs
+++ b/orchestrator/src/core/client/storage/mod.rs
@@ -37,10 +37,10 @@ pub trait StorageClient: Send + Sync {
     ///
     /// # Arguments
     /// * `key` - The key of the object to tag
-    /// * `tags` - A vector of (key, value) pairs representing the tags
+    /// * `tags` - A slice of (key, value) pairs representing the tags (passed by reference to avoid cloning)
     ///
     /// # Returns
     /// * `Ok(())` - If tagging was successful
     /// * `Err(StorageError)` - If tagging fails
-    async fn tag_object(&self, key: &str, tags: Vec<(String, String)>) -> Result<(), StorageError>;
+    async fn tag_object(&self, key: &str, tags: &[(String, String)]) -> Result<(), StorageError>;
 }
--- a/orchestrator/src/setup/aws/s3.rs
+++ b/orchestrator/src/setup/aws/s3.rs
@@ -6,9 +6,22 @@ use crate::types::params::StorageArgs;
 use crate::types::Layer;
 use crate::{OrchestratorError, OrchestratorResult};
 use async_trait::async_trait;
+use aws_sdk_s3::types::{
+    BucketLifecycleConfiguration, ExpirationStatus, LifecycleExpiration, LifecycleRule, LifecycleRuleFilter, Tag,
+};
 use aws_sdk_s3::Error as S3Error;
 use std::sync::Arc;
 use tracing::{info, warn};
+
+/// Tag key used to mark objects for expiration
+const EXPIRATION_TAG_KEY: &str = "expire-after-settlement";
+const EXPIRATION_TAG_VALUE: &str = "true";
+
+/// Number of days after object creation before S3 deletes the object
+const EXPIRATION_DAYS: i32 = 14;
+
+/// Lifecycle rule ID
+const LIFECYCLE_RULE_ID: &str = "expire-settled-artifacts";
 
 #[async_trait]
 impl Resource for InnerAWSS3 {
@@ -34,46 +47,58 @@ impl Resource for InnerAWSS3 {
     }
     /// Set up a new S3 bucket
     async fn setup(&self, _layer: &Layer, args: Self::SetupArgs) -> OrchestratorResult<Self::SetupResult> {
+        let bucket_name = match &args.bucket_identifier {
+            AWSResourceIdentifier::ARN(arn) => arn.resource.clone(),
+            AWSResourceIdentifier::Name(name) => name.clone(),
+        };
+
         // Check if the bucket already exists
-        // If it does, return the existing bucket name and location
+        // If it does, skip creation but still ensure lifecycle rule is set
         if self.check_if_exists(&args.bucket_identifier).await? {
             warn!("  S3 bucket {} already exists , skipping creation", &args.bucket_identifier);
-            return Ok(());
-        }
+        } else {
+            // s3 can have empty region in it's arn : e.g: arn:aws:s3:::mo-bucket
+            // in such scenarios we would want to default to provided region
 
-        // s3 can have empty region in it's arn : e.g: arn:aws:s3:::mo-bucket
-        // in such scenarios we would want to default to provided region
-
-        match &args.bucket_identifier {
-            AWSResourceIdentifier::ARN(arn) => {
-                // If ARN is provided in setup, we have already checked if it exists
-                tracing::info!("Bucket Arn provided, skipping setup for {}", &arn.resource);
-                Ok(())
-            }
-            AWSResourceIdentifier::Name(bucket_name) => {
-                let region =
-                    self.client().config().region().map(|r| r.to_string()).unwrap_or_else(|| "us-east-1".to_string());
-
-                info!("Creating New Bucket: {}", bucket_name);
-
-                let mut bucket_builder = self.client().create_bucket().bucket(bucket_name);
-
-                if region != "us-east-1" {
-                    let constraint = aws_sdk_s3::types::BucketLocationConstraint::from(region.as_str());
-                    let cfg =
-                        aws_sdk_s3::types::CreateBucketConfiguration::builder().location_constraint(constraint).build();
-                    bucket_builder = bucket_builder.create_bucket_configuration(cfg);
+            match &args.bucket_identifier {
+                AWSResourceIdentifier::ARN(arn) => {
+                    // If ARN is provided in setup, we have already checked if it exists
+                    tracing::info!("Bucket Arn provided, skipping setup for {}", &arn.resource);
                 }
+                AWSResourceIdentifier::Name(bucket_name) => {
+                    let region = self
+                        .client()
+                        .config()
+                        .region()
+                        .map(|r| r.to_string())
+                        .unwrap_or_else(|| "us-east-1".to_string());
 
-                let _result = bucket_builder.send().await.map_err(|e| {
-                    OrchestratorError::ResourceSetupError(format!(
-                        "Failed to create S3 bucket '{}': {:?}",
-                        bucket_name, e
-                    ))
-                })?;
-                Ok(())
+                    info!("Creating New Bucket: {}", bucket_name);
+
+                    let mut bucket_builder = self.client().create_bucket().bucket(bucket_name);
+
+                    if region != "us-east-1" {
+                        let constraint = aws_sdk_s3::types::BucketLocationConstraint::from(region.as_str());
+                        let cfg = aws_sdk_s3::types::CreateBucketConfiguration::builder()
+                            .location_constraint(constraint)
+                            .build();
+                        bucket_builder = bucket_builder.create_bucket_configuration(cfg);
+                    }
+
+                    let _result = bucket_builder.send().await.map_err(|e| {
+                        OrchestratorError::ResourceSetupError(format!(
+                            "Failed to create S3 bucket '{}': {:?}",
+                            bucket_name, e
+                        ))
+                    })?;
+                }
             }
         }
+
+        // Setup lifecycle rule for automatic expiration (idempotent - safe to run on existing buckets)
+        self.setup_lifecycle_rule(&bucket_name).await?;
+
+        Ok(())
     }
 
     // TODO: can we simplify if check_if_exists and is_ready_to_use are same ?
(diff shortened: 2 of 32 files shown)


orchestrator/src/types/constant.rs
```diff
@@ -14,6 +14,58 @@ pub const DA_SEGMENT_FILE_NAME: &str = "da_blob.json";
...
+    format!("{}", job_id)
+}
+
+// Storage cleanup constants
+/// Tag key used to mark objects for expiration
+pub const STORAGE_EXPIRATION_TAG_KEY: &str = "expire-after-settlement";
+/// Tag value used to mark objects for expiration
+pub const STORAGE_EXPIRATION_TAG_VALUE: &str = "true";
+/// Number of days after object creation before S3 deletes the object
+pub const STORAGE_EXPIRATION_DAYS: i32 = 14;
```
Mohiiit: Consider making this configurable via environment variable or config file. Different deployments may need different retention periods (e.g., 7 days for dev, 30 days for production audit requirements).

```rust
// Example:
pub fn storage_expiration_days() -> i32 {
    std::env::var("STORAGE_EXPIRATION_DAYS")
        .ok()
        .and_then(|v| v.parse().ok())
        .unwrap_or(14)
}
```

orchestrator/src/setup/aws/s3.rs
```diff
@@ -90,3 +108,67 @@ impl Resource for InnerAWSS3 {
...
+            .map_err(|e| OrchestratorError::ResourceSetupError(format!("Failed to build lifecycle rule: {:?}", e)))?;
+
+        // Create the lifecycle configuration
+        let lifecycle_config = BucketLifecycleConfiguration::builder().rules(rule).build().map_err(|e| {
+            OrchestratorError::ResourceSetupError(format!("Failed to build lifecycle configuration: {:?}", e))
+        })?;
+
+        // Apply the lifecycle configuration to the bucket
+        self.client()
+            .put_bucket_lifecycle_configuration()
```
Mohiiit, line 128: `put_bucket_lifecycle_configuration` replaces the entire lifecycle configuration, not just this rule. If the bucket has other lifecycle rules (e.g., incomplete multipart upload cleanup), they'll be deleted.

Consider fetching existing rules first and merging, or document this behavior prominently and add a test that verifies behavior with pre-existing rules.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
```diff
@@ -0,0 +1,224 @@
...
+                );
+            }
+
+            // Tag all artifacts
+            let tags = vec![(STORAGE_EXPIRATION_TAG_KEY.to_string(), STORAGE_EXPIRATION_TAG_VALUE.to_string())];
+            let mut all_tagged = true;
+            let mut tagged_count = 0;
+
+            for path in &paths_to_tag {
+                match config.storage().tag_object(path, tags.clone()).await {
```
Mohiiit: `tags.clone()` allocates a new `Vec<(String, String)>` on each iteration. Consider creating the tag strings once outside the loop and cloning only those, or modifying `tag_object` to accept `&[(&str, &str)]`.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
```diff
@@ -0,0 +1,224 @@
...
+use chrono::Utc;
+use std::sync::Arc;
+use tracing::{debug, error, info, warn};
+
+pub struct StorageCleanupTrigger;
+
+#[async_trait]
+impl JobTrigger for StorageCleanupTrigger {
+    async fn run_worker(&self, config: Arc<Config>) -> color_eyre::Result<()> {
+        info!("StorageCleanupTrigger started");
```
Mohiiit, line 22: This worker runs every 10 minutes. Consider changing most `info!` logs to `debug!` to reduce noise:
- Lines 21, 51: started/completed → `debug!`
- Lines 76, 148, 150-154, 214: per-job details → `debug!`

Keep only the final summary (line 217) at `info!` level. Other triggers don't log started/completed at entry.

Mohiiit (commented):
Nice idea to use S3 lifecycle rules for cleanup. A few questions and suggestions in the comments.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    /// Storage cleanup should always run, regardless of failed jobs
+    async fn is_worker_enabled(&self, _config: Arc<Config>) -> color_eyre::Result<bool> {
+        Ok(true)
+    }
+}
+
+impl StorageCleanupTrigger {
+    /// Process completed StateTransition jobs that haven't had their artifacts tagged
+    async fn process_completed_jobs(&self, config: &Arc<Config>) -> color_eyre::Result<()> {
```
Mohiiit, line 64: `process_completed_jobs` is 160+ lines. Consider decomposing into smaller functions for readability and testability:
- `collect_artifact_paths(job_id)` - gather paths from all 4 locations
- `tag_artifacts(paths)` - handle tagging with error recovery
- `mark_job_tagged(job)` - update metadata

This also enables unit testing each piece independently.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
heemankv: Done
info! - Started, summary, completed (3 logs per run)

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
heemankv: Valid, fixed

orchestrator/src/setup/aws/s3.rs
heemankv: Currently there are no other rules that we set,
Added a todo for this to be taken care of when we add another rule.

orchestrator/src/types/constant.rs
heemankv: Valid, done

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
heemankv: Valid, fixed and added appropriate tests

orchestrator/src/tests/workers/storage_cleanup/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    Ok(())
+}
+
+/// Test that list_files_in_dir handles pagination correctly for large directories.
+/// S3 returns max 1000 objects per request, so we need to handle pagination.
+#[rstest]
+#[tokio::test]
+async fn test_list_files_in_dir_with_pagination() -> Result<(), Box<dyn Error>> {
```
Mohiiit: Is this pagination test needed for the storage cleanup use case?

Looking at the code, `list_files_in_dir` is called on per-job directories (`artifacts/batch/{job_id}/`, `blob/batch/{job_id}/`, `{job_id}/`) which contain at most ~10-15 files (proof.json, cairo_pie.zip, snos_output.json, a few blobs, etc.).

We'll never have 1000+ files in these directories. Consider removing this test or moving it to a general S3 client test if pagination is important elsewhere.

orchestrator/src/tests/workers/storage_cleanup/mod.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    // Mock storage to return files but fail on tagging
+    storage.expect_list_files_in_dir().returning(|_| Ok(vec!["file1.json".to_string()]));
+
+    // First tag succeeds, but simulate a real S3 error (not NoSuchKey)
+    storage.expect_tag_object().returning(|_, _| {
+        Err(crate::core::client::storage::StorageError::ObjectStreamError("Connection timeout".to_string()))
+    });
+
+    // update_job should NOT be called since tagging failed
```
Mohiiit: Two suggestions:

1. Comment says "update_job should NOT be called" but there's no verification. Consider adding `database.expect_update_job().never();`

2. Consider adding a happy-path test that verifies the full successful flow: lock acquired → job fetched → artifacts tagged → `update_job` called with `storage_artifacts_tagged_at` set → lock released. Currently all 4 unit tests cover failure scenarios only.

Mohiiit (commented):
A few questions about the test coverage.

orchestrator/src/tests/workers/storage_cleanup/mod.rs
heemankv: Removed

orchestrator/src/tests/workers/storage_cleanup/mod.rs
heemankv: Done both

Mohiiit (approved):
LGTM!

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
byteZorvin: Nit: I would prefer to not have started completed as well though

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
```diff
@@ -0,0 +1,205 @@
...
+                    tagged_count += 1;
+                    debug!(job_id = %job_id, path = %path, "Tagged artifact for expiration");
+                }
+                Err(e) => {
+                    let error_str = e.to_string();
+                    // Handle non-existent files gracefully (file may have been deleted)
+                    if error_str.contains("NoSuchKey")
+                        || error_str.contains("not found")
+                        || error_str.contains("does not exist")
+                    {
```
byteZorvin: Nit: May be we can create variants for StorageErrors here
```
match e {
    StorageError::NotFound(_) => { /* skip */ }
    _ => return (tagged_count, false),
}
```

orchestrator/src/core/client/storage/mod.rs
```diff
@@ -29,4 +29,18 @@ pub trait StorageClient: Send + Sync {
...
+    /// Tags can be used for lifecycle management, cost allocation, or organization.
+    ///
+    /// # Arguments
+    /// * `key` - The key of the object to tag
+    /// * `tags` - A slice of (key, value) pairs representing the tags (passed by reference to avoid cloning)
+    ///
+    /// # Returns
+    /// * `Ok(())` - If tagging was successful
+    /// * `Err(StorageError)` - If tagging fails
+    async fn tag_object(&self, key: &str, tags: &[(String, String)]) -> Result<(), StorageError>;
```
byteZorvin: Nit: Consider using `&[(&str, &str)]` instead of `&[(String, String)]`

Since the tag keys and values are static strings (STORAGE_EXPIRATION_TAG_KEY, STORAGE_EXPIRATION_TAG_VALUE), the signature could accept &[(&str, &str)] to avoid unnecessary allocations

This would simplify the call site from:


`let tags = vec![(STORAGE_EXPIRATION_TAG_KEY.to_string(), STORAGE_EXPIRATION_TAG_VALUE.to_string())];`
To:


`let tags = [(STORAGE_EXPIRATION_TAG_KEY, STORAGE_EXPIRATION_TAG_VALUE)];`

orchestrator/src/setup/aws/s3.rs
```diff
@@ -19,61 +23,65 @@ impl Resource for InnerAWSS3 {
...
+        };
+
         // Check if the bucket already exists
-        // If it does, return the existing bucket name and location
+        // If it does, skip creation but still ensure lifecycle rule is set
         if self.check_if_exists(&args.bucket_identifier).await? {
             warn!("  S3 bucket {} already exists , skipping creation", &args.bucket_identifier);
-            return Ok(());
-        }
+        } else {
```
byteZorvin: Nit: why uncessary nesting here though? We can just shift lifecycle rule funtion above the if

orchestrator/src/tests/workers/storage_cleanup/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+    let artifact_tags = get_object_tags(&services.provider_config, &bucket_name, &artifact_file).await?;
+    assert!(
+        artifact_tags.iter().any(|(k, v)| k == STORAGE_EXPIRATION_TAG_KEY && v == STORAGE_EXPIRATION_TAG_VALUE),
+        "Artifact should have expiration tag: {:?}",
+        artifact_tags
+    );
+
+    // Step 5: Verify job is no longer in untagged list
+    let untagged_jobs_after = config.database().get_jobs_without_storage_artifacts_tagged(Some(100)).await?;
+    assert!(!untagged_jobs_after.iter().any(|j| j.internal_id == job_id), "Job should be marked as tagged");
```
byteZorvin, line 232: we can advance time and test the objects gets cleaned too.

byteZorvin (commented):
Well structured, changes.
The elaborate description helped a lot too.
Few nits

heemankv (commented):
Thanks for the detailed review! Addressed most of your nits:

**✅ Implemented:**

1. **tag_object signature** (storage/mod.rs:45): Changed from `&[(String, String)]` to `&[(&str, &str)]` to avoid allocations. Call sites now use `[(KEY, VALUE)]` instead of `vec![(KEY.to_string(), VALUE.to_string())]`

2. **StorageError::NotFound variant** (storage_cleanup.rs:174): Added `StorageError::NotFound` enum variant and replaced string-based error detection with proper match pattern:
   ```rust
   Err(e) => match e {
       StorageError::NotFound(_) => { /* skip */ }
       _ => return (tagged_count, false),
   }
   ```

3. **Remove unnecessary nesting** (s3.rs:42): Inverted the condition from `if exists { warn } else { create }` to `if !exists { create } else { warn }` to reduce nesting level

**📋 Not implementing (with reasons):**

4. **"started" and "completed" logs** (storage_cleanup.rs:21): Keeping these info-level logs. Since this is a trigger worker that runs periodically in the background, these lifecycle logs are essential for operational visibility - we need to know when the worker starts/completes each run for monitoring and debugging.

5. **Time advancement test** (tests/storage_cleanup/mod.rs:232): Our current tests already verify:
   - Lifecycle rule is created correctly (`test_setup_lifecycle_rule_creates_correct_config`)
   - Objects are tagged properly (`test_tag_object_success_and_error_cases`)
   
   The actual time-based deletion is AWS S3's lifecycle management responsibility, not ours. We're testing our integration (setting policy + tagging), not AWS's infrastructure functionality.

Let me know if you'd like me to reconsider any of these!

orchestrator/src/core/client/storage/mod.rs
heemankv: Done! Changed from `&[(String, String)]` to `&[(&str, &str)]`. This eliminated the allocations - call sites now use `[(STORAGE_EXPIRATION_TAG_KEY, STORAGE_EXPIRATION_TAG_VALUE)]` instead of the vec with `.to_string()` calls.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
heemankv: Done! Added `StorageError::NotFound` variant and replaced the string-based error detection with this match pattern. Much cleaner 👍

orchestrator/src/setup/aws/s3.rs
heemankv: Done! Inverted the condition to `if !bucket_exists { create } else { warn }` to reduce the nesting level. The lifecycle rule call now sits at the same level after the if-else block.

orchestrator/src/worker/event_handler/triggers/storage_cleanup.rs
heemankv: I'd prefer to keep these info-level logs. Since this is a trigger worker that runs periodically in the background, these lifecycle logs are essential for operational visibility - we need to know when the worker starts/completes each run for monitoring and debugging purposes.

orchestrator/src/tests/workers/storage_cleanup/mod.rs
heemankv: Our current tests already verify the critical integration points:
- `test_setup_lifecycle_rule_creates_correct_config` - verifies the lifecycle rule is created with correct config
- `test_tag_object_success_and_error_cases` - verifies objects are tagged properly

The actual time-based deletion is AWS S3's lifecycle management responsibility. We're testing our integration (setting the policy + applying tags), not AWS's infrastructure. Testing time advancement would require mocking AWS's internal lifecycle engine, which doesn't add value since we can trust their system works as documented.

byteZorvin (approved):

PR #947, opened 2026-01-14
chore: improve batching performance with batched queries

prkpndy:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

## What is the new behavior?

-
-
-

### DIFF
--- a/orchestrator/src/compression/batch_rpc.rs
+++ b/orchestrator/src/compression/batch_rpc.rs
@@ -223,13 +223,15 @@ impl BatchRpcClient {
             let value = match response {
                 Ok(val) => Self::parse_felt_result(val)?,
                 Err(err) => {
-                    // For storage, any error means we can't get the value
-                    // Log and return ZERO as a safe default (storage slots default to zero)
-                    warn!(
-                        "Failed to get storage at contract={:#x} key={:#x}: {} (code={})",
-                        contract_addr, key, err.message, err.code
+                    error!(
+                        "Failed to get pre-range storage value for contract: {}, key: {} at block {}: {:?}",
+                        contract_addr, key, block_param, err
                     );
-                    Felt::ZERO
+                    return Err(BatchRpcError::RpcError {
+                        id: idx as u64,
+                        code: err.code,
+                        message: err.message.clone(),
+                    });
                 }
             };
             results.insert((*contract_addr, *key), value);
@@ -272,11 +274,16 @@ impl BatchRpcClient {
                         None
                     } else {
                         // Unexpected error, log but return None to not block processing
-                        warn!(
-                            "Failed to get class hash for contract={:#x}: {} (code={})",
-                            contract_addr, err.message, err.code
+                        let err_message = format!(
+                            "Failed to get class hash for contract: {} at block {}: {:?}",
+                            contract_addr, block_param, err
                         );
-                        None
+                        error!("{}", &err_message);
+                        return Err(BatchRpcError::RpcError {
+                            id: idx as u64,
+                            code: err.code,
+                            message: err.message.clone(),
+                        });
                     }
                 }
             };
--- a/orchestrator/src/compression/batch_rpc.rs
+++ b/orchestrator/src/compression/batch_rpc.rs
@@ -116,6 +116,7 @@ impl BatchRpcClient {
     ///
     /// Returns a HashMap mapping (contract_address, key) to the storage value.
     /// If a storage slot doesn't exist, ZERO is returned for that slot.
+    #[allow(clippy::type_complexity)]
     pub async fn batch_get_storage_at(
         &self,
         queries: Vec<(Felt, Felt)>, // (contract_address, key)
--- a/orchestrator/src/compression/batch_rpc.rs
+++ b/orchestrator/src/compression/batch_rpc.rs
@@ -1,0 +1,417 @@
+//! Batch JSON-RPC client for efficient Starknet queries.
+//!
+//! This module provides a batch-capable JSON-RPC client that can send multiple
+//! RPC calls in a single HTTP request, significantly reducing network overhead.
+
+use futures::stream::{self, StreamExt};
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+use starknet_core::types::{BlockId, Felt};
+use std::collections::HashMap;
+use std::time::Duration;
+use thiserror::Error;
+use tracing::{debug, error, warn};
+
+/// Default batch size (number of RPC calls per HTTP request)
+pub const DEFAULT_BATCH_SIZE: usize = 100;
+/// Default number of concurrent batch requests
+pub const DEFAULT_MAX_CONCURRENT_BATCHES: usize = 10;
+/// Default retry attempts for failed batches
+pub const DEFAULT_MAX_RETRIES: u64 = 3;
+/// Default delay between retries in seconds
+pub const DEFAULT_RETRY_DELAY_SECS: u64 = 2;
+
+/// Configuration for batch RPC operations
+#[derive(Debug, Clone)]
+pub struct BatchRpcConfig {
+    /// Maximum number of RPC calls per batch request
+    pub batch_size: usize,
+    /// Maximum concurrent batch requests
+    pub max_concurrent_batches: usize,
+    /// Retry attempts for failed batches
+    pub max_retries: u64,
+    /// Delay between retries in seconds
+    pub retry_delay_secs: u64,
+}
+
+impl Default for BatchRpcConfig {
+    fn default() -> Self {
+        Self {
+            batch_size: DEFAULT_BATCH_SIZE,
+            max_concurrent_batches: DEFAULT_MAX_CONCURRENT_BATCHES,
+            max_retries: DEFAULT_MAX_RETRIES,
+            retry_delay_secs: DEFAULT_RETRY_DELAY_SECS,
+        }
+    }
+}
+
+/// Error types for batch RPC operations
+#[derive(Error, Debug)]
+pub enum BatchRpcError {
+    #[error("HTTP request failed: {0}")]
+    HttpError(#[from] reqwest::Error),
+
+    #[error("JSON serialization failed: {0}")]
+    SerializationError(#[from] serde_json::Error),
+
+    #[error("RPC error for request {id}: code={code}, message={message}")]
+    RpcError { id: u64, code: i64, message: String },
+
+    #[error("Missing response for request ID {0}")]
+    MissingResponse(u64),
+
+    #[error("All retries exhausted: {0}")]
+    RetriesExhausted(String),
+
+    #[error("Invalid response format: {0}")]
+    InvalidResponse(String),
+}
+
+/// A single JSON-RPC request
+#[derive(Serialize, Debug)]
+struct JsonRpcRequest<'a> {
+    jsonrpc: &'static str,
+    method: &'a str,
+    params: serde_json::Value,
+    id: u64,
+}
+
+/// A single JSON-RPC response
+#[derive(Deserialize, Debug)]
+struct JsonRpcResponse {
+    id: u64,
+    result: Option<serde_json::Value>,
+    error: Option<JsonRpcErrorData>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+struct JsonRpcErrorData {
+    code: i64,
+    message: String,
+}
+
+/// Starknet error code for contract not found
+const CONTRACT_NOT_FOUND_CODE: i64 = 20;
+
+/// Batch RPC client for efficient Starknet queries
+#[derive(Clone)]
+pub struct BatchRpcClient {
+    client: Client,
+    rpc_url: String,
+    config: BatchRpcConfig,
+}
+
+impl BatchRpcClient {
+    /// Create a new BatchRpcClient
+    pub fn new(rpc_url: String, config: BatchRpcConfig) -> Self {
+        Self { client: Client::new(), rpc_url, config }
+    }
+
+    /// Create a new BatchRpcClient with default configuration
+    pub fn with_defaults(rpc_url: String) -> Self {
+        Self::new(rpc_url, BatchRpcConfig::default())
+    }
+
+    /// Batch get storage values for multiple (contract, key) pairs at a specific block
+    ///
+    /// Returns a HashMap mapping (contract_address, key) to the storage value.
+    /// If a storage slot doesn't exist, ZERO is returned for that slot.
+    pub async fn batch_get_storage_at(
+        &self,
+        queries: Vec<(Felt, Felt)>, // (contract_address, key)
+        block_id: BlockId,
+    ) -> Result<HashMap<(Felt, Felt), Felt>, BatchRpcError> {
+        if queries.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        let block_param = Self::block_id_to_param(&block_id);
+
+        // Chunk queries into batches
+        let chunks: Vec<Vec<(Felt, Felt)>> = queries.chunks(self.config.batch_size).map(|c| c.to_vec()).collect();
+
+        debug!("Executing {} storage queries in {} batches", queries.len(), chunks.len());
+
+        // Execute batches concurrently
+        let results: Vec<Result<HashMap<(Felt, Felt), Felt>, BatchRpcError>> = stream::iter(chunks)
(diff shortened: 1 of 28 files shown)


Mohiiit (commented):
## Code Review Summary

Overall, this is a well-designed batching optimization that should provide significant performance improvements (~10-20x reduction in HTTP requests). The architecture is clean with good separation of concerns.

### Inline Comments

I've left a few inline comments on specific areas - mostly questions about intended behavior and a suggestion for improved testability.

### Test Coverage Gaps ⚠️

The current testing relies heavily on integration tests with real Sepolia data, which is great for the happy path. However, there are significant gaps in test coverage that would be valuable to address:

**Error Handling & Retry Logic (High Priority)**
- No tests for retry exhaustion scenarios
- No tests for partial batch failures (some requests succeed, some fail)
- No tests for HTTP timeouts and network errors
- No tests for invalid JSON responses

**Edge Cases (Medium Priority)**
- Empty input collections (`[]`, `{}`)
- Single-item batches
- Batch boundary tests (99, 100, 101 items to verify chunking)
- Very large batches (1000+ items)

**Response Parsing (Medium Priority)**
- Malformed Felt hex strings
- Missing response IDs
- Responses with neither result nor error

**Specific Functions Missing Direct Tests:**
| Function | Gap |
|----------|-----|
| `send_batch_with_retry()` | All retry scenarios |
| `send_batch()` | HTTP failure handling |
| `process_storage_diffs_batched()` | Empty inputs, all-zero values |
| `process_deployed_contracts_and_replaced_classes_batched()` | Empty inputs, unchanged class hash |

Adding these tests would provide confidence in error handling and make future refactoring safer.

Mohiiit (commented):
Overall, this is a well-designed batching optimization with clean architecture. I have a few questions about intended behavior and a suggestion for improved testability.

**Test Coverage Gaps:**
The current tests focus on the happy path with real Sepolia data. Consider adding tests for:
- Error handling paths (retry exhaustion, partial batch failures, network errors)
- Edge cases (empty inputs, batch boundaries at 99/100/101 items)
- Response parsing (malformed responses, ID validation)

Key functions missing direct tests: `send_batch_with_retry()`, `send_batch()`

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,412 @@
...
+        block_id: BlockId,
+    ) -> Result<HashMap<(Felt, Felt), Felt>, BatchRpcError> {
+        if queries.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        let block_param = Self::block_id_to_param(&block_id);
+
+        // Chunk queries into batches
+        let chunks: Vec<Vec<(Felt, Felt)>> = queries.chunks(self.config.batch_size).map(|c| c.to_vec()).collect();
```
Mohiiit, line 143: **Nit:** The `.collect()` here loads all batch results into memory before processing. For very large state updates, this could increase memory pressure. Is this intentional?

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,412 @@
...
+        let responses = self.send_batch_with_retry(&requests).await?;
+
+        // Parse responses
+        let mut results = HashMap::new();
+        for (idx, (contract_addr, key)) in queries.iter().enumerate() {
+            let response = responses.get(&(idx as u64)).ok_or(BatchRpcError::MissingResponse(idx as u64))?;
+
+            let value = match response {
+                Ok(val) => Self::parse_felt_result(val)?,
+                Err(err) => {
```
Mohiiit, line 228: Is this behavior intentional? Currently all RPC errors (including network failures) become `Felt::ZERO`. Should we distinguish "storage not found" (expected) from network failures (unexpected)?

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,412 @@
...
+            let response = responses.get(&(idx as u64)).ok_or(BatchRpcError::MissingResponse(idx as u64))?;
+
+            let value = match response {
+                Ok(val) => Some(Self::parse_felt_result(val)?),
+                Err(err) => {
+                    if err.code == CONTRACT_NOT_FOUND_CODE {
+                        // Contract doesn't exist, this is expected
+                        None
+                    } else {
+                        // Unexpected error, log but return None to not block processing
```
Mohiiit, line 279: Same question - is it intentional that all errors return `None`? Network failures would be silently treated as "contract doesn't exist."

orchestrator/src/compression/squash.rs
```diff
@@ -185,189 +171,285 @@ impl StateDiffMap {
...
+    let pre_range_values = if storage_queries.is_empty() {
+        HashMap::new()
+    } else {
+        batch_client
+            .batch_get_storage_at(storage_queries, block_id)
+            .await
+            .map_err(|e| JobError::ProviderError(format!("Failed to batch get storage values: {}", e)))?
+    };
+
+    // Step 4: Build storage diffs by comparing values
```
Mohiiit, line 253: Consider extracting the comparison logic (pre-range vs new value) into pure functions like `build_entries_for_existing_contract()`. This would allow unit testing without mocking RPC calls - important given the limited test coverage for edge cases.

orchestrator/src/compression/batch_rpc.rs
prkpndy: fair point. actually, we should return error in all cases since we are already checking for contract existence. it never returns storage not found (for non-existing address, it returns 0, which works for us). thanks for pointing it out!

orchestrator/src/compression/batch_rpc.rs
prkpndy: nopes. updated it to return None only when contract is not found

orchestrator/src/compression/batch_rpc.rs
prkpndy: umm, this is only holding the chunks, so basically vector of tuple (contract, storage). it wont be very big. tried to remove that but can't do it because chunks() method returns borrowed slices which don't satisfy Send bounds for async. have also added a comment for this

orchestrator/src/compression/squash.rs
prkpndy: done

prkpndy (commented):
resolved pr comments

byteZorvin (commented):
Starting review of batching performance improvements.

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,413 @@
...
+    ) -> Result<HashMap<u64, Result<serde_json::Value, JsonRpcErrorData>>, BatchRpcError> {
+        let body = serde_json::to_string(requests)?;
+
+        let response = self
+            .client
+            .post(self.rpc_url.as_str())
+            .header("Content-Type", "application/json")
+            .body(body)
+            .send()
+            .await?
```
byteZorvin, line 338: The HTTP client has no timeout configured. Consider adding a timeout to prevent long-running or hanging requests from blocking indefinitely:
```rust
.timeout(Duration::from_secs(30))
```

orchestrator/src/compression/squash.rs
```diff
@@ -1,32 +1,25 @@
...
+///
+/// This function uses batch JSON-RPC calls to efficiently fetch pre-range block
+/// storage values and class hashes, reducing the number of HTTP requests by ~10x.
 pub async fn squash(
     state_updates: Vec<&StateUpdate>,
     pre_range_block: Option<u64>,
-    provider: &Arc<JsonRpcClient<HttpTransport>>,
+    batch_client: &BatchRpcClient,
 ) -> Result<StateUpdate, JobError> {
+    info!("Squashing state updates");
```
byteZorvin: Nit: Consider using debug for start and end messages, could be too verbose for production loads.

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,413 @@
...
+        match block_id {
+            BlockId::Number(n) => serde_json::json!({ "block_number": n }),
+            BlockId::Hash(h) => serde_json::json!({ "block_hash": format!("{:#x}", h) }),
+            BlockId::Tag(tag) => match tag {
+                starknet_core::types::BlockTag::Latest => serde_json::json!("latest"),
+                starknet_core::types::BlockTag::L1Accepted => serde_json::json!("l1_accepted"),
+                starknet_core::types::BlockTag::PreConfirmed => serde_json::json!("pre_confirmed"),
+            },
+        }
+    }
```
byteZorvin: This manual conversion can be removed. starknet_core::types::BlockId already implements Serialize with the exact same JSON output format (see starknet-core/src/types/serde_impls.rs).

You can simplify by using serde_json::to_value(block_id)? directly instead of this function.

orchestrator/src/compression/batch_rpc.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    /// Parse a Felt from JSON-RPC response value
+    fn parse_felt_result(value: &serde_json::Value) -> Result<Felt, BatchRpcError> {
+        let hex_str =
+            value.as_str().ok_or_else(|| BatchRpcError::InvalidResponse("Expected string for Felt".to_string()))?;
+
+        Felt::from_hex(hex_str)
+            .map_err(|e| BatchRpcError::InvalidResponse(format!("Invalid Felt hex '{}': {}", hex_str, e)))
+    }
```
byteZorvin, line 365: This can also be done with starknet 
```
serde_json::from_value::<Felt>(value.clone())
    .map_err(|e| BatchRpcError::InvalidResponse(e.to_string()))```

**Review by byteZorvin — CHANGES_REQUESTED** [review]

File: orchestrator/src/compression/squash.rs [hunk 9]
```diff
@@ -39,14 +32,17 @@ pub async fn squash(
...
+
+    let state_diff = state_diff_map.get_state_diff_batched(pre_range_block, batch_client).await?;
 
     // Create the merged StateUpdate
     let mut merged_update = StateUpdate { block_hash, new_root, old_root, state_diff };
 
     // Sort the merged StateUpdate
     sort_state_diff(&mut merged_update);
 
+    info!("Squashed state updates");
```
Mohiiit: I think the logging in this function could be changed to debug

orchestrator/src/compression/squash.rs
```diff
@@ -185,43 +171,195 @@ impl StateDiffMap {
...
-    provider: &Arc<JsonRpcClient<HttpTransport>>,
-    pre_range_block_option: Option<u64>,
+/// Process storage diffs using batched RPC calls.
+///
+/// For each contract with storage changes:
+/// 1. Check if the contract existed at pre_range_block
+/// 2. For existing contracts, fetch all pre-range storage values
+/// 3. Filter out entries where the value hasn't changed
+/// 4. For new contracts (didn't exist), filter out zero values
+async fn process_storage_diffs_batched(
```
Mohiiit, line 181: I think this function can be divided in further function and it would provide better readability + ability to test things better

orchestrator/src/compression/squash.rs
```diff
@@ -185,43 +171,195 @@ impl StateDiffMap {
...
+                None
+            }
+        })
+        .collect();
+
+    if entries.is_empty() {
+        None
+    } else {
+        Some(ContractStorageDiffItem { address: contract_addr, storage_entries: entries })
+    }
```
Mohiiit:
```suggestion
    (!entries.is_empty()).then_some(ContractStorageDiffItem {
    address: contract_addr,
    storage_entries: entries,
})
```

orchestrator/src/compression/squash.rs
```diff
@@ -234,140 +372,212 @@ async fn process_deployed_contracts_and_replaced_classes(
...
+                    }],
+                }],
+                deployed_contracts: vec![],
+                declared_classes: vec![],
+                deprecated_declared_classes: vec![],
+                nonces: vec![],
+                replaced_classes: vec![],
+                migrated_compiled_classes: None,
+            },
+        };
```
Mohiiit: maybe a fixture can be sued here?

orchestrator/src/compression/squash.rs
```diff
@@ -234,140 +372,212 @@ async fn process_deployed_contracts_and_replaced_classes(
...
+    // ========================================================================
+    // Tests for pure comparison functions
+    // ========================================================================
+
+    /// Comprehensive test for build_storage_diff_item covering all edge cases:
+    /// - Existing contract: unchanged values filtered, changed values included
+    /// - New contract (no pre-range): zero values filtered, non-zero included
+    /// - Returns None when all entries filtered out
+    #[test]
+    fn test_build_storage_diff_item() {
```
Mohiiit, line 447: nice test case indeed!

orchestrator/src/compression/squash.rs
```diff
@@ -234,140 +372,212 @@ async fn process_deployed_contracts_and_replaced_classes(
...
-        ))
-    })
+    /// Comprehensive test for filter_changed_replaced_classes covering all edge cases:
+    /// - Unchanged class hash (filtered)
+    /// - Changed class hash (included)
+    /// - New contract with None prev (included)
+    /// - Contract missing from prev map (included)
+    /// - Empty input (empty output)
+    #[test]
+    fn test_filter_changed_replaced_classes() {
```
Mohiiit, line 515: I think this would be very clean with rstest, good test cases tho!

Mohiiit (commented):
good tests, no logic change requested, more on code practices!

orchestrator/src/compression/squash.rs
prkpndy: hmm.. I am not sure if we'll gain anything by breaking it down further. This function has a simple purpose that given a contract address, its storage entries and a map of storage entries in pre-range, it'll give me the diff for the contract. Have written test for it as well. breaking it down further will complicate things and make it verbose as well. wdyt?

orchestrator/src/compression/batch_rpc.rs
prkpndy: done

orchestrator/src/compression/squash.rs
prkpndy: done

orchestrator/src/compression/batch_rpc.rs
prkpndy: done

orchestrator/src/compression/batch_rpc.rs
prkpndy: done

orchestrator/src/compression/squash.rs
prkpndy: done

orchestrator/src/compression/squash.rs
prkpndy: done

orchestrator/src/compression/squash.rs
prkpndy: hmm, can do this but still to make the function generic, most of the values would needed to be passed as arguments only - block hash, roots, storage diffs, etc. also, it's being used only in 2 tests. would be an overkill imo 😅

orchestrator/src/compression/squash.rs
prkpndy: can do if we need to use fixtures

byteZorvin (approved):
GG !!

orchestrator/src/compression/squash.rs
Mohiiit: gain is not on logic side at all! it's more about having more tests and simpler functions so that it's easier to read and test

orchestrator/src/compression/squash.rs
Mohiiit: yes we should use that

orchestrator/src/compression/squash.rs
Mohiiit: I think we should follow the best practices nonetheless?

Mohiiit (approved):

orchestrator/src/compression/squash.rs
prkpndy: added helper functions to create state update

orchestrator/src/compression/squash.rs
prkpndy: no need for rstest. fixtures were getting too complicated for this. added helper functions

PR #949, opened 2026-01-14
fix: preserve VM tracebacks when error chain ends in Constructor

Mohiiit:
Fixes receipt_commitment mismatch caused by Madara filtering out Cairo VM tracebacks from `revert_reason` when the error chain ends in a `Constructor` frame. Treats `Constructor` as a final error frame (similar to `StringFrame`) rather than a filtering entry point.


heemankv (approved):
Since we've added a test for this and all tests are passing, I am good with these changes

byteZorvin (approved):

PR #950, opened 2026-01-15
feat: improve error history tracking for failed jobs

byteZorvin:
## Pull Request type

- [x] Feature

When a job fails after multiple retries, only the final error message is preserved. Previous failure reasons from earlier attempts are lost, making debugging difficult.

Resolves: #NA

## What is the new behavior?

- Preserves full error history by prepending each failure reason to existing ones
- Improved DLQ message to clarify the job exhausted retries
- SNS alerts now include the complete failure history

No

## Other information

Minor `.gitignore` update included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ benchmarking/reports/*.json
 # idea
 .idea
 .zed
+madara/.vscode/
 
 # local runs
 run.sh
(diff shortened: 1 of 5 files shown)


heemankv (commented):
As discussed offline,
Let's change to a array of failure reasons. | decided for cleanliness of failure reason management.
Also please decide on a convention, wether we want to prepend or append the new failure reason.

Mohiiit (approved):
Everything looks alright, and well done on the test cases - that really helps.

I have a single question regarding how we are inserting the reasons. Currently, we are inserting them at the beginning. However, throughout the codebase, we are never popping from the vector itself.

Maybe we can just push items into the vector instead? That logically makes sense to me, but I'm open to discussion. If it's in a linear format (for example: the first error comes at 0, then 1, then 2), I can always check the last element to see what the latest one is. Since it's just a vector, what are your views?

 Everything else looks good. GG

heemankv (approved):

PR #954, opened 2026-01-17
feat(orchestrator): improve graceful shutdown handling

byteZorvin:
## Pull Request type

Shutdown doesn't properly wait for all components to finish and errors can cause early exit.

## What is the new behavior?

- Return task handles from `setup_server` and `initialize_worker` for proper cleanup
- Collect errors from all components during shutdown instead of failing fast
- Properly abort server and await worker handle during shutdown

No

### DIFF
--- a/orchestrator/src/main.rs
+++ b/orchestrator/src/main.rs
@@ -111,25 +111,27 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
             || async {
                 let mut had_errors = false;
 
-                // 1. Trigger worker controller shutdown (cancels token and waits for workers)
+                // 1. Gracefully shutdown the server
+                if let Err(e) = server_handle.shutdown().await {
+                    warn!("Server shutdown error: {:?}", e);
+                    had_errors = true;
+                }
+
+                // 2. Trigger worker controller shutdown (cancels token and waits for workers)
                 if let Err(e) = worker_controller.shutdown().await {
-                    warn!("Worker controller shutdown error: {:?}", e);
+                    error!("Worker controller shutdown error: {:?}", e);
                     had_errors = true;
                 }
 
-                // 2. Wait for the worker task to complete
+                // 3. Wait for the worker task to complete
                 if let Err(e) = worker_handle.await {
-                    warn!("Worker task error: {:?}", e);
+                    error!("Worker task error: {:?}", e);
                     had_errors = true;
                 }
-
-                // 3. Abort the server (axum doesn't have graceful shutdown by default)
-                server_handle.abort();
-                let _ = server_handle.await; // Ignore abort error
 
                 // 4. Shutdown OTEL instrumentation
                 if let Err(e) = instrumentation.shutdown() {
-                    warn!("OTEL instrumentation shutdown error: {}", e);
+                    error!("OTEL instrumentation shutdown error: {}", e);
                     had_errors = true;
                 }
 
--- a/orchestrator/src/main.rs
+++ b/orchestrator/src/main.rs
@@ -27,7 +27,7 @@ use orchestrator::utils::signal_handler::SignalHandler;
 use orchestrator::worker::initialize_worker;
 use orchestrator::OrchestratorResult;
 use std::sync::Arc;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 #[global_allocator]
 static A: jemallocator::Jemalloc = jemallocator::Jemalloc;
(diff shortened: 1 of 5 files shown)


prkpndy (approved):
lgtm 🚀

orchestrator/src/main.rs
```diff
@@ -88,34 +88,56 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
-                worker_controller.shutdown().await?;
-
-                // Analytics Shutdown
-                instrumentation.shutdown()?;
-
-                info!("All components shutdown successfully");
+                let mut had_errors = false;
+
+                // 1. Trigger worker controller shutdown (cancels token and waits for workers)
+                if let Err(e) = worker_controller.shutdown().await {
```
heemankv, line 121: The shutdown order should be flipped - **server should stop first, then workers**.

**Reason:** The API has endpoints that can queue work (`/jobs/:id/process`, `/jobs/:id/retry`, admin bulk operations). If workers shutdown first while the server still accepts requests, those queued messages become orphaned.

Suggested order:
1. Stop server (gracefully, so in-flight requests complete)
2. Shutdown workers
3. OTEL instrumentation

orchestrator/src/main.rs
```diff
@@ -88,34 +88,56 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                    warn!("Worker controller shutdown error: {:?}", e);
```
heemankv: This should be `error!` instead of `warn!`. Shutdown failures are errors, and production alerting systems often filter by log level.

orchestrator/src/main.rs
```diff
@@ -88,34 +88,56 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                    had_errors = true;
+                }
+
+                // 2. Wait for the worker task to complete
+                if let Err(e) = worker_handle.await {
+                    warn!("Worker task error: {:?}", e);
```
heemankv: Same here - should be `error!` instead of `warn!`.

orchestrator/src/main.rs
```diff
@@ -88,34 +88,56 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                    had_errors = true;
+                }
+
+                // 3. Abort the server (axum doesn't have graceful shutdown by default)
+                server_handle.abort();
```
heemankv: Consider using axum's graceful shutdown instead of `abort()`.

With `.with_graceful_shutdown(signal)`, in-flight requests can complete before the server stops, rather than being abruptly terminated.

See: https://github.com/tokio-rs/axum/blob/main/examples/graceful-shutdown/src/main.rs

orchestrator/src/main.rs
```diff
@@ -88,34 +88,56 @@ async fn run_orchestrator(run_cmd: &RunCmd) -> OrchestratorResult<()> {
...
+                let _ = server_handle.await; // Ignore abort error
+
+                // 4. Shutdown OTEL instrumentation
+                if let Err(e) = instrumentation.shutdown() {
+                    warn!("OTEL instrumentation shutdown error: {}", e);
```
heemankv: Same here - should be `error!` instead of `warn!`.

orchestrator/src/utils/signal_handler.rs
```diff
@@ -141,17 +141,11 @@ impl SignalHandler {
...
-        let remaining_time = timeout_duration.saturating_sub(elapsed);
 
         match shutdown_result {
             Ok(Ok(())) => {
-                info!("Graceful shutdown completed successfully");
-                if remaining_time > tokio::time::Duration::from_secs(0) {
-                    info!("Waiting remaining {:?} to ensure graceful shutdown", remaining_time);
-                    tokio::time::sleep(remaining_time).await;
-                }
+                info!("Graceful shutdown completed successfully in {:?}", elapsed);
```
heemankv, line 151: TODO: Add tests for graceful shutdown behavior. This is difficult to test due to OS-level signal handling, but the timeout and error collection logic could be unit tested with mock futures.

heemankv (commented):
**Additional suggestion (not in diff):**

Consider reducing the default graceful shutdown timeout from `120` to `30` seconds in `orchestrator/src/cli/mod.rs:197`.

120 seconds is excessive for graceful shutdown - if workers can't finish in 30s, they're likely stuck. 30s aligns with Kubernetes' default `terminationGracePeriodSeconds`.

heemankv (approved):

PR #956, opened 2026-01-20
fix: improve blob transaction gas price retry logic

heemankv:
## Problem

The orchestrator was failing with the error "Gas price multiplier is too high" for StateTransition jobs. This error message was misleading and the retry logic had incorrect parameters for blob transactions.

## Root Cause

EIP-4844 blob transactions require a 100% price bump (2x multiplier) to replace a stuck transaction in the blobpool. This is different from regular transactions which only need a 10% bump.

**References:**
- [Geth blobpool config](https://github.com/ethereum/go-ethereum/blob/d0af257aa20fe9d3e244570ee4abb9a78ff3b9c4/core/txpool/blobpool/config.go#L34)
- [Reth transaction pool config](https://github.com/paradigmxyz/reth/blob/c2435ff6f8265088b9ded0014051c9a97d0d7b84/crates/transaction-pool/src/config.rs#L29)
- [Nethermind blob tx comparison](https://github.com/NethermindEth/nethermind/blob/471bcb95bac677d2ffde5bb2e882e20186841b24/src/Nethermind/Nethermind.TxPool/Comparison/CompareReplacedBlobTx.cs#L40)

## Solution

- Updated gas price multiplier constants: start at 1.1x, increment by 2.0x
- Simplified retry logic to allow max 2 attempts (1.1x -> 2.2x -> fail)
- Replaced generic `bail!` with typed `SendTransactionError::RetryLimitExceeded` for clearer error messages
- Replaced `bail!` calls with proper error propagation using `.into()` to preserve error chain
- Added structured logging:
  - **Info level**: Transaction events (sending replacement, submission, finalization)
  - **Debug level**: Multiplier details (current, next, max values)

## Testing

Unit tests added for the gas multiplier calculation logic. Full integration tests would require complex Anvil setup with blob transaction simulation.

---
Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/orchestrator/crates/settlement-clients/ethereum/src/lib.rs
+++ b/orchestrator/crates/settlement-clients/ethereum/src/lib.rs
@@ -39,8 +39,9 @@ use crate::types::{bytes_be_to_u128, convert_stark_bigint_to_u256, DefaultHttpPr
 use lazy_static::lazy_static;
 use mockall::automock;
 use tokio::time::sleep;
-#[allow(unused_imports)] // warn! is used in #[cfg(not(feature = "testing"))] block
-use tracing::{debug, error, info, warn};
+#[cfg(not(feature = "testing"))]
+use tracing::warn;
+use tracing::{debug, error, info};
 
 // For more details on state update, refer to the core contract logic
 // https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/solidity/Output.sol
@@ -60,7 +61,8 @@ const REQUIRED_BLOCK_CONFIRMATIONS: u64 = 3;
 // See: https://github.com/ethereum/go-ethereum/blob/d0af257aa20fe9d3e244570ee4abb9a78ff3b9c4/core/txpool/blobpool/config.go#L34
 // See: https://github.com/paradigmxyz/reth/blob/c2435ff6f8265088b9ded0014051c9a97d0d7b84/crates/transaction-pool/src/config.rs#L29
 // See: https://github.com/NethermindEth/nethermind/blob/471bcb95bac677d2ffde5bb2e882e20186841b24/src/Nethermind/Nethermind.TxPool/Comparison/CompareReplacedBlobTx.cs#L40
-// With 1.1x start and 2.0x increment: 1.1 → 2.2 → fail. Max 2 attempts only.
+// With 1.1x start, 2.0x increment, and default max of 2.5x: 1.1 → 2.2 → 4.4 (exceeds max, fails).
+// The max multiplier is configurable via MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR env variable.
 const GAS_PRICE_MULTIPLIER_START: f64 = 1.1; // 10% above estimated gas price
 const GAS_PRICE_INCREMENT_FACTOR: f64 = 2.0; // 2x multiplier (100% bump required for blob tx replacement)
 /// we noticed Starknet uses the same limit on the mainnet
@@ -295,34 +297,32 @@ impl SettlementClient for EthereumSettlementClient {
             let tx_envelope = self.create_transaction(program_output.clone(), state_diff.clone(), mul_factor).await?;
             let pending_transaction = match self.send_transaction(tx_envelope).await {
                 Result::Ok(pending_transaction) => pending_transaction,
-                Err(e) => match e {
-                    SendTransactionError::ReplacementTransactionUnderpriced(rpc_err) => {
-                        match self.get_next_mul_factor(mul_factor) {
-                            Some(next_mul_factor) => {
-                                info!(attempt = attempt, "Transaction underpriced, sending replacement transaction");
-                                debug!(
-                                    current_multiplier = %mul_factor,
-                                    next_multiplier = %next_mul_factor,
-                                    max_multiplier = %self.max_gas_price_mul_factor,
-                                    error = ?rpc_err,
-                                    "Increasing gas multiplier for replacement transaction"
-                                );
-                                mul_factor = next_mul_factor;
-                                attempt += 1;
-                                continue;
-                            }
-                            None => {
-                                let next_mul = GAS_PRICE_INCREMENT_FACTOR * mul_factor;
-                                return Err(eyre!(
-                                    "Transaction retry limit reached: next multiplier ({:.2}x) exceeds maximum ({:.2}x)",
-                                    next_mul,
-                                    self.max_gas_price_mul_factor
-                                ));
-                            }
+                Result::Err(SendTransactionError::ReplacementTransactionUnderpriced(rpc_err)) => {
+                    match self.get_next_mul_factor(mul_factor) {
+                        Some(next_mul_factor) => {
+                            info!(attempt = attempt, "Transaction underpriced, sending replacement transaction");
+                            debug!(
+                                current_multiplier = %mul_factor,
+                                next_multiplier = %next_mul_factor,
+                                max_multiplier = %self.max_gas_price_mul_factor,
+                                error = ?rpc_err,
+                                "Increasing gas multiplier for replacement transaction"
+                            );
+                            mul_factor = next_mul_factor;
+                            attempt += 1;
+                            continue;
+                        }
+                        None => {
+                            let next_mul = GAS_PRICE_INCREMENT_FACTOR * mul_factor;
+                            return Err(eyre!(
+                                "Transaction retry limit reached: next multiplier ({:.2}x) exceeds maximum ({:.2}x)",
+                                next_mul,
+                                self.max_gas_price_mul_factor
+                            ));
                         }
                     }
-                    SendTransactionError::Other(_) => return Err(e.into()),
-                },
+                }
+                Result::Err(e) => return Err(e.into()),
             };
 
             info!(
--- a/orchestrator/crates/settlement-clients/ethereum/src/lib.rs
+++ b/orchestrator/crates/settlement-clients/ethereum/src/lib.rs
@@ -298,7 +298,7 @@ impl SettlementClient for EthereumSettlementClient {
             let pending_transaction = match self.send_transaction(tx_envelope).await {
                 Result::Ok(pending_transaction) => pending_transaction,
                 Result::Err(SendTransactionError::ReplacementTransactionUnderpriced(rpc_err)) => {
-                    match self.get_next_mul_factor(mul_factor) {
+                    match calculate_next_gas_mul_factor(mul_factor, self.max_gas_price_mul_factor) {
                         Some(next_mul_factor) => {
                             info!(attempt = attempt, "Transaction underpriced, sending replacement transaction");
                             debug!(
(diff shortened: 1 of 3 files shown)


orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -266,27 +282,40 @@ impl SettlementClient for EthereumSettlementClient {
         );
 
         let mut mul_factor = GAS_PRICE_MULTIPLIER_START;
+        let mut attempt = 1;
```
byteZorvin, line 287: Where is the attempt being checked for not being more than 1?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -518,15 +549,8 @@ impl EthereumSettlementClient {
...
-        let min_mul_factor = GAS_PRICE_MIN_INCREMENT_PERCENTAGE * mul_factor;
-        let max_mul_factor = GAS_PRICE_INCREMENT_PERCENTAGE * mul_factor;
-
-        if min_mul_factor > self.max_gas_price_mul_factor {
-            bail!("Gas price multiplier is too high")
-        } else {
-            Ok(self.max_gas_price_mul_factor.min(max_mul_factor))
-        }
+    fn get_next_mul_factor(&self, mul_factor: f64) -> std::result::Result<f64, SendTransactionError> {
+        calculate_next_gas_mul_factor(mul_factor, self.max_gas_price_mul_factor)
```
byteZorvin: I am confused a bit, why creat a separate function, like this fn has just one line of code.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -266,27 +282,40 @@ impl SettlementClient for EthereumSettlementClient {
...
+                                    "Increasing gas multiplier for replacement transaction"
+                                );
+                                mul_factor = next_mul_factor;
+                                attempt += 1;
+                                continue;
+                            }
+                            Err(retry_err) => return Err(retry_err.into()),
+                        }
                     }
+                    SendTransactionError::RetryLimitExceeded { .. } => return Err(e.into()),
```
byteZorvin: Where is this error being returned from?

byteZorvin (commented):
The logic overral looks good, few questions on the code structure, which could be improved a bit

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -39,7 +39,8 @@ use crate::types::{bytes_be_to_u128, convert_stark_bigint_to_u256, DefaultHttpPr
 use lazy_static::lazy_static;
 use mockall::automock;
 use tokio::time::sleep;
-use tracing::{error, info, warn};
+#[allow(unused_imports)] // warn! is used in #[cfg(not(feature = "testing"))] block
```
prkpndy: let's import this with feature testing? no need to allow unused imports imo

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -55,13 +56,28 @@ const MAX_TX_FINALISATION_ATTEMPTS: usize = 30;
...
 
 // Ethereum Gas Price Estimation
-const GAS_PRICE_MULTIPLIER_START: f64 = 1.2;
-const GAS_PRICE_INCREMENT_PERCENTAGE: f64 = 1.5; // 50%
-const GAS_PRICE_MIN_INCREMENT_PERCENTAGE: f64 = 1.1; // 10%
+// For EIP-4844 blob transactions, blobpool requires a 100% price bump (2x) to replace a stuck transaction.
+// See: https://github.com/ethereum/go-ethereum/blob/d0af257aa20fe9d3e244570ee4abb9a78ff3b9c4/core/txpool/blobpool/config.go#L34
+// See: https://github.com/paradigmxyz/reth/blob/c2435ff6f8265088b9ded0014051c9a97d0d7b84/crates/transaction-pool/src/config.rs#L29
+// See: https://github.com/NethermindEth/nethermind/blob/471bcb95bac677d2ffde5bb2e882e20186841b24/src/Nethermind/Nethermind.TxPool/Comparison/CompareReplacedBlobTx.cs#L40
+// With 1.1x start and 2.0x increment: 1.1 → 2.2 → fail. Max 2 attempts only.
```
prkpndy: it won't necessarily fail after 2 attempts no? we can configure the max gas multiplier from env?

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
prkpndy: ++

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -266,27 +282,40 @@ impl SettlementClient for EthereumSettlementClient {
...
-                            error = ?e,
-                            "Transaction rejected due to low gas price, retrying with higher multiplier"
-                        );
-                        mul_factor = next_mul_factor;
-                        continue;
-                    }
-                    SendTransactionError::Other(_) => {
-                        bail!("Failed to send blob transaction: {:?}", e);
+                    SendTransactionError::ReplacementTransactionUnderpriced(rpc_err) => {
+                        match self.get_next_mul_factor(mul_factor) {
```
prkpndy: can't we use question mark here (maybe by implementing From for the error types). will be cleaner imo

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -614,3 +638,49 @@ mod test_config {
         txn_request
     }
 }
+
+#[cfg(test)]
+mod gas_multiplier_tests {
```
prkpndy, line 645: nice tests! can we also create a case for testing the gas increase feature? (like send a transaction which would fail because of gas price, then send again with increased gas. just putting this here in case you have some idea. maybe anvil has some flag for specifying min gas price?)

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -55,13 +56,28 @@ const MAX_TX_FINALISATION_ATTEMPTS: usize = 30;
...
+// With 1.1x start and 2.0x increment: 1.1 → 2.2 → fail. Max 2 attempts only.
+const GAS_PRICE_MULTIPLIER_START: f64 = 1.1; // 10% above estimated gas price
+const GAS_PRICE_INCREMENT_FACTOR: f64 = 2.0; // 2x multiplier (100% bump required for blob tx replacement)
 /// we noticed Starknet uses the same limit on the mainnet
 /// https://etherscan.io/tx/0x8a58b936faaefb63ee1371991337ae3b99d74cb3504d73868615bf21fa2f25a1
 const GAS_LIMIT_STATE_UPDATE: u64 = 5_500_000;
 
+/// Calculates the next gas price multiplier for transaction retry.
+/// Returns an error if the next multiplier would exceed the maximum allowed.
+fn calculate_next_gas_mul_factor(current_mul: f64, max_mul: f64) -> std::result::Result<f64, SendTransactionError> {
```
prkpndy: can move this logic to `get_next_mul_factor` imo

prkpndy (commented):
added comments

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: The retry limit is enforced by the multiplication factor, not attempt count. In `calculate_next_gas_mul_factor` (line 74), we return `None` when `next_mul > max_mul`. With current values (start=1.1x, increment=2.0x, max=2.5x): 1.1 → 2.2 → 4.4 (exceeds 2.5, fails). The `attempt` variable is only for logging.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: The separate pure function enables easy unit testing. Without it, testing this logic would require instantiating a full `EthereumSettlementClient` (provider, wallet, contract setup, Anvil). With the pure function, we can test directly: `calculate_next_gas_mul_factor(1.1, 2.5)` - see the unit tests at line 648-685.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Good catch - this was dead code. `send_transaction()` only returns `ReplacementTransactionUnderpriced` or `Other`, never `RetryLimitExceeded`. Refactored to use `Option<f64>` instead - now `SendTransactionError` only contains actual RPC errors, and the retry limit message is created at the call site with `eyre!()`.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Fixed. Changed to conditionally import `warn` with `#[cfg(not(feature = "testing"))]` - cleaner than `allow(unused_imports)`.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Good catch. Updated the documentation to clarify: "With 1.1x start, 2.0x increment, and default max of 2.5x: 1.1 → 2.2 → 4.4 (exceeds max, fails)." Also added a note that the max multiplier is configurable via `MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR` env variable.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: Fixed. Flattened the nested match by directly pattern matching on `SendTransactionError::ReplacementTransactionUnderpriced` in the outer match. Much cleaner now.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: That would be a great integration test, but it's quite complex - would need Anvil with custom min gas price config, transaction orchestration, and error handling. For this PR, the unit tests verify the core math logic (multiplier calculation). We can add integration tests in a follow-up if needed.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
heemankv: The pure function `calculate_next_gas_mul_factor` is intentionally kept separate from `get_next_mul_factor` to enable easy unit testing. If we move all the logic into `get_next_mul_factor`, we'd need to instantiate a full `EthereumSettlementClient` (with provider, wallet, contract setup, Anvil) just to test the multiplier calculation math. The current design allows us to test the core logic with simple function calls: `calculate_next_gas_mul_factor(1.1, 2.5)`.

orchestrator/crates/settlement-clients/ethereum/src/lib.rs
byteZorvin: You can directly call the calculate function though ??

byteZorvin (approved):

byteZorvin (approved):

prkpndy (approved):
LGTM

PR #960, opened 2026-01-25
feat(external-db): MongoDB external database for mempool transactions

Mohiiit:
## Summary

Persist mempool transactions to MongoDB via an outbox-first design:

- On tx acceptance, synchronously append a WAL entry to RocksDB (durable, local)
- A background writer drains the outbox to MongoDB asynchronously (batched)
- This enables queryability, historical archiving, and replay for recovery

## Architecture (Outbox)

```
RPC / Gateway
    |
    v
Mempool accepts tx
    |
    | 1) append WAL entry (RocksDB outbox column)
    v
RocksDB outbox (durable)
    |
    | 2) drain (batch + retry/backoff)
    v
MongoDB (at-least-once inserts)
```

Notes:
- Mongo writes are at-least-once; duplicates can exist (tx_hash is not unique).
- Optional strict mode can reject txs if the outbox write fails.

## Config / CLI

- Node CLI maps into `mc-external-db` config (`ExternalDbParams::to_config()`)
- Service is not started in `warp_update_receiver` mode (no new tx intake)

## Testing

- `cargo test -p mc-db` (unit + outbox storage tests)
- `cargo test -p mc-external-db`:
  - unit tests pass
  - `e2e_devnet_replay_from_mongo_matches_root` requires Docker (testcontainers + Mongo)

## Non-feature Changes (Drive-by)

A few small changes landed while addressing review feedback / stability:

- Make `mc-db` bloom filter property test deterministic to avoid CI flakiness.
- Add concise docs clarifying storage proof JSON normalization.
- Add module-level docs clarifying Cairo-native cache timeouts and expected behavior.
- Add short comments explaining workspace/testcontainers dependency intent.
- Small node wiring cleanup to keep `main.rs` thin.

## Follow-ups

- If we want CI to always run the E2E, ensure the runner has Docker available.


madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -57,6 +58,11 @@ pub use options::{DbWriteMode, RocksDBConfig, StatsLevel};
 
 const DB_UPDATES_BATCH_SIZE: usize = 1024;
 
+#[cfg(any(test, feature = "testing"))]
+pub(crate) fn set_external_outbox_write_failpoint(enabled: bool) {
```
Mohiiit: I see the same function in the external_outbox.rs as well, maybe we can use the same one? follow DRY?

madara/crates/client/db/src/tests/test_external_outbox.rs
```diff
@@ -0,0 +1,116 @@
...
+use mp_chain_config::ChainConfig;
+use mp_convert::Felt;
+use mp_transactions::validated::{TxTimestamp, ValidatedTransaction};
+use starknet_api::{
+    core::ContractAddress,
+    executable_transaction::AccountTransaction,
+    transaction::{InvokeTransaction, InvokeTransactionV3, TransactionHash},
+};
+
+fn make_validated_tx(seed: u64) -> ValidatedTransaction {
```
Mohiiit: we should try with more validated_tx type as well, checkout the mc-devnet tests for example of declare-deploy as well and for invoke v3 too!

madara/crates/client/external_db/src/mongodb/client.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl MongoClient {
+    /// Creates a new MongoDB client.
+    pub async fn new(config: &ExternalDbConfig) -> Result<Self> {
+        let mut options = ClientOptions::parse(&config.mongodb_uri).await?;
+        options.max_pool_size = Some(config.pool_size);
+        options.min_pool_size = Some(1);
+        options.connect_timeout = Some(Duration::from_secs(10));
+        options.server_selection_timeout = Some(Duration::from_secs(10));
```
Mohiiit: this should be config as well!

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,703 @@
...
+        return None;
+    }
+    let write_errors = failure.write_errors.as_ref()?;
+    if write_errors.is_empty() {
+        return None;
+    }
+
+    let mut duplicates = 0usize;
+    for write_error in write_errors {
+        if write_error.code == 11000 {
```
Mohiiit: write doc and reference about the error code and wriet this error code in const as well (ideally we should have the mongodb doc reference here)

madara/crates/client/external_db/tests/e2e.rs
```diff
@@ -0,0 +1,162 @@
...
+        tokio::time::sleep(Duration::from_millis(200)).await;
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn e2e_outbox_to_mongo_and_retention() {
+    let image = GenericImage::new("mongo", "7.0").with_exposed_port(27017.tcp());
+    let node = image.start().await.unwrap();
+    let port = node.get_host_port_ipv4(27017).await.unwrap();
+    let uri = format!("mongodb://127.0.0.1:{port}");
```
Mohiiit: this works fine for one test, but don't you think it will break for multiple tests? trying to have a mongo running on same port? shouldn't we have it as fixture or something? and something like get_or_init? checkout other tests you might find the pattern there as well

Mohiiit (commented):
overall we should write concise docs about function's working so that it's easy for the reviewer and reader to verify the working of the function and tests are lacking the different txn types, so let's add different txn type where-ever we are testing to ensure that we are good on all fronts, including l1 txn, the codebase have all the examples you might need!

madara/crates/client/db/src/tests/test_external_outbox.rs
```diff
@@ -0,0 +1,130 @@
...
+    let outbox: Vec<_> = backend.get_external_outbox_transactions(2).collect::<Result<Vec<_>, _>>().unwrap();
+
+    assert_eq!(outbox.len(), 2);
+
+    let all_outbox: Vec<_> = backend.get_external_outbox_transactions(10).collect::<Result<Vec<_>, _>>().unwrap();
+    assert_eq!(all_outbox.len(), 3);
+}
+
+#[tokio::test]
+async fn outbox_duplicate_write_appends_entry() {
```
Mohiiit, line 105: in this case the uuid would differ? or something like that? let's assert that too

madara/crates/client/db/src/lib.rs
```diff
@@ -573,6 +574,11 @@ impl MadaraBackend<RocksDBStorage> {
     }
 }
 
+#[cfg(any(test, feature = "testing"))]
+pub fn set_external_outbox_write_failpoint(enabled: bool) {
```
Mohiiit: we already have this, this is a duplicate I think, validate

madara/crates/client/external_db/src/mongodb/client.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub async fn new(config: &ExternalDbConfig) -> Result<Self> {
+        let mut options = ClientOptions::parse(&config.mongodb_uri).await?;
+        options.max_pool_size = Some(config.pool_size);
+        options.min_pool_size = Some(1);
+        options.connect_timeout = Some(Duration::from_secs(config.connect_timeout_secs));
+        options.server_selection_timeout = Some(Duration::from_secs(config.server_selection_timeout_secs));
+
+        let client = mongodb::Client::with_options(options)?;
+        // Ensure connection is usable at startup (not just for tests).
+        client.database("admin").run_command(doc! { "ping": 1 }).await?;
```
Mohiiit, line 25: is this for readniess check and won't create any database right?

madara/crates/client/external_db/src/mongodb/client.rs
```diff
@@ -0,0 +1,40 @@
...
+    client: mongodb::Client,
+    collection: mongodb::Collection<MempoolTransactionDocument>,
+}
+
+impl MongoClient {
+    /// Creates a new MongoDB client.
+    pub async fn new(config: &ExternalDbConfig) -> Result<Self> {
+        let mut options = ClientOptions::parse(&config.mongodb_uri).await?;
+        options.max_pool_size = Some(config.pool_size);
+        options.min_pool_size = Some(1);
```
Mohiiit: should this not be a config as well?

madara/crates/client/external_db/src/config.rs
Mohiiit, line 1: instead of creating functions for default values, can't we use serde macros for that?

madara/crates/client/db/src/test_utils.rs
Mohiiit, line 1: this is not following DRY principles, maybe we can create test_utils within crates/client and can be used by both db and external_db?

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+    validated::ValidatedTransaction, DeclareTransaction, DeployAccountTransaction, DeployTransaction,
+    InvokeTransaction, Transaction,
+};
+use mp_utils::service::ServiceContext;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use uuid::Uuid;
+
+const STATUS_PENDING: &str = "PENDING";
+// MongoDB duplicate key error code (E11000).
```
Mohiiit, line 20: let's add a doc reference for this from official documentation

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+    metrics: Arc<ExternalDbMetrics>,
+    chain_id: String,
+}
+
+impl<S: L1ConfirmationSource> RetentionScheduler<S> {
+    fn new(source: S, delay: Duration, metrics: Arc<ExternalDbMetrics>, chain_id: String) -> Self {
+        Self { source, delay, pending: std::collections::VecDeque::new(), metrics, chain_id }
+    }
+
+    async fn tick(&mut self, now: Instant, sink: &dyn ExternalDbSink) -> anyhow::Result<()> {
```
Mohiiit: I see that you are collecting all the potential txns and then filtering it right here, maybe we can move this part to poll_confirmations, where we will filter for only those blocks for which we have to delete the txns from mongo

meaning in the poll_confirmations, you will get the info on the latest_confirmed_block_on_l1 and you have the block_time as well (assuming blocks are getting created within this time, for this example take this 1 min) and we have the config about how much buffer should take (for example take a day) so now I know that I have to delte the block for latest_block - 1440 (24 * 60)

1440 because that's the total number of blocks we must have within the day

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+        let mut retention_interval = tokio::time::interval(Duration::from_secs(self.config.retention_tick_secs));
+        retention_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+
+        let mut retry_backoff = RetryBackoff::new(
+            Duration::from_millis(self.config.retry_backoff_ms),
+            Duration::from_millis(self.config.retry_backoff_max_ms),
+        );
+        let mut retry_at: Option<Instant> = None;
+
+        loop {
```
Mohiiit, line 330: so if our db is down, it will keep retrying right? and will this kill our application in case the db is not recovered within the time?

maybe we can write a test around it if not available yet

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+            let summary = self.drain_outbox_once(sink).await?;
+            total += summary.attempted;
+            if summary.attempted == 0 {
+                break;
+            }
+        }
+        Ok(total)
+    }
+
+    fn convert_to_document(&self, tx: &ValidatedTransaction) -> anyhow::Result<MempoolTransactionDocument> {
```
Mohiiit, line 445: I think this function can be divided in smaller chunks, then it's easier to test as well!

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+        test_utils::validated_transaction(
+            Transaction::Invoke(InvokeTransaction::V3(test_utils::invoke_v3(sender, Felt::from_hex_unchecked("0x1")))),
+            Felt::from_hex_unchecked(&format!("0x{seed:x}")),
+            sender,
+            None,
+        )
+    }
+
+    #[tokio::test]
+    async fn drain_outbox_inserts_to_mongo() {
```
Mohiiit, line 704: use the test_utils to test txn of each type instead of just invoke

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+        let summary = worker.drain_outbox_once(&sink).await.unwrap();
+
+        assert_eq!(summary.attempted, 2);
+        let remaining: Vec<_> = backend.get_external_outbox_transactions(10).collect::<Result<Vec<_>, _>>().unwrap();
+        assert!(remaining.is_empty());
+        assert_eq!(sink.calls.lock().unwrap().len(), 1);
+    }
+
+    #[tokio::test]
+    async fn drain_handles_duplicate_key() {
```
Mohiiit, line 760: use the test_utils to test txn of each type instead of just invoke

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,733 @@
...
+        let summary = worker.drain_outbox_once(&sink).await.unwrap();
+
+        assert_eq!(summary.attempted, 2);
+        assert_eq!(summary.duplicates, 1);
+        let remaining: Vec<_> = backend.get_external_outbox_transactions(10).collect::<Result<Vec<_>, _>>().unwrap();
+        assert!(remaining.is_empty());
+    }
+
+    #[tokio::test]
+    async fn startup_sync_drains_existing_outbox() {
```
Mohiiit, line 804: use the test_utils to test txn of each type instead of just invoke

madara/crates/client/external_db/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+homepage.workspace = true
+
+[lints]
+workspace = true
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
```
Mohiiit, line 15: shouldn't this be taken from workspace as well?

madara/crates/client/mempool/src/lib.rs
```diff
@@ -229,6 +255,19 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
     async fn add_tx(&self, tx: ValidatedTransaction, is_new_tx: bool) -> Result<(), MempoolInsertionError> {
         tracing::debug!("Accepting transaction tx_hash={:#x} is_new_tx={is_new_tx}", tx.hash);
 
+        let mut outbox_id = None;
+        if is_new_tx && self.external_outbox.enabled {
+            match self.backend.write_external_outbox(&tx) {
+                Ok(id) => outbox_id = Some(id),
+                Err(err) => {
+                    tracing::error!("Could not write external outbox transaction: {err:#}");
+                    if self.external_outbox.strict {
```
Mohiiit, line 272: so ideally we should follow a strict by default I think

madara/node/src/cli/external_db.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[arg(env = "MADARA_EXTERNAL_DB_RETRY_BACKOFF_MS", long, default_value = "1000")]
+    pub external_db_retry_backoff_ms: u64,
+
+    /// Maximum retry backoff in milliseconds.
+    /// Caps the exponential backoff to prevent excessively long waits.
+    #[arg(env = "MADARA_EXTERNAL_DB_RETRY_BACKOFF_MAX_MS", long, default_value = "60000")]
+    pub external_db_retry_backoff_max_ms: u64,
+}
+
+impl Default for ExternalDbParams {
```
Mohiiit, line 61: is this required when default_value in present in the serde above?

madara/node/src/cli/mod.rs
```diff
@@ -193,6 +195,11 @@ pub struct RunCmd {
     #[clap(flatten)]
     pub cairo_native_params: CairoNativeParams,
 
+    #[allow(missing_docs)]
```
Mohiiit, line 198: what will make this to remove? we can document it really well yeah!

madara/node/src/main.rs
```diff
@@ -473,6 +503,10 @@ async fn main() -> anyhow::Result<()> {
         app.activate(MadaraServiceId::Telemetry);
     }
 
+    if run_cmd.external_db_params.is_enabled() && !warp_update_receiver {
```
Mohiiit: why !warp_update_recieve?

madara/crates/client/external_db/tests/e2e.rs
```diff
@@ -0,0 +1,864 @@
...
+    if let Err(err) = &direct_results[5] {
+        assert!(err.contains("DeployNotSupported"), "expected deploy conversion to be unsupported, got: {err}");
+    }
+
+    ctx.cancel_global();
+    let _ = handle.await;
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn e2e_devnet_replay_from_mongo_matches_root() {
```
Mohiiit, line 188: this test is more than enough, we can remove the rest two

in this test itself, add more txn types tho, like declare and deploy account and it's good to go!

Mohiiit (commented):
new review for codex!

madara/crates/client/db/Cargo.toml
```diff
@@ -50,12 +50,14 @@ futures = { workspace = true }
...
 libc.workspace = true
 librocksdb-sys = { workspace = true }
+m-cairo-test-contracts = { workspace = true, optional = true }
 num-traits = { workspace = true }
 rayon = { workspace = true }
 rocksdb.workspace = true
 serde = { workspace = true }
 siphasher.workspace = true
 smallvec.workspace = true
+starknet-core = { workspace = true, optional = true }
```
heemankv, line 63: seems like they are only being used in testing, if that's true can we only add in dev-deps ?
- starknet-core 
- m-cairo-test-contracts

madara/crates/client/external_db/src/mongodb/indexes.rs
```diff
@@ -0,0 +1,31 @@
...
+//! These indexes are created on service startup to optimize query patterns.
+//!
+//! Indexes:
+//! - `{ "_id": 1 }` - Primary key (automatic)
+//! - `{ "sender_address": 1, "arrived_at": -1 }` - For sender queries with time ordering
+//! - `{ "arrived_at": -1 }` - For time-range queries
+//! - `{ "status": 1, "arrived_at": -1 }` - For status queries (future use)
+//! - `{ "tx_type": 1, "arrived_at": -1 }` - For transaction type queries
+//! - `{ "sender_address": 1, "nonce": 1 }` - For sender + nonce queries (replay)
+//! - `{ "tx_hash": 1 }` - For tx hash lookups (non-unique)
```
heemankv: -  (non-unique)
What does non-unique here mean ?
This is unclear to me, can you help with a scenario where the txn hash are not unique ?

madara/crates/client/external_db/src/mongodb/models.rs
heemankv, line 1: What do you think about using "starknet_api::transaction::InvokeTransactionV3" and structs as it is, instead and superset-ing all transaction types with monogdb's id field ?
On first look that seems better for mantainability.

madara/crates/client/external_db/src/config.rs
heemankv, line 1: Any reason to have these here and not in CLI ?

madara/crates/client/external_db/src/config.rs
```diff
@@ -0,0 +1,215 @@
...
+            retry_backoff_max_ms: default_retry_backoff_max_ms(),
+        }
+    }
+}
+
+impl ExternalDbConfig {
+    /// Creates a new config with the given MongoDB URI.
+    pub fn new(mongodb_uri: String) -> Self {
+        Self { enabled: true, mongodb_uri, ..Default::default() }
+    }
```
heemankv, line 156: Nitpick: IMO better way to write this is to create two functions: 
```
    fn new(mongodb_uri: String, configs: ExternalDbConfig) -> Self {}
```
and 
```
    fn new_with_defaults(mongodb_uri: String) -> Self {}
```

madara/Cargo.toml
```diff
@@ -110,11 +112,15 @@ mc-mempool = { path = "crates/client/mempool" }
...
 mc-devnet = { path = "crates/client/devnet" }
 mc-submit-tx = { path = "crates/client/submit_tx" }
+mc-external-db = { path = "crates/client/external_db" }
 
 # Madara misc
 m-cairo-test-contracts = { path = "crates/cairo-test-contracts" }
 mc-e2e-tests = { path = "crates/tests" }
 
+# Testing
+testcontainers = "0.18.0"
```
heemankv, line 124: What does this do ?

madara/node/src/main.rs
heemankv: +1

madara/node/src/main.rs
```diff
@@ -364,6 +365,31 @@ async fn main() -> anyhow::Result<()> {
...
+            config.database_name = db_name.clone();
+        }
+        config.collection_name = run_cmd.external_db_params.external_db_collection.clone();
+        config.batch_size = run_cmd.external_db_params.external_db_batch_size;
+        config.flush_interval_ms = run_cmd.external_db_params.external_db_flush_interval_ms;
+        config.retention_delay_secs = run_cmd.external_db_params.external_db_retention_secs;
+        config.retention_tick_secs = run_cmd.external_db_params.external_db_retention_tick_secs;
+        config.strict_outbox = run_cmd.external_db_params.external_db_strict_outbox;
+        config.retry_backoff_ms = run_cmd.external_db_params.external_db_retry_backoff_ms;
+        config.retry_backoff_max_ms = run_cmd.external_db_params.external_db_retry_backoff_max_ms;
```
heemankv: Would prefer to simplify this using a single fn.
```
let external_db_configs = ExternalDbConfig::from_run_cmd(run_cmd.external_db_params)
```
this will keep clean the main.rs

madara/node/src/cli/external_db.rs
heemankv, line 1: I think we can clean the `cli/external_db.rs` & `external_db/config.rs` even further. 
Have few thoughts can discuss

madara/crates/tests/src/storage_proof.rs
heemankv, line 1: Would need context on this change

madara/crates/client/cairo_native/src/cache.rs
heemankv, line 1: Context on these changes would be great.

heemankv (commented):
Overall looks good, core logic seems robust, comments usually focus on a different way to write code-pieces.

PR description could also mention : 
- DB version bumped to v10.
- Changes done in Cairo Native crate.
- Changes done in Storage Proof `fn normalize`

madara/crates/client/db/Cargo.toml
Mohiiit: Good catch. These are used in two places:
1) `cargo test -p mc-db` (so they need to be in `dev-dependencies`), and
2) downstream crates enabling `mc-db/testing` (so they need to be optional deps behind that feature).
I added a comment explaining the split (commit aa9256223).

madara/crates/client/external_db/src/mongodb/indexes.rs
Mohiiit: "Non-unique" means we intentionally allow multiple Mongo documents with the same `tx_hash`.

Reason: persistence is at-least-once. If we successfully insert into Mongo but crash/restart before deleting the outbox entry, we will retry and insert again. Also, the revert+replay flow can re-insert previously persisted txs. `_id` (UUID) is the unique key; `tx_hash` is indexed for lookup only.

I clarified this in docs in `indexes.rs` and `models.rs` (commit aa9256223).

madara/crates/client/external_db/src/mongodb/models.rs
Mohiiit: I considered storing `starknet_api::transaction::*` structs directly, but opted for:
- explicit scalar fields for query patterns (sender/nonce/type/arrived_at/etc.)
- plus `raw_transaction` (bincode) for deterministic replay

Persisting starknet_api structs "as-is" would tightly couple the DB schema to upstream serde/Rust types (and still wouldn’t remove the need for query-friendly scalar fields). We also need a stable document `_id` anyway.

I added a short rationale in `models.rs` (commit aa9256223). Happy to revisit if you prefer a different tradeoff.

madara/crates/client/external_db/src/config.rs
Mohiiit: Main reason: `ExternalDbConfig` is a service-level config used by tests and by any non-CLI callers; keeping it in `mc-external-db` avoids coupling the crate to the node’s clap layer.

The node CLI now maps `ExternalDbParams -> ExternalDbConfig` via `ExternalDbParams::to_config()` (commit aa9256223), so `main.rs` stays thin.

madara/crates/client/external_db/src/config.rs
Mohiiit: Makes sense. Right now `ExternalDbConfig::new(mongodb_uri)` is effectively "new_with_defaults" (enabled + uri + `Default::default()`).
I kept it minimal since we only need one constructor today, but I’m fine changing to `new(uri, cfg)` / `new_with_defaults(uri)` if you think it improves clarity.

madara/Cargo.toml
Mohiiit: That `testcontainers` workspace dependency is only used by the `mc-external-db` e2e tests to run MongoDB via Docker. Only crates that opt-in with `workspace = true` pull it in.
I added a comment in `madara/Cargo.toml` to make this explicit (commit aa9256223).

madara/node/src/main.rs
Mohiiit: Agreed. I moved the CLI->config mapping into `ExternalDbParams::to_config()` and simplified `main.rs` to call it (commit aa9256223).

madara/node/src/cli/external_db.rs
Mohiiit: Yep, I’m on board. In this PR I did the first cleanup step by moving mapping into `ExternalDbParams::to_config()` and adding docs about the layering/intent (commit aa9256223).
If you have a preferred direction for reducing duplication (defaults, naming, env var layout), I’m happy to iterate further in a follow-up.

madara/crates/tests/src/storage_proof.rs
Mohiiit: This is purely to make snapshot assertions deterministic: some RPC JSON responses don’t guarantee ordering of arrays/object keys, so `normalize()` sorts recursively before `assert_eq!`.
I added a doc comment on `normalize()` explaining this (commit aa9256223).

madara/crates/client/cairo_native/src/cache.rs
Mohiiit: These changes just add missing context docs around the Cairo-native cache timeouts:
- we bound memory/disk cache operations to avoid blocking tx validation
- timeouts are treated as cache misses (fall back to VM / recompilation) and recorded in metrics

I added module-level docs clarifying the behavior and why it exists (commit aa9256223).

madara/node/src/main.rs
Mohiiit: Added an explicit comment in `main.rs`: `warp_update_receiver` mode doesn’t accept/validate new txs, so we don’t start `ExternalDbService` there (commit aa9256223).

.env.test
```diff
@@ -99,7 +99,7 @@ ETHEREUM_SETTLEMENT_TXN_WAIT_SLEEP_DELAY_SECS=60
 
 #### SERVICE ####
 
-MADARA_ORCHESTRATOR_MADARA_RPC_URL=https://pathfinder-madara-ci.d.karnot.xyz
+MADARA_ORCHESTRATOR_MADARA_RPC_URL=https://pathfinder-sepolia.d.karnot.xyz
```
byteZorvin, line 102: Just curious, why was this needed

madara/crates/client/db/Cargo.toml
```diff
@@ -50,12 +50,17 @@ futures = { workspace = true }
 itertools.workspace = true
 libc.workspace = true
 librocksdb-sys = { workspace = true }
+# Optional deps used by `feature = "testing"` so downstream crates can enable
+# `mc-db/testing` without relying on `dev-dependencies`.
+m-cairo-test-contracts = { workspace = true, optional = true }
```
byteZorvin, line 55: If have both starknet-core and m-cairo-test-contracts in dev dependency, do we really need in deps too?

madara/crates/client/external_db/src/writer.rs
```diff
@@ -0,0 +1,847 @@
...
+        }
+
+        let start = Instant::now();
+        let result = sink.insert_many(documents).await?;
+        let elapsed = start.elapsed();
+        self.metrics.write_latency.record(elapsed.as_secs_f64(), &[]);
+        self.metrics.transactions_written.add(result.inserted as u64, &[]);
+
+        for entry in &entries {
+            self.backend.delete_external_outbox(entry.id).context("Deleting external outbox transaction")?;
```
byteZorvin: If the rocksdb deletion fails, then there is possiblity that multiple entry into the mongo can happen right?
Though in any case it will still not be an issue

byteZorvin (approved):

.env.test
Mohiiit: we had duplicate pathfinder running in the cloud, removed one, hence the change

madara/crates/client/db/Cargo.toml
Mohiiit: We keep them in both places intentionally:
- dev-deps: so `cargo test -p mc-db` works without requiring `--features testing`
- optional deps behind `mc-db/testing`: so downstream crates can enable `mc-db/testing` without relying on our dev-deps
Added a comment explaining this in `mc-db/Cargo.toml` (commit aa9256223).

madara/crates/client/external_db/src/writer.rs
Mohiiit: Yes, that’s expected under our at-least-once model.
If Mongo insert succeeds but RocksDB delete fails (or we crash before deleting), the outbox entry remains and will be retried, which can create duplicate Mongo docs for the same tx_hash. That’s why tx_hash is indexed but not unique, and `_id` (UUID) is the unique document key.

heemankv (approved):

PR #963, opened 2026-02-01
fix(db): disable RocksDB snapshots by default

Mohiiit:
## Summary

- Change `max_kept_snapshots` default from `None` (unlimited) to `Some(0)` (disabled)
- Prevents unbounded memory growth from accumulating snapshots

## Problem

With the previous default of `None`:
- Snapshots were created every `snapshot_interval` blocks (default: 5)
- Snapshots were **never cleaned up** (cleanup only runs when `is_some_and()` returns true)
- For a node syncing from genesis with 600k+ blocks: **120,000+ snapshots** would accumulate in memory

## Solution

Setting default to `Some(0)` disables snapshot creation entirely. Users who need snapshots for storage proofs can explicitly enable them via `--db-max-kept-snapshots`.

## Behavior Summary

| `max_kept_snapshots` | Snapshot Creation | Cleanup | Result |
|---------------------|-------------------|---------|--------|
| `None` (old default) | ✅ Yes | ❌ Never | **Unlimited accumulation** |
| `Some(0)` (new default) | ❌ No | N/A | **Disabled** |
| `Some(n)` where n > 0 | ✅ Yes | ✅ When > n | **Limited to n** |

## CI

- `test-cli`: increase step timeout to 20 minutes and wait for JSON-RPC `starknet_chainId` to return before killing the process.

## Test plan

- [ ] Verify node starts without snapshots by default
- [ ] Verify `--db-max-kept-snapshots N` enables snapshots with limit

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/madara/crates/client/db/src/rocksdb/snapshots.rs
+++ b/madara/crates/client/db/src/rocksdb/snapshots.rs
@@ -198,5 +198,4 @@ mod tests {
             assert_eq!(block_n, Some(expected_block), "Snapshot at block {} should exist", expected_block);
         }
     }
-
 }
--- a/madara/crates/client/db/src/rocksdb/snapshots.rs
+++ b/madara/crates/client/db/src/rocksdb/snapshots.rs
@@ -199,35 +199,4 @@ mod tests {
         }
     }
 
-    /// Verify snapshot_interval behavior: snapshots are only created at interval boundaries.
-    #[test]
-    fn test_snapshot_interval_behavior() {
-        let (_temp_dir, db) = create_test_db();
-
-        // Create snapshots with max_kept_snapshots = None, interval = 10
-        let snapshots = Snapshots::new(db, None, None, 10);
-
-        // Add blocks 0-35
-        for block_n in 0..36 {
-            snapshots.set_new_head(block_n);
-        }
-
-        // Should have 4 snapshots (at 0, 10, 20, 30)
-        assert_eq!(
-            snapshots.inner.read().expect("Poisoned lock").historical.len(),
-            4,
-            "Snapshots should only be created at interval boundaries"
-        );
-
-        // Verify snapshots at exact intervals
-        let (block_n, _) = snapshots.get_closest(0);
-        assert_eq!(block_n, Some(0));
-
-        let (block_n, _) = snapshots.get_closest(10);
-        assert_eq!(block_n, Some(10));
-
-        // Block 15 should return snapshot at 20 (closest >= 15)
-        let (block_n, _) = snapshots.get_closest(15);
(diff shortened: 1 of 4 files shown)


prkpndy (approved):
LGTM

madara/crates/client/db/src/rocksdb/snapshots.rs
```diff
@@ -86,3 +86,116 @@ impl Snapshots {
...
+    }
+
+    /// When `max_kept_snapshots = Some(n)`, only n snapshots should be kept.
+    /// Oldest snapshots are removed first (FIFO behavior).
+    #[test]
+    fn test_snapshots_limited_to_max() {
+        let (_temp_dir, db) = create_test_db();
+
+        // Create snapshots with max_kept_snapshots = Some(3), interval = 5
+        let snapshots = Snapshots::new(db, None, Some(3), 5);
```
byteZorvin: Here we are hardcoding the values for `max_kept_snaphosts` and `snapshot_interval` and this does not depend on the Default RocksDbConfig, and this test would pass irrespective of any Default config.
We should not create the Snapshot manually here instead instantiate through the RocksDbStorage, otherwise the purpose of test is beaten.
What do you think?

madara/crates/client/db/src/rocksdb/snapshots.rs
```diff
@@ -86,3 +86,116 @@ impl Snapshots {
...
+        assert_eq!(block_n, Some(20), "Snapshot at block 20 should exist");
+    }
+
+    /// When `max_kept_snapshots = None`, snapshots accumulate without limit.
+    #[test]
+    fn test_snapshots_unlimited_when_none() {
+        let (_temp_dir, db) = create_test_db();
+
+        // Create snapshots with max_kept_snapshots = None (unlimited), interval = 5
+        let snapshots = Snapshots::new(db, None, None, 5);
```
byteZorvin: Same as above

byteZorvin (commented):
Looks good, though I think testcases can be improved

madara/crates/client/db/src/rocksdb/snapshots.rs
Mohiiit: Good point. Updated the tests to instantiate through RocksDBStorage::open so they exercise the config defaults (and fail if defaults change). See commit 64870ea1e.

madara/crates/client/db/src/rocksdb/snapshots.rs
Mohiiit: Same fix as above. Tests now go through RocksDBStorage::open (commit 64870ea1e).

byteZorvin (approved):

PR #966, opened 2026-02-03
update: orchestrator dashboard and and metrics

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Currently the dashboard has some issues. Also the code to register metrics is not properly modularise.

## What is the new behavior?

Added some new metrics and modularise the code to register metrics. Updated the dashboard to be more clean and remove unnecessary dashboards.

No

### DIFF
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -3,8 +3,10 @@ use crate::utils::job_status_metrics::JobStatusTracker;
 use once_cell;
 use once_cell::sync::Lazy;
 use opentelemetry::global;
-use opentelemetry::metrics::{Counter, Gauge};
-use orchestrator_utils::metrics::lib::{register_counter_metric_instrument, register_gauge_metric_instrument, Metrics};
+use opentelemetry::metrics::{Counter, Gauge, Histogram};
+use orchestrator_utils::metrics::lib::{
+    register_counter_metric_instrument, register_gauge_metric_instrument, register_histogram_metric_instrument, Metrics,
+};
 use orchestrator_utils::register_metric;
 
 register_metric!(ORCHESTRATOR_METRICS, OrchestratorMetrics);
@@ -42,8 +44,8 @@ pub struct OrchestratorMetrics {
     // SLA Metrics
     pub sla_breach_count: Counter<f64>,
     pub job_age_p99: Gauge<f64>,
-    pub batching_rate: Gauge<f64>,
-    pub batch_creation_time: Gauge<f64>,
+    pub batch_creation_rate: Counter<f64>,
+    pub batch_creation_time: Histogram<f64>,
     // Job Status Tracking
     pub job_status_tracker: JobStatusTracker,
     // Atlantic Service Metrics
@@ -249,17 +251,17 @@ impl Metrics for OrchestratorMetrics {
         );
 
         // Batching Metrics
-        let batching_rate = register_gauge_metric_instrument(
+        let batch_creation_rate = register_counter_metric_instrument(
             &orchestrator_meter,
-            "batching_rate".to_string(),
-            "Number of batches created per hour".to_string(),
-            "batches/hour".to_string(),
+            "batch_creation_rate".to_string(),
+            "Total number of batches created".to_string(),
+            "1".to_string(),
         );
 
-        let batch_creation_time = register_gauge_metric_instrument(
+        let batch_creation_time = register_histogram_metric_instrument(
             &orchestrator_meter,
             "batch_creation_time".to_string(),
-            "Average time to create a batch".to_string(),
+            "Time to create a batch".to_string(),
             "s".to_string(),
         );
 
@@ -349,7 +351,7 @@ impl Metrics for OrchestratorMetrics {
             active_jobs_count,
             sla_breach_count,
             job_age_p99,
-            batching_rate,
+            batch_creation_rate,
             batch_creation_time,
             job_status_tracker,
             atlantic_api_call_duration,
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -31,6 +31,7 @@ pub struct OrchestratorMetrics {
     // Latency Metrics
     pub job_e2e_latency: Gauge<f64>,
     pub proof_generation_time: Gauge<f64>,
+    pub snos_job_processing_time: Histogram<f64>,
     pub settlement_time: Gauge<f64>,
     // Throughput Metrics
     pub jobs_per_minute: Gauge<f64>,
@@ -180,6 +181,13 @@ impl Metrics for OrchestratorMetrics {
             &orchestrator_meter,
             "proof_generation_time".to_string(),
             "Time to generate proofs".to_string(),
+            "s".to_string(),
+        );
+
+        let snos_job_processing_time = register_histogram_metric_instrument(
+            &orchestrator_meter,
+            "snos_job_processing_time".to_string(),
+            "Time to process SNOS jobs".to_string(),
             "s".to_string(),
         );
 
@@ -342,6 +350,7 @@ impl Metrics for OrchestratorMetrics {
             job_abandoned_count,
             job_e2e_latency,
             proof_generation_time,
+            snos_job_processing_time,
             settlement_time,
             jobs_per_minute,
             blocks_per_hour,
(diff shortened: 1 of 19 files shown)


orchestrator/src/utils/metrics.rs
```diff
@@ -249,17 +259,17 @@ impl Metrics for OrchestratorMetrics {
...
         // Batching Metrics
-        let batching_rate = register_gauge_metric_instrument(
+        let batch_creation_rate = register_counter_metric_instrument(
             &orchestrator_meter,
-            "batching_rate".to_string(),
-            "Number of batches created per hour".to_string(),
-            "batches/hour".to_string(),
+            "batch_creation_rate".to_string(),
+            "Total number of batches created".to_string(),
+            "1".to_string(),
```
Mohiiit: the string and the unit is not matching with the variable

the variable is batch_creation_rate, but the string says "Total number of batches created" and then "1"? (not sure about "1" here)

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -846,23 +806,61 @@ impl JobHandlerService {
...
+                        );
+                        internal_id as f64
+                    }
+                }
+            }
+            JobType::SnosRun | JobType::ProofCreation => {
+                match config.database().get_snos_batches_by_indices(vec![internal_id]).await {
+                    Ok(batches) if !batches.is_empty() => batches[0].end_block as f64,
+                    Ok(_) => {
+                        error!(
```
Mohiiit: should be a warning?

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -846,23 +806,61 @@ impl JobHandlerService {
...
-                    .map_err(|e| JobError::Other(OtherError::from(format!("Could not parse string: {e}"))))?,
-            )?
-        } else {
-            internal_id as f64
+        let block_number = match job_type {
+            JobType::Aggregator | JobType::StateTransition => {
+                match config.database().get_aggregator_batches_by_indexes(vec![internal_id]).await {
+                    Ok(batches) if !batches.is_empty() => batches[0].end_block as f64,
+                    Ok(_) => {
+                        error!(
```
Mohiiit: same here!

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -846,23 +806,61 @@ impl JobHandlerService {
...
+                        internal_id as f64
+                    }
+                    Err(e) => {
+                        error!(
+                            job_type = ?job_type,
+                            internal_id = %internal_id,
+                            error = ?e,
+                            "Failed to fetch aggregator batch for block gauge, falling back to internal_id"
+                        );
+                        internal_id as f64
```
Mohiiit: in case of L3, this will keep happening tho?

Mohiiit (commented):
nothing major, just a few questions!

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -15,7 +15,6 @@ impl MetricsRecorder {
         let attributes = [
             KeyValue::new("operation_job_type", format!("{:?}", job.job_type)),
             KeyValue::new("operation_type", "create_job"),
-            KeyValue::new("service_name", "${namespace}"),
```
byteZorvin, line 18: Why are we removing this?

orchestrator/src/utils/metrics.rs
byteZorvin: +1

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -336,6 +337,7 @@ impl AggregatorHandler {
         // Record batch creation time with starknet_version in metrics
         let duration = start_time.elapsed();
         let attributes = [
+            KeyValue::new("operation_job_type", format!("{:?}", JobType::Aggregator)),
             KeyValue::new("batch_index", index.to_string()),
             KeyValue::new("start_block", start_block.to_string()),
             KeyValue::new("bucket_id", bucket_id.to_string()),
```
byteZorvin: Although the attributes(`batch_index`, `start_block`, `bucket_id`) are not added now, but should be removed. They are high cardinality and should not be in prometheus metrics.
If really want them, then loki query is a better solution. This method is done in close block metrics for madara dashboard, can take a look for reference

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
```diff
@@ -345,7 +347,7 @@ impl AggregatorHandler {
 
         // Update batching rate (batches per hour)
         // This is a simple counter that will be used to calculate rate in Grafana
-        ORCHESTRATOR_METRICS.batching_rate.record(1.0, &attributes);
+        ORCHESTRATOR_METRICS.batch_creation_rate.add(1.0, &attributes);
```
byteZorvin: Batching rate would be a query in dashboard, looks a bit confusing to me.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -25,6 +24,52 @@ impl MetricsRecorder {
...
+                KeyValue::new("from_state", from_state.to_string()),
+                KeyValue::new("to_state", to_state.to_string()),
+                KeyValue::new("operation_job_type", format!("{:?}", job_type)),
+            ],
+        );
+    }
+
+    pub fn record_job_status(job: &JobItem, status: &JobStatus) {
+        ORCHESTRATOR_METRICS.job_status_tracker.update_job_status(
+            job.internal_id,
```
byteZorvin, line 62: That creates a new time series for every job and every block (effectively unbounded/high-cardinality). It will blow up metrics memory/cardinality and backend cost

orchestrator/src/utils/metrics.rs
prkpndy: `1` is a standard unit for unitless metrics in OTEL. I have updated the name and the description.

orchestrator/src/utils/metrics_recorder.rs
prkpndy: `${namespace}` is a Grafana dashboard variable, not a runtime value in Rust. IMO `service_name` should come from OTEL resource attributes (set via instrumentation config), not hardcoded per metric call.

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: have removed start_block and bucket_id. batch_index will be needed. we can update the dashboard later to use loki for metrics.

orchestrator/src/worker/event_handler/service.rs
prkpndy: done

orchestrator/src/worker/event_handler/service.rs
prkpndy: done

orchestrator/src/worker/event_handler/service.rs
prkpndy: fixed

orchestrator/src/worker/event_handler/triggers/batching/aggregator.rs
prkpndy: updated the comment

orchestrator/src/utils/metrics_recorder.rs
prkpndy: fixed

byteZorvin (approved):

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -299,6 +263,10 @@ impl JobHandlerService {
             }
         }
 
+        // Calculate and record queue wait time after confirming we will process the job
+        let queue_wait_time = Utc::now().signed_duration_since(job.created_at).num_seconds() as f64;
```
Mohiiit: I think this metric is emitted too early. `record_job_processing_started` runs before `check_ready_to_process`, so jobs that are requeued due to unmet dependencies still get counted as “processing started.” Could we move this after readiness succeeds (ideally after transitioning to `LockedForProcessing`) so the metric reflects actual processing starts?

Mohiiit (approved):

PR #967, opened 2026-02-05
fix: restore L1->L2 messages with correct fee during chain revert

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

When reverting the chain to a previous block (during reorgs), consumed L1->L2 messages were not being restored to the pending state. Additionally, if they were restored, the `paid_fee_on_l1` was lost because it was not persisted after the message moved from pending to confirmed.

Resolves: #NA

## What is the new behavior?

- L1 handler transactions are now restored to pending state during chain reverts
- Added a new RocksDB column (`l1_to_l2_paid_fee_by_nonce`) to persist the `paid_fee_on_l1` when messages are confirmed
- During revert, messages are restored with the correct fee from storage
- Falls back to a default fee for messages confirmed before this fix was deployed
- Added comprehensive tests for revert functionality

No

This adds a new RocksDB column which will be automatically created on startup. Existing confirmed messages that predate this change will use a fallback fee value if reverted.

### Database Schema Changes

New column added:
- `l1_to_l2_paid_fee_by_nonce`: Maps `core_contract_nonce (8 bytes)` → `paid_fee_on_l1 (u128 as 16 bytes big-endian)`

This column stores the fee paid on L1 when a message is confirmed, so it can be recovered during reverts.

### DIFF
--- a/madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
+++ b/madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
@@ -12,19 +12,34 @@ pub const L1_TO_L2_PENDING_MESSAGE_BY_NONCE: Column =
     Column::new("l1_to_l2_pending_message_by_nonce").set_point_lookup();
 /// <core_contract_nonce 8 bytes> => txn hash
 pub const L1_TO_L2_TXN_HASH_BY_NONCE: Column = Column::new("l1_to_l2_txn_hash_by_nonce").set_point_lookup();
+/// <core_contract_nonce 8 bytes> => paid_fee_on_l1 (u128 as 16 bytes big-endian)
+pub const L1_TO_L2_PAID_FEE_BY_NONCE: Column = Column::new("l1_to_l2_paid_fee_by_nonce").set_point_lookup();
+
+// Fallback fee used when re-inserting pending L1->L2 messages during a revert
+// when the fee is not found in storage (for messages confirmed before this fix).
+const REVERT_PENDING_FEE_ON_L1_FALLBACK: u128 = 1_000_000_000_000;
 
 impl RocksDBStorageInner {
     /// Also removed the given txns from the pending column.
-    pub(super) fn messages_to_l2_write_trasactions<'a>(
+    /// Stores the paid_fee_on_l1 from pending messages so it can be recovered during revert.
+    pub(super) fn messages_to_l2_write_transactions<'a>(
         &self,
         txs: impl IntoIterator<Item = (&'a L1HandlerTransaction, &'a L1HandlerTransactionReceipt)>,
     ) -> Result<()> {
         let mut batch = WriteBatchWithTransaction::default();
         let pending_cf = self.get_column(L1_TO_L2_PENDING_MESSAGE_BY_NONCE);
         let on_l2_cf = self.get_column(L1_TO_L2_TXN_HASH_BY_NONCE);
+        let fee_cf = self.get_column(L1_TO_L2_PAID_FEE_BY_NONCE);
 
         for (txn, receipt) in txs {
             let key = txn.nonce.to_be_bytes();
+
+            // Read the pending message to get the paid_fee_on_l1 before deleting
+            if let Some(pending_msg) = self.get_pending_message_to_l2(txn.nonce)? {
+                // Store the fee so we can recover it during revert
+                batch.put_cf(&fee_cf, key, pending_msg.paid_fee_on_l1.to_be_bytes());
+            }
+
             batch.delete_cf(&pending_cf, key);
             batch.put_cf(&on_l2_cf, key, receipt.transaction_hash.to_bytes_be());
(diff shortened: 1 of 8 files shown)


Mohiiit (commented):
I have a couple of questions in the inline comments.

madara/crates/client/db/src/tests/test_block.rs
```diff
@@ -1,7 +0,0 @@
-#[cfg(test)]
```
Mohiiit, line 1: Why was this test removed? If it’s redundant, can we add a short note in the PR description (or move it to another test module) so we don’t silently lose coverage.

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
```diff
@@ -82,6 +97,52 @@ impl RocksDBStorageInner {
...
+
+            // Re-insert as pending with the actual paid_fee_on_l1
+            let pending = L1HandlerTransactionWithFee::new(tx.clone(), paid_fee_on_l1);
+            batch.put_cf(&pending_cf, nonce_bytes, super::serialize(&pending)?);
+        }
+
+        Ok(())
+    }
+
     pub(super) fn message_to_l2_remove_txns(
```
Mohiiit, line 146: Now that `messages_to_l2_restore_pending` exists, do we still need `message_to_l2_remove_txns`? If it is still needed for block removal, should it also delete `L1_TO_L2_PAID_FEE_BY_NONCE` to avoid leaving stale fee entries?

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
```diff
@@ -82,6 +97,52 @@ impl RocksDBStorageInner {
...
+        let on_l2_cf = self.get_column(L1_TO_L2_TXN_HASH_BY_NONCE);
+        let fee_cf = self.get_column(L1_TO_L2_PAID_FEE_BY_NONCE);
+
+        for tx in txs {
+            let nonce_bytes = tx.nonce.to_be_bytes();
+
+            // Get the stored paid_fee_on_l1, falling back to default if not found
+            // (for messages confirmed before this fix was implemented)
+            let paid_fee_on_l1 =
+                self.get_l1_handler_paid_fee_by_nonce(tx.nonce)?.unwrap_or(REVERT_PENDING_FEE_ON_L1_FALLBACK);
```
Mohiiit, line 130: Is the fallback value safe for pre-fix messages? If not, consider making it configurable/derived, or at least logging when the fallback path is used.

madara/crates/client/db/src/tests/test_block.rs
prkpndy: this exact test is present in test_open.rs. also this file was not present in `db/src/tests.rs` => test was not being run anyways

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
prkpndy: we mostly have no other option than to use a default value. another approach was to delete all the pending messages and remove the unneeded messages from processed column and reset the l1 sync tip, but that would have been much more complicated and would have required a restart. so that's why i chose this route. will add a log and make it configurable with reasonable default value.

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
prkpndy: regarding if this function is needed => the above function removes from completed and add in pending while this removes from completed and don't put it anywhere.
regarding if we should delete `L1_TO_L2_PAID_FEE_BY_NONCE` => yes. we should delete it. will do!

prkpndy (commented):
reply to pr comments

madara/crates/client/db/src/tests/mod.rs
```diff
@@ -1,3 +1,4 @@
 #![cfg(any(test, feature = "testing"))]
```
byteZorvin, line 1: All files in the module are #![cfg(test)], why was feature="testing" added?

madara/crates/client/db/src/tests/test_revert_l1_messages.rs
```diff
@@ -0,0 +1,145 @@
+#![cfg(test)]
```
byteZorvin, line 1:
```suggestion
```
Nit: If in mod.rs we decide to remove `feature="testing"` we can remove this?

byteZorvin (changes requested):
As discussed offline, 
- lets remove the pending logic
- update the testcase to have at least 2 blocks
- pre hook of cancellation before reverto
- kill madara on revert to

byteZorvin (approved):
+ revert get_l1_handler_txn_hash_by_nonce

PR #970, opened 2026-02-10
feat: add across block exec cache for storage and class hash

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Currently, we only have write cache (in Layered State Adapter, which is also cleared when the block is persisted) + we have same block cache kept by the Blockifier.

## What is the new behavior?

Now, we have added an in-memory cache option for hot contracts in the LSA layer itself. This cache will be persisted in the memory across blocks.

NOPES

## Other information

Benchmarking this gives ~2x performance on Paradex Trading!


madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -482,6 +482,9 @@ impl ExecutorThread {
                 execution_state.executor.bouncer.lock().expect("Bouncer lock poisoned").get_bouncer_weights()
             );
             tracing::debug!("Block now full: {:?}", block_full);
+            if let Some(block_state) = execution_state.executor.block_state.as_mut() {
+                block_state.state.evict_read_cache_if_needed();
```
Mohiiit, line 486: Is it intentional to run read-cache eviction both here and again in LayeredStateAdapter::finish_block? If both run every block, this may be redundant lock/queue work. Could we keep one eviction cadence and document it?

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -130,22 +497,61 @@ impl<D: MadaraStorageRead> StateReader for LayeredStateAdapter<D> {
         {
             return Ok(*el);
         }
-        self.inner.get_storage_at(contract_address, key)
+        if let Some(read_cache) = &self.read_cache {
```
Mohiiit, line 167: get_storage_at / get_nonce_at / get_class_hash_at all follow the same L0 -> L1 -> DB -> backfill flow with similar hit/miss handling. Could we extract a helper so policy + metrics stay centralized?

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -29,6 +31,359 @@ struct CacheByBlock {
...
+        size_of::<CacheEntry<CompiledClassHash>>() + size_of::<ClassEntryKey>();
+    const ORDER_ENTRY_SIZE: usize = size_of::<CacheQueueEntry>();
+
+    fn next_seq(&mut self) -> u64 {
+        let seq = self.next_seq;
+        self.next_seq = self.next_seq.wrapping_add(1);
+        seq
+    }
+
+    fn insert_storage(&mut self, contract_address: ContractAddress, key: StorageKey, value: Felt) {
```
Mohiiit: The insert_* methods in ExecutionReadCacheInner still repeat sequence/replace/accounting/queue bookkeeping. Could we centralize that in a shared helper to reduce drift risk?

Mohiiit (commented):
A few follow-ups on maintainability and test depth for the execution read cache.

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -293,4 +733,98 @@ mod tests {
...
+        adaptor.finish_block(StateMaps::default(), Default::default(), Default::default()).unwrap();
+
+        {
+            let cache = adaptor.read_cache.as_ref().unwrap();
+            let guard = cache.inner.read().unwrap();
+            assert_eq!(guard.storage.get(&(contract_address, storage_key)).unwrap().value, Felt::THREE);
+        }
+    }
+
+    #[tokio::test]
```
Mohiiit, line 435: Nice cache coverage additions. Could we extend tests beyond storage to include nonce, class_hash, compiled_class_hash, and compiled_class_hash_v2 paths (including allowlist/all-contracts modes)?

madara/crates/client/exec/src/metrics.rs
```diff
@@ -58,7 +66,28 @@ impl ExecutionMetrics {
             "ms".to_string(),
         );
 
-        Self { tx_execution_time_histogram }
+        let read_cache_hits_counter = register_counter_metric_instrument(
```
Mohiiit, line 95: Would it make sense to add a test metrics harness (similar to cairo-native test_counters) so read-cache hit/miss/size metrics can be asserted in unit tests?

madara/crates/client/block_production/src/executor/thread.rs
prkpndy: yeah. here, the cache is cleared after every batch. in finish block, it's getting cleared after every block. have put it at finish block as well, since we update the read cache from write cache (if anything is evicted from LSA). so not redundant

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -121,31 +482,75 @@ impl<D: MadaraStorageRead> LayeredStateAdapter<D> {
...
+
+    fn get_contract_scoped_with_cache<T, FLayered, FReadCache, FDb, FBackfill>(
+        &self,
+        contract_address: ContractAddress,
+        cache_kind: &'static str,
+        layered_lookup: FLayered,
+        read_cache_lookup: FReadCache,
+        db_lookup: FDb,
+        read_cache_backfill: FBackfill,
+    ) -> StateResult<T>
```
Mohiiit, line 155: [P3] Reduce duplication in StateReader getters

get_storage_at, get_nonce_at, get_class_hash_at, get_compiled_class_hash, and get_compiled_class_hash_v2 follow the same cache-or-fetch pattern. A shared helper would cut duplication and keep hit/miss semantics consistent.

Mohiiit (commented):
I have one maintainability suggestion in the inline comment.

madara/crates/client/db/src/lib.rs
```diff
@@ -323,6 +324,18 @@ pub struct MadaraBackend<DB = RocksDBStorage> {
     pub custom_header: Mutex<Option<CustomHeader>>,
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct ExecutionReadCacheConfig {
```
Mohiiit, line 328: Consider simplifying the config to just enabled + contracts: Option<Vec<ContractAddress>> (where None => all, Some(list) => allowlist). Also, during startup/config load, if enabled == false but contracts.is_some() (or other cache params are set), emit a warn! that config validation failed / cache is configured but disabled so we don't silently run with caching off.

madara/node/src/cli/backend.rs
```diff
@@ -195,6 +218,12 @@ impl BackendParams {
             save_preconfirmed: !self.no_save_preconfirmed,
             unsafe_starting_block: self.unsafe_starting_block,
             skip_migration_backup: self.skip_migration_backup,
+            execution_read_cache: mc_db::ExecutionReadCacheConfig {
+                enabled: self.exec_read_cache_enabled,
+                all_contracts: self.exec_read_cache_all_contracts,
+                contracts: self.exec_read_cache_contracts.clone(),
+                max_memory_bytes: self.exec_read_cache_max_memory_mib * MiB,
```
Mohiiit: max_memory_bytes: self.exec_read_cache_max_memory_mib * MiB can overflow usize for large inputs and silently wrap. Consider checked_mul (with an explicit error) or saturating_mul/clamp.

madara/crates/client/exec/src/metrics.rs
```diff
@@ -68,6 +123,27 @@ impl ExecutionMetrics {
             &[KeyValue::new("tx_type", tx_type.to_string()), KeyValue::new("context", context.to_string())],
         );
     }
+
+    pub fn record_read_cache_hit(&self, kind: &str) {
```
Mohiiit: record_read_cache_{hit,miss} takes kind: &str but callsites pass static kinds. Consider switching to kind: &'static str and avoiding kind.to_string() so we don't allocate on every cache hit/miss.

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -29,6 +31,358 @@ struct CacheByBlock {
...
+    fn with_write_recording<R>(&self, f: impl FnOnce(&mut ExecutionReadCacheInner) -> R) -> R {
+        let mut guard = self.inner.write().expect("Poisoned execution read cache lock");
+        let out = f(&mut guard);
+        let bytes = guard.current_bytes as u64;
+        drop(guard);
+        exec_metrics().record_read_cache_size_bytes(bytes);
+        out
+    }
+
+    fn from_config(config: &ExecutionReadCacheConfig) -> Option<Self> {
```
Mohiiit: The ExecutionReadCache implementation is now large enough that keeping it in layered_state_adapter.rs makes the adapter harder to scan/maintain. Consider moving it to a dedicated module/file (for example execution_read_cache.rs or read_cache.rs) and keeping LayeredStateAdapter focused on orchestration.

madara/crates/client/exec/src/layered_state_adapter.rs
```diff
@@ -29,6 +31,358 @@ struct CacheByBlock {
...
+                        return true;
+                    }
+                }
+            }
+            CacheKey::ClassHash(contract_address) => {
+                if let Some(existing) = self.class_hashes.get(&contract_address) {
+                    if existing.seq == entry.seq {
+                        let existing = self.class_hashes.remove(&contract_address).expect("entry exists");
+                        self.current_bytes = self.current_bytes.saturating_sub(Self::CLASS_HASH_ENTRY_SIZE);
+                        if self.cached_class_hashes.remove(&existing.value) {
```
Mohiiit: cached_class_hashes is a HashSet, and eviction of one contract's class-hash entry unconditionally removes the class hash from the set. If multiple cached contracts share a class hash, this can flip should_cache_class_hash to false even though another contract still references it, reducing compiled-hash caching. Is this intended behavior?

Mohiiit (commented):
Left a few inline comments around execution read-cache config validation, metrics hot-path behavior, and keeping layered_state_adapter.rs maintainable.

madara/crates/client/exec/src/layered_state_adapter.rs
prkpndy: this was a conscious design decision actually. the code was getting too complicated otherwise - had to maintain ref count for each contract for which we had to store compiled class. IRL, this won't matter much since the class hash would be eligible for caching again once we get get class hash call for the contract

madara/node/src/cli/backend.rs
prkpndy: updated

madara/crates/client/db/src/lib.rs
prkpndy: done

madara/crates/client/exec/src/metrics.rs
prkpndy: done

madara/crates/client/exec/src/layered_state_adapter.rs
prkpndy: done

madara/node/src/cli/backend.rs
heemankv: great catch !

heemankv (approved):
Looks solid and well thought of, LGTM.
All my doubts have been answered in comments by @Mohiiit | gg on that as well.

Mohiiit (approved):

PR #971, opened 2026-02-12
rpc: implement starknet_getMessagesStatus (v0.9.0, v0.10.0)

Mohiiit:
## Summary
Implements `starknet_getMessagesStatus` for RPC user versions **v0.9.0** and **v0.10.0**.

Key design choices:
- DB-backed response path (no nested RPC calls, no on-demand L1 fetch at request time).
- Metadata is written during settlement sync and finalized during L1-handler block consumption.
- `v0.8.1` is intentionally out of scope.

### Before
```mermaid
flowchart TD
    A["L1 emits message"] --> B["Settlement sync stores pending by nonce"]
    B --> C["Block import stores nonce -> L2 tx hash"]
    D["getMessagesStatus(l1_tx_hash)"] --> E["No local l1_tx_hash index"]
    E --> F["Cannot resolve status from DB"]
```

### After
```mermaid
flowchart TD
    A["L1 emits message event"] --> B["Settlement sync receives (l1_block, l1_tx_hash, message)"]
    B --> C["Write nonce -> l1_tx_hash"]
    C --> D["Insert (l1_tx_hash||nonce) -> empty marker"]
    D --> E["Validate pending/not-cancelled"]
    E --> F["Store pending by nonce"]

    G["Block import consumes L1 handler"] --> H["Write nonce -> l2_tx_hash"]
    H --> I["Write (l1_tx_hash||nonce) -> l2_tx_hash"]

    J["getMessagesStatus(l1_tx_hash)"] --> K["Prefix-scan (l1_tx_hash||*)"]
    K --> L["Resolve consumed tx receipts/status"]
    L --> M["Return statuses in nonce order"]
```

## Message Lifecycle
```mermaid
stateDiagram-v2
    [*] --> Unknown
    Unknown --> Seen: L1 event finalized and indexed
    Seen --> Pending: validity checks pass
    Pending --> Consumed: L1 handler included in L2 block
    Seen --> Consumed: consumed hash backfilled
    Pending --> Dropped: cancelled/invalid
```

RPC behavior:
- No keys for `l1_tx_hash` => `TXN_HASH_NOT_FOUND`.
- Keys exist but only empty markers => `[]`.
- Consumed entries exist => return status objects from L2 tx receipts.

## Data Model
Secondary index introduced for lookup by L1 transaction hash:
- `nonce -> l1_tx_hash`
- `(l1_tx_hash||nonce) -> empty | l2_tx_hash`

This allows efficient prefix scans per `l1_tx_hash` and avoids O(N) runtime scans.

## Migration
- DB version bumped to **11**.
- Added **v10 -> v11** no-op migration (new columns are lazily created by RocksDB).

## Scope
- Supported in this PR: **v0.9.0**, **v0.10.0**.
- Not changed: **v0.8.1**.


madara/crates/primitives/rpc/src/v0_9_0/mod.rs
```diff
@@ -130,3 +132,161 @@ fn block_id_to_l1_accepted() {
...
     assert_eq!(s, "\"l1_accepted\"");
 }
+
+/// The hash of an Ethereum transaction.
+///
+/// Spec: `L1_TXN_HASH` is `NUM_AS_HEX` (`^0x[a-fA-F0-9]+$`). We accept 1..=64 hex digits
+/// after the `0x` prefix and left-pad to 32 bytes. We always serialize to a 32-byte
+/// canonical form (`0x` + 64 lowercase hex digits).
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct L1TxnHash(pub [u8; 32]);
```
byteZorvin, line 146: Will this work with L1TxnHash as starknet txn in case of L3s?

byteZorvin (approved):
Great work on this, super clean and was so easy to review.

madara/crates/primitives/rpc/src/v0_9_0/mod.rs
heemankv: Yes — it will work for L3s where the settlement chain is Starknet, because L1TxnHash explicitly accepts Starknet‑style transaction hashes (shorter than 32 bytes) and left‑pads to 32 bytes.

heemankv (commented):
Overall looks good, 
Need Clarifications on :  
1. `message_to_l2_remove_txns` (was written for re-org but has been commented) only deletes L1_TO_L2_TXN_HASH_BY_NONCE  but leaves the secondary index, if say the the l2 was supposed to skip that message after re-org, it will still be present in the secondary index right ? that will lead to the get_message_status to return an empty vector ( [] ) rather than a TXN_NOT_FOUND, is this being handled in @prkpndy 's l1 revert fix pr ? am I understanding this correctly?

heemankv (approved):

PR #972, opened 2026-02-12
test(cairo-native): make memory cache timeout deterministic

Mohiiit:
## What / Why

`test_handle_sierra_class_memory_cache_timeout` in `mc-class-exec` is intended to validate the behavior when the cairo-native memory-cache worker is slow and a request times out.

The test was flaky because the worker thread and the timeout are racing: depending on scheduling, the worker could send its response before the timeout trips, making the test outcome non-deterministic.

## Changes

- Add a **test-only** delay hook in the memory-cache worker send path (`#[cfg(test)]`) so the test can deterministically force the timeout.
- Add a small guard in the test module to always reset the delay back to `0ms`, preventing cross-test contamination.
- No production behavior changes (the delay code is compiled only for tests).

## Testing

```bash
cd madara
cargo test -p mc-class-exec test_handle_sierra_class_memory_cache_timeout -- --nocapture
cargo clippy -p mc-class-exec --all-features --tests --no-deps -- -D warnings
```

### DIFF
--- a/madara/crates/client/cairo_native/src/cache.rs
+++ b/madara/crates/client/cairo_native/src/cache.rs
@@ -9,11 +9,26 @@ use dashmap::DashMap;
 use mp_convert::ToFelt;
 use starknet_api::core::ClassHash;
 use std::path::PathBuf;
+#[cfg(test)]
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, Instant};
 
 use super::config;
 use super::native_class::NativeCompiledClass;
 use mp_class::SierraConvertedClass;
+
+#[cfg(test)]
+static TEST_MEMORY_CACHE_SEND_DELAY_MS: AtomicU64 = AtomicU64::new(0);
+
+#[cfg(test)]
+pub fn set_test_memory_cache_send_delay(delay: Duration) {
+    TEST_MEMORY_CACHE_SEND_DELAY_MS.store(delay.as_millis() as u64, Ordering::Relaxed);
+}
+
+#[cfg(test)]
+fn test_memory_cache_send_delay() -> Duration {
+    Duration::from_millis(TEST_MEMORY_CACHE_SEND_DELAY_MS.load(Ordering::Relaxed))
+}
 
 /// Timeout for loading compiled classes from disk.
 ///
@@ -324,6 +339,13 @@ pub(crate) fn try_get_from_memory_cache(
                 "memory_hit"
             );
 
+            #[cfg(test)]
+            {
+                let delay = test_memory_cache_send_delay();
+                if !delay.is_zero() {
+                    std::thread::sleep(delay);
+                }
+            }
             let _ = tx.send(Some(runnable));
         } else {
             // Memory cache miss - log it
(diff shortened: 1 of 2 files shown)


heemankv (approved):
Looks good, 
The _delay_guard Drop to reset it nice!

Small nitpick : 
```rust
#[cfg(test)]
{
    let delay = test_memory_cache_send_delay();
    if !delay.is_zero() {
        std::thread::sleep(delay);
    }
}
let _ = tx.send(Some(runnable));
```
These code-statements can be taken out of the condition and written after it.
We would have to define a return_val before the condition for it.
but this will allow us to not write the delay and the tx.send logic twice, your call here.

byteZorvin (approved):

madara/crates/client/cairo_native/src/cache.rs
```diff
@@ -9,12 +9,27 @@ use dashmap::DashMap;
...
+
+#[cfg(test)]
+pub fn set_test_memory_cache_send_delay(delay: Duration) {
+    TEST_MEMORY_CACHE_SEND_DELAY_MS.store(delay.as_millis() as u64, Ordering::Relaxed);
+}
+
+#[cfg(test)]
+fn test_memory_cache_send_delay() -> Duration {
+    Duration::from_millis(TEST_MEMORY_CACHE_SEND_DELAY_MS.load(Ordering::Relaxed))
+}
```
byteZorvin, line 31: Nit: We don't need it to be atomic as, anyway, all the Cairo Native tests share a mutex guard, so they are serialized and not running in parallel.

byteZorvin (commented):
Looks good, one nit

madara/crates/client/cairo_native/src/execution.rs
```diff
@@ -961,6 +969,10 @@ mod tests {
                 .build(),
         );
 
+        // Delay the memory-cache worker so the timeout fires after it records a miss.
+        cache::set_test_memory_cache_send_delay(Duration::from_millis(5));
+        let _delay_guard = MemoryCacheDelayGuard;
```
byteZorvin, line 974: Nit: Also in your PR description, you mentioned that you want to avoid cross-test contamination, but making the delay atomic isn't enough by itself. You would have to use a mutex guard here, as another thread might read the value between line 973 and 974. Although the risk is contained here, since these tests are running in serialized form by taking the shared mutex guard in the first line.

madara/crates/client/cairo_native/src/cache.rs
Mohiiit: Good callout. We can simplify this. I used `AtomicU64` because the delay is written by the test thread and read by the spawned memory-cache worker thread, so lock-free cross-thread visibility is enough for this test hook. Since Cairo Native tests are serialized via `test_counters::acquire_and_reset()`, this does not rely on parallel test safety. I can also switch to a `Mutex<Duration>` if you prefer consistency, but functionally this path is already safe.

madara/crates/client/cairo_native/src/execution.rs
Mohiiit: Agreed on the general point. In this specific test, we already take the shared test mutex at the start (`test_counters::acquire_and_reset()`), so another test cannot interleave between setting the delay and creating the guard. The guard is there to guarantee reset at test end and avoid leakage. So the current setup avoids cross-test contamination for this module. If useful, I can add a short inline comment to make this synchronization assumption explicit.

PR #975, opened 2026-02-13
feat(reorg): coordinated revert with L1 message sync rewind

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

`revertTo` only reverted L2 chain state and did not coordinate service shutdown or L1 message sync state rewind.
This could leave in-memory services stuck after tip rollback and forced operators to restart the node manually.
L1 message metadata was also not fully reverted, which could cause inconsistencies across processed/pending message state and replay cursor.

Resolves: #NA

## What is the new behavior?

- Replaces admin revert flow with `revertToAndShutdown` and coordinates stop requests for non-admin services, waits for *actual* service shutdown, reverts, then shuts down node.
- Adds requested vs actual service status plumbing and exposes admin `serviceStatus` RPC for shutdown acknowledgements.
- Reverts L1 messaging state alongside chain revert:
  - removes reverted L1 handler nonces from processed (`L1_TO_L2_TXN_HASH_BY_NONCE`) and pending (`L1_TO_L2_PENDING_MESSAGE_BY_NONCE`) stores,
  - tracks `nonce -> source_l1_block` in a new DB column,
  - rewinds `LAST_SYNCED_L1_EVENT_BLOCK` from reverted nonce metadata.
- Adds optional `l1_messages_rewind_hint` parameter to `revertToAndShutdown` for safe fallback when historical source metadata is missing.
- Adds broad maintainable tests for:
  - service shutdown gating before revert,
  - L1 message state cleanup + sync-tip rewind,
  - hint/no-hint behavior,
  - post-revert shutdown signal.

Yes.

- Admin RPC method `revertTo` is removed; use `revertToAndShutdown`.
- `revertToAndShutdown` signature now includes optional `l1_messages_rewind_hint`.
- DB schema adds a new column family: `l1_to_l2_source_l1_block_by_nonce`.

Migration path:
- Update admin clients to call `madara_revertToAndShutdown`.
- If reverting old data with missing L1 source metadata, pass `l1_messages_rewind_hint`.
- Add `db-migration` label on this PR.

## Other information

Validation run:

- `cargo check`
- `cargo clippy --workspace --all-features --tests --no-deps -- -D warnings`
- `cargo test -p mc-db test_reorg_l1_messages -- --nocapture`
- `cargo test -p mc-rpc revert_waits_for_actual_service_shutdown_before_reverting_and_cancels_node -- --nocapture`
- `cargo test -p mc-rpc revert_requires_hint_when_source_mapping_missing_but_succeeds_with_hint -- --nocapture`


madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -11,6 +11,21 @@ use mp_rpc::v0_9_0::{
     ClassAndTxnHash, ContractAndTxnHash,
 };
 use mp_transactions::{L1HandlerTransactionResult, L1HandlerTransactionWithFee};
+use mp_utils::service::{MadaraServiceId, MadaraServiceStatus, SERVICE_GRACE_PERIOD};
+use std::time::Duration;
+use tokio::time::Instant;
+
+fn services_to_stop_for_revert() -> [MadaraServiceId; 7] {
+    [
+        MadaraServiceId::Database,
```
Mohiiit: Is `MadaraServiceId::Database` expected to be controlled by the service monitor lifecycle here? If not, can we remove it from this stop list to avoid implying it participates in ack-based shutdown?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -121,12 +136,91 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+            tracing::info!(
+                target: "rpc::admin",
+                "revertToAndShutdown: shutdown requested for service={} (was_requested={})",
+                svc,
+                prev
+            );
+        }
+
+        // 2) Wait until all services are *actually* down.
+        let timeout = SERVICE_GRACE_PERIOD + Duration::from_secs(5);
```
Mohiiit: Could we centralize these timing literals into named constants (`+5s`, log interval, poll interval, delayed cancel) so tuning is explicit and consistent?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -121,12 +136,91 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+                break;
+            }
+
+            if Instant::now() >= deadline {
+                tracing::error!(
+                    target: "rpc::admin",
+                    "revertToAndShutdown: timed out waiting for services to stop; still_up={:?}",
+                    still_up.iter().map(|s| s.to_string()).collect::<Vec<_>>()
+                );
+                return Err(StarknetRpcApiError::ErrUnexpectedError {
```
Mohiiit, line 202: Timeout returns an error after services were already requested `Off`. Should this path fail-closed (for example force global cancellation) instead of leaving the process in a potentially half-stopped state?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -121,12 +136,91 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
+
+        // 3) Revert DB state, then refresh backend chain tip broadcast.
+        tracing::info!(
+            target: "rpc::admin",
+            "revertToAndShutdown: reverting chain to block_hash={:#x} (block_number={}, l1_messages_rewind_hint={:?})",
+            block_hash,
+            target_block_n,
+            l1_messages_rewind_hint
+        );
+        self.backend.revert_to(&block_hash, l1_messages_rewind_hint).map_err(StarknetRpcApiError::from)?;
```
Mohiiit: If `backend.revert_to(...)` fails here, we return an error after stop requests were sent and without restoring prior service state. Can we add a rollback path (reactivate previously-on services) or force process shutdown on failure?

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -333,7 +333,10 @@ impl PipelineSteps for GatewaySyncSteps {
                                         } else {
                                             // Normal reorg - found common ancestor
                                             tracing::info!("🔄 Triggering reorg to common ancestor hash={:#x}", common_ancestor_hash);
-                                            self._backend.revert_to(&common_ancestor_hash)?;
+                                            // Gateway-synced historical blocks do not always have per-nonce
+                                            // L1 source block metadata. Use a conservative rewind hint so
+                                            // automatic reorg remains available.
+                                            self._backend.revert_to(&common_ancestor_hash, Some(0))?;
```
Mohiiit: Using `Some(0)` is safe but very conservative because it can rewind L1-message sync to genesis. Can we derive a tighter hint from available reorg context to reduce replay cost?

Mohiiit (commented):
I have a few questions about shutdown/revert failure handling in this flow. Also, I could not find an automated e2e that covers revertToAndShutdown across restart with L1-handler activity and DB-state validation, which seems important for this path.

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -583,7 +597,7 @@ impl MadaraStorageWrite for RocksDBStorage {
     /// * This is a destructive operation - all blocks after the target block are permanently removed.
     /// * The function is atomic - if any step fails, the database may be in an inconsistent state.
     /// ```
-    fn revert_to(&self, new_tip_block_hash: &Felt) -> Result<(u64, Felt)> {
+    fn revert_to(&self, new_tip_block_hash: &Felt, l1_messages_rewind_hint: Option<u64>) -> Result<(u64, Felt)> {
```
byteZorvin: Nit: Can update the Arguments of the function doc, also add about the shutdown side-effect.

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -621,6 +635,58 @@ impl MadaraStorageWrite for RocksDBStorage {
             anyhow::bail!("Cannot revert to block_n={target_block_n} which is > current tip={current_tip}");
         }
 
+        // Preflight L1 messaging rewind before any destructive write.
+        let reverted_l1_handler_nonces = self
+            .inner
+            .collect_reverted_l1_handler_nonces(target_block_n, current_tip)
+            .context("Collecting reverted L1 handler nonces")?;
+
+        let reverted_l1_handler_nonces: HashSet<u64> = reverted_l1_handler_nonces.into_iter().collect();
```
byteZorvin: Why is a Hashset conversion required here if we just want to iterate later which is already returned by `reverted_l1_handler_nonces`

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -621,6 +635,58 @@ impl MadaraStorageWrite for RocksDBStorage {
...
+                missing_source_block_nonces.len()
+            );
+        }
+
+        let rewind_from_l1_block = match (min_source_l1_block, l1_messages_rewind_hint) {
+            (Some(db_min), Some(hint)) => Some(db_min.min(hint)),
+            (Some(db_min), None) => Some(db_min),
+            (None, Some(hint)) => Some(hint),
+            (None, None) => None,
+        };
```
byteZorvin: Nit: can be simplified like this
```suggestion
    let rewind_from_l1_block = [min_source_l1_block, l1_messages_rewind_hint]
    .into_iter()
    .flatten()
    .min();
```

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -621,6 +635,58 @@ impl MadaraStorageWrite for RocksDBStorage {
...
+        let l1_messaging_sync_tip_after_revert = rewind_from_l1_block.map(|b| b.saturating_sub(1));
```
byteZorvin, line 786: Going directly one block might not be safe, as one may process a deposit twice

madara/crates/client/sync/src/gateway/blocks.rs
byteZorvin: I am skeptical that this is not safe as well, as this may lead to processing messages twice

byteZorvin: Also I would propose that hint should'nt used, or be a param

byteZorvin (commented):
Great work on this, few concerns around the l1_block_hint

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: fair point. restarting the service won't be possible tbh. i'll add the code to stop the node completely!

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: yeah, restarting the services won't be possible. i'll shut down the node completely!

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: done

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: done

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: but on a second thought, this would anyways lead to an inconsistent state. probably we should have some fallback if reorg fails 🤔 ?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/mod.rs
prkpndy: -1 is intentional for safety at boundary/inclusive stream semantics, and replay is expected to be idempotent. Our major concern is missing a message here. If a message is duplicated, there are safety checks at both message sync and block production

madara/crates/client/db/src/rocksdb/mod.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/mod.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/mod.rs
prkpndy: done

Mohiiit (commented):
A few inline naming and maintainability nits.

madara/crates/client/db/src/lib.rs
```diff
@@ -985,6 +985,9 @@ impl<D: MadaraStorageRead> MadaraBackend<D> {
     pub fn get_l1_handler_txn_hash_by_nonce(&self, core_contract_nonce: u64) -> Result<Option<Felt>> {
         self.db.get_l1_handler_txn_hash_by_nonce(core_contract_nonce)
     }
+    pub fn get_l1_message_source_l1_block_by_nonce(&self, core_contract_nonce: u64) -> Result<Option<u64>> {
```
Mohiiit: Naming nit: can we make this consistently l1_handler_* and drop source? For example get_l1_handler_l1_block_by_nonce / write_l1_handler_l1_block_by_nonce. Main ask is consistency across backend methods, storage traits, and RocksDB helpers.

Mohiiit (commented):
Inline nits in comments.

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
```diff
@@ -82,6 +85,24 @@ impl RocksDBStorageInner {
         Ok(())
     }
 
+    pub(super) fn get_l1_message_source_l1_block_by_nonce(&self, core_contract_nonce: u64) -> Result<Option<u64>> {
```
Mohiiit: Same naming nit at the storage layer: if we rename the backend API, it’d be good to rename the RocksDB inner get_* / write_* helpers and the constant name to match the l1_handler_* naming (keeping the CF string stable if we want to avoid migrations).

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -156,10 +156,11 @@ impl RocksDBStorageInner {
                     .collect::<Result<_>>()?;
 
                 self.events_remove_block(block_n, &mut batch)?;
-                self.message_to_l2_remove_txns(
-                    transactions.iter().filter_map(|v| v.transaction.as_l1_handler()).map(|tx| tx.nonce),
-                    &mut batch,
-                )?;
+                let l1_handler_nonces: Vec<u64> =
+                    transactions.iter().filter_map(|v| v.transaction.as_l1_handler().map(|tx| tx.nonce)).collect();
+                self.message_to_l2_remove_txns(l1_handler_nonces.iter().copied(), &mut batch)?;
```
Mohiiit: We now do the same L1-message cleanup here (remove_all_blocks_starting_from) and also in block_db_revert in rocksdb/blocks.rs. Seems intentional (both delete blocks), but it’s easy for them to drift. Consider factoring the remove message columns for l1_handler_nonces into a shared helper.

madara/crates/client/db/src/lib.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/l1_to_l2_messages.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/mod.rs
prkpndy: done

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
prkpndy: Thought some more about it. If the revert fails, ideally, we'll get an error response. In that case, imo we should not restart the node, since it can restart some services which we earlier stopped successfully.
Regarding this endpoint's idempotency, that'll require some work. Have created a task for that. Will handle that in a seperate PR!

Mohiiit (approved):

byteZorvin (approved):

PR #981, opened 2026-02-15
feat(parallel-merkle): add state-root parity replay e2e

Mohiiit:
## What changed
- Reworked the parity e2e to run sequential and parallel nodes in lockstep with synchronized block closes.
- Kept both runs on fresh DBs (no DB cloning helper), while enforcing identical transaction sequence and tx hashes across 100 non-empty blocks.
- Kept focus on state correctness: per-block state diff equality and per-block state root equality.

## Config used in parallel scenario
- `--parallel-merkle-flush-interval 3`
- `--parallel-merkle-max-inflight 10`
- `--parallel-merkle-trie-log-mode checkpoint`

## Code updates
- Removed the test-only `clone_db_from` helper from `madara/crates/tests/src/lib.rs`.
- Reverted `madara/crates/client/block_production/src/util.rs` to the existing timestamp behavior.
- Updated `madara/crates/tests/src/parallel_merkle_state_root_parity.rs` for lockstep dual-node replay.


madara/crates/client/block_production/src/finalizer.rs
```diff
@@ -20,14 +21,55 @@ pub struct FinalizedBlockPayload {
     pub bouncer_weights: BouncerWeights,
 }
 
+const MADARA_PARALLEL_MERKLE_FINALIZER_DEBUG_DELAY_MS: &str = "MADARA_PARALLEL_MERKLE_FINALIZER_DEBUG_DELAY_MS";
```
apoorvsadana: why not from CLI?

PR #985, opened 2026-02-24
docs(readme): refresh core docs for current capabilities

Mohiiit:
## Summary
- Keeps the existing README structure and section flow intact while updating outdated content in-place.
- Refreshes RPC/version coverage and method support to match current node behavior.
- Clarifies websocket subscription status: subscribe methods are placeholders and currently return `UnimplementedMethod`.
- Corrects `starknet_getStorageProof` documentation:
  - marked as currently unavailable in the default node profile
  - documents default retention limits in `configs/args/config.json`
  - explains default config-loading behavior when no explicit CLI/config overrides are provided
- Expands SnapSync documentation with concrete operational behavior:
  - activation threshold (`>= 1000` blocks to target)
  - per-block trielog/storage-proof caveats for snap-synced ranges
  - revert boundary behavior tied to `snap_sync_latest_block`
- Expands Cairo Native documentation with implementation details:
  - single enable flag (`--enable-native-execution`)
  - async vs blocking compilation flow
  - persisted native cache reuse across restarts and retry behavior
- Expands L3 documentation with settlement-layer specifics:
  - `Eth` vs `Starknet` settlement modes
  - endpoint interpretation and gas unit differences
  - execution-mode implications when running with Starknet settlement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,11 @@ Current migration metadata is tracked in [`.db-versions.yml`](.db-versions.yml):
 > Backups are created before migrations by default. Use
 > `--skip-migration-backup` only when you already have external backups.
 
+> [!CAUTION]
+> Database migrations are forward-oriented. After migrating to a newer schema
+> version, you cannot safely reuse the same database with an older Madara
+> binary. To roll back, restore from a backup taken before migration.
+
 To migrate your database, you have two options:
 
 1. Start the node and let automatic in-place migration run
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <!-- markdownlint-disable -->
 <div align="center">
-  <img src="https://github.com/keep-starknet-strange/madara-branding/blob/main/logo/PNGs/Madara%20logomark%20-%20Red%20-%20Duotone.png?raw=true" width="500">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -526,61 +564,56 @@ You should receive something like the following:
...
+> Backups are created before migrations by default. Use
+> `--skip-migration-backup` only when you already have external backups.
 
 To migrate your database, you have two options:
 
-1. Use Madara's **warp update** feature (recommended)
-2. Re-synchronize from genesis (not recommended)
+1. Start the node and let automatic in-place migration run
+2. Use Madara's **warp update** feature (recommended for fast local migration)
+3. Re-synchronize from genesis (not recommended)
```
prkpndy, line 622: nit: can add a note that we can't go back to a previous version once migrated

prkpndy (approved):
gg

byteZorvin (approved):

README.md
Mohiiit: Added this note in the Database Migration section: after migrating to a newer schema version, downgrading to an older Madara binary is not supported on the same DB; restore from pre-migration backup to roll back.

PR #992, opened 2026-03-04
Stop S3 directory listing in storage cleanup; tag known paths directly and treat 404 as NotFound

heemankv:
## Pull Request type
- [x] Bugfix
- [x] Testing
- [ ] Feature
- [ ] Build-related changes

- Storage cleanup lists S3 directories to discover artifacts, which is costly and error‑prone.
- Missing objects during tagging can surface as generic errors due to string matching.
- Cleanup behavior for SNOS root artifacts is implicit via listing.

Resolves: #NA

## What is the new behavior?
- Storage cleanup **never lists S3 directories**. It tags only **known, explicit paths** for batch artifacts, blob files, state updates, and SNOS root artifacts.
- L2 SNOS root tagging now directly targets **three files**: `cairo_pie.zip`, `snos_output.json`, `program_output.txt`.
- L3 SNOS root tagging now directly targets **six files**: `cairo_pie.zip`, `snos_output.json`, `program_output.txt`, `blob_data.txt`, `proof.json`, `proof_part2.json`.
- Missing objects are treated as `NotFound` via SDK metadata/HTTP status (`404`, `NoSuchKey`) and are **not considered failures** for cleanup.

No.

## Other information
**Tests added/updated**
- Unit tests assert direct tagging for L2/L3 without directory listing.
- Tagging test now expects `StorageError::NotFound` for missing objects.

**Notes**
- `list_files_in_dir` was removed from the storage trait and S3 implementation since it is no longer used.

### DIFF
--- a/orchestrator/src/core/client/storage/s3.rs
+++ b/orchestrator/src/core/client/storage/s3.rs
@@ -4,12 +4,10 @@ use crate::core::client::storage::StorageError;
 use crate::types::params::AWSResourceIdentifier;
 use async_trait::async_trait;
 use aws_config::SdkConfig;
+use aws_sdk_s3::error::{ProvideErrorMetadata, SdkError};
 use aws_sdk_s3::operation::put_object_tagging::PutObjectTaggingError;
 use aws_sdk_s3::types::{Tag, Tagging};
 use aws_sdk_s3::Client;
-use aws_smithy_runtime_api::client::result::SdkError;
-use aws_smithy_runtime_api::http::Response as SmithyResponse;
-use aws_smithy_types::error::metadata::ProvideErrorMetadata;
 use bytes::Bytes;
 
 /// AWSS3 is a struct that represents an AWS S3 client.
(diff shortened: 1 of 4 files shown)


orchestrator/src/core/client/storage/s3.rs
```diff
@@ -65,6 +67,23 @@ impl AWSS3 {
...
+
+    if let Some(raw) = err.raw_response() {
+        let status: u16 = raw.status().into();
+        if status == 404 {
+            return true;
+        }
+    }
+
+    false
+}
```
Mohiiit, line 80:
```suggestion
fn is_not_found_tagging_error(err: &SdkError<PutObjectTaggingError>) -> bool {
    let service_error = err
        .as_service_error()
        .and_then(|e| e.code())
        .map(|code| matches!(code, "NoSuchKey" | "NotFound"))
        .unwrap_or(false);

    let status_404 = err
        .raw_response()
        .map(|raw| raw.status().as_u16() == 404)
        .unwrap_or(false);

    service_error || status_404
}
```

Mohiiit (approved):
only nit, no blocker!

prkpndy (approved):
LGTM

PR #993, opened 2026-03-09
feat: add blob size buffer for aggregator batching

prkpndy:
## Pull Request type

- [x] Feature

The maximum blob size for aggregator batching is calculated as `max_num_blobs * 4096` with no margin. The prover (Atlantic) adds encryption overhead (e.g. 7 felts for `n_keys + sn_public_keys + encrypted_symmetric_keys`) to the compressed state diff, which can push the final blob count beyond the Ethereum limit of 6 blobs per transaction.

Resolves: #NA

## What is the new behavior?

- Added `--blob-size-buffer` CLI argument / `MADARA_ORCHESTRATOR_BLOB_SIZE_BUFFER` env var (default: 15) that specifies a buffer (in number of felts) subtracted from the maximum blob capacity.
- The effective max blob size is now `(max_num_blobs * 4096) - blob_size_buffer`.
- With defaults (6 blobs, 15 buffer): `24576 - 15 = 24561` felts.
- Added unit tests to verify the buffer calculation with default, zero, and custom values.

No. The default buffer of 15 is applied automatically, which is a stricter (safer) limit than before. Existing deployments will batch slightly more conservatively, avoiding the "too many blobs" error.

## Other information

The 15-felt default buffer accounts for the 7-felt encryption overhead added by the aggregator's Cairo program (`n_keys + n sn_public_keys + n encrypted_symmetric_keys` where n=3 DA committee members), with additional margin for safety.

### DIFF
--- a/orchestrator/src/tests/config.rs
+++ b/orchestrator/src/tests/config.rs
@@ -781,7 +781,11 @@ pub(crate) fn get_env_params(test_id: Option<&str>) -> EnvParams {
         .parse::<u64>()
         .unwrap(),
         max_num_blobs,
-        max_blob_size: max_num_blobs * BLOB_LEN,
+        blob_size_buffer: get_env_var_or_default("MADARA_ORCHESTRATOR_BLOB_SIZE_BUFFER", "15")
+            .parse::<usize>()
+            .unwrap(),
+        max_blob_size: max_num_blobs * BLOB_LEN
+            - get_env_var_or_default("MADARA_ORCHESTRATOR_BLOB_SIZE_BUFFER", "15").parse::<usize>().unwrap(),
         default_empty_block_proving_gas: get_env_var_or_default(
             "MADARA_ORCHESTRATOR_DEFAULT_EMPTY_BLOCK_PROVING_GAS",
             "1500000",
--- a/orchestrator/src/types/params/batching.rs
+++ b/orchestrator/src/types/params/batching.rs
@@ -10,11 +10,13 @@ pub struct BatchingParams {
     pub fixed_blocks_per_snos_batch: Option<u64>,
     pub max_snos_batches_per_aggregator_batch: u64,
     pub max_num_blobs: usize,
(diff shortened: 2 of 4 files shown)


orchestrator/src/types/params/batching.rs
```diff
@@ -34,9 +36,51 @@ impl From<BatchingCliArgs> for BatchingParams {
...
+            max_blocks_per_snos_batch: 10,
+            fixed_blocks_per_snos_batch: None,
+            max_snos_batches_per_aggregator_batch: 50,
+            default_empty_block_proving_gas: 1500000,
+            max_batch_processing_size: 10,
+        }
+    }
+
+    #[test]
+    fn test_max_blob_size_with_default_buffer() {
```
Mohiiit, line 67: rstest would be nice to user here!

orchestrator/src/tests/config.rs
```diff
@@ -781,7 +781,11 @@ pub(crate) fn get_env_params(test_id: Option<&str>) -> EnvParams {
         .parse::<u64>()
         .unwrap(),
         max_num_blobs,
-        max_blob_size: max_num_blobs * BLOB_LEN,
+        blob_size_buffer: get_env_var_or_default("MADARA_ORCHESTRATOR_BLOB_SIZE_BUFFER", "15")
+            .parse::<usize>()
+            .unwrap(),
+        max_blob_size: max_num_blobs * BLOB_LEN
+            - get_env_var_or_default("MADARA_ORCHESTRATOR_BLOB_SIZE_BUFFER", "15").parse::<usize>().unwrap(),
```
Mohiiit, line 788: putting it in variable and using it twice is much cleaner IMO

Mohiiit (approved):
non blocker nits + it would be better to add a comment about why buffer is really required, like in our case we have encryption keys (although we have it in the config and we can calculate the exact buffer we need, but I think having buffer is still better in case we have more complicated addition to the blob size) + the maths behind the buffer (to find the right values)

heemankv (approved):

PR #994, opened 2026-03-09
feat(node): add parallel Merkle and Cairo execution optimizations

Mohiiit:
Block production can continue executing while Merkle roots are computed concurrently. The finalizer publishes blocks and durable trie checkpoints in canonical order, with explicit execution and confirmation heads so mempool, replay intake, and full-node sync can follow runahead safely.

This combined change also includes configurable Cairo execution optimizations: bounded execution read and hash caches, Cairo Native integration, executor metrics, and jemalloc. Parallel Merkle and the execution caches remain opt-in.

Recovery starts from the newest checkpoint whose root verifies against its block header. An unchanged class trie no longer forces replay from its last mutation. Durable recovery intent, atomic replay checkpoints, and coordinated reorg completion make interrupted recovery resumable while preserving independent full-node cursors during ordinary startup.

```mermaid
sequenceDiagram
    participant E as Executor
    participant W as Root workers
    participant F as Ordered finalizer
    participant D as Database
    E->>E: Execute with configured Cairo caches
    E->>W: Submit close jobs with checkpoint and state diffs
    W-->>F: Return completed roots and overlays
    loop Canonical block order
        F->>D: Persist block parts
        opt Checkpoint boundary
            F->>D: Commit trie overlays and checkpoint atomically
        end
        F->>D: Advance confirmed head
    end
```

```mermaid
flowchart TD
    A[Confirmed trie needs repair] --> B[Flush recovery intent]
    B --> C[Newest checkpoint at or below confirmed head]
    C --> D[Roll back tries ahead of candidate]
    D --> E{Checkpoint root matches header?}
    E -- No --> F{Older checkpoint within retained history?}
    F -- Yes --> K[Select previous checkpoint]
    K --> D
    F -- No --> G[Stop and preserve recovery intent]
    E -- Yes --> H[Replay remaining blocks in order]
    H --> I[Verify each root and atomically publish its checkpoint]
    I --> J[Flush repaired state and clear recovery markers]
```

Validation covers idle tries, partial rollback, process exit and reopen with WAL on and off, replay beyond the original log-retention floor, nested reorg recovery, and independent sync cursors. Failure-path regressions verify that finalizer shutdown joins dispatched work and that replay-boundary retention stays bounded without evicting active entries. The E2E matrix runs sequential and parallel Merkle separately and verifies the selected finalizer mode from runtime logs.


madara/crates/client/block_production/src/finalizer/parallel.rs
```diff
@@ -0,0 +1,341 @@
...
+            commit,
+        }
+    }
+
+    /// Runs until the input is closed and every accepted block is committed.
+    /// Root preparation may finish out of order, while commit delivery remains strictly ordered.
+    async fn run(mut self) -> Result<()> {
+        loop {
+            self.dispatch_waiting_roots();
+            self.commit_ready_in_order().await?;
```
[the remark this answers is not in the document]
Mohiiit: Fixed in 621d2c24f. The scheduler closes intake on failure, joins every dispatched root, and explicitly fails all remaining completions before returning the original error. It also checks ordered failures before dispatching more queued work. Regression cases cover both root and commit errors with a real blocking task: the finalizer join stays pending until that task exits, later work is not committed, and every completion receives an error. Both cases fail against the previous implementation and pass with the fix.

madara/crates/client/db/src/backend.rs
```diff
@@ -0,0 +1,577 @@
...
+        Some(entry.to_status())
+    }
+
+    /// Marks an existing replay boundary closed and returns its final status snapshot.
+    /// Missing boundaries remain a no-op represented by `None`.
+    pub fn replay_boundary_mark_closed(&self, block_n: u64) -> Option<ReplayBlockBoundaryStatus> {
+        let mut guard = self.replay_boundaries.lock().expect("Poisoned lock");
+        let entry = guard.get_mut(&block_n)?;
+        entry.closed = true;
+        Some(entry.to_status())
```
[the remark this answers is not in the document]
Mohiiit: Fixed in 84ca174eb. Replay-boundary retention is capped at 1,024 entries. New entries evict the oldest closed status; when all entries are active, intake returns an error instead of evicting executor state. Existing boundaries can still be replaced at capacity. Regression coverage exercises sustained closed-history churn, preserving active entries, replacement at capacity, and rejection when no closed entry can be reclaimed. The RPC documentation describes the retention behavior.

.github/workflows/task-test-end-to-end.yml
heemankv: Should we not have two separate E2E ? 
one with parallel merkle and one without ?

configs/args/config.json
heemankv: these changes are already in main, will have to rebase

heemankv (commented):
Great work on this PR! 🎯 The ordered finalizer and checkpoint-recovery design are strong. I’ve left two focused inline suggestions: make the sequential validation fail closed on divergence, and pin the RocksDB fork to an immutable revision.

madara/crates/client/db/src/rocksdb/backend.rs
```diff
@@ -0,0 +1,768 @@
...
+        if let Some(compare_snapshot) = compare_snapshot {
+            let sequential = compute_root_from_snapshot_sequential(
+                self,
+                snapshot_block,
+                compare_snapshot,
+                block_n,
+                state_diff,
+                protocol_version,
+            )?;
+            tracing::debug!(
```
heemankv, line 567: **P2 — fail closed on comparator divergence:** When sequential comparison is enabled, this only emits the root equality results at `debug` level and then unconditionally returns the parallel computation. The caller subsequently writes `parallel.state_root` and confirms the block. Since this option is documented as a correctness check, please return an error before any database writes when the contract, class, or state roots differ. If continuing is intentional, the option and documentation should instead describe this as diagnostic-only comparison.

Cargo.toml
```diff
@@ -340,7 +340,16 @@ time = "0.3"
 zip = "4.3.0"
 
 [patch.crates-io]
+cairo-native = { git = "https://github.com/karnotxyz/cairo_native", rev = "eb59cc0af7d854e60451fe6c6696ead71d2fada7" }
 jsonrpsee-core = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 jsonrpsee-types = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 rocksdb = { git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
```
heemankv: **P3 — pin the database fork to an immutable revision:** Both RocksDB patches currently follow a movable branch. `Cargo.lock` pins `75f13c78…` for locked builds today, but a dependency update can silently advance the branch. Because this fork supplies snapshot behavior used by consensus-critical state computation, please use `rev = "75f13c78fdf970b0afd2f21f52caf3317341341c"` until the required API is released upstream.

heemankv (changes requested):

.github/workflows/task-test-end-to-end.yml
heemankv, line 1: Should we not have two E2E, one with and one without parallel merkle ?

configs/args/config.json
heemankv, line 41: Already in main, needs rebasing

heemankv (changes requested):

madara/crates/client/sync/src/gateway/blocks.rs
```diff
@@ -38,13 +48,186 @@ pub fn block_with_state_update_pipeline(
...
+        .collect();
+
+    tracing::debug!(
+        "Gateway preconfirmed block sync: skip_first_n={skip_first_n} {header:?} {executed:?}, {candidates:?}"
+    );
+    match mode {
+        PreconfirmedUpdateMode::Replace => {
+            let block = PreconfirmedBlock::new_with_content(header, executed, candidates);
+            if backend.has_preconfirmed_block() {
+                backend.write_access().replace_preconfirmed(block)?;
```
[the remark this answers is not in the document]
Mohiiit: Addressed in 4356577a7bd3c33b4f356d89b314bdcd995635d0. The database replacement guard remains intact: replacing a block beneath dependent execution is unsafe.

The reachable case is a full node opening a database that still contains a sequencer's multi-block preconfirmed suffix. Before upstream polling/import begins, startup now reconciles the materialized trie to the unchanged confirmed head, discards that suffix, and flushes the cleanup. Persisted preconfirmed rows are removed even if the new process has disabled saving them. Ordinary single-block recovery still honors disable_reorg_preconfirmed.

The restart regression covers both persistence settings and both replacement-policy settings, abandoned block/state-history rows, a trie ahead of confirmation, upstream replacement, a second restart, and subsequent confirmation. Additional cases cover an empty confirmed head and preservation of a single preconfirmed block when replacements are disabled. The regression failed without startup normalization; all 309 database and sync tests now pass (two existing tests ignored).

madara/crates/client/sync/src/sync.rs
```diff
@@ -125,6 +126,7 @@ impl<P: ForwardPipeline> SyncController<P> {
     }
 
     pub async fn run(&mut self, mut ctx: mp_utils::service::ServiceContext) -> anyhow::Result<()> {
+        self.discard_inherited_execution_suffix().await?;
```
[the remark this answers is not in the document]
Mohiiit: Fixed in 0b303e524bbf1a089ebaab32dfb3ce9e403d9396. This was a lifecycle gap in my preceding startup-recovery change: awaiting Rayon did not protect that await from the outer service timeout.

L2 sync now uses the existing service_loop_with_graceful_shutdown path, already used by block production. It continues awaiting startup recovery after the grace-period warning, so the monitor cannot report sync stopped before reconciliation, cleanup, and flushing finish. The admin revert path already waits for actual service shutdown and returns an error on timeout before mutating the database.

The production diff is one runner selection plus its explanatory comment. Parallel Merkle scheduling, checkpoint recovery, the internal/external head model, and preconfirmed replacement policy are unchanged.

A regression holds an actual Rayon startup task past the service shutdown deadline and checks that service completion is withheld until that task finishes. The negative control fails under the ordinary force-cancel runner. All 48 sync and utility tests pass, with one existing ignored test.

madara/crates/client/sync/src/sync.rs
```diff
@@ -147,6 +149,33 @@ impl<P: ForwardPipeline> SyncController<P> {
         Ok(())
     }
 
+    /// Full-node sync imports upstream state and cannot resume a sequencer's dependent execution suffix.
+    /// Normalize it before polling or importing; retain ordinary single-block preconfirmed recovery.
+    async fn discard_inherited_execution_suffix(&mut self) -> anyhow::Result<()> {
+        let head = self.backend.chain_head_state();
+        let (Some(external_tip), Some(internal_tip)) = (head.external_preconfirmed_tip, head.internal_preconfirmed_tip)
```
heemankv: **P1 — Reconcile the confirmed trie even when there is no inherited execution suffix.** These early returns also skip trie reconciliation. Parallel production can leave a confirmed non-boundary head ahead of the latest durable Merkle checkpoint without any multi-block preconfirmed suffix—the existing `create_non_durable_confirmed_head` test fixture constructs this state. Opening that database as a full node returns here, while the gateway pipelines begin at `confirmed_tip + 1`; the next upstream diff is therefore applied on the older checkpoint and state-root validation fails. Please reconcile `head.confirmed_tip` unconditionally at sync startup, keep only `clear_preconfirmed` conditional, and add a confirmed-only handoff regression.

madara/crates/client/db/src/rocksdb/storage_write.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn flush(&self) -> Result<()> {
+        tracing::debug!("Flushing");
+        self.inner.flush().context("Flushing RocksDB database")?;
+        self.backup.backup_if_enabled(&self.inner).context("Backing up RocksDB database")
+    }
+
+    fn on_new_confirmed_head(&self, block_n: u64) -> Result<()> {
+        tracing::debug!("on_new_confirmed_head block_n={block_n}");
+        let started_at = Instant::now();
+        self.snapshots.set_new_head(block_n);
```
heemankv, line 304: **P1 — Do not label a non-durable trie snapshot as this confirmed block.** Parallel Merkle intentionally skips the overlay flush for non-boundary blocks, but this still publishes a generic head snapshot under every confirmed height. `starknet_getStorageProof` resolves `get_transactional_state(block_n)` through that snapshot; for the current head, Bonsai applies no revisions and reads the previous checkpoint's trie while the response uses the newer block hash and state data. This can return a proof that is invalid for the requested block during normal sequencing. Please preserve proof-capable state for each confirmed height or explicitly reject non-durable proof heights, with a `Latest` proof test after a non-boundary block.

madara/crates/primitives/utils/src/service.rs
```diff
@@ -1393,3 +1419,94 @@ impl ServiceMonitor {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
heemankv, line 1423: **P2 — Keep the default-feature unit-test build compiling.** This module calls `ServiceContext::new_for_testing()`, but that constructor is gated by `feature = "testing"`. Consequently, `cargo test -p mp-utils --lib` fails with `E0599` at all three new tests; CI does not catch it because the test archive enables that feature. Please gate this module with `#[cfg(all(test, feature = "testing"))]` or use a constructor available under default features.

heemankv (changes requested):
Great work on this PR! 🎯 The latest updates resolve the earlier dependency-pin, comparator-documentation, E2E-matrix, and shutdown-lifecycle concerns. I excluded the transaction-replay stack from this pass. I found two remaining live-path correctness issues and one default-feature test-build issue inline.

madara/crates/primitives/utils/src/service.rs
Mohiiit: Fixed in fc122a46a9f1e4f59ab8bd46a7918ff6f49eadc6. The three shutdown regressions now use ServiceContext::new(), so they remain enabled under default features. Their cancellation-token and join behavior does not require the feature-gated service-mask helper. All seven mp-utils library tests pass both with default features and with testing enabled; formatting and Clippy pass.

madara/crates/client/sync/src/sync.rs
Mohiiit: Fixed for full-node startup in d3f82e3fe21780a5e3302a4ab857208307236079. Trie reconciliation now also runs when the applied-trie cursor is behind confirmation, independently of whether an inherited multi-block suffix exists. Only that suffix is cleared; a single preconfirmed block retains its replacement policy. Sequencer startup is unchanged.

The reconciliation is intentionally not unconditional: ordinary full-node trie application can run ahead of other import stages, and that progress must be preserved. A regression covers this case. The confirmed-only handoff and single-preconfirmed handoff regressions both failed before the fix and now pass. They reopen the producer database, run real sync startup, reopen again to verify the flushed repair, and apply the next empty upstream block with state-root verification enabled.

The sync and utility suites pass 51 tests, with one existing ignored test. Full workspace and test Clippy, Rust/Prettier/Taplo formatting, and markdown lint pass.

madara/crates/client/db/src/rocksdb/storage_write.rs
Mohiiit: this madara won't be serving the storage proof and hence this is not fixed!

heemankv (approved):

PR #995, opened 2026-03-11
fix: populate trace state diff fields per Starknet spec

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

`starknet_traceTransaction` and `starknet_traceBlockTransactions` return state diffs with `deployed_contracts`, `replaced_classes`, `declared_classes`, and `deprecated_declared_classes` always hardcoded to empty arrays. This does not match the Starknet RPC spec.

Resolves: #NA

## What is the new behavior?

- Properly distinguish deployed contracts vs replaced classes by checking initial state reads (if prior class hash was zero, it's a deployment; otherwise it's a replacement)
- Populate `declared_classes` from `class_hash_to_compiled_class_hash` in the state diff
- Populate `deprecated_declared_classes` for Cairo 0 class declarations (DeclareV0/V1)
- Enable blockifier's `reexecution` feature to access `get_initial_reads()` on transactional state
- Fix applies to all RPC versions (v0.7.1, v0.8.1, v0.9.0) and all trace methods (`traceTransaction`, `traceBlockTransactions`, `simulateTransactions`)
- Address PR #993 review comments: use rstest for parametrized tests, extract duplicate env var parsing, add documentation for blob size buffer

No. The trace responses now include data that was previously omitted. No existing fields are removed or modified.

## Other information

- The approach matches pathfinder's implementation (checking initial class hash == zero for deployment detection)
- Changes only affect the RPC re-execution path, zero impact on sequencer block production performance
- `get_initial_reads()` returns data already tracked by blockifier's `StateCache` internally, so no additional state tracking overhead

### DIFF
--- a/madara/crates/client/exec/src/execution.rs
+++ b/madara/crates/client/exec/src/execution.rs
@@ -88,7 +88,7 @@ impl<D: MadaraStorageRead> ExecutionContext<D> {
                     .state_maps
                     .class_hashes
                     .keys()
-                    .filter(|addr| initial_reads.class_hashes.get(addr).map_or(true, |ch| *ch == ClassHash::default()))
+                    .filter(|addr| initial_reads.class_hashes.get(addr).is_none_or(|ch| *ch == ClassHash::default()))
                     .map(|addr| addr.to_felt())
                     .collect();
 
--- a/madara/crates/client/exec/src/execution.rs
+++ b/madara/crates/client/exec/src/execution.rs
@@ -8,11 +8,12 @@ use blockifier::transaction::objects::{HasRelatedFeeType, TransactionInfoCreator
 use blockifier::transaction::transaction_execution::Transaction;
 use blockifier::transaction::transactions::ExecutableTransaction;
 use mc_db::MadaraStorageRead;
-use mp_convert::ToFelt;
+use mp_convert::{Felt, ToFelt};
 use mp_receipt::RevertErrorExt;
 use starknet_api::block::FeeType;
 use starknet_api::contract_class::ContractClass;
 use starknet_api::core::{ClassHash, ContractAddress, Nonce};
+use std::collections::HashSet;
 use starknet_api::executable_transaction::{AccountTransaction as ApiAccountTransaction, TransactionType};
 use starknet_api::transaction::fields::{GasVectorComputationMode, Tip};
 use starknet_api::transaction::{TransactionHash, TransactionVersion};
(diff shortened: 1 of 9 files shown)


madara/crates/client/exec/src/execution.rs
```diff
@@ -75,8 +76,32 @@ impl<D: MadaraStorageRead> ExecutionContext<D> {
...
+                    .get_initial_reads()
+                    .map_err(TransactionExecutionError::StateError)
+                    .map_err(make_reexec_error)?;
+                let deployed_contracts: HashSet<Felt> = state_diff
+                    .state_maps
+                    .class_hashes
+                    .keys()
+                    .filter(|addr| initial_reads.class_hashes.get(addr).is_none_or(|ch| *ch == ClassHash::default()))
+                    .map(|addr| addr.to_felt())
+                    .collect();
```
heemankv, line 93: Nitpick: Consider adding a unit test that exercises the deployment-vs-replacement classification. A test with a `CommitmentStateDiff` containing addresses both in and out of the `deployed_contracts_set` would give confidence this logic is correct and catch regressions. The `to_state_diff` function in `trace.rs` would also benefit from a test showing the split between `deployed_contracts` and `replaced_classes`.

madara/crates/client/exec/src/trace.rs
```diff
@@ -34,13 +34,13 @@ pub struct FunctionInvocation {
 #[derive(thiserror::Error, Debug)]
 pub enum ConvertCallInfoToExecuteInvocationError {
     #[error(transparent)]
-    GetFunctionInvocation(#[from] TryFuntionInvocationFromCallInfoError),
+    GetFunctionInvocation(#[from] TryFunctionInvocationFromCallInfoError),
```
heemankv, line 37: Nitpick: Good typo fix (`TryFuntionInvocationFromCallInfoError` → `TryFunctionInvocationFromCallInfoError`). Since this is a public type, just confirming no external crates reference the old name — grep shows no other references, so this is safe.

prkpndy: yeah, i have updated the typo in all the places it was being used

madara/crates/client/exec/src/execution.rs
prkpndy: added

heemankv: Nice tests! The deployed-vs-replaced classification test and the deprecated/declared class tests cover the key logic paths well. 👍

madara/crates/client/exec/src/trace.rs
heemankv: Sounds good, thanks for confirming!

heemankv (approved):
LGTM! All feedback addressed — tests for the state diff logic look good, CI is green. Nice fix for the trace spec compliance.

madara/crates/client/exec/src/trace.rs
```diff
@@ -552,7 +555,34 @@ fn resources_mapping(
...
+            deployed_contracts
+                .push(mp_rpc::v0_7_1::DeployedContractItem { address: address_felt, class_hash: class_hash_felt });
+        } else {
+            replaced_classes
+                .push(mp_rpc::v0_7_1::ReplacedClass { contract_address: address_felt, class_hash: class_hash_felt });
+        }
+    }
+
+    let declared_classes = commitment_state_diff
+        .class_hash_to_compiled_class_hash
```
Mohiiit, line 586: Nit: `class_hash_to_compiled_class_hash` is not declarations-only here. The block production path already treats this map as "declared + SNIP-34 migrated" and filters migrations out before populating `declared_classes`. Mapping it 1:1 in the trace path means transactions that trigger a migration can still surface that migrated class under `declared_classes` in traces. Since this only affects trace output it is not critical, but it still looks like a spec mismatch.

Mohiiit (commented):
One small follow-up on the trace state-diff conversion.

Mohiiit (approved):

PR #996, opened 2026-03-11
fix: historical L1→L2 messages missed on node restart

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

`messages_to_l2_stream()` in the Ethereum settlement client uses alloy's `.watch()` which calls `eth_newFilter` + `eth_getFilterChanges`. This is **forward-only** from the moment the filter is created on the RPC node. When Madara restarts after downtime, any L1→L2 messages emitted during that period are silently lost because the filter only tracks events *after* its creation — historical events between `from_l1_block_n` and the current tip are never fetched.

Resolves: #NA

## What is the new behavior?

- **Historical catchup via `eth_getLogs`**: Before starting the live watcher, the code now fetches all historical events from `from_l1_block_n` to the current finalized block using paginated `eth_getLogs` queries (10k block chunks). This ensures no events are missed on restart.
- **Ordering to prevent gaps**: The live watcher (`.watch()`) is created *before* querying the finalized block number. This guarantees the watcher's start point ≥ the historical query's upper bound, ensuring overlap (never a gap). Duplicates from the overlap are handled by the existing nonce-based dedup in `check_message_to_l2_validity`.
- **Finalized block as upper bound**: The historical query uses the `finalized` block tag (not `latest`) to stay consistent with the message sync finality guarantees and avoid processing events from potentially reorgable blocks.
- **E2e regression test**: Added `e2e_test_historical_messages_catchup` that fires a `LogMessageToL2` event *before* starting the sync worker, then verifies the historical event is picked up. This test would fail with the old `.watch()`-only implementation.
- **Anvil finality config**: Test Anvil instances now use `--slots-in-an-epoch 1` so the finalized tip advances quickly (finalized = latest - 2 instead of default latest - 64).

No

## Other information

The consumer side (`process_finalized_events` in `messaging.rs`) is unchanged — it continues to enforce its own finality threshold and nonce dedup, making the historical/live stream transition transparent.

### DIFF
--- a/madara/crates/client/settlement_client/src/eth/mod.rs
+++ b/madara/crates/client/settlement_client/src/eth/mod.rs
@@ -401,7 +401,9 @@ impl SettlementLayerProvider for EthereumClient {
         let live_stream = EthereumEventStream::new(event_poller);
 
         // Step 2: Get finalized block number AFTER creating the watcher.
-        // This is the upper bound for the historical query.
+        // This is the upper bound for the historical query. We use `finalized` (not `latest`)
+        // to stay consistent with the rest of the message sync logic — we only process events
+        // that are finalized on L1 to avoid issues with chain reorgs.
         let finalized_block = self
             .provider
             .get_block_by_number(BlockNumberOrTag::Finalized, alloy::network::primitives::BlockTransactionsKind::Hashes)
@@ -700,7 +702,12 @@ mod l1_messaging_tests {
     #[fixture]
     async fn setup_test_env() -> TestRunner {
         // Start Anvil instance
-        let anvil = Anvil::new().block_time(1).chain_id(1337).try_spawn().expect("failed to spawn anvil instance");
+        let anvil = Anvil::new()
+            .block_time(1)
+            .chain_id(1337)
+            .args(["--slots-in-an-epoch", "1"])
+            .try_spawn()
+            .expect("failed to spawn anvil instance");
         let chain_config = Arc::new(ChainConfig::madara_test());
         // Initialize database service
         let db = MadaraBackend::open_for_testing(chain_config.clone());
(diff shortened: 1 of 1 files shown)


madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -336,19 +380,57 @@ impl SettlementLayerProvider for EthereumClient {
...
+        let live_stream = EthereumEventStream::new(event_poller);
 
-        Ok(EthereumEventStream::new(event_stream).boxed())
+        // Step 2: Get finalized block number AFTER creating the watcher.
+        // This is the upper bound for the historical query. We use `finalized` (not `latest`)
+        // to stay consistent with the rest of the message sync logic — we only process events
+        // that are finalized on L1 to avoid issues with chain reorgs.
+        let finalized_block = self
+            .provider
+            .get_block_by_number(BlockNumberOrTag::Finalized, alloy::network::primitives::BlockTransactionsKind::Hashes)
```
heemankv: Nitpick: `alloy::network::primitives::BlockTransactionsKind::Hashes` could be imported at the top of the file alongside the other `alloy::` imports for readability.

madara/crates/client/settlement_client/src/eth/mod.rs
```diff
@@ -336,19 +380,57 @@ impl SettlementLayerProvider for EthereumClient {
...
+        // historical query and buffering any new events.
+        let event_poller = self
+            .l1_core_contract
+            .event_filter::<LogMessageToL2>()
+            .from_block(from_l1_block_n)
+            .to_block(BlockNumberOrTag::Finalized)
+            .watch()
+            .await
+            .map_err(|e| -> SettlementClientError {
+                EthereumClientError::ArchiveRequired(format!("Could not create event watcher: {e}")).into()
```
heemankv, line 400: Nitpick: `ArchiveRequired` is slightly misleading for a watcher creation failure (could be a network error, rate limit, etc.), but the old code used the same variant for the same `.watch()` call, so this is consistent. No action needed unless you want to add a more specific variant later.

heemankv (commented):
Great work on this PR! 🎯 The historical catchup approach is clean and well-reasoned — creating the watcher before querying the finalized block is the right ordering to guarantee no gaps. The regression test is a solid addition that directly validates the fix. Two minor nitpicks in the inline comments.

heemankv (approved):
LGTM! Import nitpick addressed, CI was green on the prior commit and this change is trivial. Clean fix for the historical L1→L2 message catchup — the ordering guarantee and regression test are solid.

Mohiiit (approved):
is this behaviour same for starknet l1 message tho? do we need to make the changes over there as well?

as per the current scope, things looks alright!

PR #998, opened 2026-03-18
feat: file-based secret loading & settlement CLI fix

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Feature

1. **Secrets are only configurable via environment variables**, which leak through `/proc/pid/environ`, crash dumps, `kubectl describe pod`, and logging frameworks. No support for file-based secret injection (K8s CSI Secrets Store Driver).

2. **Settlement CLI `requires_all` group is broken**: if any Ethereum settlement env var is set in the pod environment, clap demands all Ethereum settlement fields — even when running with `--settle-on-starknet`. This causes startup failures when both sets of env vars coexist.

Resolves: #NA

## What is the new behavior?

- **File-based secret loading**: For 8 sensitive env vars, the orchestrator now checks for a corresponding `_FILE` suffix variant (e.g. `MADARA_ORCHESTRATOR_ETHEREUM_PRIVATE_KEY_FILE=/mnt/secrets/eth-key`). If set, the secret is read from the file at that path. `_FILE` takes precedence over the direct env var. Fully backward compatible.

  Supported `_FILE` variants:
  - `MADARA_ORCHESTRATOR_ETHEREUM_PRIVATE_KEY_FILE`
  - `MADARA_ORCHESTRATOR_STARKNET_PRIVATE_KEY_FILE`
  - `MADARA_ORCHESTRATOR_ATLANTIC_API_KEY_FILE`
  - `MADARA_ORCHESTRATOR_MONGODB_CONNECTION_URL_FILE`
  - `MADARA_ORCHESTRATOR_ETHEREUM_SETTLEMENT_RPC_URL_FILE`
  - `MADARA_ORCHESTRATOR_STARKNET_SETTLEMENT_RPC_URL_FILE`
  - `MADARA_ORCHESTRATOR_ETHEREUM_DA_RPC_URL_FILE`
  - `MADARA_ORCHESTRATOR_STARKNET_DA_RPC_URL_FILE`

- **Settlement CLI fix**: Removed `requires_all` from `EthereumSettlementCliArgs` and `StarknetSettlementCliArgs`. Validation is already handled in `TryFrom<RunCmd>` for `SettlementConfig` (`src/types/params/settlement.rs`), so the clap-level constraint was redundant and caused false errors.

No. All existing env var configurations continue to work unchanged. The `_FILE` variants are additive. The `requires_all` removal is purely a bug fix — the actual validation logic is unchanged.

## Other information

- `resolve_secret_from_file()` utility added in `crates/utils/src/env_utils.rs` with 4 unit tests
- Follows the `_FILE` suffix convention used by official Docker images (MySQL, PostgreSQL, Redis)
- Designed for K8s CSI Secrets Store Driver: secrets mounted as tmpfs files never enter etcd, env vars, or process environment

### DIFF
--- a/orchestrator/crates/utils/src/env_utils.rs
+++ b/orchestrator/crates/utils/src/env_utils.rs
@@ -1,4 +1,5 @@
 use std::env::VarError;
+use std::path::Path;
 
 pub fn get_env_var(key: &str) -> Result<String, VarError> {
     std::env::var(key)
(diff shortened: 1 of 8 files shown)


orchestrator/crates/utils/src/env_utils.rs
```diff
@@ -25,3 +26,95 @@ pub fn get_env_var_optional(key: &str) -> Result<Option<String>, VarError> {
...
+    use std::io::Write;
+
+    #[test]
+    fn test_resolve_secret_from_file_reads_file() {
+        let mut tmp = tempfile::NamedTempFile::new().unwrap();
+        write!(tmp, "  my-secret-value\n  ").unwrap();
+
+        let env_var = "TEST_RESOLVE_SECRET_READS";
+        let file_env_var = format!("{}_FILE", env_var);
+        std::env::set_var(&file_env_var, tmp.path().to_str().unwrap());
```
heemankv, line 78: Nitpick: `std::env::set_var` / `remove_var` are not thread-safe (process-global state). Tests use unique env var names which avoids conflicts, but if these tests ever run in parallel within the same binary, you could get flaky behavior. Consider using `serial_test` crate or a test mutex for extra safety. Not blocking — this is a common pattern and the unique names make collisions unlikely.

orchestrator/src/types/params/da.rs
```diff
@@ -19,18 +21,42 @@ impl TryFrom<RunCmd> for DAConfig {
...
+                    Some(url_str) => Url::parse(&url_str).map_err(|e| {
+                        OrchestratorError::SetupCommandError(format!(
+                            "Invalid Ethereum DA RPC URL from secret file: {}",
+                            e
+                        ))
+                    })?,
+                    None => run_cmd.ethereum_da_args.ethereum_da_rpc_url.ok_or_else(|| {
+                        OrchestratorError::SetupCommandError("Ethereum RPC URL is missing".to_string())
+                    })?,
+                };
```
heemankv, line 38: Nitpick: The `match resolve_secret_from_file(...) { Some(url_str) => Url::parse(...), None => fallback }` pattern appears 4 times across `da.rs` and `settlement.rs`. A small helper like `resolve_url_secret(env_var, fallback: Option<Url>) -> Result<Url, OrchestratorError>` could reduce this duplication. Fine as-is for 4 occurrences though — just flagging for awareness.

heemankv (approved):
Great work on this PR! 🎯 The file-based secret loading follows the well-established `_FILE` suffix convention from official Docker images, and the settlement CLI fix correctly delegates validation to `TryFrom`. Clean implementation with good test coverage. A couple minor nitpicks in the inline comments.

orchestrator/crates/utils/src/env_utils.rs
prkpndy: each test uses a unique env var name (TEST_RESOLVE_SECRET_READS, TEST_RESOLVE_SECRET_UNSET, etc.) so there's no actual conflict. we can make the test run serially, but that would add an extra dependency. can take lite for now ig

orchestrator/src/types/params/da.rs
prkpndy: hmm, would reduce ~10 lines of code. let's leave it for now ig. if the pattern grows, we'll extract it 😅

PR #1003, opened 2026-03-25
feat: add --unsafe-no-l1-handler-tx-creation-from-message flag

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

When L1 sync is enabled, the messaging worker writes L1→L2 messages to both the metadata tables (for `starknet_getMessagesStatus`) and the pending table (for block production to consume). There is no way to have L1 sync populate metadata without also having messages automatically queued for block production.

This is problematic when an external orchestrator (e.g. a transaction replay service) submits L1 handler transactions exclusively via the admin RPC `addL1HandlerMessage` — the L1 sync messaging worker creates duplicate entries in the pending table that compete with the admin-submitted transactions.

Resolves: #NA

## What is the new behavior?

- Adds `--unsafe-no-l1-handler-tx-creation-from-message` CLI flag (env: `MADARA_UNSAFE_NO_L1_HANDLER_TX_CREATION_FROM_MESSAGE`)
- When enabled:
  - L1 sync messaging worker runs normally and persists all metadata (nonce → L1 tx hash, seen marker, nonce → L1 block number) but does NOT write messages to the pending table
  - L1 handler transactions can only enter blocks via the admin RPC (since we don't populate the pending column in rocks DB from where the block production creates L1 handler transactions)
  - State update and gas price workers continue running normally
- The flag cannot be combined with `--no-l1-sync` (enforced via clap `conflicts_with`)
- A prominent warning is logged when the flag is enabled
- The admin RPC `addL1HandlerMessage` is reverted to its original `L1HandlerTransactionWithFee` parameter — L1 origin metadata is now populated exclusively by the L1 sync messaging worker, with backfill support for messages executed before the corresponding L1 event is discovered

No. The new flag is opt-in and defaults to `false`. Existing behavior is unchanged.

The admin RPC `addL1HandlerMessage` parameter type is reverted to `L1HandlerTransactionWithFee` (from the short-lived `L1HandlerMessageWithOrigin`), but since that change was never released, this is not a breaking change.

## Other information

When `--unsafe-no-l1-handler-tx-creation-from-message` is enabled and an L1 handler transaction is submitted via admin RPC before L1 sync discovers the corresponding L1 event, the nonce → L2 tx hash mapping is written at block commit time. When L1 sync later catches up, `process_finalized_events` detects the existing mapping and backfills the L1 origin metadata (L1 tx hash, seen marker, L1 block number) without writing to the pending table. This ensures `starknet_getMessagesStatus` eventually returns complete data.

### DIFF
--- a/configs/args/config.json
+++ b/configs/args/config.json
@@ -73,6 +73,7 @@
       "nanos": 0
     },
     "unsafe_skip_l1_message_consumed_check": false,
+    "unsafe_no_l1_handler_tx_creation_from_message": false,
     "settlement_layer": "Eth"
   },
   "analytics_params": {
--- a/configs/args/config.json
+++ b/configs/args/config.json
@@ -73,7 +73,7 @@
       "nanos": 0
     },
     "unsafe_skip_l1_message_consumed_check": false,
-    "unsafe_no_l1_handler_tx_creation_from_message": false,
+    "unsafe_l1_handler_metadata_only": false,
     "settlement_layer": "Eth"
   },
   "analytics_params": {
--- a/madara/crates/client/settlement_client/src/messaging.rs
+++ b/madara/crates/client/settlement_client/src/messaging.rs
@@ -94,6 +94,7 @@ pub async fn sync(
     notify_consumer: Arc<Notify>,
     mut ctx: ServiceContext,
     unsafe_skip_l1_message_consumed_check: bool,
+    metadata_only: bool,
 ) -> Result<(), SettlementClientError> {
     // sync inner is cancellation safe.
     ctx.run_until_cancelled(sync_inner(
@@ -101,6 +102,7 @@ pub async fn sync(
         backend,
         notify_consumer,
         unsafe_skip_l1_message_consumed_check,
+        metadata_only,
     ))
     .await
     .transpose()?;
@@ -112,6 +114,7 @@ async fn sync_inner(
     backend: Arc<MadaraBackend>,
     notify_consumer: Arc<Notify>,
     unsafe_skip_l1_message_consumed_check: bool,
+    metadata_only: bool,
 ) -> Result<(), SettlementClientError> {
     // Note: It's fine to reprocess events - duplicates are filtered during block production.
 
@@ -129,6 +132,7 @@ async fn sync_inner(
             replay_max_duration,
             finality_blocks,
             unsafe_skip_l1_message_consumed_check,
+            metadata_only,
         )
         .await
         {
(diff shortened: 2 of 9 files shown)


prkpndy (commented):
LGTM

madara/crates/client/settlement_client/src/messaging.rs
```diff
@@ -94,13 +94,15 @@ pub async fn sync(
     notify_consumer: Arc<Notify>,
     mut ctx: ServiceContext,
     unsafe_skip_l1_message_consumed_check: bool,
+    metadata_only: bool,
```
heemankv, line 97: Nitpick: The parameter is called `unsafe_no_l1_handler_tx_creation_from_message` at the CLI/config level but `metadata_only` here. Consider adding a brief comment like `// aka --unsafe-no-l1-handler-tx-creation-from-message` to make the mapping obvious when reading this code in isolation.

heemankv (approved):
Great work on this PR! 🎯 The metadata-only approach is clean and well-tested. The separation of concerns — L1 sync handles metadata, admin RPC handles tx submission — is well thought out. The backfill logic for messages executed before L1 sync catches up is a nice touch.

**Nitpick:** The PR description mentions "Block production receives an empty L1 message consumer stream (`L1SyncDisabledClient`)" but this was removed in the refactoring commits (`3052767de`, `096c32d14`). The actual mechanism is simpler: `metadata_only=true` skips `write_pending_message_to_l2`, so the consumer naturally finds nothing in the pending table. Consider updating the description to match the final implementation.

configs/args/config.json
```diff
@@ -73,6 +73,7 @@
       "nanos": 0
     },
     "unsafe_skip_l1_message_consumed_check": false,
+    "unsafe_no_l1_handler_tx_creation_from_message": false,
```
Mohiiit: Would `unsafe_l1_handler_metadata_only` be a clearer name here? It reads closer to the operator-visible behavior than `unsafe_no_l1_handler_tx_creation_from_message`, which feels more implementation-shaped.

Mohiiit (commented):
I have one naming suggestion in the inline comment.

Mohiiit (approved):

madara/crates/client/settlement_client/src/messaging.rs
prkpndy: updated the cli flag to `unsafe_l1_handler_metadata_only`. no confusion now

configs/args/config.json
prkpndy: done

PR #1004, opened 2026-03-25
feat: 0.14.2 bump + 0.10.2 RPC

Mohiiit:
## Summary
- add direct Madara JSON-RPC `0.10.2` support and register it as the latest user RPC version
- add the `0.10.2` RPC primitive/schema layer and REST method updates for storage, state update, events, messages status, and trace plumbing
- align transaction hashing and add sequencer/pathfinder-backed regression coverage for `0.10.2`

## Validation
- cargo fmt --all
- cargo clippy --workspace --all-targets --locked -- -D warnings
- cargo test -p mp-transactions compute_hash --lib
- cargo test -p mp-transactions compute_hash::tests::test_compute_hash_matches_sequencer_vectors --lib


orchestrator/src/worker/utils/fact_topology.rs
```diff
@@ -65,7 +65,7 @@ pub fn get_fact_topology(cairo_pie: &CairoPie, output_size: usize) -> Result<Fac
 
 /// Returns the sizes of the program output pages, given the pages dictionary that appears
 /// in the additional attributes of the output builtin.
-pub fn get_page_sizes(pages: &HashMap<usize, PublicMemoryPage>, output_size: usize) -> Result<Vec<usize>, FactError> {
+pub fn get_page_sizes(pages: &BTreeMap<usize, PublicMemoryPage>, output_size: usize) -> Result<Vec<usize>, FactError> {
```
prkpndy, line 68: why did we change this?

madara/crates/primitives/transactions/src/lib.rs
```diff
@@ -1106,6 +1108,7 @@ mod tests {
             account_deployment_data: vec![Felt::from(10), Felt::from(11)],
             nonce_data_availability_mode: DataAvailabilityMode::L1,
             fee_data_availability_mode: DataAvailabilityMode::L2,
+            proof_facts: None,
```
prkpndy, line 1122: should we have some value here? assuming this is being used for testing purposes!

madara/crates/primitives/rpc/src/lib.rs
```diff
@@ -2,6 +2,7 @@ mod custom_serde;
 
 pub mod admin;
 pub mod v0_10_0;
+pub mod v0_10_2;
```
prkpndy, line 5: btw, don't we have v0.10.1?

heemankv (commented):
(review withdrawn)

madara/crates/client/block_production/src/executor/thread.rs
```diff
@@ -136,23 +135,45 @@ impl ExecutorThread {
...
-                }
-                _ = OptionFuture::from(deadline.map(tokio::time::sleep_until)) => {
-                    tracing::debug!("Waiting for batch timed out.");
-                    WaitTxBatchOutcome::Batch(Default::default())
-                }
-                el = self.incoming_batches.recv() => match el {
-                    Some(el) => {
-                        tracing::debug!("Got new batch with {} transactions.", el.len());
-                        WaitTxBatchOutcome::Batch(el)
+            match deadline {
```
prkpndy, line 139: Let's mention in the function doc that passing `None` => no timeout, wait indefinitely. was a little confusing for me in the first read!

madara/crates/client/block_production/src/lib.rs
```diff
@@ -761,12 +761,31 @@ impl BlockProductionTask {
 
                 // Check if pre-confirmed block exists (it shouldn't at this point)
                 // Create new preconfirmed block
+                let preconfirmed_header = exec_ctx.into_header();
+                tracing::info!(
```
prkpndy: should we make this debug?

madara/crates/client/block_production/src/lib.rs
```diff
@@ -761,12 +761,31 @@ impl BlockProductionTask {
...
+
                 let backend = self.backend.clone();
                 global_spawn_rayon_task(move || {
-                    backend.write_access().new_preconfirmed(PreconfirmedBlock::new(exec_ctx.into_header()))
+                    backend.write_access().new_preconfirmed(PreconfirmedBlock::new(preconfirmed_header))
                 })
                 .await?;
 
+                if let Some(preconfirmed) = self.backend.chain_tip.borrow().as_preconfirmed() {
+                    tracing::info!(
```
prkpndy: same

madara/crates/client/db/src/rocksdb/blocks.rs
```diff
@@ -27,6 +32,114 @@ struct StateDiffV8 {
...
+    nonce: Felt,
+    resource_bounds: ResourceBoundsMapping,
+    tip: u64,
+    paymaster_data: Vec<Felt>,
+    account_deployment_data: Vec<Felt>,
+    nonce_data_availability_mode: DataAvailabilityMode,
+    fee_data_availability_mode: DataAvailabilityMode,
+}
+
+impl From<LegacyTransactionWithReceipt> for TransactionWithReceipt {
```
prkpndy, line 83: Let's add tests for all the From/Into impl from Legacy -> Current, and vice-versa.
Also, let's test the normal/current-schema path of deserialize_transaction_with_receipt and where a current invoke-v3 row with proof_facts: Some(...) is decoded.

madara/crates/client/db/src/test_utils.rs
```diff
@@ -68,6 +68,7 @@ pub fn invoke_v3(sender: Felt, nonce: Felt) -> InvokeTransactionV3 {
         account_deployment_data: vec![Felt::from_hex_unchecked("0xbeef")],
         nonce_data_availability_mode: DataAvailabilityMode::L1,
         fee_data_availability_mode: DataAvailabilityMode::L1,
+        proof_facts: None,
```
prkpndy, line 71: should we add proof_facts here to test with that?

madara/crates/client/db/src/view/mod.rs
```diff
@@ -62,37 +62,52 @@ impl<D: MadaraStorageRead> MadaraBackend<D> {
...
-                            &quote,
-                            parent_block_info.header.gas_prices.strk_l2_gas_price,
-                            parent_block_info.total_l2_gas_used,
-                        )?
+                let (block_timestamp, gas_prices) =
+                    if let Some(custom_header) = self.get_custom_header(parent_block_number + 1) {
+                        // Convert Unix timestamp (seconds since Jan 1, 1970) to SystemTime
+                        let block_timestamp = UNIX_EPOCH + Duration::from_secs(custom_header.timestamp);
+                        let gas_prices = custom_header.gas_prices;
+                        tracing::info!(
```
prkpndy: should we make this debug?

madara/crates/client/mempool/src/inner/tests/mod.rs
```diff
@@ -64,6 +64,7 @@ impl From<TestTx> for ValidatedTransaction {
                         account_deployment_data: Default::default(),
                         nonce_data_availability_mode: Default::default(),
                         fee_data_availability_mode: Default::default(),
+                        proof_facts: None,
```
prkpndy: nit: let's use default here to be consistent

madara/crates/client/rpc/src/types.rs
```diff
@@ -74,4 +110,32 @@ mod tests {
...
+            event_index_in_block: 0,
+            in_preconfirmed: false,
+        }
+    }
+
+    #[rstest]
+    #[case(&[0, 0, 0], 2, ContinuationToken { block_number: 0, event_n: 0 }, Some(ContinuationToken { block_number: 0, event_n: 2 }))]
+    #[case(&[0, 0, 4], 2, ContinuationToken { block_number: 0, event_n: 0 }, Some(ContinuationToken { block_number: 4, event_n: 0 }))]
+    #[case(&[4, 6, 6], 2, ContinuationToken { block_number: 0, event_n: 2 }, Some(ContinuationToken { block_number: 6, event_n: 1 }))]
+    fn continuation_token_from_page_tracks_block_boundaries(
```
prkpndy, line 134: continuation_token_from_page has useful coverage for the main pagination cases, but it is missing a few edge cases. In particular, page_size = 0 currently appears to produce a non-advancing continuation token (previous_token.event_n + 0) while build_events_chunk returns an empty page, which can make clients loop forever. Can we either reject chunk_size = 0 during request validation, or add an explicit test proving zero-sized pages make forward progress? I’d also add a simple direct test for events_infos.len() <= page_size returning None, since that is an important branch of this helper.

madara/crates/client/rpc/src/versions/user/common.rs
```diff
@@ -0,0 +1,66 @@
+use crate::StarknetRpcApiError;
+use mp_convert::Felt;
+
+pub(crate) fn convert_storage_keys_for_v0_8_1(
```
prkpndy, line 18: This helper fixes the previous `unwrap_or(Felt::ZERO)` behavior, which is good, but the tests only cover a single valid key and a single invalid key. Can we add table coverage for `None`, empty input, multiple contract entries, multiple storage keys with order preserved, and mixed valid/invalid keys? I’d also like one `getStorageProof`-level test for an invalid storage key so we lock the RPC error shape. Also, since an invalid storage key is client input, returning `ErrUnexpectedError`/code 63 feels misleading; can we return an invalid-params style error or add a dedicated validation error instead?

madara/crates/primitives/transactions/src/to_blockifier.rs
```diff
@@ -156,6 +156,17 @@ impl IntoStarknetApiExt for mp_rpc::v0_8_1::BroadcastedTxn {
     }
 }
 
+impl IntoStarknetApiExt for mp_rpc::v0_10_2::BroadcastedTxn {
+    fn into_starknet_api(
+        self,
+        chain_id: Felt,
+        starknet_version: StarknetVersion,
+    ) -> Result<(ApiAccountTransaction, Option<ConvertedClass>), ToBlockifierError> {
+        let tx: mp_rpc::v0_8_1::BroadcastedTxn = self.into();
```
prkpndy: This downcast drops v0.10.2 invoke-v3 `proof` / `proof_facts` before blockifier conversion. `simulateTransactions` and `estimateFee` use this path, so a v0.10.2 client can get results for a different transaction hash/context than the one it submitted.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -272,7 +272,8 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
             .into());
         }
 
-        self.backend.set_custom_header(custom_block_headers);
+        self.backend.set_custom_header(custom_block_headers).map_err(StarknetRpcApiError::from)?;
```
prkpndy, line 275: This returns success after staging the custom header, but if that block is already live as preconfirmed the live header and executor context are not updated; block close later only consumes the staged header for expected-hash comparison.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/read/get_events.rs
```diff
@@ -0,0 +1,307 @@
...
+    )?;
+
+    Ok(build_events_chunk(events_infos, chunk_size, &requested_continuation_token))
+}
+
+fn validate_filter_limits(chunk_size: usize, keys: Option<&[Vec<Felt>]>) -> StarknetRpcResult<usize> {
+    if keys.map(|patterns| patterns.iter().map(|pattern| pattern.len()).sum::<usize>()).unwrap_or(0) > MAX_EVENTS_KEYS {
+        return Err(StarknetRpcApiError::TooManyKeysInFilter);
+    }
+    if chunk_size > MAX_EVENTS_CHUNK_SIZE {
```
prkpndy: `chunk_size = 0` is accepted here. With zero, the collector can fetch one event, `build_events_chunk` truncates the response to an empty page, and `continuation_token_from_page` can return `previous_token.event_n + 0`. That can produce a non-advancing continuation token.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/starknet_unsubscribe.rs
```diff
@@ -0,0 +1,9 @@
+use crate::errors::StarknetRpcResult;
+use crate::Starknet;
+
+/// Unsubscribes from a WebSocket subscription by ID.
+pub async fn starknet_unsubscribe(_starknet: &Starknet, _subscription_id: u64) -> StarknetRpcResult<bool> {
```
prkpndy, line 5: This implementation ignores the subscription id and always returns `Ok(true)`, so invalid ids are reported as successful and active subscriptions are not closed.

prkpndy (commented):
some comments/questions. most importantly, i would suggest that we update the rpc test suite for testing the 14.2 implementation

orchestrator/src/worker/utils/fact_topology.rs
Mohiiit: was a upstream change in the cairoPIE itself, couldn't be helped!

madara/crates/primitives/transactions/src/lib.rs
Mohiiit: presenting right value would be hard, because it's a proof, will look for some txns if present on the sequencer repo!

madara/crates/primitives/rpc/src/lib.rs
Mohiiit: 0.10.1 was introduced but within the day 0.10.2 was released, so 0.10.2 is 0.10.1 + additional changes (which will be used and no one will use 0.10.1)

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/starknet_unsubscribe.rs
Mohiiit: ws is not implemented yet, so doesn't matter yet!

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
Mohiiit: that's on the txn replay service to ensure the header is always sent first, and for the existing preconfirmed (shouldn't be the case because executor gets created only after first batch is passed)

for replay, this is fine IMO

madara/crates/client/block_production/src/util.rs
```diff
@@ -191,12 +192,17 @@ pub(crate) fn create_execution_context(
...
     previous_l2_gas_used: u128,
 ) -> anyhow::Result<BlockExecutionContext> {
-    let (block_timestamp, gas_prices) = if let Some(custom_header) =
-        backend.get_custom_header().filter(|h| h.block_n == block_n)
-    {
+    let (block_timestamp, gas_prices) = if let Some(custom_header) = backend.get_custom_header(block_n) {
         // Convert Unix timestamp (seconds since Jan 1, 1970) to SystemTime
         let block_timestamp = UNIX_EPOCH + Duration::from_secs(custom_header.timestamp);
         let gas_prices = custom_header.gas_prices;
+        tracing::info!(
```
heemankv: Do we change this to debug as well ? 
(following up on @prkpndy's request to turn custom header logs to debug)

I am fine with either debug or info, just looking for consistency

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/read/get_storage_proof.rs
```diff
@@ -0,0 +1,29 @@
...
+use mp_rpc::v0_10_0::BlockId;
+use mp_rpc::v0_10_2::{ContractStorageKeysItem, GetStorageProofResult};
+
+pub fn get_storage_proof(
+    starknet: &Starknet,
+    block_id: BlockId,
+    class_hashes: Option<Vec<Felt>>,
+    contract_addresses: Option<Vec<Felt>>,
+    contracts_storage_keys: Option<Vec<ContractStorageKeysItem>>,
+) -> StarknetRpcResult<GetStorageProofResult> {
```
heemankv: Madara will return the storage proofs at the latest confirmed block when we enter preconfirmed as the block tag, whereas it should give "Invalid block id"

heemankv (commented):
Overall looks good, had some nitpicks but I've chosen to defer them for now, one comment added.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/starknet_unsubscribe.rs
prkpndy: Got it. Let's mention this in a comment then?

madara/crates/primitives/rpc/src/lib.rs
prkpndy: got it

madara/crates/client/sync/src/import.rs
```diff
@@ -400,25 +400,33 @@ impl BlockImporterCtx {
...
+                        return Err(BlockImportError::CompiledClassHash {
+                            class_hash,
+                            got: gateway_hash,
+                            expected: hashes.blake_hash,
+                        });
+                    }
+                    (None, Some(gateway_hash))
+                } else {
+                    if !self.config.no_check && hashes.poseidon_hash != gateway_hash {
+                        tracing::debug!(
```
heemankv: **Nitpick:** This should be `warn!` rather than `debug!` — a compiled class hash mismatch is worth operator visibility even if you're intentionally accepting the gateway value.

heemankv (commented):
Good job, review and re-review done.

.prettierignore
```diff
@@ -14,3 +14,4 @@ orchestrator/build/
 orchestrator/node_modules/
 lib/
 s3/
+bootstrapper-v2/contracts/ethereum/src/starkware/
```
prkpndy, line 17: why do we need to add this?

orchestrator/src/utils/mod.rs
```diff
@@ -8,7 +8,3 @@ pub mod metrics_recorder;
 pub mod preflight;
 pub mod rest_client;
 pub mod signal_handler;
-
-/// TODO: This is super Awkward to have this code here
-/// but will try to remove this and move it to the config from the root path
-pub const COMPILED_OS: &[u8] = include_bytes!("../../../build-artifacts/cairo_lang/os_latest.json");
```
prkpndy, line 14: This was not being used anywhere, right? We can also update the build artifacts logic to not build/include this in the artifacts zip. That is redundant now.
If we are not doing that in this PR, let's add a TODO for that to pick later?

prkpndy (approved):
LGTM. No blockers.
Regarding chain config, I would suggest to update the version of that since the actual yaml has changed. We can later update the code itself to own the types used in the `ChainConfig` struct so that we can detect any changes at compile time only.
Also, I would suggest to make greptile score 5/5 before merging (or mention why the issues raised by it are not relevant)

orchestrator/src/utils/mod.rs
Mohiiit: sure!

PR #1010, opened 2026-03-27
feat(test): comprehensive JSON-driven RPC spec compliance test suite

prkpndy:
## Pull Request type

- [x] Testing
- [x] Feature

The existing `tests/js_tests/` suite has only 4 basic tests (declare, deploy, transfer, deploy_account) that validate RPC responses implicitly through starknet.js parsing. There is no source of truth — Madara could return empty arrays, zero values, or structurally valid but semantically wrong responses, and all tests would pass.

## What is the new behavior?

A comprehensive, JSON-driven RPC test suite that validates **all 32 Starknet v0.10.0 RPC methods** with:

- **75 tests**, 0 skipped, 0 failed
- **Field-level value assertions** via JSON fixtures (`state_setup.json`, `read_assertions.json`, `error_assertions.json`) with `$ref`, `$any`, `$gt`, `$length`, `$computed` operators
- **AJV schema validation** against the official OpenRPC spec (downloaded with SHA-256 integrity checks)
- **Error case testing** (9 tests): `-32601`, `-32602`, `BLOCK_NOT_FOUND`, `TXN_HASH_NOT_FOUND`, `CONTRACT_NOT_FOUND`, `CLASS_HASH_NOT_FOUND`, `INVALID_TXN_INDEX`, `CLASS_ALREADY_DECLARED`, `MESSAGES_STATUS_NOT_FOUND`
- **JSON-RPC 2.0 envelope validation** on every call (version, id, result XOR error)
- **10 cross-endpoint consistency checks** (block hash, tx count, class lookup, storage vs call, nonce, etc.)
- **Full L1→L2 messaging test via Anvil**: deploys DummyContract on Anvil, fires `LogMessageToL2`, waits for Madara L1 sync, validates `getMessagesStatus`
- **Method surface coverage check**: verifies all spec methods are exposed and tested
- **Deterministic block composition** via admin `closeBlock()` + `block_time=1h` override
- Raw JSON-RPC for all assertions (not starknet.js abstraction)

### Architecture

```
tests/js_tests/
  fixtures/
    v0_10_0/
      state_setup.json         # 15 write steps defining deterministic chain state
      read_assertions.json     # 48 assertions covering all 32 RPC methods
      error_assertions.json    # 9 error case tests
    DummyContract.sol          # Solidity contract for L1→L2 messaging via Anvil
  src/
    state-executor.ts          # Processes state_setup.json write steps
    assertion-runner.ts        # Processes read_assertions.json with deep matching + AJV
    rpc-caller.ts              # Raw JSON-RPC with envelope validation
    admin-client.ts            # Admin port (closeBlock, addL1HandlerMessage)
    block-tracker.ts           # Block number/hash tracking after closeBlock
    ref-resolver.ts            # $ref, $computed, $self token resolution
    spec/                      # OpenRPC spec download, caching, AJV compilation
  tests/
    rpc.test.ts                # Single entry point: setup → read → write → error → coverage → cross-validation
scripts/
  setup-anvil.sh               # Shared Anvil + DummyContract bootstrap
  js-tests.sh                  # Full test runner (Anvil optional)
```

### Madara change

One small fix in `madara/node/src/main.rs`: devnet mode was unconditionally disabling L1 sync even when `--l1-endpoint` was explicitly provided. Now it only disables L1 sync in devnet when no endpoint is given.

### CI changes

- Installs Foundry (Anvil) via `foundry-rs/foundry-toolchain@v1`
- Shared `scripts/setup-anvil.sh` used by both CI and local `js-tests.sh`
- Madara starts with L1 sync from Anvil for full messaging coverage

No. The Madara change is backward-compatible — devnet without `--l1-endpoint` behaves exactly as before.

## Other information

- New dependencies: `viem` (Anvil interaction), `ajv`/`ajv-formats` (OpenRPC schema validation)
- New Cairo contract: `DeclareTestContract` (reserved for `addDeclareTransaction` test)
- New Solidity contract: `DummyContract` (messaging + state stubs for Anvil)
- Spec files cached in `.cache/specs/` (gitignored) with SHA-256 verification


tests/js_tests/src/admin-client.ts
```diff
@@ -0,0 +1,64 @@
...
+      tx: {
+        version: params.version || "0x0",
+        nonce: params.nonce,
+        contract_address: params.contract_address,
+        entry_point_selector: params.entry_point_selector,
+        calldata: params.calldata,
+      },
+      paid_fee_on_l1: parseInt(params.paid_fee_on_l1),
+    };
+    return this.call("madara_addL1HandlerMessage", [message]);
```
heemankv, line 65: Nitpick: `parseInt(params.paid_fee_on_l1)` will lose precision for fee values > `Number.MAX_SAFE_INTEGER` (2^53 - 1). Since `paid_fee_on_l1` is a u128 on the Madara side, this is a latent issue for anyone extending the test suite with realistic fee values. Current fixtures use `"0x0"` so it works today.

Consider keeping it as a string/hex if the RPC accepts that, or using `BigInt(params.paid_fee_on_l1)` and letting the JSON serializer handle it.

tests/js_tests/src/admin-client.ts
```diff
@@ -0,0 +1,64 @@
...
+        jsonrpc: "2.0",
+        id,
+        method,
+        params,
+      }),
+    });
+
+    const json = await response.json();
+    if (json.error) {
+      throw new Error(
```
heemankv, line 28: Nitpick: The error message drops `json.error.data`, which often contains useful debugging context (e.g., which field was invalid). Consider:
```ts
throw new Error(
  `Admin RPC error [${method}]: ${json.error.message} (code: ${json.error.code}, data: ${JSON.stringify(json.error.data)})`,
);
```

tests/js_tests/tests/rpc.test.ts
```diff
@@ -0,0 +1,394 @@
...
+        createPublicClient,
+        createWalletClient,
+        http,
+        parseAbi,
+        encodeFunctionData,
+      } = await import("viem");
+      const { foundry } = await import("viem/chains");
+
+      const abi = parseAbi([
+        "function fireEvent() public",
```
heemankv, line 92: Nitpick: The hardcoded `sleep(20000)` for L1 finality + sync is understandable for v1, but could be a source of flakiness if CI is slow or Anvil config changes. A polling loop (e.g., checking `getMessagesStatus` or block number advancement) would be more resilient. Fine for now — just flagging for future robustness.

heemankv (commented):
Great work on this PR! 🎯 This is a massive improvement to the test infrastructure — going from 4 implicit tests to 75 field-level assertions covering all 32 RPC methods is impressive. The JSON-driven architecture with `$ref`/`$computed` token resolution is clean and extensible, and the deterministic block composition via `closeBlock()` eliminates the timing-based flakiness from the old suite. The SHA-256 integrity checks on spec files and the cross-validation tests are excellent practices.

A few minor suggestions in the inline comments — nothing blocking.

.github/workflows/task-test-js.yml
```diff
@@ -50,26 +53,46 @@ jobs:
...
-      - name: Setup dev chain and run tests
+      - name: Setup Anvil and deploy DummyContract
+        run: |
+          ANVIL_PORT=8545
+          anvil --port $ANVIL_PORT --block-time 1 --chain-id 1337 --slots-in-an-epoch 1 --silent > /dev/null 2>&1 &
+          echo "ANVIL_PID=$!" >> $GITHUB_ENV
+          echo "ANVIL_PORT=$ANVIL_PORT" >> $GITHUB_ENV
+          sleep 2
+
+          BYTECODE=$(cat tests/js_tests/fixtures/DummyContract.bytecode)
```
Mohiiit: Could we pull the Anvil + dummy core-contract bootstrap into a shared helper? It is now open-coded here and in `scripts/js-tests.sh`, so any future tweak to the L1 harness has to be kept in sync in two places. Since this setup already includes sleeps and a raw `eth_sendTransaction`, it feels like the kind of environment logic that will drift unless both entrypoints share the same script.

.github/workflows/task-test-js.yml
```diff
@@ -50,26 +53,46 @@ jobs:
...
+            --rpc-admin \
             --devnet \
             --preset devnet \
             --l1-gas-price 0 \
             --blob-gas-price 0 \
             --strk-per-eth 1 \
-            --no-l1-sync \
-            --rpc-pre-v0-9-preconfirmed-as-pending &
+            --l1-endpoint "http://127.0.0.1:${{ env.ANVIL_PORT }}" \
+            --unsafe-skip-l1-message-consumed-check \
```
Mohiiit: Do we still need `--unsafe-skip-l1-message-consumed-check` in the default JS test harness? CI starts from a fresh Anvil instance and deploys the dummy core contract per run, so the L1 message existence / consumed checks should already be deterministic. Keeping the unsafe bypass in the normal path makes it harder to tell whether the suite is validating real integration behavior or a relaxed one.

.github/workflows/task-test-js.yml
```diff
@@ -50,26 +53,46 @@ jobs:
...
+            --rpc-pre-v0-9-preconfirmed-as-pending \
```
Mohiiit: This suite is explicitly targeting RPC `v0.10.0`, but this flag is about pre-v0.9 preconfirmed/pending semantics. I don't see a v0.10 assertion here that obviously depends on it, so carrying it in the default harness reads like unrelated environment skew. If there is a concrete dependency, it would help to document it here.

tests/js_tests/src/assertion-runner.ts
```diff
@@ -0,0 +1,562 @@
...
+  const rpcCaller = new RpcCaller(ctx.rpcUrl);
+
+  // Resolve $ref and $computed in params
+  const resolvedParams = resolveValue(assertion.params, ctx.results);
+
+  // Build transaction if construct hint is present
+  let finalParams = resolvedParams;
+  let actual: any;
+
+  if (assertion.construct) {
```
Mohiiit, line 43: When `buildConstructedParams()` returns `_raw_result`, `runAssertion()` never goes through `RpcCaller.call()`, and the later `!assertion.construct` guard also skips OpenRPC result validation. That means constructed coverage for `starknet_estimateFee`, `starknet_simulateTransactions`, `starknet_addInvokeTransaction`, `starknet_addDeclareTransaction`, and `starknet_addDeployAccountTransaction` is currently exercising `starknet.js` responses rather than the raw JSON-RPC path this suite is supposed to verify. I think the construct step should only build/sign the payload, and the final request should still be sent through the raw caller so the real RPC request/response shape is what gets validated.

Mohiiit (commented):
A few follow-ups on the harness design, the environment defaults, and how maintainable the spec-fixture setup will be as this grows.

tests/js_tests/src/assertion-runner.ts
```diff
@@ -0,0 +1,562 @@
...
+  assertion: Assertion,
+  hint: ConstructHint,
+  resolvedParams: any,
+  ctx: TestContext,
+  provider: RpcProvider,
+): Promise<any> {
+  const deployer = new Deployer(UDC_ADDRESS, "deployContract");
+  const account = new Account({
+    provider,
+    address: hint.sender_address || DEFAULT_ACCOUNT_ADDRESS,
```
Mohiiit: This builder accepts `hint.sender_address`, but always signs with `DEFAULT_PRIVATE_KEY`. It works with the current fixtures because every constructed invoke uses the default predeployed account, but the API surface is misleading: any future fixture that sets a different sender address will silently build an invalid transaction. Either bind the signer to the hinted sender, or make it explicit that only the default account is supported.

tests/js_tests/src/assertion-runner.ts
```diff
@@ -0,0 +1,562 @@
...
+}
+
+async function buildDeclareV3Params(
+  _assertion: Assertion,
+  hint: ConstructHint,
+  resolvedParams: any,
+  _ctx: TestContext,
+  provider: RpcProvider,
+): Promise<any> {
+  // Close any pending block from prior write assertions so nonce is canonical
```
Mohiiit: This helper is doing two jobs: preparing a constructed transaction and mutating chain state by closing a block and sleeping. That makes the builder path stateful and harder to reason about, and it also bakes timing assumptions into a place that looks like pure parameter construction. I'd keep block orchestration in the test/setup layer and make the construct helpers only build the transaction payload. The same pattern appears again in `buildDeployAccountV3Params()` below.

tests/js_tests/tests/rpc.test.ts
```diff
@@ -0,0 +1,394 @@
...
+import {
+  StateSetup,
+  ReadAssertions,
+  ErrorAssertions,
+  TestContext,
+} from "../src/types";
+import { createSpecRegistry } from "../src/spec/spec-registry";
+
+// Load fixture JSONs
+const FIXTURE_DIR = path.resolve(__dirname, "../fixtures/v0_10_0");
```
Mohiiit, line 17: The harness is version-aware in a few places (`getRpcUrl(stateSetup.version)`, spec registry by tag), but fixture loading is still hardcoded to `fixtures/v0_10_0`. That means adding a new RPC version is not yet a data-only change. Pushing fixture discovery through the declared version would make multi-version growth much easier.

tests/js_tests/fixtures/DummyContract.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity >=0.7.0 <0.9.0;
+
+contract DummyContract {
```
Mohiiit, line 4: There is already a `DummyContract`-style core-contract mock embedded in `settlement_client` ETH tests with overlapping surface (`LogMessageToL2`, `l1ToL2Messages`, `stateBlockNumber`, etc.). Is there a reason this needs to be a second source of truth instead of being shared or generated? Otherwise the JS harness and Rust settlement tests can drift in what they consider a valid core-contract stub.

tests/js_tests/fixtures/v0_10_0/error_assertions.json
```diff
@@ -0,0 +1,81 @@
+{
+  "version": "0.10.0",
```
Mohiiit, line 2: I checked these error codes against the official Starknet v0.10 OpenRPC and they line up. Since this file is meant to be a spec-compliance fixture, would it make sense to document or derive that provenance from the locked spec files? Right now a future change to one of these numeric codes would look like an arbitrary fixture edit rather than a deliberate spec update.

tests/js_tests/fixtures/v0_10_0/read_assertions.json
```diff
@@ -0,0 +1,755 @@
...
+            "nonce": "$any",
+            "signature": "$length_gte:1",
+            "resource_bounds": "$any"
+          }
+        ]
+      },
+      "description": "Full tx body for DECLARE. Validates class_hash, sender, signature present."
+    },
+    {
+      "id": "get_block_with_txs_deploy_account",
```
Mohiiit, line 123: This suite is broad, but some of the high-value cases are still pretty loosely constrained. For example `get_block_with_txs_deploy_account` only checks `status` + `transactions.length >= 1`, and parts of `get_block_with_receipts` and later assertions rely heavily on `$any`. That still catches gross regressions, but it leaves some v0.10-specific semantic bugs underconstrained. Tightening a few representative cases would make the suite much harder to accidentally satisfy.

tests/js_tests/src/contract-loader.ts
```diff
@@ -0,0 +1,36 @@
+import * as fs from "fs";
+import * as path from "path";
+import { CompiledSierra, CompiledSierraCasm } from "starknet";
+
+const BUILD_ARTIFACTS_DIR = path.resolve(
```
Mohiiit, line 5: Worth discussing whether these tests should depend on checked-in `build-artifacts/js_tests` snapshots or invoke the existing artifact pipeline on demand. The repo already has `scripts/artifacts.sh` generating this directory, so the current approach works, but it also means fixture correctness depends on committed generated JSON staying in sync with the Cairo sources.

tests/js_tests/src/admin-client.ts
prkpndy: our tests are not using that high values tbh. I have added a comment specifying the limitation

tests/js_tests/src/admin-client.ts
prkpndy: done

tests/js_tests/tests/rpc.test.ts
prkpndy: hmm.. it'll add extra code for `getMessageStatus` to maintain. ig it's okay for testing purposes. since the block time in anvil is 1 seconds and madara waits for 10 blocks, 20 secs is enough time

tests/js_tests/tests/rpc.test.ts
```diff
@@ -0,0 +1,394 @@
...
+        address: DEFAULT_ACCOUNT_ADDRESS,
+        signer: DEFAULT_PRIVATE_KEY,
+        deployer,
+      });
+
+      const sierra = loadContractSierra("madara_contracts_HelloStarknet");
+      const casm = loadContractCasm("madara_contracts_HelloStarknet");
+
+      try {
+        await account.declare({ contract: sierra, casm });
```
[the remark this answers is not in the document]
prkpndy: done

tests/js_tests/src/rpc-caller.ts
```diff
@@ -0,0 +1,133 @@
...
+function assertValidEnvelope(json: any, expectedId: number): void {
+  if (json.jsonrpc !== "2.0") {
+    throw new Error(
+      `Invalid JSON-RPC version: expected "2.0", got "${json.jsonrpc}"`,
+    );
+  }
+
+  if (json.id !== expectedId) {
+    throw new Error(
+      `JSON-RPC id mismatch: expected ${expectedId}, got ${json.id}`,
```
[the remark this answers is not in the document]
prkpndy: done

tests/js_tests/src/spec/spec-manifest.ts
```diff
@@ -0,0 +1,47 @@
...
+      {
+        file: "starknet_trace_api_openrpc.json",
+        sourceUrl:
+          "https://raw.githubusercontent.com/eqlabs/pathfinder/main/specs/rpc/v10/starknet_trace_api_openrpc.json",
+        sha256:
+          "ba07e54565e2de6ef25d64e4ba2454c2dbfd0133555557b75c14e5d8f22db993",
+      },
+    ],
+  },
+];
```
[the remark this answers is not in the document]
prkpndy: i think it's okay. we'll explicitly know and fix it when the rpc spec changes

tests/js_tests/tests/rpc.test.ts
```diff
@@ -0,0 +1,394 @@
...
+      if (txCount === undefined || !blockTxHashes) return;
+
+      expect(txCount).toBe(blockTxHashes.transactions.length);
+    });
+
+    it("class lookup consistency: getClass and getClassAt return same entry points", async () => {
+      const byHash = ctx.assertionResults.get("get_class");
+      const byAddr = ctx.assertionResults.get("get_class_at");
+      if (!byHash || !byAddr) return;
+
```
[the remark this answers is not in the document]
prkpndy: fixed

.github/workflows/task-test-js.yml
prkpndy: done

.github/workflows/task-test-js.yml
prkpndy: yeah, fair. have removed it

.github/workflows/task-test-js.yml
prkpndy: removed

tests/js_tests/src/assertion-runner.ts
prkpndy: fair. removed the `sender_address` from the Hint objecy

tests/js_tests/src/assertion-runner.ts
prkpndy: simplified

tests/js_tests/fixtures/DummyContract.sol
prkpndy: The Rust test's DummyContract is embedded as bytecode inside a sol! macro - it can't be directly shared with JS. Our contract is a superset (combines messaging from one Rust DummyContract + state stubs from another). We compiled it separately because the Rust contracts were split across two separate sol! macros and couldn't be reused as-is.

tests/js_tests/fixtures/v0_10_0/error_assertions.json
prkpndy: hmm, i would put it as a good to have. we'll know when the spcs are changed (because of hash lock) and when they do, we can manually update it. wdyt?

tests/js_tests/src/contract-loader.ts
prkpndy: hmm, can do but it'll complicate the ci pipeline (fetch or build the artifacts). it's just test contracts so won't change very ofter

Mohiiit (approved):

PR #1012, opened 2026-03-28
Implement Starknet websocket subscriptions

heemankv:
## Summary
- implement Starknet WebSocket subscriptions for RPC `v0.10.0` and `v0.10.2`
- intentionally do not expose WebSocket subscriptions for RPC `v0.8.1` or `v0.9.0`; legacy WS calls are unregistered and return standard JSON-RPC method-not-found (`-32601`)
- align `v0.10.0` `subscribeNewHeads` with the OpenRPC shape by making `block_id` optional and defaulting it to `latest`, matching `v0.10.2`
- add backend reorg notifications and emit `starknet_subscriptionReorg` from the supported WS paths
- add contract and handler tests for the supported versioned WS surface, including reorg/resume, many-client cleanup, legacy method-not-found coverage, tx-status no-known-status handling, unsubscribe stream closure, preconfirmed receipt reorg priority, and preconfirmed event append delivery
- make websocket unsubscribe cancellation stateful so late waiters cannot miss the close signal
- prioritize pending reorg notifications before receipt emission for both confirmed and preconfirmed receipt blocks
- wire dedicated tx-status and new-transaction watcher abstractions into the RPC layer and node service
- keep common v0.10 WS handlers shared where the protocol types are identical, including receipt subscriptions and unsubscribe, while retaining version-specific paths for transaction payload differences such as `include_proof_facts`
- clean up stale WS docs/comments and remove unsupported legacy WS implementations
- skip the Atlantic live prover-client test path when the live account has insufficient credits

## Supported WS RPC Versions
- `v0.10.0`: `subscribeNewHeads`, `subscribeEvents`, `subscribeTransactionStatus`, `subscribeNewTransactions`, `subscribeNewTransactionReceipts`, `unsubscribe`
- `v0.10.2`: `subscribeNewHeads`, `subscribeEvents`, `subscribeTransactionStatus`, `subscribeNewTransactions`, `subscribeNewTransactionReceipts`, `unsubscribe`

## Unsupported WS RPC Versions
- `v0.8.1` and `v0.9.0` HTTP/read/write/trace RPCs remain registered as before.
- `v0.8.1` and `v0.9.0` WebSocket methods are not registered.
- Calling legacy WS methods such as `starknet_V0_8_1_subscribeNewHeads` or `starknet_V0_9_0_subscribeEvents` returns JSON-RPC method-not-found (`-32601`). This is covered by `legacy_ws_methods_return_method_not_found`.

## jsonrpsee Fork
This PR pins `jsonrpsee-core` and `jsonrpsee-types` to the Madara-owned `madara-alliance/jsonrpsee` fork at `52c76441b922b9d1c6b260fcb29d62e94a8ca57c`, based on upstream jsonrpsee `v0.22.5`. The fork exists because jsonrpsee's built-in subscription framing always emits `params.subscription` + `params.result`, while Starknet OpenRPC expects `params.subscription_id` + `params.result` and specific `starknet_subscription*` method names. The core patch exposes the existing raw complete-message path publicly so Madara can keep jsonrpsee for connection/subscription lifecycle while emitting the Starknet-spec public wire shape. The types patch accepts `subscription_id` as a deserialize-only alias so jsonrpsee-based tests and diagnostic clients can consume Starknet-shaped notifications. The fork rationale is documented in `MADARA_FORK.md` in that branch.

## Verification
- `cargo fmt --all -- --check`
- `cargo clippy -p mc-rpc --lib --tests -- -D warnings`
- `cargo test -p mc-mempool topic_pubsub --lib`
- `cargo test -p mc-rpc subscribe_transaction_status --lib`
- `cargo test -p mc-rpc unsubscribe --lib`
- `cargo test -p mc-rpc subscribe_new_heads_defaults_to_latest_when_block_id_missing_v0_10_0 --lib`
- `cargo test -p mc-rpc subscribe_events_preconfirmed_append_emits_only_new_events_v0_10_2 --lib`
- `cargo test -p mc-rpc subscribe_new_transaction_receipts_l1_confirmed_block --lib`
- `cargo test -p mc-rpc subscribe_new_transaction_receipts_preconfirmed_reorg_wins --lib`
- `cargo test -p mc-rpc subscribe_new_transaction_receipts --lib`
- `cargo test -p mc-rpc ws_contract_tests --lib`
- `cargo test -p mc-rpc ws --lib`
- `cargo check -p mc-rpc`

## Notes
- The Atlantic live prover-client test now treats insufficient live-account credits as a skipped external-service condition instead of failing CI.
- I could not find a repo-local `GPE` command or workflow name to run directly. The standard PR CI for this repository includes the Rust lint/fmt and code-style workflows, so this PR should exercise the expected checks there.
- `cargo test -p mc-db --lib -- --nocapture` still hits the existing unrelated `rocksdb::events_bloom_filter::tests::test_false_positive_rate_with_size` property test failure, which is outside this WS change set.


heemankv:
@greptile please re-review the PR

Mohiiit (commented):
### Final review: Starknet WebSocket subscriptions

The implementation now covers the v0.10.x WebSocket surface with explicit reorg handling and good protocol-level test coverage. I left a small set of maintainability and scope comments below. The one item I strongly recommend addressing before merge is removing the generated `build-artifacts/js_tests` output from this PR.

Local validation completed successfully: the full Madara binary builds in the debug/dev profile using the configured SSD target, and `git diff --check origin/main...HEAD` passes.

build-artifacts/js_tests/madara_contracts.starknet_artifacts.json
```diff
@@ -0,0 +1 @@
...contracts","contract_name":"StateUpdateContract","module_path":"madara_contracts::appchain_test::StateUpdateContract","artifacts":{"sierra":"madara_contracts_StateUpdateContract.contract_class.json","casm":"madara_contracts_StateUpdateContract.compiled_contract_class.json"}},{"id":"duf7pspu76tbe","package_name":"madara_contracts","contract_name":"DeclareTestContract","module_path":"madara_contracts::declare_test::DeclareTestContract","artifacts":{"sierra":"madara_contracts_DeclareTestContract.contract_class.json","casm":"madara_contracts_DeclareTestContract.compiled_contract_class.json"}},{"id":"kp7flgguc0c3g","package_name":"madara_contracts","contract_name":"HelloStarknet","module_path":"madara_contracts::hello::HelloStarknet","artifacts":{"sierra":"madara_contracts_HelloStarknet.contract_class.json","casm":"madara_contracts_HelloStarknet.compiled_contract_class.json"}},{"id":"qvgdjlbsj664q","package_name":"madara_contracts","contract_name":"MessagingContract","module_path":"madara_contracts::messaging_test::MessagingContract","artifacts":{"sierra":"madara_contracts_MessagingContract.contract_class.json","casm":"madara_contracts_MessagingContract.compiled_contract_class.json"}}]}
```
Mohiiit: Please remove the generated `build-artifacts/js_tests` files from this PR. These nine one-line JSON artifacts add roughly 466 KB and are unrelated to the WebSocket implementation; they should be regenerated through the established artifact workflow when needed.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/mod.rs
```diff
@@ -1,34 +1,64 @@
...
+pub mod subscribe_transaction_status;
 
-// FIXME(subscriptions): Re-add subscriptions once they're working.
-// pub mod subscribe_events;
-// pub mod subscribe_new_heads;
-// pub mod subscribe_pending_transactions;
-// pub mod subscribe_transaction_status;
-
-// FIXME(subscriptions): Remove this #[allow(unused)] once subscriptions are back.
 #[allow(unused)]
```
Mohiiit: `BLOCK_PAST_LIMIT` and `ADDRESS_FILTER_LIMIT` are both used now that these subscriptions are enabled, so the `#[allow(unused)]` suppressions are stale. Please remove them. In the v0.10.0 module, `BLOCK_PAST_LIMIT` can be removed entirely if it remains unused because the relevant handlers delegate to v0.10.2.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
```diff
@@ -0,0 +1,682 @@
...
+
+    fn matches(&self, event_info: &EventWithInfo) -> bool {
+        self.allowed_addresses
+            .as_ref()
+            .map(|addresses| addresses.contains(&event_info.event.from_address))
+            .unwrap_or(true)
+    }
+}
+
+pub async fn subscribe_events(
```
Mohiiit, line 69: `subscribe_events` currently combines parameter validation, initial cursor resolution, historical backfill, live streaming, cancellation, and reorg recovery in one roughly 157-line state machine. Could we extract at least the initial cursor, backfill phase, and live phase into named helpers? That would make the reorg transitions substantially easier to review and unit-test.

madara/crates/client/rpc/src/versions/user/ws_contract_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+    assert!(!methods.contains("starknet_V0_9_0_unsubscribe"));
+    assert!(!methods.contains("starknet_V0_9_0_subscribePendingTransactions"));
+}
+
+#[tokio::test]
+async fn legacy_ws_methods_return_method_not_found() {
+    let (_, starknet) = rpc_test_setup();
+    let module = rpc_api_user(&starknet).expect("Building user RPC module");
+
+    for method in [
```
Mohiiit: These surface-contract tests are useful, but the legacy method names are duplicated between two tests and have already drifted: this raw-request list omits `starknet_V0_9_0_subscribePendingTransactions`, while the registry test includes it. Please define one `LEGACY_WS_METHODS` constant and reuse it in both tests.

madara/crates/client/rpc/src/lib.rs
```diff
@@ -988,20 +1159,195 @@ pub(crate) struct WsSubscriptionGuard {
     id: u64,
     // FIXME(subscriptions): Remove this #[allow(unused)] once subscriptions are back.
     #[allow(unused)]
-    handle: std::sync::Arc<tokio::sync::Notify>,
-    map: std::sync::Arc<dashmap::DashMap<u64, std::sync::Arc<tokio::sync::Notify>>>,
+    handle: std::sync::Arc<WsSubscriptionHandle>,
```
Mohiiit: `subscription_register()` and `cancelled()` are now actively used, so their `#[allow(unused)]` and FIXME markers are stale. The guard's `handle` field itself is never read after registration; if it is intentionally retained only to pin the sender lifetime, please rename it to `_handle` and document that ownership role, otherwise remove the extra `Arc` from the guard.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/mod.rs
heemankv: Done in 912f4d077. Removed the stale allow(unused) suppressions in v0.10.2 and removed the unused v0.10.0 BLOCK_PAST_LIMIT constant.

madara/crates/client/rpc/src/versions/user/ws_contract_tests.rs
heemankv: Done in 912f4d077. Added one LEGACY_WS_METHODS constant and reused it in both legacy WS surface tests, including the v0.9.0 subscribePendingTransactions case.

madara/crates/client/rpc/src/lib.rs
heemankv: Done in 912f4d077. Removed the stale FIXME/allow markers, renamed the retained guard field to _handle, and documented that it keeps the registered handle alive until drop.

build-artifacts/js_tests/madara_contracts.starknet_artifacts.json
heemankv: Done in 912f4d077. Removed the nine PR-added build-artifacts/js_tests files so the WebSocket PR no longer carries generated JS test artifacts.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Done in 912f4d077. Split subscribe_events into named initial cursor, backfill, and live-stream helpers while keeping the existing reorg-first behavior and cursor rewind semantics.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_heads.rs
```diff
@@ -0,0 +1,156 @@
...
+use mp_rpc::v0_10_2::{BlockHeader, BlockId, BlockTag};
+
+use super::BLOCK_PAST_LIMIT;
+
+pub async fn subscribe_new_heads(
+    starknet: &crate::Starknet,
+    subscription_sink: jsonrpsee::PendingSubscriptionSink,
+    block_id: BlockId,
+) -> Result<(), StarknetWsApiError> {
+    let sink = subscription_sink.accept().await.or_internal_server_error("Failed to establish websocket connection")?;
```
Mohiiit, line 21: Could we validate `block_id` before calling `accept()`? I may be missing a jsonrpsee detail, but errors after acceptance seem to become a successful subscription followed by EOF. Also, should `L1Accepted` be allowed when `SUBSCRIPTION_BLOCK_ID` excludes it?

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
```diff
@@ -0,0 +1,795 @@
...
+        match stream_live_events(starknet, &mut state).await? {
+            LiveResult::Restart => continue,
+            LiveResult::Closed => return Ok(()),
+        }
+    }
+}
+
+fn initial_event_block_n(starknet: &crate::Starknet, block_id: Option<BlockId>) -> Result<u64, StarknetWsApiError> {
+    let Some(block_id) = block_id else {
+        return Ok(starknet.backend.latest_confirmed_block_n().map_or(0, |block_n| block_n.saturating_add(1)));
```
Mohiiit: Should this start from the latest block rather than `latest + 1`? Could we add a test where the latest block already contains an event before the subscription starts?

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
```diff
@@ -0,0 +1,795 @@
...
+    Ok(None)
+}
+
+fn take_pending_reorg(reorgs: &mut ReorgStream) -> Result<Option<mc_db::ReorgNotification>, StarknetWsApiError> {
+    crate::try_recv_live_reorg(reorgs, super::missed_reorg_notifications_error)
+}
+
+fn subscription_allows_finality(requested_finality: &FinalityStatus, event_finality: TxnFinalityStatus) -> bool {
+    match requested_finality {
+        FinalityStatus::PreConfirmed => matches!(event_finality, TxnFinalityStatus::PreConfirmed),
```
Mohiiit: Was `PRE_CONFIRMED` intended to emit only the preconfirmed version? The spec seems to suggest that later finality updates may also be emitted. Could we add a preconfirmed-to-confirmed test to clarify this?

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_transactions.rs
```diff
@@ -0,0 +1,901 @@
...
+
+    let ctx = starknet
+        .ws_handles
+        .subscription_register(sink.subscription_id(), crate::metrics::SUBSCRIBE_NEW_TRANSACTIONS)
+        .await;
+    let allowed_statuses =
+        finality_status.unwrap_or_else(|| vec![TxnStatusWithoutL1::AcceptedOnL2]).into_iter().collect::<HashSet<_>>();
+    let sender_address = crate::normalize_sender_address_filter(sender_address);
+    let include_proof_facts = tags.as_ref().is_some_and(|tags| tags.contains(&SubscriptionTag::IncludeProofFacts));
+    let mut emitted = HashSet::<(Felt, TxnStatusWithoutL1)>::new();
```
Mohiiit, line 52: Could this deduplication set be bounded or pruned? I may be missing another reset path, but it looks like it could grow with total chain traffic for every long-lived subscriber.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_transaction_status.rs
```diff
@@ -0,0 +1,126 @@
...
+    snapshot: crate::TxStatusSnapshot,
+) -> Result<(), crate::errors::StarknetWsApiError> {
+    let finality_status = match snapshot {
+        crate::TxStatusSnapshot::Received => mp_rpc::v0_10_2::TxnStatus::Received,
+        crate::TxStatusSnapshot::Candidate => mp_rpc::v0_10_2::TxnStatus::Candidate,
+        crate::TxStatusSnapshot::PreConfirmed => mp_rpc::v0_10_2::TxnStatus::PreConfirmed,
+        crate::TxStatusSnapshot::AcceptedOnL2 => mp_rpc::v0_10_2::TxnStatus::AcceptedOnL2,
+        crate::TxStatusSnapshot::AcceptedOnL1 => mp_rpc::v0_10_2::TxnStatus::AcceptedOnL1,
+    };
+    let status = mp_rpc::v0_10_2::WsTxnStatusResult { execution_status: None, finality_status, failure_reason: None };
```
Mohiiit: Was `execution_status` intended to remain empty after the transaction has executed? Could we include `SUCCEEDED`, `REVERTED`, and the failure reason once that information is available?

tests/starknetgo_ws/ws_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+func closeProbes(probes []activeProbe) {
+	for _, p := range probes {
+		if p.close != nil {
+			p.close()
+		}
+	}
+}
+
+func assertResults(t *testing.T, results []result) {
```
Mohiiit: Could the required probes also assert the expected transaction hash, finality, and filter result? I wonder if an incorrect notification could currently pass as long as something arrives.

madara/crates/client/rpc/src/metrics.rs
```diff
@@ -0,0 +1,148 @@
...
+    subscriptions_opened: Counter<u64>,
+    subscriptions_closed: Counter<u64>,
+    notifications_sent: Counter<u64>,
+    notification_send_failures: Counter<u64>,
+    reorg_notifications_sent: Counter<u64>,
+    lagged_notifications: Counter<u64>,
+    subscription_duration: Histogram<f64>,
+}
+
+impl WsMetrics {
```
Mohiiit, line 29: Could we document how these metrics are expected to determine WebSocket health? Would dashboards or alerts for send failures, lagged notifications, and active subscriptions, together with a synthetic `subscribeNewHeads` probe, be useful?

Mohiiit (commented):
Hey, I re-reviewed the latest changes. I left a few questions inline around subscription behavior, memory usage, testing, and metrics.

One general ask: could we add short Rustdoc comments to the subscription entry points and state-machine helpers? It would help explain each function's role and its cursor, finality, and reorg assumptions, making the implementation easier to validate while reading it.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_heads.rs
heemankv: Addressed in dac2b6d5f: subscribeNewHeads now resolves or rejects block_id before accept(), and L1Accepted is rejected with the existing unsupported-tag error path. Added coverage for both PreConfirmed and L1Accepted rejection.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Addressed in dac2b6d5f: kept latest + 1 intentionally for default subscriptions because the default path is new events after subscription creation. Added subscribe_events_default_starts_after_latest_v0_10_2 to lock that behavior, and block_id resolution now happens before accept().

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Addressed in dac2b6d5f: PRE_CONFIRMED event subscriptions now allow later finality updates too. Added subscribe_events_preconfirmed_filter_emits_confirmed_update_v0_10_2.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_transactions.rs
heemankv: Addressed in dac2b6d5f: the transaction dedup set is cleared when a confirmed head advances, so it is bounded to the active live/preconfirmed window rather than total subscriber lifetime. Applied the same parity cleanup to the v0.10.0 transaction path and receipt path.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_transaction_status.rs
heemankv: Addressed in dac2b6d5f: subscribeTransactionStatus now looks up the executed receipt when available and emits execution_status as SUCCEEDED or REVERTED, including failure_reason for reverted transactions. Added tests for both branches.

tests/starknetgo_ws/ws_test.go
heemankv: Addressed in dac2b6d5f: the StarknetGo matrix now asserts the observed transaction hash, finality, event from_address, and sender-filtered hash/sender expectations instead of passing on any notification.

madara/crates/client/rpc/src/metrics.rs
heemankv: Addressed in dac2b6d5f: added metric documentation for websocket health usage: active subscriptions by method, send failures, lagged notifications, and synthetic subscribeNewHeads probing.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_heads.rs
Mohiiit: Verified on the latest head. This looks addressed, thanks.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
Mohiiit: Hey, this still looks pending to me. The v0.10.2 spec says `block_id` defaults to `latest` and events are emitted from the specified block through latest. Could we start from the current latest block instead of `latest + 1`, or point to upstream behavior that intentionally treats the default differently?

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
Mohiiit: Verified on the latest head. This looks addressed, thanks.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_new_transactions.rs
Mohiiit: Verified on the latest head. This looks addressed, thanks.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_transaction_status.rs
Mohiiit: Verified on the latest head. This looks addressed, thanks.

tests/starknetgo_ws/ws_test.go
Mohiiit: Verified on the latest head. This looks addressed, thanks.

madara/crates/client/rpc/src/metrics.rs
Mohiiit: Hey, this still looks pending to me. The Rustdoc explains the intended signals, but I could not find these metrics in the operator docs, dashboard, alerts, or an actual synthetic probe. Could we wire or document those so operators can verify delivery health?

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
```diff
@@ -0,0 +1,904 @@
...
+fn validate_keys(keys: &Option<Vec<Vec<Felt>>>) -> Result<(), StarknetWsApiError> {
+    let total_keys = keys.as_ref().map(|patterns| patterns.iter().map(Vec::len).sum::<usize>()).unwrap_or(0);
+    if total_keys > MAX_EVENTS_KEYS {
+        return Err(StarknetWsApiError::TooManyKeysInFilter);
+    }
+
+    Ok(())
+}
+
+async fn send_block_events(
```
Mohiiit, line 314: Could we add a short doc comment here and to the remaining non-trivial helpers? For this function, could the comment explain view selection, reorg checks, finality filtering, and deduplication so its invariants are clear before reading the body?

Mohiiit (commented):
Hey, I did one final pass. Most of the earlier comments look addressed. I kept two existing threads open and left one documentation follow-up inline.

madara/crates/client/rpc/src/metrics.rs
heemankv: Done in c8fcb96c7: added operator-facing WebSocket subscription health docs in observability/README.md with metric list, PromQL panel/alert queries, alert guidance, and a synthetic subscribeNewHeads probe command.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Done in c8fcb96c7: subscribeEvents now defaults to the current latest block, matching OpenRPC, and the default-cursor test now asserts an already-existing latest block event is replayed.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Done in c8fcb96c7: documented send_block_events and the helper invariants around view selection, reorg checks, finality filtering, and phase-scoped deduplication.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_events.rs
heemankv: Still addressed on latest head c8fcb96c7: subscribe_events is split into initial_event_block_n, backfill_events, stream_live_events, send_block_events, and smaller validation/finality helpers, with the reorg transitions isolated in the named phases.

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/ws/subscribe_transaction_status.rs
```diff
@@ -0,0 +1,155 @@
...
+) -> Result<Option<SubscriptionUpdate>, crate::errors::StarknetWsApiError> {
+    if allow_current {
+        if let Some(snapshot) = watch.take_current() {
+            return Ok(Some(SubscriptionUpdate::Snapshot(snapshot)));
+        }
+    }
+
+    tokio::select! {
+        _ = sink.closed() => Ok(None),
+        _ = ctx.cancelled() => Err(crate::errors::StarknetWsApiError::SubscriptionClosed),
```
Mohiiit, line 93: Could we treat `ctx.cancelled()` as a clean `Ok(None)` here and in the other subscription handlers? I may be missing a jsonrpsee detail, but after the subscription is accepted, returning `SubscriptionClosed` is converted into an error notification, while `starknet_unsubscribe` seems intended to close the stream normally. Could we also assert that unsubscribe emits no terminal error notification? This does not need to block the PR.

Mohiiit (approved):
Looks good to me overall. I left one non-blocking follow-up around unsubscribe shutdown behavior.

PR #1017, opened 2026-04-02
Consolidate OTEL into mc-telemetry

heemankv:
## Summary
Consolidate Madara OTEL into `mc-telemetry` and remove the old split between `mc-analytics` and the websocket telemetry crate.

## What Changed
- removed the `mc-analytics` crate and moved the OTEL setup and metric registration helpers into `mc-telemetry`
- removed Prometheus exporter/endpoint support and the old websocket telemetry path
- switched node CLI/config to the OTEL-only `TelemetryParams` surface
- updated metric consumers to depend on `mc-telemetry`
- unified node startup on a single `TelemetryService`
- renumbered the service IDs to keep them contiguous after removing `Analytics`, and fixed `MadaraServiceMask::active_set()` to enumerate the highest service IDs correctly
- preserved the old OTEL per-signal naming split so traces, metrics, and logs still appear as separate services downstream

## Why
Madara no longer needs the Prometheus-specific and websocket telemetry paths. This change leaves a single telemetry crate focused on OTEL while keeping the downstream OTEL service naming behavior consistent with the old analytics implementation.

## Impact
- removes `--analytics-*`, `MADARA_ANALYTICS_*`, `--telemetry`, and `--telemetry-url`
- adds the OTEL-only telemetry surface (`--otel-service-name`, `--otel-collector-endpoint`, `--otel-export-metrics`, `--otel-export-traces`, `--otel-export-logs`)
- removes the old external websocket telemetry broadcast and Prometheus endpoint support

## Validation
- `cargo test -p mc-telemetry`
- `cargo test -p mp-utils`
- `cargo test -p madara`


prkpndy (approved):
lgtm

PR #1022, opened 2026-04-08
fix: state root calculation for starknet version >= 0.14.0

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Testing

`calculate_state_root` unconditionally short-circuits when `class_trie_root == 0`, returning `contract_trie_root` directly without hashing. This is correct for Starknet < 0.14.0 (where the class trie didn't exist), but **wrong for >= 0.14.0** where the formula is always `Poseidon(STARKNET_STATE_V0, contract_root, class_root)` — even when `class_root` is zero.

This causes madara to compute incorrect state roots for any >= 0.14.0 chain where the class trie is empty (e.g. early blocks of a new L3 chain), which in turn produces wrong block hashes and fails sync verification.

## What is the new behavior?

- `calculate_state_root` now takes a `StarknetVersion` parameter and gates the short-circuit on `version < 0.14.0`, matching `calculate_global_state_root` from the Starknet OS (`commitment.cairo`) and pathfinder's `StateCommitment::calculate`.
- `apply_to_global_trie` accepts `last_block_protocol_version` and passes it through. Callers (`write_new_confirmed_inner`, sync `import.rs`, snap sync `apply_state.rs`) each provide the version from the block header.
- Intermediate state roots in a batch use the last block's version — a comment documents that these can be imprecise across version boundaries, but only the final root is returned and verified.

### Test coverage

- **5 unit test cases** for `calculate_state_root` covering the full matrix:
  - `(non-zero, non-zero)` — hash
  - `(non-zero, zero)` with >= 0.14.0 — hash (regression test)
  - `(zero, non-zero)` — hash
  - `(zero, zero)` — zero
  - `(non-zero, zero)` with < 0.14.0 — short-circuit (pre-0.14.0 legacy behavior)
- **End-to-end test** applying a real 0.14.1 genesis state diff through `apply_to_global_trie` and verifying the returned root.
- **Block hash test** pinning `Header::compute_hash` against a real Starknet 0.14.1 genesis block with known-good values.
- Updated `test_update_tries::case_1_success` expected state root to match the >= 0.14.0 formula.

No. The `apply_to_global_trie` function signature gains a `StarknetVersion` parameter, but this is an internal API — no external consumers are affected.

## Other information

- The fix matches the canonical implementation in `calculate_global_state_root` from `starkware/starknet/core/os/state/commitment.cairo` in the sequencer repo.
- Pre-existing build break in `mp-block/src/commitments.rs` (`starknet_api::felt!` no longer re-exported) was fixed incidentally to unblock `cargo test -p mp-block`.

### DIFF
--- a/madara/crates/client/db/src/rocksdb/global_trie/mod.rs
+++ b/madara/crates/client/db/src/rocksdb/global_trie/mod.rs
@@ -69,7 +69,7 @@ pub fn apply_to_global_trie<'a>(
     state_diffs: impl IntoIterator<Item = &'a StateDiff>,
     last_block_protocol_version: StarknetVersion,
 ) -> Result<(Felt, MerklizationTimings)> {
-    let mut last_trie_roots = None;
+    let mut state_root = None;
     let mut timings = MerklizationTimings::default();
 
     for (block_n, state_diff) in (start_block_n..).zip(state_diffs) {
@@ -113,8 +113,14 @@ pub fn apply_to_global_trie<'a>(
         let (contract_trie_root, contract_trie_timings) = contract_result?;
         let (class_trie_root, class_trie_timings) = class_result?;
 
-        // Keep the trie roots from the last iteration — state root is computed once after the loop.
-        last_trie_roots = Some((contract_trie_root, class_trie_root));
+        // NOTE: We compute the state root inside the loop (rather than only after it) so that
+        // the per-block timing metrics below include the cost of `calculate_state_root`.
+        //
+        // Because we use `last_block_protocol_version` for every iteration, intermediate state
+        // roots may be incorrect if the batch spans a protocol version boundary (e.g. 0.13.x →
+        // 0.14.0). This is harmless today — only the final root is returned and verified — but
+        // should be kept in mind if intermediate roots are ever used in the future.
+        state_root = Some(calculate_state_root(contract_trie_root, class_trie_root, last_block_protocol_version));
 
         // Capture timings
         timings.contract_trie_root = contract_duration;
@@ -129,11 +135,8 @@ pub fn apply_to_global_trie<'a>(
         metrics().apply_to_global_trie_last.record(block_secs, &[]);
     }
 
-    let (contract_trie_root, class_trie_root) =
-        last_trie_roots.context("Applying an empty batch to the global trie")?;
-    let state_root = calculate_state_root(contract_trie_root, class_trie_root, last_block_protocol_version);
-
-    Ok((state_root, timings))
+    let root = state_root.context("Applying an empty batch to the global trie")?;
+    Ok((root, timings))
 }
 
 /// "STARKNET_STATE_V0"
--- a/madara/crates/client/db/src/rocksdb/global_trie/mod.rs
+++ b/madara/crates/client/db/src/rocksdb/global_trie/mod.rs
@@ -158,6 +158,7 @@ mod tests {
     use super::*;
     use crate::MadaraBackend;
     use mp_chain_config::ChainConfig;
+    use mp_state_update::{ContractStorageDiffItem, StorageEntry};
     use rstest::*;
     use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


madara/crates/client/db/src/rocksdb/global_trie/mod.rs
```diff
@@ -122,13 +122,23 @@ pub fn apply_to_global_trie<'a>(
...
-        contracts_trie_root
-    } else {
-        Poseidon::hash_array(&[STARKNET_STATE_PREFIX, contracts_trie_root, classes_trie_root])
+    tracing::trace!("global state root calc contracts={contracts_trie_root:#x} classes={classes_trie_root:#x}");
+
+    if contracts_trie_root == Felt::ZERO && classes_trie_root == Felt::ZERO {
+        return Felt::ZERO;
     }
+
+    Poseidon::hash_array(&[STARKNET_STATE_PREFIX, contracts_trie_root, classes_trie_root])
```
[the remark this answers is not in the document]
prkpndy: done

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
```diff
@@ -158,33 +168,27 @@ mod tests {
...
+    // NOT fire — the result is Poseidon(STARKNET_STATE_V0, contract_root, 0), not contract_root.
     #[case::zero_class_trie_root(
-        Felt::from_hex_unchecked("0x123456"),  // Non-zero contracts trie root
-        Felt::from_hex_unchecked("0x0"),       // Zero classes trie root
-        Felt::from_hex_unchecked("0x123456")   // Expected result: same as contracts trie root
+        Felt::from_hex_unchecked("0x3c538d437670f4c6f72dd799f215a007720ec7d19bc64195c96399145d8746f"),
+        Felt::ZERO,
+        Felt::from_hex_unchecked("0x68bcf9e9257ab6bffd9425833a208aaab6b85649fd21c787a546cb7cb9abf")
     )]
+    #[case::both_zero(Felt::ZERO, Felt::ZERO, Felt::ZERO)]
```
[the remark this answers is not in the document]
prkpndy: done

Mohiiit (commented):
A couple of follow-ups in the inline comments, mostly around observability.

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
```diff
@@ -115,30 +129,49 @@ pub fn apply_to_global_trie<'a>(
         metrics().apply_to_global_trie_last.record(block_secs, &[]);
     }
 
-    let root = state_root.context("Applying an empty batch to the global trie")?;
-    Ok((root, timings))
+    let (contract_trie_root, class_trie_root) =
+        last_trie_roots.context("Applying an empty batch to the global trie")?;
+    let state_root = calculate_state_root(contract_trie_root, class_trie_root, last_block_protocol_version);
```
Mohiiit: I think this timing is a bit misleading today. We record `apply_to_global_trie_duration` before `calculate_state_root()` runs, so for batched calls it ends up looking more like "time spent on the last trie update" than "time spent merklizing the batch". Since this is likely to be used operationally, would it make sense to either record it after the final state-root calculation or rename it so the scope is clearer?

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
```diff
@@ -115,30 +129,49 @@ pub fn apply_to_global_trie<'a>(
...
-fn calculate_state_root(contracts_trie_root: Felt, classes_trie_root: Felt) -> Felt {
-    tracing::trace!("global state root calc {contracts_trie_root:#x} {classes_trie_root:#x}");
-    if classes_trie_root == Felt::ZERO {
-        contracts_trie_root
-    } else {
-        Poseidon::hash_array(&[STARKNET_STATE_PREFIX, contracts_trie_root, classes_trie_root])
+/// Computes the global state root from the contract and class trie roots.
+///
+/// Mirrors pathfinder's `StateCommitment::calculate` (`pathfinder/crates/common/src/lib.rs`):
+///
```
Mohiiit, line 151: Small doc suggestion: could we point to the Sequencer / Starknet OS implementation as the canonical reference here, and keep Pathfinder as a compatibility reference? That feels closer to the actual source of truth for this behavior.

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
prkpndy: discussed offline. since the state root calculation time is very small compared to trie roots calculation, we decided to skip this!

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/global_trie/mod.rs
prkpndy: resolved. reverted the optimization for better metrics

Mohiiit (approved):

PR #1024, opened 2026-04-10
fix(settlement): enforce L1 message finality client-side with reorg protection

prkpndy:
## Pull Request type

- [x] Bugfix
- [x] Testing

The L1→L2 message sync uses `eth_newFilter` with `toBlock: "finalized"` for the live watcher and `eth_getLogs` with `toBlock: "finalized"` for the historical query. This has three problems:

1. **Not portable across providers**: Geth rejects `toBlock: "finalized"` in `eth_newFilter` with `errInvalidBlockRange`. Alchemy accepts it but silently ignores it — events are delivered at chain-tip latency, not finality latency.
2. **Coverage gap**: events in the `(finalized, latest]` window at filter creation time are missed by both the historical query and the live watcher.
3. **No reorg protection**: if an event's L1 block is reorged out, the message is processed anyway and the nonce mapping is poisoned — preventing a legitimate re-emission on the new canonical chain from being picked up.

Resolves: #NA

## What is the new behavior?

- **Live watcher uses `toBlock: "latest"`** — works on Geth, Reth, and all hosted providers.
- **Historical query covers up to `latest`** — closes the coverage gap between finalized and latest at filter creation time.
- **Client-side finality enforcement** in `process_finalized_events`:
  1. Confirmation depth check (`l1_messages_finality_blocks`, default 10 blocks / ~2 min).
  2. Canonical block hash check — compares the hash captured at event observation time against the current canonical hash. Drops reorged events without poisoning nonce metadata and without advancing the sync tip (so reconnection re-scans the reorged region).
  3. Transient RPC failures during the canonical check leave the event safely in the queue (peek-before-pop pattern) for retry on the next poll.
- **Default `l1_messages_finality_blocks` bumped to 10** — clears the deepest observed post-PoS reorg (7 blocks, May 2022) with a 3-block margin.

### Tests

- **`e2e_test_reorg_drops_event_and_processes_re_emission`** (Anvil): fires event, triggers `anvil_rollback`, fires a second event with the same nonce on the new canonical chain. Verifies event #1 is dropped, event #2 is processed, and the nonce mapping points to event #2.
- **`test_drops_event_when_block_does_not_exist`** (mock): `get_block_n_hash` returns `None` → event dropped, nonce not poisoned, sync tip unchanged.
- **`test_drops_event_when_block_hash_mismatches`** (mock): `get_block_n_hash` returns a different hash → same assertions.
- **`test_missing_block_hash`** (unit): verifies the new `block_hash` field requirement on event conversion.

No. The `l1_messages_finality_blocks` config field already existed; only the default changed (0 → 10). The new `get_block_n_hash` trait method is additive.

## Other information

The `--slots-in-an-epoch 1` Anvil flag was removed from the test setup — it was only needed for the `Finalized` block tag, which is no longer used in the message sync path.

### DIFF
--- a/madara/crates/client/settlement_client/src/client.rs
+++ b/madara/crates/client/settlement_client/src/client.rs
@@ -110,6 +110,16 @@ pub trait SettlementLayerProvider: Send + Sync {
     /// * Block timestamp in seconds
     async fn get_block_n_timestamp(&self, l1_block_n: u64) -> Result<u64, SettlementClientError>;
 
+    /// Return the canonical block hash at the given block number, or `None` if no block exists
+    /// at that height.
+    ///
+    /// Used by the message sync worker to detect L1 reorgs: when a `LogMessageToL2` event is
+    /// received, its `l1_block_hash` is captured at observation time. Before processing the
+    /// message, the worker calls this function with the event's block number — if the returned
+    /// canonical hash differs from the observed hash, the event's block was reorged out and the
+    /// message must be discarded.
+    async fn get_block_n_hash(&self, l1_block_n: u64) -> Result<Option<[u8; 32]>, SettlementClientError>;
+
     // ============================================================
     // Stream Implementations :
     // ============================================================
--- a/madara/crates/client/settlement_client/src/eth/event.rs
+++ b/madara/crates/client/settlement_client/src/eth/event.rs
@@ -20,6 +20,12 @@ impl TryFrom<(LogMessageToL2, Log)> for MessageToL2WithMetadata {
             l1_block_number: log.block_number.ok_or_else(|| -> SettlementClientError {
                 EthereumClientError::MissingField("block_number in Ethereum log").into()
             })?,
+            l1_block_hash: log
+                .block_hash
+                .ok_or_else(|| -> SettlementClientError {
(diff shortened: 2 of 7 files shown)


Mohiiit (commented):
Nothing blocking from my side. A few non-blocking readability follow-ups if you touch this area again:

- `madara/crates/client/settlement_client/src/messaging.rs`: `process_finalized_events` now owns confirmation gating, canonical-hash checks, origin metadata writes, validity filtering, and queueing. The behavior looks coherent, but most of the readability comes from the comments rather than the structure of the code. It might be worth splitting it into a few small helpers so the peek-before-pop contract and the happy-path vs drop-path logic are easier to scan.
- `madara/crates/client/settlement_client/src/eth/mod.rs`: `messages_to_l2_stream` has a clean flow, but that flow currently lives mostly in the long explanatory comment. Pulling watcher creation, latest-tip snapshot, and historical catchup into named helpers would make the pipeline more self-describing.
- `madara/crates/client/settlement_client/src/eth/mod.rs`: the reorg e2e test is useful and covers the important behavior, but it reads like one long scenario script. Small helpers for emit / rollback / wait / assert phases would make it easier to revisit later.

No blocker from me on these.

Mohiiit (approved):

PR #1027, opened 2026-04-15
feat: add mock prover for mocknet/dev environments

prkpndy:
## Pull Request type

- [x] Feature

The orchestrator supports only `--sharp` and `--atlantic` prover backends. There's no way to run the full state-transition pipeline end-to-end against a local/mocknet environment without hitting a real proving service, which makes dev and integration workflows cumbersome.

Resolves: #NA

## What is the new behavior?

Adds a `--mock` prover backend, L2-only, that runs the full pipeline without producing real proofs:

- **New `orchestrator-mock-service` crate** implementing `ProverClient`. Child proving jobs and `CreateBucket` are no-ops; `RunAggregationWithPie` optionally registers the aggregator fact hash on an L1 `MockGpsVerifier` so the Starknet core contract's `isValid(fact)` check passes during state transition.
- **New `orchestrator/contracts/mock-verifier/`** — Foundry project with `MockGpsVerifier.sol` (anyone can `registerFact`; `isValid` reflects registration state). README carries a clear "never deploy to production" warning.
- **`gps-fact-checker` consolidated** — single inlined `sol!` binding for `isValid` + `registerFact` works for both the real Starkware GPS verifier and our mock (the real one just reverts on `registerFact`, which the mock prover never triggers). Adds a `FactRegistrar` type with:
  - `ensure_registered(fact)` — idempotent: pre-checks `isValid`, sends `registerFact` only if not already registered, **waits for the receipt**, cross-checks `isValid` again, and returns the tx hash (or `B256::ZERO` when no tx was needed).
  - `is_registered(fact)` — thin `isValid` call.
- **`ProverClient` trait** — adds `ApplicativeJobInfo.fact_hash` + `get_aggregation_artifacts`; unifies aggregation task types into `RunAggregation(bucket_id)` (Atlantic) and `RunAggregationWithPie` (SHARP / Mock).
- **Aggregator job handler** branches on `config.prover_kind()`:
  - Atlantic → close bucket (existing).
  - SHARP / Mock → run local aggregator via a new `orchestrator-aggregator-runner` crate, store artifacts, submit the aggregator PIE.
  - Mock additionally stamps the fact hex into `metadata.ensure_on_chain_registration` at process time so `verify_job` can cross-check `isValid` on-chain during verification.
- **CLI** — `--mock` flag; optional `MADARA_ORCHESTRATOR_MOCK_VERIFIER_ADDRESS` env var. When unset, registration is skipped entirely (useful when the core contract points at an always-true verifier stub). Ethereum RPC + key reused from the existing settlement config.

No. `--mock` is additive and mutually exclusive with `--sharp` / `--atlantic`; existing flows are unchanged. L3 + `--mock` is rejected at CLI validation time.

## Other information

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-features --tests --no-deps -- -D warnings` clean.
- `MockGpsVerifier.registerFact` is idempotent on-chain; `FactRegistrar::ensure_registered` additionally short-circuits off-chain by checking `isValid` first.
- No database schema changes.


orchestrator/contracts/mock-verifier/src/IFactRegistry.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.24;
+
+/// @notice Minimal IFactRegistry interface consumed by Starkware's Starknet core contract.
+/// @dev Mirrors the shape at:
```
Mohiiit: Could we make the source of truth a bit clearer for this interface? I’m mostly wondering about drift over time if we keep a local copy of a very small upstream contract shape.

orchestrator/crates/aggregator-runner/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "orchestrator-aggregator-runner"
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+apollo_starknet_os_program = { git = "https://github.com/karnotxyz/sequencer", tag = "v0.14.1-alpha.1" }
```
Mohiiit: Nit: would it help to move both sequencer git deps to workspace deps so the tag is pinned in one place?

orchestrator/crates/aggregator-runner/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+/// 1. Build the bootloader output vector from pre-stored program outputs
+/// 2. Run the aggregator program via `starknet_os::runner::run_aggregator`
+/// 3. Read the DA segment from the temp file where the aggregator wrote it
+/// 4. Validate the output CairoPIE
+pub fn run_local_aggregator(input: AggregatorRunnerInput) -> Result<AggregatorRunnerOutput, AggregatorRunnerError> {
+    if input.child_program_outputs.is_empty() {
+        return Err(AggregatorRunnerError::NoChildOutputs);
+    }
+
+    tracing::info!(
```
Mohiiit, line 61: Would it make sense to add a few dedicated metrics around this local aggregation path too? Right now I only see logs here. A duration/error metric, plus maybe child count or artifact-size metrics, would make Grafana and alerts much more useful.

orchestrator/crates/aggregator-runner/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+        num_children = input.child_program_outputs.len(),
+        "Building aggregator input from child program outputs"
+    );
+
+    // 1. Build bootloader output from pre-stored program outputs
+    let bootloader_output = build_bootloader_output(&input.child_program_outputs);
+
+    // 2. Create temp file for DA segment output.
+    //    The aggregator writes the DA segment to this path during execution.
+    //    TODO: Upstream PR to starknet_os to support in-memory DA segment output.
```
Mohiiit: Could we keep the TODOs in the `TODO(author date): ...` shape here? I find them easier to track later when they include ownership and a timestamp.

Mohiiit (commented):
Overall I like the direction here, and most of my comments are clarifications / maintainability nits rather than pushback on the design itself. The main thing I still feel strongly about is observability: the new local aggregation path looks useful, but right now I mostly see logs and not much dedicated metric coverage around it. I also checked the storage-cleanup path and it looks compatible with the current batch artifact keys, so my note there is more about locking that in with tests / explicitness than a confirmed leak.

orchestrator/crates/aggregator-runner/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+///
+/// This is used in the SHARP L2 flow where the orchestrator must create the
+/// aggregator PIE itself (unlike Atlantic which does it remotely).
+///
+/// # Process
+/// 1. Build the bootloader output vector from pre-stored program outputs
+/// 2. Run the aggregator program via `starknet_os::runner::run_aggregator`
+/// 3. Read the DA segment from the temp file where the aggregator wrote it
+/// 4. Validate the output CairoPIE
+pub fn run_local_aggregator(input: AggregatorRunnerInput) -> Result<AggregatorRunnerOutput, AggregatorRunnerError> {
```
Mohiiit, line 56: Could we add a couple of focused unit tests around this runner too? The E2E coverage helps, but I’d still like small tests for things like empty input and the bootloader output layout.

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -310,17 +336,6 @@ impl ProverClient for AtlanticProverService {
...
-    /// The type of artifacts to be fetched is defined by the `file_name` parameter.
-    ///
-    /// Calls the `get_artifacts` method of `AtlanticClient` defined in `client.rs`
-    /// # Arguments
-    /// `task_id` - the ID of the task for which the artifacts need to be fetched
-    /// `file_name` - the name of the file which is to be fetched
-    ///
-    /// # Returns
-    /// The artifact as a byte array if the request is successful, otherwise an error is returned
     async fn get_task_artifacts(&self, task_id: &str, file_name: &str) -> Result<Vec<u8>, ProverClientError> {
```
Mohiiit, line 341: Could we keep a short comment for this helper too? The old one was a bit long, but I still found the extra context useful when reading the artifact flow.

orchestrator/crates/prover-clients/gps-fact-checker/src/lib.rs
```diff
@@ -38,12 +58,12 @@ impl FromStr for SettlementLayer {
 }
 
 pub struct FactChecker {
-    fact_registry: Option<FactRegistry::FactRegistryInstance<ProviderT>>,
+    fact_registry: Option<GpsVerifier::GpsVerifierInstance<ReadProvider>>,
```
Mohiiit: Small naming nit: this is `GpsVerifier`, but the field is still called `fact_registry`. Would `gps_verifier` be clearer so the terminology stays consistent?

orchestrator/crates/prover-clients/mock-service/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+use orchestrator_gps_fact_checker::{FactCheckerError, FactRegistrar};
+use orchestrator_prover_client_interface::{
+    AggregationArtifacts, ApplicativeJobInfo, ProverClient, ProverClientError, Task, TaskStatus, TaskType,
+};
+use url::Url;
+use uuid::Uuid;
+
+#[derive(Debug, Clone)]
+pub struct MockValidatedArgs {
+    /// When `Some`, the aggregator fact hash is registered on this contract during
```
Mohiiit, line 25: Could we spell out the `None` case a bit more here? I think the behavior is intentional, but without extra context it reads a little like the orchestrator can move forward without fact registration by accident.

orchestrator/crates/prover-clients/mock-service/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+impl MockProverService {
+    pub fn new_with_args(args: &MockValidatedArgs) -> Self {
+        let fact_registrar = args
+            .verifier_address
+            .map(|addr| FactRegistrar::new(args.ethereum_rpc_url.clone(), args.ethereum_signer.clone(), addr));
+        Self { fact_registrar }
+    }
+}
+
+fn parse_fact_hex(fact_hex: &str) -> Result<[u8; 32], ProverClientError> {
```
Mohiiit, line 53: Could we add direct tests for this service too? I’m mainly thinking about `parse_fact_hex`, the no-verifier path, and the bad fact / not-registered cases.

orchestrator/crates/prover-clients/mock-service/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    async fn submit_l2_query(
+        &self,
+        _task_id: &str,
+        _fact: &str,
+        _n_steps: Option<usize>,
+    ) -> Result<String, ProverClientError> {
+        Err(ProverClientError::TaskInvalid("Mock prover does not support L2 queries".to_string()))
+    }
+
+    /// Mock aggregates locally in the handler; artifacts are already written to storage
```
Mohiiit: Just checking my understanding: returning default artifacts is intentional here because Mock already writes the PIE / program output / DA segment during `process_job`, right? If yes, maybe a slightly more explicit comment here would help the next reader.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -71,10 +71,13 @@ impl ProverClient for SharpProverService {
                 let response = self.sharp_client.create_bucket().await?;
                 Ok(response.bucket_id.to_string())
             }
-            Task::CloseBucket(bucket_id) => {
+            Task::RunAggregation(bucket_id) => {
                 self.sharp_client.close_bucket(&bucket_id).await?;
                 Ok(bucket_id)
             }
+            Task::RunAggregationWithPie(_) => Err(ProverClientError::TaskInvalid(
```
Mohiiit, line 78: Is `--sharp` intentionally in a temporary state on this branch, with the real PIE-based aggregation coming in a follow-up PR? If so, would it be safer to reject that path earlier rather than leave it partially wired?

orchestrator/src/types/params/prover.rs
```diff
@@ -1,114 +1,170 @@
...
+            ProverConfig::Sharp(_) => ProverKind::Sharp,
+            ProverConfig::Atlantic(_) => ProverKind::Atlantic,
+            ProverConfig::Mock(_) => ProverKind::Mock,
+        }
+    }
 }
 
 impl TryFrom<RunCmd> for ProverConfig {
     type Error = OrchestratorError;
     fn try_from(run_cmd: RunCmd) -> Result<Self, Self::Error> {
```
Mohiiit, line 45: Would it help to split this into a few prover-specific validator helpers? The logic makes sense, it just feels a bit harder to test and scan all in one `try_from` now.

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -31,31 +33,36 @@ impl JobHandlerTrait for AggregatorJobHandler {
         Ok(job_item)
     }
 
-    /// Note: We confirm before creating an Aggregator job that
-    /// 1. All its child jobs are completed (i.e., ProofCreation jobs for all the blocks in the batch have status as Completed)
-    /// 2. Batch status is Closed (i.e., no new blocks will be added in the batch now)
+    /// Process the aggregator job.
```
Mohiiit, line 36: Minor readability thing: could we call out here that Mock eventually takes the `RunAggregationWithPie` path from `process_job_local_agg` below? I first read this as only `RunAggregation` existing in the handler.

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -65,110 +72,80 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
             }
             TaskStatus::Succeeded => {
-                // Get the aggregator query ID
-                let aggregator_query_id =
-                    config.prover_client().get_aggregator_task_id(&bucket_id).await.map_err(|e| {
-                        error!(
-                            error = %e,
-                            "Failed to get aggregator query ID from prover client"
-                        );
+                // Atlantic returns bytes; SHARP/Mock return empty (already stored during process_job).
```
Mohiiit: Am I reading this right that verification can now succeed even if `get_aggregation_artifacts()` comes back empty? That seems fine for Mock because the handler already stored the PIE / program output / DA segment during `process_job`, but I wasn’t sure if it is also safe for every other prover kind. Would it help to make the expected artifact source explicit here and fail if the required files were never written?

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -204,33 +176,177 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
+        // 4. Store DA segment.
+        config.storage().put_data(bytes::Bytes::from(aggregator_output.da_segment), &metadata.da_segment_path).await?;
+
+        // 5. Convert PIE -> zip bytes (consumed/dropped before bytes are buffered).
+        let pie_bytes = crate::worker::utils::pie::cairo_pie_to_zip_bytes(aggregator_output.aggregator_cairo_pie)
+            .await
+            .map_err(|e| JobError::Other(OtherError(eyre!(e))))?;
+
+        // 6. Persist zip to storage.
+        config.storage().put_data(pie_bytes.clone(), &metadata.cairo_pie_path).await?;
```
Mohiiit, line 328: I checked the storage-cleanup flow and this looks compatible with the current batch artifact keys, which is good. Would it still be worth adding a small cleanup-oriented test for the local aggregation path, just so a future path change doesn’t accidentally leave these objects untagged in S3?

orchestrator/src/worker/utils/pie.rs
```diff
@@ -0,0 +1,52 @@
+//! Helpers for serializing CairoPIEs into bytes.
+//!
+//! These functions are shared by the SNOS and aggregator job handlers to avoid
```
Mohiiit, line 3: Small follow-up: this says the helper is shared by SNOS and aggregator, but `snos.rs` still keeps its own temp-file helper. Would it make sense to dedupe that too while this lands?

orchestrator/src/worker/utils/pie.rs
```diff
@@ -0,0 +1,52 @@
...
+use tempfile::NamedTempFile;
+use tokio::io::AsyncReadExt;
+
+use crate::types::constant::BYTE_CHUNK_SIZE;
+
+/// Convert a [`CairoPie`] into zip bytes via a temp file.
+///
+/// The PIE is consumed and dropped before the bytes are streamed back,
+/// so only one in-memory representation of the PIE exists at a time.
+pub async fn cairo_pie_to_zip_bytes(cairo_pie: CairoPie) -> Result<Bytes> {
```
Mohiiit, line 21: Could we add a small test or two for these helpers as well? A round-trip case and a read-error case would make me more comfortable since they sit on the artifact serialization path.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: this is implemented in the sharp pr

orchestrator/contracts/mock-verifier/src/IFactRegistry.sol
prkpndy: done

orchestrator/crates/aggregator-runner/Cargo.toml
prkpndy: done

orchestrator/crates/aggregator-runner/src/lib.rs
prkpndy: oki

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
prkpndy: done

orchestrator/crates/prover-clients/gps-fact-checker/src/lib.rs
prkpndy: done

orchestrator/crates/prover-clients/mock-service/src/lib.rs
prkpndy: done

orchestrator/crates/prover-clients/mock-service/src/lib.rs
prkpndy: understanding is correct. updated the comment

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: Yeah, we need to do this to keep all provers under the same trait. we are doing the assertion in the job code, to always verify that the expected artifact is received. added some more comment to make this more clear

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: fair point. can also do this in a separate PR. this looks out of scope from the current PR imo. wdyt?

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: done

orchestrator/src/types/params/prover.rs
prkpndy: done

orchestrator/src/worker/utils/pie.rs
prkpndy: done

orchestrator/src/worker/utils/pie.rs
prkpndy: done in sharp pr

orchestrator/crates/prover-clients/mock-service/src/lib.rs
prkpndy: done in sharp pr

orchestrator/crates/aggregator-runner/src/lib.rs
prkpndy: done in sharp pr 🙏

orchestrator/crates/aggregator-runner/src/lib.rs
prkpndy: done in sharp pr

orchestrator/crates/aggregator-runner/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Build the bootloader_output Vec<Felt> from pre-stored program outputs.
+///
+/// Format: `[num_children, output_size_1, program_hash, ...child_1_output..., ...]`
+///
+/// For each child:
+/// - `output_size` = child output length + 2 (for the size and hash fields)
+/// - `program_hash` = pre-computed SNOS program hash from PROGRAM_HASHES
+/// - followed by the child's program output felts
+fn build_bootloader_output(child_outputs: &[Vec<[u8; 32]>]) -> Vec<Felt> {
```
Mohiiit, line 117: Could we add a couple of focused unit tests around the bootloader output construction here? The local aggregation E2E gives happy-path coverage, but it would still be useful to pin the exact `[num_children, output_size, program_hash, ...]` layout and empty-input behavior close to this crate.

orchestrator/src/types/params/prover.rs
```diff
@@ -1,114 +1,184 @@
...
+        atlantic_sharp_prover: args
+            .atlantic_sharp_prover
+            .ok_or_else(|| OrchestratorError::RunCommandError("Atlantic sharp prover is required".to_string()))?,
+        atlantic_artifacts_base_url: args
+            .atlantic_artifacts_base_url
+            .ok_or_else(|| OrchestratorError::RunCommandError("Atlantic artifacts base URL is required".to_string()))?,
+    })
+}
+
+fn validate_mock(
```
Mohiiit, line 154: Could we add focused tests for the new prover validation matrix here? I don’t see coverage for the selection rules (`none` / multiple selected), required Sharp/Atlantic fields, Atlantic L3 requiring `cairo_verifier_program_hash`, or the Mock-specific cases like L3 rejection, missing RPC/private key, invalid private key, and invalid mock verifier address.

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -204,33 +181,177 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
-        let artifact = config.prover_client().get_task_artifacts(task_id, file_name).await.map_err(|e| {
-            error!(error = %e, "Failed to download {}", file_name);
-            JobError::Other(OtherError(eyre!(e)))
+    /// Run the aggregator locally and submit the resulting PIE to the Mock prover.
+    ///
+    /// Computes the aggregator PIE's fact hash via `get_fact_info`, stamps the hex
+    /// into `metadata.ensure_on_chain_registration` (so `verify_job` can cross-check
+    /// `isValid` on-chain), then threads the fact into `ApplicativeJobInfo.fact_hash`
+    /// so the mock prover can register it on the L1 `MockGpsVerifier`.
+    async fn process_job_local_agg(&self, config: &Arc<Config>, job: &mut JobItem) -> Result<String, JobError> {
```
Mohiiit, line 204: Could we split this into smaller helpers? This function now loads child outputs, extracts child job keys, builds aggregator inputs, runs local aggregation, mutates metadata, stores artifacts, and submits the applicative job. Breaking it into load / run / persist / submit helpers would make the behavior easier to test and also line up with the stacked SHARP branch where this path becomes shared.

orchestrator/src/worker/event_handler/triggers/mod.rs
```diff
@@ -11,10 +11,63 @@ pub(crate) mod update_state;
...
+///
+/// The buffer mechanism prevents a stalled low-internal-id job from letting the
+/// pipeline run away ahead of it: newer jobs can be created until the window
+/// (latest - oldest_incomplete + 1) hits `buffer_size`, at which point creation
+/// pauses until the stall clears. Critical for job types whose downstream
+/// consumers must run in strict internal_id order (e.g. state updates).
+///
+/// Shared by [`snos::SnosJobTrigger`] and [`aggregator::AggregatorJobTrigger`];
+/// any trigger whose `internal_id` is monotonically increasing can reuse it.
+pub(crate) async fn calculate_jobs_to_create(
```
Mohiiit, line 47: The pure buffer math is tested below, but could we also cover the DB-backed helper or the aggregator trigger behavior? The new behavior depends on `calculate_jobs_to_create()` correctly reading latest/oldest jobs and then capping `get_aggregator_batches_by_status`, so a small mocked trigger test would give more confidence than the pure math cases alone.

orchestrator/src/worker/event_handler/triggers/aggregator.rs
```diff
@@ -8,26 +8,42 @@ use crate::types::jobs::metadata::{AggregatorMetadata, CommonMetadata, JobMetada
...
+        info!("Creating max {} {:?} jobs", max_to_create, JobType::Aggregator);
+
+        // Fetch up to `max_to_create` closed batches — no point pulling more
+        // than we can legitimately turn into jobs this tick.
         let closed_batches = config
             .database()
             .get_aggregator_batches_by_status(
                 AggregatorBatchStatus::Closed,
-                Some(10),
+                Some(max_to_create as i64),
```
Mohiiit: Small safety nit: `max_to_create` is a `u64`, then it is cast with `as i64` before going into the Mongo limit. Normal values are tiny, but an env misconfiguration above `i64::MAX` would wrap into a negative limit. Could we either bound the config value or use `try_into()` so this fails validation instead of silently wrapping?

Mohiiit (commented):
I have a few minor change requests and follow-up questions after rechecking the latest head.

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: this is already done in sharp pr

orchestrator/crates/aggregator-runner/src/lib.rs
prkpndy: done

orchestrator/src/types/params/prover.rs
prkpndy: will add this in the prover env pr since this code is changed there

orchestrator/src/worker/event_handler/triggers/mod.rs
prkpndy: we already have tests for the db helper functions being used here. so in essence the complete flow is being tested

orchestrator/src/worker/event_handler/triggers/aggregator.rs
prkpndy: this looks like a very extreme env misconfig tbh. there can be many such env misconfigs that can't really be predicted during startuo. i have updated the code in the aggregator/snos code itself to check for overflow

Mohiiit (approved):

PR #1028, opened 2026-04-16
feat(orchestrator): add SHARP gateway and local aggregation flow

prkpndy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

## Summary

Brings the SHARP prover path on par with the Atlantic/Mock paths: the orchestrator can now submit child jobs and applicative aggregation jobs to the current `/v1/gateway/` SHARP API, cross-verify facts on the GPS verifier, and drive the full Atlantic → SHARP switchover without any code changes outside of env.

### What's new

- **SHARP gateway client.** Rewritten around the current `/v1/gateway/` API: `add_job`, `add_applicative_job`, `get_status`, `get_proof`. Uses native-tls (OpenSSL on Linux, SecureTransport on macOS) — rustls rejects SHARP's self-signed CA-as-leaf server cert. mTLS identity is built from PKCS#8 PEM cert + key. On macOS the server cert needs a one-time `security add-trusted-cert` into login keychain (documented in the client source).
- **Shared local-aggregation path.** Both SHARP and Mock drive the aggregator through the same `run_and_submit_with_local_aggregation` flow: load child outputs, run the aggregator, store PIE/DA-segment/program-output to S3, submit `RunAggregationWithPie`. Only Atlantic keeps the bucket-close flow.
- **File-based mTLS config.** Secrets are passed as file paths, not inline PEM, so cert material never hits process args / `/proc/<pid>/environ` / CI logs. Paths are read eagerly during config validation — bad paths fail at startup.
- **Metrics + observability.** SHARP API calls emit a duration histogram (for real p50/p95/p99 per operation), retry counters, idempotency hits, and bounded-label status counters. The aggregator path gets stage-level failure metrics (`load_children`, `run_aggregator`, `fact_hash`, `store_artifacts`, `submit_prover`) plus per-run artifact sizes.
- **Aggregator buffer.** `MADARA_ORCHESTRATOR_AGGREGATOR_JOB_BUFFER_SIZE` gates creation on the `[oldest-incomplete, latest]` internal-id window, mirroring the existing SNOS buffer — stops Created jobs piling up behind a stalled lower-block aggregator.
- **Offchain-proof toggle.** `MADARA_ORCHESTRATOR_SHARP_OFFCHAIN_PROOF` selects between onchain and offchain proof generation.

## Breaking changes (SHARP config only)

Migration:

- `MADARA_ORCHESTRATOR_SHARP_USER_CRT` → `MADARA_ORCHESTRATOR_SHARP_USER_CRT_FILE` (path)
- `MADARA_ORCHESTRATOR_SHARP_USER_KEY` → `MADARA_ORCHESTRATOR_SHARP_USER_KEY_FILE` (path)
- `MADARA_ORCHESTRATOR_SHARP_SERVER_CRT` → `MADARA_ORCHESTRATOR_SHARP_SERVER_CRT_FILE` (path)
- `MADARA_ORCHESTRATOR_SHARP_PROOF_LAYOUT` is no longer read by the SHARP path
- `MADARA_ORCHESTRATOR_SHARP_OFFCHAIN_PROOF` is new (required for SHARP)
- `MADARA_ORCHESTRATOR_AGGREGATOR_JOB_BUFFER_SIZE` is new (defaults to 5)

## Validation

- End-to-end locally against the `gw1.public-testnet.provingservice.io` SHARP gateway with real mTLS certs: child jobs + applicative aggregation + L1 fact registration succeed.
- `cargo clippy --workspace --all-features --tests --no-deps -- -D warnings` clean; orchestrator lib + SHARP service unit tests pass.


orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -185,8 +187,142 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
+                    JobError::Other(OtherError(eyre!(
+                        "ProofCreation job not found for SNOS batch {}",
+                        snos_batch.index
+                    )))
+                })?;
+
+            let proving_metadata: ProvingMetadata = proving_job.metadata.specific.try_into().inspect_err(|e| {
+                error!(error = %e, "Invalid metadata for ProofCreation job");
+            })?;
+            let _ = proving_metadata;
```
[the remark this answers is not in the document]
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/types.rs
```diff
@@ -41,15 +51,7 @@ pub struct SharpGetStatusResponse {
...
-}
-
-/// **IMPORTANT NOTE: THIS IS A MOCK RESPONSE FOR E2E TEST**
-#[derive(Default, Debug, Clone, Serialize, Deserialize)]
-pub struct SharpGetAggTaskIdResponse {
-    pub task_id: String,
+#[derive(Default, Debug, Clone, Deserialize)]
+pub struct SharpGetProofResponse {
+    pub code: Option<String>,
 }
```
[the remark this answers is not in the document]
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -53,125 +41,152 @@ impl SharpClient {
...
             .client
             .request()
             .method(Method::POST)
             .path("add_job")
-            .query_param("cairo_job_key", &cairo_key.to_string())
-            .query_param("offchain_proof", "true")
-            .query_param("proof_layout", proof_layout)
-            .body(encoded_pie)
+            .query_param("cairo_job_key", cairo_job_key)
+            .query_param("offchain_proof", "true") // TODO: don't hardcode it
```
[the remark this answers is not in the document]
prkpndy: we need to hardcode this. removed the TODO

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -109,6 +109,11 @@ impl AtlanticClient {
     /// Retries network-level errors (timeouts, incomplete messages, connection issues)
     /// Does NOT retry API-level errors (4xx, 5xx with proper Atlantic responses) - fast fail
     ///
+    // TODO(@prakhar): Replace this inline retry logic with the shared
```
Mohiiit: Nit: could we keep TODOs in the `TODO(author date): ...` shape here too? I find them easier to track later when they include ownership and a timestamp.

orchestrator/crates/prover-clients/atlantic-service/src/transport.rs
```diff
@@ -3,6 +3,9 @@
 //! This module provides HTTP transport utilities for the Atlantic API client:
 //! authentication header and error classification.
 //!
+//! TODO(@prakhar): The `HttpResponseClassifier` here is a duplicate of
```
Mohiiit: Same nit here for the TODO format.

orchestrator/crates/prover-clients/prover-client-interface/src/http.rs
```diff
@@ -0,0 +1,121 @@
...
+
+            return !has_api_keywords;
+        }
+
+        false
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Mohiiit, line 62: Nit: these tests look like a good `rstest` candidate. It would make the classifier cases a bit more compact and easier to scan.

Mohiiit (commented):
Most of my inline comments are clarification and maintainability questions, but there are a few things I’d want to tighten before merge: observability around the new SHARP local-aggregation path, the new L2 audit-proof download behavior with Mock, and a couple of retry/error-handling details in the SHARP client. One stacked follow-up from the base PR that still feels important is that the shared local aggregator runner also only emits logs today, so the end-to-end observability story for local aggregation still feels a bit thin.

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -1,177 +1,217 @@
...
-            .decode(sharp_params.sharp_server_crt.clone())
-            .expect("Failed to decode sharp server certificate");
+        let cert = sharp_params.sharp_user_crt.as_bytes();
+        let key = sharp_params.sharp_user_key.as_bytes();
+        let server_cert = sharp_params.sharp_server_crt.as_bytes();
 
         let customer_id = sharp_params.sharp_customer_id.clone();
 
+        // rustls wants cert + key combined in a single PEM buffer.
+        let mut combined_pem = Vec::with_capacity(cert.len() + key.len() + 1);
```
Mohiiit: Would it help to pull the PEM-combining logic into a tiny helper? It feels like a reusable little unit now that the mTLS setup is more involved.

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -1,177 +1,217 @@
...
+        let body = SharpAddJobRequest { cairo_pie_encoded: cairo_pie_b64 };
+
+        retry_with_exponential_backoff("sharp_add_job", cairo_job_key, self.retry_config, || async {
+            let response = self
+                .client
+                .request()
+                .method(Method::POST)
+                .path("add_job")
+                .query_param("cairo_job_key", cairo_job_key)
+                .query_param("offchain_proof", CHILD_JOB_OFFCHAIN_PROOF)
```
Mohiiit, line 88: Just checking my understanding: `offchain_proof = true` here is intentional for real SHARP child jobs, not only for a mock/test flow, right?

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -1,177 +1,217 @@
...
+                .query_param("program_hash_function", SHARP_PROGRAM_HASH_FUNCTION)
```
Mohiiit, line 89: Curious if `program_hash_function` is meant to stay a protocol constant here, or if you expect environments where it should be configurable.

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -1,177 +1,217 @@
...
+                .header(CONTENT_TYPE, HeaderValue::from_static("application/json"))
+                .body(&body)
+                .map_err(|e| SharpError::SerializationError(e.into()))?
+                .send()
+                .await
+                .map_err(SharpError::AddJobFailure)?;
+
+            match response.status() {
```
Mohiiit, line 97: I might be missing something, but do we lose the response body here before applying the new infrastructure-error classification? If so, SHARP would still fail fast on LB / gateway-style 404 or HTML responses instead of retrying them like the shared classifier intends.

orchestrator/crates/prover-clients/sharp-service/src/client.rs
```diff
@@ -1,177 +1,217 @@
...
+                .send()
+                .await
+                .map_err(SharpError::AddApplicativeJobFailure)?;
+
+            match response.status() {
+                StatusCode::OK => response.json().await.map_err(SharpError::AddApplicativeJobFailure),
+                code => {
+                    let url = response.url().to_string();
+                    let body = response.text().await.unwrap_or_default();
+                    tracing::debug!(status = %code, url = %url, body = %body, "SHARP non-2xx response");
```
Mohiiit: Curious about the log level here: is `debug!` enough for a non-2xx response, or would `warn!` be more useful operationally? I can also see the case for keeping this quiet since the retry helper already warns.

orchestrator/crates/prover-clients/sharp-service/src/constants.rs
```diff
@@ -1,2 +1,8 @@
-pub const SHARP_FETCH_ARTIFACTS_BASE_URL: &str =
-    "https://raw.githubusercontent.com/madara-alliance/madara-test-artifacts/add/data/PR632";
+pub const SHARP_PROGRAM_HASH_FUNCTION: &str = "PEDERSEN";
+
+/// Child proof-creation jobs always use offchain proof.
+/// Core contract verifies Applicative job's proof.
+pub const CHILD_JOB_OFFCHAIN_PROOF: &str = "true";
```
Mohiiit, line 5: Just checking: SHARP is intentionally L2-only on this branch, so this constant is not expected to vary with any L3 flow, right?

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -39,201 +38,243 @@ pub struct SharpValidatedArgs {
 pub struct SharpProverService {
     sharp_client: SharpClient,
     fact_checker: FactChecker,
-    proof_layout: LayoutName,
+}
+
+/// Encode a [`CairoPie`] struct via the temp-file dance.
```
Mohiiit, line 44: Nice note. Would it make sense to reuse the shared PIE-to-zip helper here too, or does the base64-encoding step make that awkward enough to keep local?

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -39,201 +38,243 @@ pub struct SharpValidatedArgs {
...
-            .sharp_client
-            .get_artifacts(format!("{}/queries/{}/{}", SHARP_FETCH_ARTIFACTS_BASE_URL, task_id, file_name))
-            .await?)
-    }
-
-    #[allow(dead_code)]
-    async fn get_aggregator_task_id(&self, bucket_id: &str) -> Result<String, ProverClientError> {
-        Ok(self.sharp_client.get_aggregator_task_id(bucket_id).await?.task_id)
+        let proof = if include_proof {
+            let proof_json = self.get_proof(external_id).await?;
```
Mohiiit, line 214: Do we actually need to fetch and persist the SHARP proof here, or is this only for the audit-artifacts / debugging path?

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -243,6 +284,25 @@ impl SharpProverService {
...
+    /// `get_task_status` → `TaskStatus::Failed`, letting the operator investigate
+    /// rather than silently retrying with the same inputs.
+    ///
+    /// Returns `None` only when SHARP reports `Unknown` (job not found).
+    async fn check_existing_job(&self, cairo_job_key: &str) -> Result<Option<String>, ProverClientError> {
+        let res = self.sharp_client.get_job_status(cairo_job_key).await?;
+        match res.status {
+            CairoJobStatus::Unknown => Ok(None),
+            status => {
+                tracing::info!(cairo_job_key = %cairo_job_key, status = ?status, "Job already exists on SHARP, skipping submission");
```
Mohiiit, line 320: Minor logging question: when we short-circuit because the job already exists, should this stay `info!`, or would `warn!` make more sense for some states like `Failed` / `Invalid`?

orchestrator/crates/prover-clients/sharp-service/tests/lib.rs
```diff
@@ -71,25 +78,23 @@ async fn prover_client_submit_task_works() {
 #[case(CairoJobStatus::NotCreated)]
 #[case(CairoJobStatus::Processed)]
 #[ignore]
```
Mohiiit: Could we remove the `#[ignore]` here now, or at least extract the status mapping into a smaller unit-testable helper? It would boost confidence in the new SHARP status semantics.

orchestrator/src/tests/config.rs
```diff
@@ -48,6 +50,14 @@ use std::str::FromStr as _;
 use std::sync::Arc;
 use url::Url;
 use uuid::Uuid;
+
+/// Decode a historically base64-wrapped PEM fixture into raw PEM text.
+fn pem_from_env(name: &str) -> String {
```
Mohiiit: Small test-utils nit: `pem_from_env()` now exists in more than one place. Could we move it to a shared test helper?

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -185,8 +183,147 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
 
 // =============================================================================
-// Local-aggregation path (Mock prover only on this branch; SHARP will join once
-// `feat/sharp-integration` lands).
+// SHARP-specific process_job logic
+// =============================================================================
+
+impl AggregatorJobHandler {
+    /// Run the aggregator locally and submit the applicative job to SHARP.
+    async fn process_job_sharp(&self, config: &Arc<Config>, job: &mut JobItem) -> Result<String, JobError> {
```
Mohiiit: Just checking my understanding: the 'all child jobs are completed' guard already happens in the aggregator trigger before we get here, right? Also, this function feels like a good candidate to split into load / aggregate / persist / submit helpers.

Mohiiit: Would it make sense to add dedicated metrics around this SHARP local-aggregation path too? I only see logs here right now. A duration / error metric, plus maybe child count or artifact-size metrics, would make Grafana and alerts much more useful.

orchestrator/src/worker/event_handler/triggers/proving.rs
```diff
@@ -54,8 +54,12 @@ impl JobTrigger for ProvingJobTrigger {
                                     continue;
                                 }
                                 Some(start_snos_batch) => (
-                                    None,
-                                    None,
+                                    if config.params.store_audit_artifacts {
```
Mohiiit, line 57: I think this now breaks the Mock path when `store_audit_artifacts = true`. The L2 recursion branch sets `download_proof`, and `ProvingJobHandler::verify_job()` then calls `get_proof()` after success, but `MockProverService::get_proof()` still returns `TaskInvalid`. Would it make sense to gate this by prover kind, or make Mock return a no-op proof in audit mode?

orchestrator/crates/prover-clients/prover-client-interface/src/retry.rs
```diff
@@ -0,0 +1,137 @@
...
+                        operation = operation_name,
+                        error_type = error.error_type(),
+                        retry_count = outcome.retry_count(),
+                        error = %error,
+                        "API call failed (non-retryable)"
+                    );
+                    return Err(RetryFailure { error, outcome });
+                }
+
+                if outcome.elapsed >= config.timeout {
```
Mohiiit, line 109: Would it make sense to clamp the backoff sleep to the remaining timeout budget here? Right now we check the timeout before sleeping, so a 120s budget can still overshoot by a full delay chunk.

orchestrator/crates/prover-clients/sharp-service/src/client.rs
prkpndy: yes. since sharp doesn't have a mock flow and we don't settle without AR in L2, we would never have a case when we need to register fact for a child job and would always need the fact registered for an applicative job!

orchestrator/crates/prover-clients/sharp-service/src/client.rs
prkpndy: I think this is for computing the fact hash, since that requires program hash. In our code also, we are using pedersen (sequencer repo exports pedersen hashes for programs). Don't think this will change with every chain or env.

orchestrator/crates/prover-clients/sharp-service/src/constants.rs
prkpndy: yes

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: Two problems here:
1. that util doesn't accept a reference
2. that util is present in `src`. we can't access it from `sharp-service` crate :(

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: yes, it's fetched only when download audit artifacts is true from env. check aggregator trigger, we set the download path only when the env var is true!

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: imo a failed job will be detected immediately during verification step of the job (which will happen immediately after this since we'll return the external id directly). this function should only have the responsibility to classify sent/not-sent and log if it's sent. also, info makes more sense imo. wdyt?

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: > Also, this function feels like a good candidate to split into load / aggregate / persist / submit helpers.

Yup. Will split

> Just checking my understanding: the 'all child jobs are completed' guard already happens in the aggregator trigger before we get here, right?

Yes

orchestrator/src/worker/event_handler/triggers/proving.rs
prkpndy: yeah, good point. will add the gating during startup to make sure audit is false with mock prover

orchestrator/crates/prover-clients/prover-client-interface/src/retry.rs
prkpndy: fair. will do

orchestrator/crates/prover-clients/atlantic-service/src/client.rs
prkpndy: done

orchestrator/crates/prover-clients/atlantic-service/src/transport.rs
prkpndy: done

orchestrator/crates/prover-clients/prover-client-interface/src/http.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/client.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/client.rs
prkpndy: The retry wrapper already logs at warn! on retryable failures and warn! on timeout/fast-fail. The debug! here is the raw response body dump, intentionally quiet to avoid polluting logs with large SHARP responses at higher levels.

orchestrator/src/tests/config.rs
prkpndy: done

orchestrator/crates/prover-clients/prover-client-interface/src/retry.rs
prkpndy: done

orchestrator/src/worker/event_handler/triggers/proving.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/tests/lib.rs
prkpndy: i am not sure if testing with sharp is possible. we'll need real certificates for that in the ci. not sure if adding them to gh secrets is safe 🤔

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/tests/lib.rs
prkpndy: have added a new mapping function and tested against that. removed the ignored test. it wanted to make call to sharp

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: done

prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/client.rs
prkpndy: fixed this. checking if the 404 is coming from infra level

orchestrator/crates/prover-clients/sharp-service/tests/lib.rs
```diff
@@ -71,25 +79,24 @@ async fn prover_client_submit_task_works() {
...
```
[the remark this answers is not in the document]
prkpndy: can't call sharp directly since it requires proper certs. created a map function - a util which contains the main logic and does not call sharp. this is being used by the sharp client now. added tests for this function.

Mohiiit (commented):
The earlier runtime issues look addressed on the latest head. I have a few follow-up comments around metric naming and making the new Grafana panels more useful for percentiles and alerts.

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
+use once_cell::sync::Lazy;
+use opentelemetry::global;
+use opentelemetry::metrics::{Counter, Gauge};
+use orchestrator_utils::metrics::lib::{register_counter_metric_instrument, register_gauge_metric_instrument, Metrics};
+use orchestrator_utils::register_metric;
+
+register_metric!(MOCK_METRICS, MockMetrics);
+
+/// Metrics for Mock prover operations.
+pub struct MockMetrics {
```
Mohiiit: Small naming nit: could we call this `MockProverMetrics` instead? It reads a bit clearer next to `SharpMetrics`, since this module is specifically for the mock prover service.

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Duration of fact registration (ensure_registered) calls.
+    pub fact_registration_duration_seconds: Gauge<f64>,
+    /// Fact already registered on-chain (no tx sent).
+    pub fact_already_registered_total: Counter<f64>,
+    /// New fact registered on-chain (tx mined).
+    pub fact_registered_total: Counter<f64>,
```
Mohiiit: I checked `FactRegistrar::ensure_registered()`, and `B256::ZERO` does mean the fact was already registered and no tx was sent, so the split above is correct. Could we make this metric name a bit more explicit though, maybe `fact_newly_registered_total` or `fact_registration_tx_total`? `fact_registered_total` can read like it includes both already-registered and newly-registered facts.

orchestrator/crates/prover-clients/sharp-service/src/metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// Metrics for SHARP API calls exported to OTEL.
+///
+/// Labels:
+/// - `operation`: API endpoint (`add_job`, `add_applicative_job`, `get_status`, `get_proof`)
+/// - `success`: `"true"` / `"false"`
+/// - `error_type`: error variant name when `success=false`, `"none"` otherwise
+pub struct SharpMetrics {
+    /// Duration of SHARP API calls in seconds.
+    pub api_duration_seconds: Gauge<f64>,
```
Mohiiit: The bounded `operation` label looks right, but I checked the Grafana panel and it currently does `avg by (operation)` over this duration gauge. If we want operation-level p50/p95/p99 for `add_job`, `add_applicative_job`, `get_status`, and `get_proof`, could this be a histogram instead, with Grafana using `histogram_quantile(... by (operation, le))`?

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Fact registration errors.
+    pub fact_registration_errors_total: Counter<f64>,
```
Mohiiit, line 18: Since we’re adding Mock prover metrics, could we also expose the error and `isValid` signals in Grafana? I see panels for registration duration / newly registered / already on-chain, but not for `mock_fact_registration_errors_total` or `mock_is_valid_check_duration_seconds`, which are probably the more useful alert/debug signals.

orchestrator/src/utils/metrics.rs
```diff
@@ -331,6 +335,28 @@ impl Metrics for OrchestratorMetrics {
             "bytes".to_string(),
         );
 
+        // Local aggregation metrics (SHARP / Mock paths)
+        let aggregator_local_run_duration = register_gauge_metric_instrument(
```
Mohiiit, line 354: I checked the Grafana updates: local aggregation duration and run count are shown, but `aggregator_child_count` does not seem to be on the dashboard, and failures are only split by `success` without a stage or error type. Could we add the child-count panel and either a failure-stage/error-type label or separate failure metric? Artifact-size metrics for PIE / DA segment / program output would also help explain why a local aggregation run regressed.

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
prkpndy: if is `isValid` is false, the whole job will fail. we would see that in the job failures (it won't be a very useful metric imo, since this shouldn't happen at all)

prkpndy: added metric/panel for fact registration errors (we'll add retry and other gas issue handling in a separate pr. will be useful then)

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
prkpndy: done

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/metrics.rs
prkpndy: done

orchestrator/crates/prover-clients/mock-service/src/metrics.rs
prkpndy: done

orchestrator/src/utils/metrics.rs
prkpndy: done

Mohiiit (approved):

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -39,201 +39,251 @@ pub struct SharpValidatedArgs {
...
+        TaskType::Job => match res.status {
+            CairoJobStatus::Failed => TaskStatus::Failed(res.error_log.clone().unwrap_or_default()),
+            CairoJobStatus::Invalid => {
+                TaskStatus::Failed(format!("Job is invalid: {:?}", res.invalid_reason.clone().unwrap_or_default()))
+            }
+            CairoJobStatus::Unknown => TaskStatus::Failed(format!("Job not found: {}", job_key)),
+            CairoJobStatus::Processed => TaskStatus::Succeeded,
+            CairoJobStatus::InProgress => {
+                if res.validation_done == Some(true) {
+                    TaskStatus::Succeeded
```
heemankv, line 255: **Nitpick (semantic clarification):** Returning `Succeeded` while SHARP still reports `InProgress` (gated on `validation_done=true`) is surprising at first read.

**Why:** If `validation_done=true` means "proof validated but not yet finalized/aggregated server-side," we may mark the child proof-creation job complete before SHARP treats it as fully done — potentially racing downstream aggregator submission against SHARP's own accounting.

**How to apply:** Could you add a one-line comment citing the SHARP doc/contract that says `validation_done=true` at `InProgress` is safe to treat as terminal-success for child jobs? That way the next person touching this doesn't have to go re-read SHARP docs to decide whether to preserve the shortcut.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
```diff
@@ -39,201 +39,251 @@ pub struct SharpValidatedArgs {
...
-        );
-        let job_key = Uuid::from_str(job_key)
-            .map_err(|e| ProverClientError::InvalidJobKey(format!("Failed to convert {} to UUID {}", job_key, e)))?;
-        let res = self.sharp_client.get_job_status(&job_key).await?;
+        let res = self.sharp_client.get_job_status(job_key).await?;
+        let task_label = match task {
+            TaskType::Job => "job",
+            TaskType::Aggregation => "aggregation",
+        };
+        metrics::SHARP_METRICS.record_job_status(task_label, &format!("{:?}", res.status));
```
heemankv: **Nitpick:** Using `format!("{:?}", res.status)` as a metric label value means Rust's `Debug` output becomes a stable external API for your Grafana dashboards and alerts.

**Why:** Renaming a `CairoJobStatus` variant (a routine refactor) would silently change the label values and break any alert querying them, with no compile-time warning.

**How to apply:** Consider an explicit mapping to stable strings, e.g.:
```rust
let status_label = match res.status {
    CairoJobStatus::Unknown => "unknown",
    CairoJobStatus::NotCreated => "not_created",
    CairoJobStatus::InProgress => "in_progress",
    CairoJobStatus::Processed => "processed",
    CairoJobStatus::Invalid => "invalid",
    CairoJobStatus::Failed => "failed",
};
```
Or `impl Display for CairoJobStatus` and use `{}`. Either decouples the metric label surface from the Rust Debug representation.

orchestrator/src/worker/event_handler/jobs/aggregator.rs
```diff
@@ -54,16 +53,24 @@ impl JobHandlerTrait for AggregatorJobHandler {
...
                 })?
             }
-            ProverKind::Mock => self.process_job_local_agg(&config, job).await?,
+            ProverKind::Sharp | ProverKind::Mock => {
+                let prover_label = match config.prover_kind() {
+                    ProverKind::Sharp => "sharp",
+                    ProverKind::Mock => "mock",
+                    _ => unreachable!(),
+                };
+                self.run_and_submit_with_local_aggregation(&config, job, prover_label).await?
```
heemankv: **Nitpick:** The inner `match config.prover_kind()` is redundant — we're already inside the outer `ProverKind::Sharp | ProverKind::Mock` arm, and the `_ => unreachable!()` is a code smell hinting at exactly that.

**How to apply:** A small helper on `ProverKind` would clean this up and also help elsewhere (e.g. metric labels):
```rust
impl ProverKind {
    pub fn label(self) -> &'static str {
        match self {
            ProverKind::Sharp => "sharp",
            ProverKind::Mock => "mock",
            ProverKind::Atlantic => "atlantic",
        }
    }
}
```
Then: `let prover_label = config.prover_kind().label();` before the outer match, and no `unreachable!()`.

orchestrator/src/utils/metrics.rs
```diff
@@ -331,6 +342,56 @@ impl Metrics for OrchestratorMetrics {
...
```
heemankv, line 354: **Nitpick:** `aggregator_local_run_duration` (and the sibling `aggregator_child_count`, `aggregator_program_output_bytes`, `aggregator_da_segment_bytes`, `aggregator_pie_zip_bytes` just below) are all `Gauge<f64>` — `.record()` *overwrites* the previous value per label set, so we only ever see the most recent run.

**Why:** For per-run distributions (latency percentiles, size SLOs) you generally want a histogram. Notably, `sharp-service/src/metrics.rs` gets this right — `api_duration_seconds` is a `Histogram<f64>`. Keeping these as gauges makes percentile SLOs ("p99 aggregator run duration") impossible to query correctly.

**How to apply:** Not blocking — it follows the existing Atlantic gauge pattern — but worth filing a small follow-up to migrate the per-run aggregator metrics to histograms (ideally alongside `atlantic_api_call_duration` in the same sweep).

heemankv (commented):
Great work on this PR! 🎯 Really thoughtful rewrite — the SHARP gateway migration, local aggregation path, and shared CairoPIE serialization hang together cleanly, and a few things stood out as especially well done:

A few things below — mostly description/implementation mismatches plus some nitpicks inline. Nothing blocking.

---

### Description-vs-implementation mismatches

**1. `MADARA_ORCHESTRATOR_SHARP_OFFCHAIN_PROOF` doesn't exist**

The PR description says:
> `MADARA_ORCHESTRATOR_SHARP_OFFCHAIN_PROOF` is added to control whether SHARP should generate offchain proofs

But it's actually hardcoded in `sharp-service/src/constants.rs:5-8` (`CHILD_JOB_OFFCHAIN_PROOF="true"`, `APPLICATIVE_JOB_OFFCHAIN_PROOF="false"`) with no CLI arg or env var.

The hardcoded semantics are actually *correct* (child offchain; applicative onchain for fact registration) — just remove the claim from the description, or drop the `sharp_offchain_proof` mention from the "new behavior" section.

**2. `.env.example` still lists removed `MADARA_ORCHESTRATOR_SHARP_PROOF_LAYOUT`**

Description says it's no longer used, but `.env.example:36` still has: `MADARA_ORCHESTRATOR_SHARP_PROOF_LAYOUT=       # SHARP proof layout`. Worth removing the line so operators don't set a value that has no effect.

---

### Nits not in the diff (can't be inline)

**3. Duplicate `#[error]` message in `ProverClientError`** — `prover-client-interface/src/lib.rs:129-132`

```rust
#[error("Failed to write file: {0}")]
FailedToCreateTempFile(String),      // ← wrong message
#[error("Failed to write file: {0}")]
FailedToWriteFile(String),
```

Pre-existing, but this PR actively uses `FailedToCreateTempFile` via `sharp-service/src/lib.rs:49`. When temp-file creation fails in production, logs will say "Failed to write file" and misdirect triage. Trivial fix to `"Failed to create temp file: {0}"`.

**4. Dead variant `ProverClientError::InvalidJobKey`** — `prover-client-interface/src/lib.rs:126`

`grep -r InvalidJobKey` returns only the definition. Pre-existing dead code, but since this PR touches the file, a small cleanup pass removing unused variants would be nice — not blocking.

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: done

orchestrator/crates/prover-clients/sharp-service/src/lib.rs
prkpndy: done

orchestrator/src/worker/event_handler/jobs/aggregator.rs
prkpndy: done

orchestrator/src/utils/metrics.rs
prkpndy: done

heemankv (approved):
LGTM 🚀 Thanks for the quick turnaround on the review comments!

heemankv (approved):

PR #1030, opened 2026-04-17
feat(orchestrator): add SLA support for alerting service

heemankv:
## Summary
- remove orchestrator-owned SLA metric computation and keep supporting settlement state APIs only
- add block-to-SNOS and block-to-aggregator reverse lookups with a Mongo backfill migration
- expose raw block settlement state for the alerting service and add route/database coverage

## Validation
- skipped for final PR publish per request


prkpndy (approved):
lgtm!!!

PR #1032, opened 2026-04-19
feat(orchestrator): select prover via MADARA_ORCHESTRATOR_PROVER env

prkpndy:
## Summary

Replace the three mutually-exclusive `--sharp` / `--atlantic` / `--mock` bool flags with a single typed `--prover` argument (env: `MADARA_ORCHESTRATOR_PROVER`). Clap's `ValueEnum` on the existing `ProverKind` enforces "exactly one of sharp|atlantic|mock" at parse time, replacing the bespoke bool-counting + `ArgGroup` construction we had before.

Also rolls in two related cleanups:
- `MADARA_ORCHESTRATOR_ETHEREUM_PRIVATE_KEY_FILE` now overrides the non-`_FILE` env for the mock prover, mirroring the pattern Atlantic already uses for its API key.
- Every in-tree call site of the old bool flags — tests, Makefile, k8s manifests, docs, VS Code launch config, e2e harness — updated to the new enum form.


e2e-tests/src/node.rs
```diff
@@ -65,7 +65,7 @@ impl Orchestrator {
         if is_run_mode {
             command.arg("--settle-on-ethereum");
             command.arg("--da-on-ethereum");
-            command.arg("--sharp");
+            command.arg("--prover=sharp");
             command.arg("--mongodb");
             command.arg("--madara-version=0.13.3");
```
Mohiiit, line 70: side-track: this harness still pins `0.13.3`, while the newer e2e config defaults to `0.14.1`.

e2e/src/services/orchestrator/config.rs
```diff
@@ -331,14 +331,32 @@ impl OrchestratorConfig {
...
+        // `self.atlantic` booleans into `--prover <kind>`.
+        //
+        // We match exhaustively and panic on invalid combinations here rather
+        // than emitting bad args and letting the orchestrator sort it out:
+        // - `(true, true)` would emit both `--prover sharp` and `--prover atlantic`;
+        //   clap silently takes the last one (atlantic), hiding the misconfig.
+        // - `(false, false)` emits nothing; the orchestrator then fails at
+        //   startup with a generic "required arg missing" that's harder to
+        //   trace back to the test harness.
+        match (self.sharp, self.atlantic) {
```
Mohiiit: Should this builder model prover selection as an enum now? With only `sharp`/`atlantic` booleans, there is no way to run the orchestrator e2e harness with `--prover mock`, even though the orchestrator CLI now supports it.

orchestrator/src/cli/mod.rs
```diff
@@ -145,7 +140,13 @@ pub struct RunCmd {
     #[clap(flatten)]
     pub starknet_settlement_args: settlement::starknet::StarknetSettlementCliArgs,
 
-    // Prover
+    // Prover: single enum selects which backend to use. The per-prover struct
+    // below that matches the selection is validated in `TryFrom<RunCmd> for
+    // ProverConfig` — the others are ignored. Accepted values: sharp, atlantic,
+    // mock (case-insensitive).
+    #[arg(env = "MADARA_ORCHESTRATOR_PROVER", long, value_enum)]
```
Mohiiit: Can we add focused parser tests for this? The old `ArgGroup` made the selection rule explicit; now the guarantee depends on the required `ProverKind` field plus `required_if_eq("prover", ...)` across flattened structs. Tests for missing `--prover`, invalid value, and selected-prover required args would lock this down.

Mohiiit: The comment says accepted values are case-insensitive, but this arg does not set `ignore_case = true`; clap's `ValueEnum` parser is case-sensitive unless that is enabled. Either add `ignore_case = true` if uppercase env values like `MADARA_ORCHESTRATOR_PROVER=SHARP` should work, or drop the claim so operators don't assume uppercase values are accepted.

Mohiiit (commented):
I have a couple of follow-ups around the new prover selector: one test coverage ask for the clap contract, one e2e harness gap for mock prover support, and a small comment mismatch.

e2e-tests/src/node.rs
prkpndy: 2 things here
1. the artifacts are for 13.3 only
2. this test doesn't run now afaik. we have a new e2e now

e2e/src/services/orchestrator/config.rs
prkpndy: done

orchestrator/src/cli/mod.rs
prkpndy: done

prkpndy: done

e2e-tests/src/node.rs
Mohiiit: so we can remove it then?

Mohiiit (approved):

PR #1035, opened 2026-04-24
feat(db): crash-safe block hash validation with staged trie roots

heemankv:
## Summary
- Adds two-phase staged trie root computation for block hash validation when custom headers are set via admin RPC
- When the computed block hash doesn't match the expected hash, Madara logs an error and shuts down immediately — no data is persisted to disk
- Uses bonsai-trie's new `root_hash_staged()` API ([bonsai-trie PR #48](https://github.com/madara-alliance/bonsai-trie/pull/48)) to compute Merkle roots from in-memory nodes without committing

## How it works
1. **Phase 1 (staged)**: Trie changes are applied in-memory and the global state root is computed via `root_hash_staged()` — nothing is written to RocksDB
2. **Hash check**: If a custom header is set, the computed block hash is validated against the expected hash. On mismatch → `tracing::error!` + `std::process::exit(1)`. The staged tries are dropped, DB stays untouched.
3. **Phase 2 (commit)**: Only on match (or no custom header), the staged tries are committed to RocksDB and the block is saved

**Note**: For full crash safety, run with `--no-save-preconfirmed` (or `MADARA_NO_SAVE_PRECONFIRMED=true`) to prevent preconfirmed blocks from being persisted before the hash check runs.

## Test plan
- [x] Tested with replay tool: block with matching hash closes normally, block with mismatching hash triggers error log and clean shutdown
- [x] Verified on restart after mismatch that no data was persisted for the rejected block
- [x] Verified with `--no-save-preconfirmed` that restart resumes from last confirmed block

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ rocksdb = "0.22"
 librocksdb-sys = "0.17.0"
 
 # Bonsai trie dependencies
-bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "update-starknet-types-core", features = [
+bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "feat/root-hash-staged", features = [
   "std",
 ] }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ rocksdb = "0.22"
 librocksdb-sys = "0.17.0"
 
 # Bonsai trie dependencies
-bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "feat/root-hash-staged", features = [
+bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", tag = "v1.0.0", features = [
   "std",
 ] }
 
--- a/madara/crates/client/db/src/lib.rs
+++ b/madara/crates/client/db/src/lib.rs
@@ -843,10 +843,6 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
             block.header.protocol_version,
             block.header.block_number,
         )?;
-        let merklization_duration = merklization_start.elapsed();
-        let merklization_secs = merklization_duration.as_secs_f64();
-        metrics().apply_to_global_trie_duration.record(merklization_secs, &[]);
-        metrics().apply_to_global_trie_last.record(merklization_secs, &[]);
 
         let header =
             block.header.clone().into_confirmed_header(parent_block_hash, commitments.clone(), global_state_root);
@@ -873,7 +869,10 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
                     block.header.block_number, custom.expected_block_hash, block_hash,
                 );
                 tracing::error!("{msg}");
-                eprintln!("{msg}");
+                #[allow(clippy::print_stderr)]
+                {
+                    eprintln!("{msg}");
+                }
                 std::process::exit(1);
             }
         }
@@ -882,6 +881,13 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
         let contract_trie_root_duration = staged_tries.contract_trie_root_duration;
         let class_trie_root_duration = staged_tries.class_trie_root_duration;
         let (contract_trie_timings, class_trie_timings) = staged_tries.commit(block.header.block_number)?;
+
+        // Record total merklization duration (Phase 1 + Phase 2) to match the sync path's
+        // apply_to_global_trie metric which also covers both compute and commit.
+        let merklization_duration = merklization_start.elapsed();
+        let merklization_secs = merklization_duration.as_secs_f64();
+        metrics().apply_to_global_trie_duration.record(merklization_secs, &[]);
+        metrics().apply_to_global_trie_last.record(merklization_secs, &[]);
 
         // Record per-trie root metrics (histogram + gauge)
         let contract_root_secs = contract_trie_root_duration.as_secs_f64();
--- a/madara/crates/client/db/src/lib.rs
+++ b/madara/crates/client/db/src/lib.rs
@@ -834,16 +834,19 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
         metrics().block_commitments_compute_duration.record(commitments_secs, &[]);
         metrics().block_commitments_compute_last.record(commitments_secs, &[]);
 
-        let (global_state_root, merklization_timings) =
-            self.apply_to_global_trie(block.header.block_number, [&block.state_diff], block.header.protocol_version)?;
-
-        // Copy merklization timings
-        timings.merklization = merklization_timings.total;
-        timings.contract_trie_root = merklization_timings.contract_trie_root;
-        timings.class_trie_root = merklization_timings.class_trie_root;
-        timings.contract_storage_trie_commit = merklization_timings.contract_trie.storage_commit;
-        timings.contract_trie_commit = merklization_timings.contract_trie.trie_commit;
-        timings.class_trie_commit = merklization_timings.class_trie.trie_commit;
+        // Phase 1: Compute the global state root from staged (uncommitted) trie changes.
+        // Nothing is persisted to disk yet — if the custom header hash check fails,
+        // the staged tries are dropped and the DB remains untouched.
+        let merklization_start = Instant::now();
+        let (global_state_root, staged_tries) = self.inner.db.compute_global_trie_staged(
+            &block.state_diff,
+            block.header.protocol_version,
+            block.header.block_number,
+        )?;
+        let merklization_duration = merklization_start.elapsed();
+        let merklization_secs = merklization_duration.as_secs_f64();
+        metrics().apply_to_global_trie_duration.record(merklization_secs, &[]);
+        metrics().apply_to_global_trie_last.record(merklization_secs, &[]);
 
         let header =
             block.header.clone().into_confirmed_header(parent_block_hash, commitments.clone(), global_state_root);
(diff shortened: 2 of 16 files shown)


madara/crates/client/db/src/lib.rs
```diff
@@ -857,13 +860,32 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
+                    computed = %block_hash,
+                    "Block hash mismatch: computed block hash does not match expected. \
+                     No data has been persisted. Shutting down."
+                );
+                std::process::exit(1);
             }
         }
 
+        // Hash check passed (or no custom header). Consume the custom header.
+        self.inner.get_custom_header_with_clear(true);
```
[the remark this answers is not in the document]
heemankv: Fixed — merged the check and clear into a single `guard.take()` so the lock is held across both operations. See a8583b35f.

madara/crates/client/db/src/lib.rs
```diff
@@ -857,13 +860,32 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
+        if let Some(custom) = self.inner.get_custom_header() {
+            if !custom.is_block_hash_as_expected(&block_hash) {
+                tracing::error!(
+                    block_n = block.header.block_number,
+                    expected = %custom.expected_block_hash,
+                    computed = %block_hash,
+                    "Block hash mismatch: computed block hash does not match expected. \
+                     No data has been persisted. Shutting down."
+                );
+                std::process::exit(1);
```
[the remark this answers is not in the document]
heemankv: Good catch — added `eprintln!` with the same message before `process::exit(1)` so the diagnostic hits stderr even if the tracing subscriber buffers. See a8583b35f.

madara/crates/client/db/src/lib.rs
```diff
@@ -857,13 +860,32 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
+        self.inner.get_custom_header_with_clear(true);
+
+        // Phase 2: Persist the staged trie changes to RocksDB.
+        let (contract_trie_timings, class_trie_timings) = staged_tries.commit(block.header.block_number)?;
+
+        // Record merklization timings
+        timings.merklization = merklization_duration;
+        timings.contract_storage_trie_commit = contract_trie_timings.storage_commit;
+        timings.contract_trie_commit = contract_trie_timings.trie_commit;
+        timings.class_trie_commit = class_trie_timings.trie_commit;
```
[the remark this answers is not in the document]
heemankv: Fixed — added timing instrumentation to `compute_global_trie_staged` (wraps each rayon branch with `Instant::now()`/`elapsed()`) and now populating `timings.contract_trie_root` and `timings.class_trie_root`. Also recording the per-trie root metrics (histogram + gauge) that `apply_to_global_trie` records. See a8583b35f.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ rocksdb = "0.22"
 librocksdb-sys = "0.17.0"
 
 # Bonsai trie dependencies
-bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "update-starknet-types-core", features = [
+bonsai-trie = { default-features = false, git = "https://github.com/madara-alliance/bonsai-trie", branch = "feat/root-hash-staged", features = [
```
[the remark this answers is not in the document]
heemankv: Expected — this will be updated to the merged branch (or a tag/SHA) once [bonsai-trie PR #48](https://github.com/madara-alliance/bonsai-trie/pull/48) is merged.

madara/crates/client/db/src/lib.rs
```diff
@@ -834,16 +834,19 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
+        let merklization_start = Instant::now();
+        let (global_state_root, staged_tries) = self.inner.db.compute_global_trie_staged(
+            &block.state_diff,
+            block.header.protocol_version,
+            block.header.block_number,
+        )?;
+        let merklization_duration = merklization_start.elapsed();
+        let merklization_secs = merklization_duration.as_secs_f64();
+        metrics().apply_to_global_trie_duration.record(merklization_secs, &[]);
+        metrics().apply_to_global_trie_last.record(merklization_secs, &[]);
```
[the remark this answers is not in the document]
heemankv: Fixed — moved the `apply_to_global_trie_duration` metric recording to after `staged_tries.commit()` completes (line 882), so it now covers both Phase 1 (staged compute) and Phase 2 (commit), matching what `apply_to_global_trie` records on the sync path. See 82abc1524.

orchestrator/src/worker/event_handler/triggers/batching/replay_bounds.rs
```diff
@@ -0,0 +1,71 @@
...
+    let reference_hash = match reference_result {
+        Ok(MaybePreConfirmedBlockWithTxHashes::Block(block)) => block.block_hash,
+        Ok(MaybePreConfirmedBlockWithTxHashes::PreConfirmedBlock(_)) => {
+            tracing::warn!(block_number, "Replay bounds: block is still pending on reference node");
+            return Ok(());
+        }
+        Err(e) => {
+            tracing::error!(block_number, error = %e, "Replay bounds: failed to fetch block from reference node");
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
heemankv: Fixed — validation now fails closed. RPC errors return `ReplayBoundsError::FetchFailed` and pending blocks return `ReplayBoundsError::BlockPending`, both of which halt the batching loop (same as a hash mismatch). No block passes validation without an actual hash comparison. See 82abc1524.

madara/crates/client/db/src/lib.rs
```diff
@@ -857,13 +856,55 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
+                    "Block hash mismatch at block #{}: expected={}, computed={}. \
+                     No data has been persisted. Shutting down.",
+                    block.header.block_number, custom.expected_block_hash, block_hash,
+                );
+                tracing::error!("{msg}");
+                #[allow(clippy::print_stderr)]
+                {
+                    eprintln!("{msg}");
+                }
+                std::process::exit(1);
```
Mohiiit: Could this return an explicit error instead of exiting the process from DB/write code? Since the staged trie changes have not been committed yet, it seems like the caller could halt block closing while still allowing normal shutdown, telemetry flushing, and tests for this branch.

madara/crates/client/db/src/lib.rs
```diff
@@ -857,13 +856,55 @@ impl<D: MadaraStorage> MadaraBackendWriter<D> {
...
-        if let Some(header) = self.inner.get_custom_header_with_clear(true) {
-            let is_valid = header.is_block_hash_as_expected(&block_hash);
-            if !is_valid {
-                tracing::warn!("Block hash not as expected for {}", block.header.block_number);
+        // Atomically check and clear the custom header to avoid TOCTOU races.
+        let maybe_custom = {
+            let mut guard = self.inner.custom_header.lock().expect("Poisoned lock");
+            guard.take()
+        };
+        if let Some(custom) = maybe_custom {
```
Mohiiit: Should we validate `custom.block_n` before consuming this header? The execution path only applies custom headers when the block number matches, so would it be safer to return an error like `custom header is for block #X but closing block #Y` and leave the header untouched unless it matches?

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
```diff
@@ -61,6 +62,15 @@ impl JobTrigger for AggregatorBatchingTrigger {
             let mut state = state_handler.load_batch_state().await?;
 
             for block_num in start_block..=end_block {
+                if let Some(ref_client) = config.replay_bounds_client() {
+                    if let Err(e) =
+                        replay_bounds::validate_block_hash(config.madara_rpc_client(), ref_client, block_num).await
+                    {
+                        error!(block_num, "Replay bounds: {}, stopping aggregator batching", e);
+                        break;
```
Mohiiit, line 72: Should this propagate the replay-bound failure instead of only breaking the loop? With the current flow, a hash mismatch or reference RPC failure logs an error but the worker can still save partial state and return `Ok(())`, which makes fail-closed validation look successful to job-level monitoring.

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
```diff
@@ -95,6 +96,15 @@ impl JobTrigger for SnosBatchingTrigger {
                     }
                 };
 
+                if let Some(ref_client) = config.replay_bounds_client() {
+                    if let Err(e) =
+                        replay_bounds::validate_block_hash(config.madara_rpc_client(), ref_client, block_num).await
+                    {
+                        error!(block_num, "Replay bounds: {}, stopping SNOS batching", e);
+                        break;
```
Mohiiit, line 106: Same question here: should a replay-bound failure be returned as a worker error after any intentional state preservation? Otherwise pending/reference failures stop this loop but still fall through the normal success path.

orchestrator/src/worker/event_handler/triggers/batching/replay_bounds.rs
```diff
@@ -0,0 +1,77 @@
...
+        Ok(MaybePreConfirmedBlockWithTxHashes::Block(block)) => block.block_hash,
+        Ok(MaybePreConfirmedBlockWithTxHashes::PreConfirmedBlock(_)) => {
+            return Err(ReplayBoundsError::BlockPending { block_number, source: "madara".into() });
+        }
+        Err(e) => {
+            return Err(ReplayBoundsError::FetchFailed { block_number, source: "madara".into(), error: e.to_string() });
+        }
+    };
+
+    let reference_hash = match reference_result {
```
Mohiiit: Could we factor the repeated `get_block_with_tx_hashes` result handling into a small helper, maybe similar in shape to the existing block-version fetch path? Both Madara and reference branches map `Block`, `PreConfirmedBlock`, and provider errors the same way, so a helper would make the mismatch check easier to scan and test.

Mohiiit (commented):
I had a few questions around failure handling and replay-bound validation. Could we also add focused tests for the custom-header mismatch path and the replay-bounds worker behavior, especially since the current `process::exit` path is hard to exercise safely?

madara/crates/client/db/src/lib.rs
heemankv: Fixed — replaced `process::exit(1)` with `anyhow::bail!` so the caller handles shutdown gracefully. Since staged tries aren't committed, the DB remains untouched and the error propagates naturally. Also makes this path testable. See b368f92d1.

madara/crates/client/db/src/lib.rs
heemankv: Good catch — now validating `custom.block_n == block.header.block_number` before consuming the header. If mismatched, we bail with a clear error and leave the header untouched in the mutex for the correct block. See b368f92d1.

orchestrator/src/worker/event_handler/triggers/aggregator_batching.rs
heemankv: Fixed — replay bounds failure is now captured, partial state is saved first, then the error is propagated as `Err(eyre!("Replay bounds validation failed: ..."))` so job-level monitoring sees the failure. See b368f92d1.

orchestrator/src/worker/event_handler/triggers/snos_batching.rs
heemankv: Same fix applied here — replay bounds error is saved, valid partial state is persisted, then the error is returned so the worker reports failure instead of silent success. See b368f92d1.

orchestrator/src/worker/event_handler/triggers/batching/replay_bounds.rs
heemankv: Done — extracted an `extract_block_hash(result, block_number, source)` helper that handles the `Block`/`PreConfirmedBlock`/`Err` mapping for both Madara and reference results. See b368f92d1.

Cargo.toml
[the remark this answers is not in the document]
heemankv: Now pinned to `tag = "v1.0.0"` — see 8898d5e13.

PR #1036, opened 2026-04-24
fix(cairo-native): guard native cache by host fingerprint

Mohiiit:
## Summary
Adds host/ABI metadata for persisted Cairo Native artifacts and clears the existing native disk cache once during DB migration so old artifacts are recompiled with metadata.

```mermaid
flowchart TD
    A[DB upgrade to v13] --> B[Clear native_classes cache]
    B --> C[Runtime native lookup]
    C --> D{Metadata matches host?}
    D -->|yes| E[Load native executor]
    D -->|missing / invalid / mismatch| F[Record metric + remove stale artifact]
    F --> G[Return cache miss]
    G --> H{Compilation mode}
    H -->|async| I[Return VM now + compile native in background]
    H -->|blocking| J[Compile native before returning]
    I --> K[Write artifact + metadata]
    J --> K
```

## Validation
- Format check
- Workspace clippy
- Cairo Native tests
- DB migration tests


madara/crates/client/cairo_native/src/host_fingerprint.rs
```diff
@@ -0,0 +1,237 @@
...
+    {
+        read_linux_cpuinfo_value("vendor_id").unwrap_or_else(|| "unknown".to_string())
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    {
+        // Production nodes share Cairo Native disk cache only on Linux today.
+        // Keep non-Linux builds working for local development and tests without
+        // adding platform-specific probes that do not affect deployed nodes.
+        "unknown".to_string()
```
prkpndy, line 154: nit: should we add one for MacOS as well, since we do dev on it?

prkpndy (approved):
LGTM. GG :rocket
Ques: should we add an ENV for the user to specify if he wants to delete the compiled artifacts on fingerprint mismatch or crash?

madara/crates/client/cairo_native/src/host_fingerprint.rs
Mohiiit: it doesn't support afaik

PR #1044, opened 2026-05-04
test(e2e): support configurable prover

Mohiiit:
## Summary
- Defaults the bridge E2E flow to the local mock prover while keeping prover selection configurable.
- Removes the Atlantic-only secret requirement from the reusable E2E workflow.
- Keeps local E2E binary/profile handling configurable so debug builds can be used outside CI.
- Uses the official KSS SNOS v0.14.2 release tag for generate-pie.

## Validation
- Formatting and focused linting passed locally.
- Bridge deposit/withdraw E2E passed locally with the mock prover before the release-tag update.
- E2E workflow passed on the branch after the release-tag update.


prkpndy (approved):
LGTM. No Blockers.
One thing I realized while reviewing the PR: The make command for running the E2E doesn't work in Linux. Also, we are duplicating the logic for running the E2E in the workflow and the Makefile. Was thinking, if we can have a bash script that can be called from both Makefile and the workflow - would be much easier to test the exact flow that runs on CI, locally on an EC2. Can also do this in a separate PR tho with a TODO in this PR!

e2e/src/setup/config.rs
```diff
@@ -13,12 +13,26 @@ use crate::services::{
...
     mock_verifier::{MockVerifierDeployerConfig, MockVerifierDeployerConfigBuilder, MockVerifierDeployerError},
     mongodb::{MongoConfig, MongoError},
-    orchestrator::{OrchestratorConfig, OrchestratorConfigBuilder, OrchestratorError, OrchestratorMode},
+    orchestrator::{OrchestratorConfig, OrchestratorConfigBuilder, OrchestratorError, OrchestratorMode, ProverKind},
     pathfinder::{PathfinderConfig, PathfinderConfigBuilder, PathfinderError},
 };
 use std::env;
 use std::time::Duration;
 
+fn e2e_prover_kind() -> Result<ProverKind, SetupError> {
```
prkpndy, line 22: Can we implement From for ProverKind?

prkpndy (approved):
Minor suggestion

PR #1046, opened 2026-05-05
Fix trie reverts and repair missed legacy mainnet classes

heemankv:
## Summary

This PR fixes two issues we hit while replaying Starknet mainnet blocks against a snap-synced Madara DB, and adds an operational tool to repair already-synced databases.

The first issue was transaction replay failing with `Class with hash ... is not declared` for historical legacy Cairo 0 classes, even though the node had the correct state root and correct contract-to-class-hash mappings.

The second issue was chain revert / reorg failing with:

`Failed to revert contract trie: GoTo("current_id (...) is not the latest")`

when the Bonsai trie log heads had advanced beyond the confirmed chain tip.

### 1. Fix trie reverts when Bonsai trie heads diverge from the confirmed tip

Madara previously assumed that the confirmed chain tip was always the latest Bonsai trie revision. In practice that invariant can be false: the trie log head may already be ahead of the confirmed head.

This PR changes the revert path to:
- read the latest persisted log id for each Bonsai trie independently
- revert each trie from its actual latest persisted head instead of from the confirmed tip
- skip revert for tries that have no persisted logs
- skip revert for tries whose latest log head is already older than the requested target
- revert and commit on the same trie handle, instead of reverting one handle and committing a fresh one
- commit only tries that actually reverted, so skipped paths do not fabricate a new trie revision at the target block
- log trie-head state before the revert so divergence is visible in diagnostics

This fixes both admin-triggered reverts and sync-triggered reorg handling when trie and chain heads are temporarily out of sync.

### 2. Repair sync-time fetching of historical missing legacy classes

We found that some historical Starknet mainnet gateway state updates omitted legacy class declarations from the state diff. Because of that, Madara could import the correct state, compute the correct state root, and still miss the actual legacy class blob in `class_info`.

This PR updates the gateway class-repair logic to:
- apply `MISSED_CLASS_HASHES` by exact block lookup instead of the old `block_n < 2597` gate
- keep the workaround explicit and data-driven by block number
- fetch the missing legacy classes whenever sync reaches one of the known affected blocks
- keep the repair gated to `ChainId::Mainnet`

This PR adds many known missing historical declarations like:
- block `2930` -> `0x03131fa018d520a037686ce3efddeab8f28895662f019ca3ca18a626650f7d1e` (deployed by tx `0x735807ed81f50a4bcfcaf1cd10324c330dad51c00a7e12a1e56f0d7379ce7c5`)
- block `5982` -> `0x026fe8ea36ec7703569cfe4693b05102940bf122647c4dbf0abc0bb919ce27bd` (`FactoryProxy`)
- block `5985` -> `0x07197021c108b0cc57ae354f5ad02222c4b3d7344664e6dd602a0e2298595434` (`RouterProxy`)

The class-hash repair table tests are updated accordingly.

### 3. Add an offline backfill helper for already-synced DBs

Changing sync logic is enough for fresh syncs or resyncs, but it does not repair a DB that has already synced past the affected historical blocks.

To handle that case, this PR adds an offline helper:

`cargo run --example backfill_legacy_class --release -- --block-number <N> --class-hash <HASH>`

The helper:
- opens the Madara RocksDB directly
- fetches a legacy class from the Starknet mainnet gateway using `(block_number, class_hash)`
- verifies that the fetched legacy class actually hashes to the requested class hash before writing it
- writes the class into the local DB
- verifies that the class is visible from the latest confirmed state
- exits early if the class is already present

This gives us a fast operational recovery path for existing mainnet DBs without requiring a full resync.

## Root Cause

The missing-class issue was not a bad state root and not a corrupt DB.

The root cause was that historical pre-`0.13.2` feeder/gateway state diffs could omit legacy class declarations. Madara still imported the correct contract state and class-hash mappings, so commitments and state roots remained correct, but execution and `starknet_getClass` failed when they needed a class blob that had never been fetched and stored.

The revert issue was separate: the DB revert code assumed `confirmed_head == latest_bonsai_trie_head`, which is not always true.

## Validation

This work was driven by replaying mainnet block `9274729`, specifically transaction:

`0x41271033c388ac367c50753bc3a23584461907ff1d12dbefe762a77e0dc49fe`

Observed progression:
- initially replay failed because class `0x07197021...` (`RouterProxy`) was missing
- after repairing that class, replay advanced and then failed because class `0x026fe8ea...` (`FactoryProxy`) was missing
- after backfilling the second class, replay advanced past the original failing block

A third missing class was discovered later during continued replay:
- class `0x03131fa0...` at block `2930`, deployed by transaction `0x735807ed...`
- this was also backfilled using the offline helper

This confirms that:
- the missing-class root cause was real
- the three added historical class repairs are on the canonical execution path
- the offline backfill helper is sufficient to repair already-synced DBs
- the trie-revert path now has targeted tests for same-handle revert+commit and for skipped paths not fabricating target revisions


heemankv:
@greptile can you re-review?

heemankv (commented):
Self review done, lgtm!

prkpndy (commented):
Review focused on fatal bugs, best practices, and test coverage across the three parts (trie revert, class-repair table, offline backfill). The class-repair part is clean and correctly chain-guarded; findings are concentrated in the revert path and the backfill helper. Inline comments cover the specifics; collecting test-coverage gaps here.

### Test coverage gaps

- **The actual revert path is not integration-tested.** Only the pure decision function `trie_revert_action` (4 branches) and the `latest_bonsai_log_id` reader are covered. The headline fix — `revert_single_trie` + the unconditional `commit(target_id)` when a trie head diverges from the confirmed tip — has no end-to-end test. A test that builds a trie with its log head ahead of (and another behind) the target, runs the revert, and asserts the resulting revisions/state would cover both the fix and the commit-on-skip risk flagged inline.
- **No test for the mainnet chain guard** — that `fixup_missed_mainnet_classes` is *not* applied under a non-mainnet chain config.
- The backfill example is untested (acceptable for an example, but the inline class-hash-verification suggestion would be worth a small assertion).

Nice unit coverage on the pieces that are tested (`trie_revert_action`, `latest_bonsai_log_id`, `fixup_missed_mainnet_classes`, and the `MISSED_CLASS_HASHES` count/spot-checks).

### Note on the class-repair part

Verified the (now table-driven) `fixup_missed_mainnet_classes` is correctly gated on `chain_id == ChainId::Mainnet` (classes.rs:105, unchanged context so not inline-commentable), so the mainnet class hashes can't be injected into Sepolia/devnet/custom syncs at the same block numbers — the main correctness risk in dropping the `block_n < 2597` gate. Worth a small test asserting the fixup is a no-op under a non-mainnet chain config so this guard can't regress.

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -833,38 +937,52 @@ impl MadaraStorageWrite for RocksDBStorage {
...
-        self.class_trie()
-            .revert_to(target_id, current_id)
-            .map_err(|e| anyhow::anyhow!("Failed to revert class trie: {e:?}"))?;
-        tracing::info!("✅ REORG: Class trie reverted successfully");
+        let mut class_trie = self.class_trie();
+        revert_single_trie("class", &mut class_trie, trie_log_heads.class, target_block_n)?;
 
         tracing::info!("💾 REORG: Committing tries after revert...");
-        self.contract_trie()
+        contract_trie
```
prkpndy: **[Medium — please verify] `commit(target_id)` runs unconditionally, even for tries whose revert was skipped.** `revert_single_trie` no-ops on the `AlreadyAtTarget` / `OlderThanTarget` / `Missing` branches, but all three tries are still committed at `target_id` here. Per branch:

- `AlreadyAtTarget(8)` + `commit(8)` re-commits a `BasicId` that already exists as the trie head — depending on Bonsai semantics this can overwrite the existing trie-log entry at `target_id` with an empty changeset, which would silently break a *future* revert that needs to cross that block.
- `Missing` + `commit(8)` commits an empty trie at the target, fabricating a revision for a trie that had no logs.
- `OlderThanTarget {current:5, target:8}` + `commit(8)` fabricates revision 8 from the block-5 state.

The `Revert` case genuinely needs the commit (and the original code committed `target_id` after `revert_to`, so that's presumably valid for Bonsai). But committing on the *skipped* paths is new behavior on a data-integrity path. Suggest having `revert_single_trie` return whether it actually reverted and committing only those, or confirm with the Bonsai API that re-committing an existing/non-greater id is a safe no-op. Either way, please add an integration test for the divergent-head case.

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -833,38 +937,52 @@ impl MadaraStorageWrite for RocksDBStorage {
...
+                );
+            }
+        }
 
         tracing::debug!("🌳 REORG: Reverting contract trie...");
-        self.contract_trie()
-            .revert_to(target_id, current_id)
-            .map_err(|e| anyhow::anyhow!("Failed to revert contract trie: {e:?}"))?;
-        tracing::info!("✅ REORG: Contract trie reverted successfully");
+        let mut contract_trie = self.contract_trie();
```
prkpndy, line 964: Worth calling out: reverting and then committing on the **same** `contract_trie` handle (vs. the original code, which committed on a *fresh* `self.contract_trie()`) looks like a real part of the fix — if the handle carries in-memory staging, the old code could have committed a handle that never saw the revert. Good change; please make sure the PR description mentions it and that a test locks it in, since it's easy to regress back to a fresh handle.

madara/crates/client/db/src/rocksdb/mod.rs
```diff
@@ -173,6 +177,33 @@ impl RocksDBStorageInner {
 
         Ok(())
     }
+
+    /// Bonsai trie log keys are ordered by the committed revision id first.
+    ///
+    /// Madara uses `bonsai_trie::id::BasicId`, and bonsai serializes that id as a big-endian
+    /// `u64` (`BasicId::to_bytes`). That means the lexicographically-last key in a trie-log
+    /// column belongs to the latest committed revision for that trie.
+    fn latest_bonsai_log_id(&self, column: Column) -> anyhow::Result<Option<u64>> {
```
prkpndy, line 186: **[Low] Coupled to Bonsai's internal key layout.** This assumes (a) the first 8 bytes of every trie-log key are the big-endian `BasicId` and (b) `IteratorMode::End` therefore yields the highest revision. True for the current `BasicId::to_bytes`, and the new test pins it — good. But it reaches into Bonsai's on-disk encoding with no public-API guarantee; a Bonsai upgrade that changes key framing (e.g. adds a column/trie prefix) would silently return wrong heads and mis-drive the revert. Consider sourcing the latest id via a Bonsai-provided accessor if one exists, so a version bump fails loudly instead of silently. The doc-comment is a good mitigation in the meantime.

madara/examples/backfill_legacy_class.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let provider = GatewayProvider::starknet_alpha_mainnet();
+    let class = provider
+        .get_class_by_hash(class_hash, BlockId::Number(args.block_number))
+        .await
+        .with_context(|| format!("Fetching legacy class {class_hash:#x} at block {}", args.block_number))?;
+
+    let converted = match class {
+        mp_class::ContractClass::Legacy(contract_class) => {
+            ConvertedClass::Legacy(LegacyConvertedClass { class_hash, info: LegacyClassInfo { contract_class } })
```
prkpndy, line 82: **[Medium] No class-hash verification before writing to the DB.** This wraps the gateway response using the user-supplied `class_hash` without recomputing/asserting it. The normal sync path verifies class hashes (`verify_compile_classes`); this tool writes straight to RocksDB and skips that. A wrong/buggy/malicious gateway response (or a typo'd `--class-hash`) would store a class blob under a hash that doesn't match its contents, and execution / `starknet_getClass` would then trust it. `ContractClass::compute_class_hash()` is available — recompute the legacy class hash from `contract_class` and `bail!` unless it equals `class_hash` before calling `write_classes`. For a direct DB-repair tool this guard matters.

madara/examples/backfill_legacy_class.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+        mp_class::ContractClass::Sierra(_) => {
+            bail!("Class {class_hash:#x} is Sierra, but this helper only backfills legacy classes")
+        }
+    };
+
+    rayon::ThreadPoolBuilder::new()
+        .num_threads(1)
```
prkpndy, line 92: **[Low]** `setup_rayon_threadpool()` already built a global pool sized to `available_parallelism()`, and the write then runs inside this separate 1-thread local pool. If the global pool is required by backend internals, fine; otherwise one of the two is redundant. A one-line comment on why the local single-thread pool is needed (presumably to serialize the single `write_classes`) would clarify intent.

madara/crates/client/db/src/rocksdb/mod.rs
heemankv: Fixed in `0ac3fce34`.

I updated the PR description to call out the same-handle revert/commit behavior, and added a focused regression test for that path in [mod.rs](/Users/heemankverma/Work/Carnot/repos/madara/worktrees/fix-revert-and-missed-class/madara/crates/client/db/src/rocksdb/mod.rs:1108).

madara/examples/backfill_legacy_class.rs
heemankv: Fixed in `0ac3fce34`.

The backfill helper now recomputes the fetched legacy class hash and bails on mismatch before writing to RocksDB in [backfill_legacy_class.rs](/Users/heemankverma/Work/Carnot/repos/madara/worktrees/fix-revert-and-missed-class/madara/examples/backfill_legacy_class.rs:74).

madara/examples/backfill_legacy_class.rs
heemankv: Fixed in `0ac3fce34`.

I added a comment explaining why the dedicated single-thread pool is intentional here in [backfill_legacy_class.rs](/Users/heemankverma/Work/Carnot/repos/madara/worktrees/fix-revert-and-missed-class/madara/examples/backfill_legacy_class.rs:89).

madara/crates/client/db/src/rocksdb/mod.rs
heemankv: Fixed in `0ac3fce34`.

`revert_single_trie` now returns whether a real revert happened, and [mod.rs](/Users/heemankverma/Work/Carnot/repos/madara/worktrees/fix-revert-and-missed-class/madara/crates/client/db/src/rocksdb/mod.rs:979) only commits tries that actually reverted. I also added trie-level tests covering both same-handle revert+commit and skipped paths not fabricating target revisions in [mod.rs](/Users/heemankverma/Work/Carnot/repos/madara/worktrees/fix-revert-and-missed-class/madara/crates/client/db/src/rocksdb/mod.rs:1108).

prkpndy (approved):

PR #1048, opened 2026-05-11
refactor: compile replay support behind cargo features

heemankv:
## Summary
- replace the earlier runtime replay switch with a compile-time `replay` Cargo feature on the Madara and orchestrator workspaces
- move replay-only admin RPCs, custom header plumbing, and replay-bounds wiring behind `#[cfg(feature = "replay")]` so they are not compiled into default builds
- keep the default build clean while preserving replay behavior and tests when the feature is explicitly enabled

## Testing
- `RUST_BUILD_DOCKER=1 cargo check -p madara`
- `RUST_BUILD_DOCKER=1 cargo check -p madara --features replay`
- `RUST_BUILD_DOCKER=1 cargo test -p mc-db --features replay test_custom_header`
- `RUST_BUILD_DOCKER=1 cargo test -p mc-rpc --features replay versions::admin::v0_1_0::methods::write::tests`
- `RUST_BUILD_DOCKER=1 cargo test -p orchestrator prover_selection_parser`
- `PATH="/tmp/sequencer_venv/bin:$PATH" RUST_BUILD_DOCKER=1 cargo test -p orchestrator --features replay prover_selection_parser`


madara/crates/client/block_production/src/util.rs
```diff
@@ -206,6 +209,16 @@ pub(crate) fn create_execution_context(
         (SystemTime::now(), gas_prices)
     };
 
+    #[cfg(not(feature = "replay"))]
+    let (block_timestamp, gas_prices) = {
+        let l1_gas_quote = backend
+            .get_last_l1_gas_quote()
+            .context("No L1 gas quote available. Ensure that the L1 gas quote is set before calculating gas prices.")?;
```
heemankv: We can make this better written by doing the code of #[cfg(not(feature = "replay"))] explicitly first and then doing the code of #[cfg(feature = "replay")] later

so : 

Code for that runs by default, regardless replay no replay ->
let l1_gas_quote = backend
            .get_last_l1_gas_quote()
            .context("No L1 gas quote available. Ensure that the L1 gas quote is set before calculating gas prices.")?;
 let gas_prices = backend.calculate_gas_prices(&l1_gas_quote, previous_l2_gas_price, previous_l2_gas_used)?;
        (SystemTime::now(), gas_prices)
    };

and then the code for : 
#[cfg(feature = "replay")]
only the if case is needed to be written here

madara/crates/client/db/src/view/mod.rs
```diff
@@ -96,6 +99,20 @@ impl<D: MadaraStorageRead> MadaraBackend<D> {
...
+                            &quote,
+                            parent_block_info.header.gas_prices.strk_l2_gas_price,
+                            parent_block_info.total_l2_gas_used,
+                        )?
+                    } else {
+                        parent_block_info.header.gas_prices
+                    };
+                    (SystemTime::now(), gas_prices)
+                };
+
```
heemankv, line 117: same as above here !

madara/crates/client/db/src/lib.rs
```diff
@@ -311,16 +314,7 @@ pub struct MadaraBackend<DB = RocksDBStorage> {
...
-    ///
-    /// When replaying a block, we must match the exact timestamp and gas configuration
-    /// from the original block to reproduce the expected block hash. These per-block
-    /// overrides are applied during transaction validation and execution, along with the
-    /// expected block hash to validate against after block creation.
-    /// # Important Notes
-    /// - Custom headers are keyed by block number because replay can prepare future blocks ahead of time
-    /// - **Must verify** that the block number matches before use
-    /// - **Must clear** the matching block entry after use to prevent reuse across different blocks
-    /// - Access is thread-safe via Mutex to allow concurrent operations
```
heemankv, line 323: why did you remove these comments, they are informative

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
heemankv, line 1: whole file looks very unclean now that we've added #[cfg(feature = "replay")] in so many places.
Any better way to write this ?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
```diff
@@ -249,20 +280,7 @@ impl MadaraWriteRpcApiV0_1_0Server for Starknet {
...
-    ) -> RpcResult<L1HandlerTransactionResult> {
-        Ok(self
-            .block_prod_handle
-            .as_ref()
-            .ok_or(StarknetRpcApiError::UnimplementedMethod)?
-            .submit_l1_handler_transaction(l1_handler_message)
-            .await
-            .map_err(StarknetRpcApiError::from)?)
-    }
-
```
heemankv, line 265: why removed ?

madara/crates/client/db/src/view/mod.rs
heemankv: Done. I applied the same cleanup here: the fake preconfirmed default path is now the main helper, and replay only overrides it when a staged custom header exists.

madara/crates/client/db/src/lib.rs
heemankv: Done. I restored the explanatory comment block on `custom_headers`; the field stays feature-gated, but the docs are still useful context when replay is compiled in.

madara/crates/client/block_production/src/util.rs
heemankv: Done. I pulled the default gas/timestamp path into a shared helper and now apply the replay custom-header override afterwards, so the non-replay path reads first and replay only overrides it when present.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
heemankv: Not removed, just separated. The replay-only methods moved into `write_replay.rs`, while the common admin write path stays in `write.rs`; the split keeps the same behavior but makes the shared file cleaner.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/write.rs
heemankv: Done. I split the replay-only admin RPC implementation into `methods/write_replay.rs` and left `write.rs` with only the common admin write methods. That removes most of the `#[cfg(feature = "replay")]` noise from the shared file.

PR #1055, opened 2026-05-18
Clear stale runtime exec config across mode changes

heemankv:
## Summary
- clear saved runtime execution config on full-node startup so sequencer-only recovery metadata is not carried across mode changes
- add a sequencer startup flag to discard an open preconfirmed block and resave the current runtime config instead of re-executing it
- add focused coverage for runtime-config clearing and discard-on-startup recovery

## Testing
- cargo test -p mc-db test_clear_runtime_exec_config
- cargo test -p mc-block-production test_discard_preconfirmed_on_startup_replaces_runtime_exec_config

## Notes
- I stopped a broad `cargo check -p madara --bin madara` after the focused tests passed because it was still chewing through the full dependency graph; the node wiring changed in `main.rs`, `BlockProductionService`, and the block-production CLI params.

### DIFF
--- a/madara/crates/client/block_production/src/lib.rs
+++ b/madara/crates/client/block_production/src/lib.rs
@@ -322,6 +322,7 @@ pub struct BlockProductionTask {
     l1_client: Arc<dyn SettlementClient>,
     bypass_tx_input: Option<mpsc::Receiver<ValidatedTransaction>>,
     no_charge_fee: bool,
+    discard_preconfirmed_on_startup: bool,
 }
 
 impl BlockProductionTask {
@@ -330,6 +331,8 @@ impl BlockProductionTask {
     /// # Parameters
     ///
     /// * `no_charge_fee`: Determines whether fees are charged during transaction execution.
+    /// * `discard_preconfirmed_on_startup`: Drops any recovered preconfirmed block instead of
+    ///   re-executing and closing it during sequencer startup.
     ///
     /// # TODO(mohit 18/11/2025): Update the code to use config same as pre-close
     pub fn new(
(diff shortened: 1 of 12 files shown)


madara/crates/client/block_production/src/lib.rs
```diff
@@ -660,6 +665,31 @@ impl BlockProductionTask {
             return Ok(());
         }
 
+        if self.discard_preconfirmed_on_startup {
+            let preconfirmed_view =
+                self.backend.block_view_on_preconfirmed().context("Getting preconfirmed block view")?;
+            let block_number = preconfirmed_view.block_number();
+            let n_txs = preconfirmed_view.num_executed_transactions();
+
+            tracing::warn!(
```
prkpndy, line 680: nit: should we also log the transaction hashes, since once lost we mostly don't have any reference for them! can be useful in case something goes wrong

madara/crates/client/block_production/src/lib.rs
```diff
@@ -660,6 +665,31 @@ impl BlockProductionTask {
...
+            let backend = self.backend.clone();
+            global_spawn_rayon_task(move || {
+                backend.write_access().clear_preconfirmed().context("Discarding preconfirmed block on startup")
+            })
+            .await?;
+
+            self.save_current_runtime_exec_config()
+                .context("Saving runtime execution config after discarding preconfirmed block")?;
+
+            tracing::info!("🧹 Discarded preconfirmed block #{} on startup", block_number);
```
prkpndy, line 696: emojis okay?

prkpndy (approved):
lgtm

madara/crates/client/block_production/src/lib.rs
heemankv: Done! I now log the discarded transaction hashes on the warning itself as a structured field (`discarded_transaction_hashes`) so we retain a reference to exactly what was dropped.

madara/crates/client/block_production/src/lib.rs
heemankv: Keeping the emoji here. We already use emoji-prefixed startup logs in this codepath, so this stays consistent with the existing logging style.

PR #1058, opened 2026-05-18
Add SNOS batch details endpoint for alert consumers

Mohiiit:
## Summary
- add a dedicated SNOS batch details endpoint that returns SNOS batch gas metrics and the parent aggregator batch blob length
- expose the endpoint through the orchestrator router so external alert consumers can read stable batch metadata without recomputing batch math
- add server coverage for parented and standalone SNOS batches

## Notes
- This PR is intended to support a companion alerting-service change that consumes the new endpoint.
- Local Rust verification was environment-blocked in this worktree because the Cairo 0 toolchain on PATH is `0.14.0.1` while the build expects `0.14.1a0`.

### DIFF
--- a/orchestrator/src/core/client/database/mod.rs
+++ b/orchestrator/src/core/client/database/mod.rs
@@ -15,6 +15,31 @@ use async_trait::async_trait;
 pub use error::DatabaseError;
 use std::time::Instant;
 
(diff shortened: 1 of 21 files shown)


heemankv (approved):
Reviewed the updated implementation. The new batch query and settlement-status path look good to me.

PR #1061, opened 2026-05-18
fix(external-db): persist retention cursor safely

Mohiiit:
## Summary
- persist the external-db retention cursor in RocksDB metadata
- advance the cursor only after successful Mongo retention deletes
- fail loudly when a confirmed block expected by retention is missing from local DB state
- use rstest for the retention cursor roundtrip coverage

## Notes
- retention cleanup deletes Mongo documents by confirmed block tx hashes and chain id
- local focused test verification is still pending because Cargo lock contention blocked completion in this workspace

### DIFF
--- a/.db-versions.yml
+++ b/.db-versions.yml
@@ -18,7 +18,7 @@
 # Current database schema version.
 # This is the version that new databases will be created with,
 # and the version that existing databases will be migrated to.
-current_version: 13
+current_version: 14
 
 # Minimum version that can be migrated from.
 # Databases older than this must be deleted and re-synced from scratch.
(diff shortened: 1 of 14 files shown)


heemankv (commented):
Changes look good overall. While reviewing, I realized the external DB outbox/worker can still be enabled in full node mode today.

We might want to gate this behind `is_sequencer()` / `should_run_mempool()` the same way other sequencer-only features are gated.

Right now:
- the external DB worker service can be configured and activated independently of node role
- the outbox producer path is tied to mempool acceptance, which is really a sequencer/devnet concern
- on a full node this is at best unnecessary and at worst a confusing mode to support

I do not think this should block this PR if we are happy with the current behavior, but it seems worth tightening either in this PR or in a small follow-up by rejecting/ignoring external DB in full node mode explicitly.

heemankv (approved):

heemankv (approved):
Looks good now. The full-node gating concern is addressed, and the retention cursor persistence/migration path looks solid on the current head.

PR #1066, opened 2026-05-19
fix(db): use leveled compaction for bonsai *_log CFs

prkpndy:
Universal compaction defers L0->L1 work via size-ratio and size-amplification heuristics that do not fire on append-only `bonsai_*_log` CFs (Bonsai TrieLog: one large entry per block, monotonic keys, no dedup potential). L0 SST files accumulated past the per-CF stop trigger, and because `atomic_flush(true)` is enabled the stall on one CF blocked all writes DB-wide, hanging block production at the first put_cf_opt on the meta column.

Switch the three `bonsai_*_log` CFs (contract / contract-storage / class) to leveled compaction, which drains L0->L1 continuously as soon as L0 exceeds level0_file_num_compaction_trigger=4. Other CFs keep universal. Also re-apply the per-CF L0 slowdown/stop triggers after optimize_universal_style_compaction() since that call resets them to RocksDB's hardcoded defaults regardless of the global Options / CLI flags / env vars.

Verified on a real DB: bonsai_contract_storage_log L0 dropped from 33 to 0 within ~5 minutes of running the patched binary, with no stall and block production starting normally.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Resolves: #NA

### DIFF
--- a/madara/crates/client/db/src/rocksdb/options.rs
+++ b/madara/crates/client/db/src/rocksdb/options.rs
@@ -184,11 +184,13 @@ impl std::fmt::Display for DbWriteMode {
 
 #[derive(Debug, Clone)]
 pub struct RocksDBConfig {
-    /// Enable statistics. Statistics will be put in the `LOG` file in the db folder. This can have an effect on performance.
+    /// Enable statistics for compaction/stall monitoring. Uses ExceptDetailedTimers by default
+    /// which has minimal overhead but enables all counter-based metrics (write amplification,
+    /// compaction throughput, stall time). Disable only if profiling shows measurable impact.
     pub enable_statistics: bool,
     /// Dump statistics every `statistics_period_sec`.
     pub statistics_period_sec: u32,
-    /// Statistics level. This can have an effect on performance.
+    /// Statistics level. ExceptDetailedTimers is recommended for production.
     pub statistics_level: StatsLevel,
     /// Memory budget for blocks-related columns
     pub memtable_blocks_budget_bytes: usize,
@@ -242,9 +244,9 @@ pub struct RocksDBConfig {
 impl Default for RocksDBConfig {
     fn default() -> Self {
         Self {
-            enable_statistics: false,
+            enable_statistics: true,
             statistics_period_sec: 60,
-            statistics_level: StatsLevel::All,
+            statistics_level: StatsLevel::ExceptDetailedTimers,
             // TODO: these might not be the best defaults at all
             memtable_blocks_budget_bytes: 1 * GiB,
             memtable_contracts_budget_bytes: 128 * MiB,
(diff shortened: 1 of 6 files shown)


madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -505,26 +513,45 @@ impl Column {
...
+            // for read-heavy or balanced CFs). Sets write_buffer_size, memtable
+            // counts, L0 thresholds, and num_levels for the given memory budget.
+            options.optimize_universal_style_compaction(budget_bytes);
         }
 
+        // IMPORTANT: `optimize_universal_style_compaction` resets the L0 stall
+        // triggers to RocksDB's hardcoded defaults (slowdown=20, stop=36),
+        // ignoring whatever we set globally / via CLI / via env. Re-apply them
+        // here so our configured values actually take effect per-CF. Setting
+        // them on the leveled branch is harmless (matches the explicit values).
```
Mohiiit: Would keep this per-CF.

These thresholds are CF-scoped, and this DB is opened with explicit CF descriptors, so setting them only in the DB-wide builder is not enough for the named CFs.

The comment may just need rewording: the issue is CF scoping, not `optimize_universal_style_compaction()` resetting the values.

madara/crates/client/db/src/rocksdb/options.rs
```diff
@@ -505,26 +513,45 @@ impl Column {
...
-                options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
-                options.optimize_universal_style_compaction(config.memtable_other_budget_bytes);
-            }
+        let budget_bytes = match self.budget_tier {
+            ColumnMemoryBudget::Blocks => config.memtable_blocks_budget_bytes,
+            ColumnMemoryBudget::Contracts => config.memtable_contracts_budget_bytes,
+            ColumnMemoryBudget::Other => config.memtable_other_budget_bytes,
+        };
+
+        options.set_memtable_prefix_bloom_ratio(config.memtable_prefix_bloom_filter_ratio);
```
Mohiiit, line 516: Do we need all of these settings, or only the ones that change behavior?

`set_compaction_style(Level)` makes sense, and I can also see the case for `write_buffer_size`, `max_write_buffer_number`, and the slowdown/stop thresholds.

A few others look redundant here:
- `min_write_buffer_number_to_merge(2)` is sanitized back to `1` with `atomic_flush=true`
- `level_zero_file_num_compaction_trigger(4)` looks like the default
- `target_file_size_base(64 MiB)`, `target_file_size_multiplier(1)`, `num_levels(7)`, and `max_bytes_for_level_multiplier(10.0)` also look like defaults

If `max_bytes_for_level_base(512 MiB)` is intentional, a short note would help.

Mohiiit (commented):
Looks good overall.\n\nTwo small nits from my side:\n- the per-CF slowdown/stop thresholds seem worth keeping\n- some of the leveled-compaction settings either restate defaults or do not take effect with atomic_flush=true\n\nI think this gets clearer if we trim to the knobs that actually change behavior.

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: done

madara/crates/client/db/src/rocksdb/options.rs
prkpndy: done

Mohiiit (approved):

PR #1067, opened 2026-05-20
feat(mempool): add reject-new policy and admin RPCs

heemankv:
## Summary
- add a new optional `mempool_full_policy` chain-config setting for full-mempool behavior
- support two policies: `evict_less_desirable` (legacy default) and `reject_new`
- accept legacy `mempool_full_mode` as a backward-compatible alias for `mempool_full_policy`
- wire the new policy through chain config, `--chain-config-override`, and mempool insertion logic
- fix `chain_config_overrides` so overridden mempool fields are applied when rebuilding `ChainConfig`
- add admin mempool inspection RPCs for listing hashes or full transactions currently in mempool
- cap `getMempoolTxnHashes` and `getMempoolTxns` at 100 results by default, with explicit `limit` and `offset` pagination controls, while keeping oldest-first arrival order
- add an admin mempool flush RPC with filters for `all`, `contract_address`, `transaction_hashes`, and nonce ranges
- make `flushMempoolTxns` match and remove transactions under a single mempool write lock
- use direct hash-index lookups for explicit `transaction_hashes` flushes without nonce filters
- keep mempool iteration used by admin RPC resilient to stale timestamp index keys by warning and skipping instead of panicking
- surface local and forwarded JSON-RPC mempool-full rejections with the same Sequencer-compatible capacity-exceeded shape
- add focused tests for full-policy behavior, outer mempool config wiring, and the new admin RPC paths

## Config
Example:
```yaml
mempool_max_transactions: 10000
mempool_full_policy: reject_new
```

Supported values:
- `evict_less_desirable`: current behavior, evict a less desirable tx if the incoming tx is better
- `reject_new`: reject incoming txs once `mempool_max_transactions` is reached

Backward compatibility:
- `mempool_full_mode` is still accepted as an alias for `mempool_full_policy`.
- Existing configs can omit `mempool_full_policy`; the default remains `evict_less_desirable`.

## User-Facing Mempool-Full Errors
For direct JSON-RPC add-transaction calls and full-node forwarded sequencer responses, mempool-full rejections return the Sequencer-compatible JSON-RPC shape: code `63`, message `An unexpected error occurred`, and `data` containing `Transaction rejected: mempool capacity exceeded.`.

For gateway/deprecated gateway responses, mempool-full remains gateway-style `StarknetErrorCode.TRANSACTION_LIMIT_EXCEEDED` with the same capacity-exceeded message.

## Admin RPC
Added under admin `v0_1_0`:
- `getMempoolTxnHashes`
  - returns oldest-first arrival order
  - optional `include_ttl: bool`
  - optional `offset: usize` with a default of `0`
  - optional `limit: usize` with a default of `100`
  - optional `nonce_after` / `nonce_before`
- `getMempoolTxns`
  - returns oldest-first arrival order
  - optional `include_ttl: bool`
  - optional `offset: usize` with a default of `0`
  - optional `limit: usize` with a default of `100`
  - optional `nonce_after` / `nonce_before`
  - returns full transactions and, when requested, `remaining_ttl_ms`
- `flushMempoolTxns`
  - base filters:
    - `all: true`
    - `contract_address` for sender/account address matching
    - `transaction_hashes: [...]`
  - optional narrowing filters:
    - `nonce_after`
    - `nonce_before`

Nonce filtering is strict and applies only to account transactions:
- `nonce_after = x` means `tx.nonce > x`
- `nonce_before = y` means `tx.nonce < y`
- nonce filters are invalid unless one explicit base flush filter is also provided
- L1 handler message nonces and deploy zero nonces are intentionally not matched by nonce filters

Nonce-range flushes are surgical by design:
- only transactions matching the explicit filter are removed
- if an operator removes nonce `N` and leaves higher account nonces pending, the missing nonce can be resubmitted by the user/operator
- to remove an account from a starting nonce onward, use `contract_address` plus the strict nonce range, for example `nonce_after: "0x2"` to remove nonce `3` and above

Examples:
```json
{"include_ttl": true, "offset": 100, "limit": 250, "nonce_after": "0x5", "nonce_before": "0xa"}
```

```json
{"include_ttl": true, "nonce_after": "0x5"}
```

```json
{"all": true, "nonce_after": "0x5"}
```

```json
{"contract_address": "0x123", "nonce_before": "0xa"}
```

## Notes
- `getMempoolTxnHashes` applies nonce filtering, `offset`, and `limit` while iterating instead of snapshotting the full mempool first.
- `getMempoolTxns` mirrors the hash endpoint by supporting optional TTL reporting and by capping results at 100 by default with explicit `offset` and `limit` controls.
- `reject_new` never evicts existing transactions when the configured mempool transaction limit is reached.
- Default `evict_less_desirable` policy entries are intentionally not repeated in every preset config; the field remains documented in the example config and changelog.

## Validation
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p mc-rpc get_mempool_txn -- --nocapture`
- `cargo test -p mc-rpc flush_mempool_txns -- --nocapture`
- `cargo test -p mc-rpc mempool_limit -- --nocapture`
- `cargo test -p mc-rpc forwarded_transaction_limit_maps_to_unexpected_rpc_error -- --nocapture`
- `cargo test -p mc-rpc nonce_filter_matches_only_account_transaction_nonces -- --nocapture`
- `cargo test -p mc-mempool mempool_reject_new_policy_is_applied_from_chain_config -- --nocapture`
- `cargo clippy -p mp-rpc -p mc-mempool -p mc-rpc -p mc-gateway-server --lib --no-deps -- -D warnings`
- latest review-fix head: `b327cda2a38c3f17a6624f4dbaf9b4bdcb3f7ae9`; PR is draft while review fixes settle


madara/crates/primitives/chain_config/src/chain_config.rs
```diff
@@ -280,6 +288,8 @@ pub struct ChainConfigV2 {
 
     #[serde(default)]
     pub mempool_mode: MempoolMode,
+    #[serde(default, alias = "mempool_full_mode")]
```
prkpndy, line 291: we should create a new chain config version

madara/crates/primitives/rpc/src/admin/mod.rs
```diff
@@ -20,3 +23,96 @@ impl BroadcastedDeclareTxnV0 {
...
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct GetMempoolTxnsParams {
+    /// Include the remaining TTL for each transaction when the node has a mempool TTL configured.
+    pub include_ttl: bool,
+    /// Maximum number of transactions to return.
+    pub limit: usize,
```
prkpndy, line 44: should we also include an offset for pagination?

madara/crates/primitives/transactions/src/validated.rs
```diff
@@ -140,7 +140,7 @@ impl ValidatedTransaction {
                 let tx = tx.try_into().map_err(|_| ValidatedToBlockifierTxError::InvalidContractAddress)?;
 
                 starknet_api::executable_transaction::Transaction::Account(ApiAccountTransaction::Invoke(
-                    InvokeTransaction { tx, tx_hash },
+                    ApiInvokeTransaction { tx, tx_hash },
```
prkpndy, line 143: just to be clear, this is just a name change, right?

madara/crates/client/submit_tx/src/error.rs
```diff
@@ -63,8 +63,8 @@ pub enum RejectedTransactionErrorKind {
     NotPermittedContract,
     #[error("UndeclaredClass")]
     UndeclaredClass,
-    #[error("TransactionLimitExceeded")]
-    TransactionLimitExceeded,
+    #[error("MempoolLimitReached")]
+    MempoolLimitReached,
```
prkpndy, line 67: This is for internal use, right?

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/mod.rs
```diff
@@ -1,4 +1,13 @@
+use mp_rpc::admin::MempoolNonceFilter;
+use mp_transactions::validated::ValidatedTransaction;
+
 pub mod read;
 pub mod services;
 pub mod status;
 pub mod write;
+
+pub(super) fn matches_nonce_filter(transaction: &ValidatedTransaction, nonce_filter: MempoolNonceFilter) -> bool {
```
prkpndy: should we put this function in a better place?

madara/node/src/service/rpc/mod.rs
```diff
@@ -27,6 +28,7 @@ pub struct RpcService {
     server_handle: Option<ServerHandle>,
     rpc_type: RpcType,
     block_prod_handle: Option<BlockProductionHandle>,
+    mempool: Option<Arc<Mempool>>,
```
prkpndy, line 31: this is just for read purpose, right? should we impl a read only trait for mempool, and give that to the rpc services?

prkpndy (commented):
did an initial review. some questions/suggestions!

madara/crates/primitives/chain_config/src/chain_config.rs
heemankv: Keeping the config version because `mempool_full_policy` is optional/defaulted and `mempool_full_mode` remains accepted as a backward-compatible alias, so existing v2 configs continue to load without migration, wdyt ?

madara/crates/primitives/rpc/src/admin/mod.rs
heemankv: Added `offset` to both `GetMempoolTxnHashesParams` and `GetMempoolTxnsParams`, defaulting to `0`. The listing path now applies nonce filtering, then skips `offset`, then takes `limit`, preserving oldest-first pagination.

madara/crates/primitives/transactions/src/validated.rs
heemankv: Yes, this is only a name-disambiguation import alias. There are multiple `InvokeTransaction` types in scope, so `ApiInvokeTransaction` makes the constructed Starknet API type explicit.

madara/crates/client/submit_tx/src/error.rs
heemankv: Yes, this is an internal classification used between submit-tx, RPC, and gateway mapping layers. User-facing JSON-RPC still maps local mempool-full to the Sequencer-compatible code 63 shape, and gateway maps it to `TRANSACTION_LIMIT_EXCEEDED`.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/mod.rs
heemankv: Moved it out of `methods/mod.rs` into a dedicated admin mempool helper module so the shared nonce-filter helper is no longer sitting in the module barrel.

madara/node/src/service/rpc/mod.rs
heemankv: This handle is used by both admin read methods and the admin `flushMempoolTxns` write method, so RPC needs write access in this PR. A read-only trait would only make sense if we split read/write mempool handles across the admin RPC surface in a separate cleanup.

prkpndy (commented):
Review focused on fatal bugs, efficiency of the new mempool paths, and test coverage. Inline comments cover the specific issues; collecting the smaller best-practices items here.

### Best practices / smaller notes

- **Changelog framing** (`CHAIN_CONFIG_CHANGELOG.md`): `CURRENT_CONFIG_VERSION` is already `2` (from the earlier gas-price refactor) and this PR does not change version gating. `mempool_full_policy` is correctly *additive / backward-compatible*, so re-describing v1 as suddenly "failing to load" reads as if this PR changed config-version enforcement. Suggest scoping the entry to "added optional `mempool_full_policy` field (default `evict_less_desirable`)" rather than restating the whole version-2 breaking-change narrative.
- **Default duplicated 3x**: `MempoolFullPolicy` default is repeated across `ChainConfigV2`, `ChainConfig`, and `ChainConfigOverridesInner`. Consistent today, but three sources of truth for one default drift easily — consider reusing a single `Default`/constructor.
- **Redundant guard**: `get_mempool_txns` double-guards `include_ttl` — `snapshot_matching` already nulls `remaining_ttl` when `include_ttl == false`, then the mapping re-checks `if params.include_ttl`. Harmless, just redundant.

### Test coverage gaps (also flagged inline where they have a home)

- `FlushMode::from_params` "more than one base filter" error branch is untested (only nonce-only rejection is covered).
- `contract_address` + `nonce_filter` combination untested (only `contract_address` alone, and `all` + nonce).
- `mempool == None` path (user RPC) returning `UnimplementedMethod` for the new methods — untested.
- Edge cases: `offset` beyond length → empty; `limit == 0` early-return → empty.

Overall the feature is well-structured; the blocking items are the orphaned-nonce flush behavior, the double O(n) flush scan, and the RPC-reachable `.expect` panics.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -417,6 +429,94 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
...
+        self.snapshot_matching(offset, limit, include_ttl, predicate, |transaction, remaining_ttl| {
+            MempoolTransactionHashSnapshot { transaction_hash: transaction.hash, remaining_ttl }
+        })
+        .await
+    }
+
+    /// Remove all matching transactions while holding a single inner write lock so selection and
+    /// removal observe the same mempool state.
+    ///
+    /// Warning: this removes only the transactions that match the predicate. If an operator uses a
```
prkpndy: **[High] Flush can permanently orphan an account's higher nonces.** This is documented as a "Warning" but left entirely to the operator. If a non-tail nonce is flushed (e.g. `{contract_address, nonce_before: N}` or `{all, nonce_after/before}`), every higher nonce for that account stays in the mempool as a pending follower that can *never* become ready — the gap is never filled because the node itself dropped the predecessor. They're stuck until the user resubmits the missing nonce or the on-chain account nonce advances past the gap by some other path.

For an incident-time cleanup tool, silently bricking an account's later txs is a serious footgun. Recommend one of:
- cascade-remove all higher nonces for the account when removing a non-tail nonce, or
- reject a nonce-range flush that would split an account's chain, or
- at minimum return the set of accounts left with a nonce gap in `FlushMempoolTxnsResult`.

Also currently **untested** — please pin the chosen behavior with a test.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -417,6 +429,94 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
...
+    /// higher nonces for that same account remain in the mempool as pending followers until they are
+    /// explicitly flushed or the account nonce catches up through some other path.
+    pub async fn flush_transactions_matching(
+        &self,
+        mut predicate: impl FnMut(&ValidatedTransaction) -> bool,
+    ) -> Vec<ValidatedTransaction> {
+        let mut removed_txs = smallvec::SmallVec::<[ValidatedTransaction; 1]>::new();
+        let summary = {
+            let mut lock = self.inner.write().await;
+            let tx_hashes =
```
prkpndy: **[Efficiency] Flush does two full O(n) scans + a throwaway `HashSet<Felt>`.** Pass 1 here scans `transactions_by_arrival().filter(predicate)` and keeps only hashes; `remove_transactions_by_hashes` (line 507) then re-iterates the *entire* `timestamp_queue` and re-matches each tx by hash. The `TxKey`s identified in pass 1 are discarded and recomputed.

Collect `Vec<TxKey>` directly in a single pass and remove those — eliminates the second full scan, the `HashSet` allocation, and the *n* hash comparisons. (Needs a key-exposing sibling of `transactions_by_arrival`, or fold the predicate into the removal helper.)

madara/crates/client/mempool/src/lib.rs
```diff
@@ -417,6 +429,94 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
...
+            return Vec::new();
+        }
+
+        let now = TxTimestamp::now();
+        let ttl = self.ttl;
+        let lock = self.inner.read().await;
+
+        lock.transactions_by_arrival()
+            .filter(|transaction| predicate(transaction))
+            .skip(offset)
```
prkpndy, line 451: **[Efficiency] Offset pagination is O(offset) per page → O(n²) to page a large mempool.** `.filter(predicate).skip(offset).take(limit)` re-scans from the front on every call. Fine for the default top-100 view, but this is exposed as a general pagination API. Either document that `offset` is not intended for deep scans, or offer cursor-based pagination (last-seen `arrived_at` + hash) for O(limit) pages.

madara/crates/client/mempool/src/lib.rs
```diff
@@ -417,6 +429,94 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
...
+
+    pub async fn snapshot_transactions_matching(
+        &self,
+        offset: usize,
+        limit: usize,
+        include_ttl: bool,
+        predicate: impl FnMut(&ValidatedTransaction) -> bool,
+    ) -> Vec<MempoolTransactionSnapshot> {
+        self.snapshot_matching(offset, limit, include_ttl, predicate, |transaction, remaining_ttl| {
+            MempoolTransactionSnapshot { transaction: transaction.clone(), remaining_ttl }
```
prkpndy, line 474: **[Minor]** Full `ValidatedTransaction`s are cloned while holding the inner read lock; block production / inserts block on this lock during the clone+collect. Bounded by `limit` (default 100) so acceptable, but the stall scales if an operator raises `limit`.

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -324,6 +332,43 @@ impl InnerMempool {
         self.accounts.get_transaction(contract_address, nonce).map(|tx| &tx.inner)
     }
 
+    pub fn transactions_by_arrival(&self) -> impl Iterator<Item = &ValidatedTransaction> {
+        self.timestamp_queue.iter().map(|tx_key| {
+            &self
+                .accounts
+                .get_tx_by_key(tx_key)
+                .expect("Invariant violation: timestamp queue key must resolve to a transaction")
```
prkpndy: **[Medium] RPC-reachable panic.** This `.expect` (and the one at line 361) is now on the call path of `getMempoolTxns`, `getMempoolTxnHashes`, and `flushMempoolTxns`. Any desync between `timestamp_queue` and `accounts` turns an operator RPC call into a node-level panic — precisely the calls an operator makes when the mempool is already misbehaving. Prefer `filter_map` + `tracing::warn!` (skip the stale key) so a structural inconsistency degrades gracefully instead of taking down the process.

madara/crates/client/mempool/src/inner/mod.rs
```diff
@@ -324,6 +332,43 @@ impl InnerMempool {
...
+                .inner
+        })
+    }
+
+    pub fn remove_transactions_by_hashes(
```
prkpndy: **[Efficiency] `transaction_hashes` flush ignores the existing O(1) `by_tx_hash` index.** The mempool already maintains `ByTxHashIndex` (`by_tx_hash`, hash → `TxKey`, used by `get_transaction_by_hash`). Flushing K explicit hashes currently scans all N entries here (twice, combined with the caller). It could be O(K) direct lookups via `by_tx_hash` instead of O(N) — a big difference for "flush these few stuck hashes" on a large mempool.

madara/crates/client/rpc/src/errors.rs
```diff
@@ -185,6 +187,7 @@ impl From<&StarknetRpcApiError> for i32 {
             StarknetRpcApiError::InsufficientMaxFee { .. } => 53,
             StarknetRpcApiError::InsufficientAccountBalance { .. } => 54,
             StarknetRpcApiError::ValidationFailure { .. } => 55,
+            StarknetRpcApiError::MempoolLimitReached { .. } => 67,
```
prkpndy: **[Medium] Inconsistent code + non-standard value for "mempool full".** A full node *forwarding* to a sequencer gateway that returns `TransactionLimitExceeded` maps to this `MempoolLimitReached => 67` and is returned to that node's RPC user — whereas a *direct* submission rejection surfaces as `ErrUnexpectedError` code `63` (validation.rs → `ErrUnexpectedError`). Same logical condition, two different user-facing codes depending on topology.

Also, code `67` is not a Starknet JSON-RPC spec error code and numerically collides with `StarknetWsApiError::TooManyAddressesInFilter => 67` on the same node. Pick one representation for "mempool full" across both paths, or document why the forwarded path intentionally differs.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/mempool.rs
```diff
@@ -0,0 +1,8 @@
+use mp_rpc::admin::MempoolNonceFilter;
+use mp_transactions::validated::ValidatedTransaction;
+
+pub(super) fn matches_nonce_filter(transaction: &ValidatedTransaction, nonce_filter: MempoolNonceFilter) -> bool {
```
prkpndy, line 4: **[Low/Med] Nonce filter applies to tx types whose `nonce()` is not an account nonce.** `transaction.transaction.nonce()` returns the **L1 message nonce** for `L1Handler` and **`Felt::ZERO`** for `Deploy` (see `mp-transactions/src/lib.rs:199` — "warning on what this nonce is for l1 handler txs"). So `flush {all, nonce_before: 5}` also matches `L1Handler` txs whose message nonce < 5 and every `Deploy` tx (nonce 0); same for the read endpoints. Almost certainly unintended — consider restricting nonce filters to account txs (`sender_contract_address().is_some()`) or documenting the cross-type semantics. Untested.

madara/crates/client/mempool/src/notify.rs
```diff
@@ -47,6 +47,7 @@ impl MempoolInnerWithNotify {
                     MempoolMode::Timestamp => ScoreFunction::Timestamp,
                     MempoolMode::Tip => ScoreFunction::Tip { min_tip_bump: config.mempool_min_tip_bump },
                 },
+                full_policy: config.mempool_full_policy,
```
prkpndy, line 50: **[Test coverage] `reject_new` is not tested end-to-end.** Only the `InnerMempool` unit test exercises `RejectNew`; this wiring (`config.mempool_full_policy → InnerMempoolConfig.full_policy`) is untested. Add a test that builds a `Mempool` from a chain config with `mempool_full_policy: reject_new` and asserts an incoming tx is rejected (not evicting) when the mempool is full — that covers the plumbing this line introduces.

madara/crates/client/mempool/src/lib.rs
heemankv: This is intentional admin behavior. `flushMempoolTxns` nonce filters are surgical by design: they remove only transactions matching the explicit filter, so an operator can remove a specific nonce and allow that nonce to be resubmitted. If the operator wants to remove from a starting nonce onward, they can use the strict range semantics with `contract_address` (`nonce_after: start - 1`). I updated the admin RPC docs and PR description to make this explicit.

madara/crates/client/mempool/src/lib.rs
heemankv: Fixed. Generic predicate-based flush now collects matching `TxKey`s in one pass while holding the mempool write lock, then removes those keys directly. This avoids the previous hash collection plus second full timestamp queue scan.

madara/crates/client/mempool/src/lib.rs
heemankv: Keeping offset pagination for this PR because it was requested as a simple bounded admin inspection API. The endpoints default to `limit = 100`, and the PR description now frames this as offset/limit inspection, not deep cursor scanning. Cursor pagination is a larger API change that I’d keep out of this PR.

madara/crates/client/mempool/src/lib.rs
heemankv: Agreed on the tradeoff. This remains bounded by `limit` with a default of 100, so it is acceptable for the current admin inspection use case. Operators can raise `limit`, but that is explicit. I’m not changing this further in this PR.

madara/crates/client/mempool/src/inner/mod.rs
heemankv: Fixed. `transactions_by_arrival` now uses `filter_map` and logs a warning if the timestamp queue contains a stale key instead of panicking. The removal selection path also skips stale keys with a warning.

madara/crates/client/mempool/src/inner/mod.rs
heemankv: Fixed. `transaction_hashes` flush without nonce narrowing now uses the existing `by_tx_hash` index, so flushing K explicit hashes resolves keys through direct lookups instead of scanning the whole mempool.

madara/crates/client/rpc/src/errors.rs
heemankv: Fixed. Removed the dedicated JSON-RPC `MempoolLimitReached` code path. Local mempool-full and forwarded sequencer `TransactionLimitExceeded` now both map to the Sequencer-compatible JSON-RPC shape: code `63`, message `An unexpected error occurred`, and `data: "Transaction rejected: mempool capacity exceeded."`. Gateway/deprecated gateway still maps to `TRANSACTION_LIMIT_EXCEEDED`.

madara/crates/client/rpc/src/versions/admin/v0_1_0/methods/mempool.rs
heemankv: Fixed. Nonce filters now apply only to account transactions. If a nonce filter is provided, transactions without a sender/account address, such as L1 handlers, are not matched by the nonce predicate. Added a focused test for this.

madara/crates/client/mempool/src/notify.rs
heemankv: Fixed. Added an outer `Mempool` test that builds from a chain config with `mempool_full_policy: reject_new`, fills the mempool, and verifies a better incoming transaction is rejected without evicting the existing transaction.

prkpndy (approved):

PR #1071, opened 2026-05-21
feat(orchestrator): add workload activity metrics

heemankv:
## Summary
- add a workload tracker and workload-specific OTEL metrics for orchestrator activity
- instrument queue-backed processing, verification, worker triggers, capacity slots, and stale-job healing
- add unit coverage for workload label mapping and active-slot tracking

## Metrics Added
- `workload_active_slots`
- `workload_busy_seconds_total`
- `workload_capacity_slots`
- `healed_jobs_total`

## Notes
- labels are bounded to `work_kind`, `work_phase`, `source_job_type`, and `outcome`
- queue-backed workers emit configured capacity slots so Grafana can compare busy time versus local capacity per pod
- healing is emitted as maintenance work under `work_kind="Healing"` with the original job type in `source_job_type`

## Testing
- `cargo fmt --manifest-path Cargo.toml -p orchestrator`
- `cargo test --manifest-path Cargo.toml -p orchestrator metrics_recorder -- --nocapture` *(blocked by existing upstream dependency failure in `size-of 0.1.5` on `aarch64-apple-darwin` due to unsupported ABIs)*

### DIFF
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -1,9 +1,10 @@
 use crate::core::client::database::constant::JOBS_COLLECTION;
 use crate::utils::job_status_metrics::JobStatusTracker;
+use crate::utils::metrics_recorder::register_workload_active_slots_observer;
 use once_cell;
 use once_cell::sync::Lazy;
 use opentelemetry::global;
-use opentelemetry::metrics::{Counter, Gauge, Histogram};
+use opentelemetry::metrics::{Counter, Gauge, Histogram, ObservableGauge};
 use orchestrator_utils::metrics::lib::{
     register_counter_metric_instrument, register_gauge_metric_instrument, register_histogram_metric_instrument, Metrics,
 };
@@ -78,7 +79,7 @@ pub struct OrchestratorMetrics {
     pub cleanup_artifacts_tagged: Counter<f64>,
     pub cleanup_failures_total: Counter<f64>,
     // Workload metrics
-    pub workload_active_slots: Gauge<f64>,
+    pub workload_active_slots: ObservableGauge<u64>,
     pub workload_busy_seconds_total: Counter<f64>,
     pub workload_runs_total: Counter<f64>,
     pub workload_duration_seconds: Histogram<f64>,
@@ -443,12 +444,7 @@ impl Metrics for OrchestratorMetrics {
             "errors".to_string(),
         );
 
-        let workload_active_slots = register_gauge_metric_instrument(
-            &orchestrator_meter,
-            "workload_active_slots".to_string(),
-            "Current active workload slots by kind, phase, and class".to_string(),
-            "slots".to_string(),
-        );
+        let workload_active_slots = register_workload_active_slots_observer(&orchestrator_meter);
 
         let workload_busy_seconds_total = register_counter_metric_instrument(
             &orchestrator_meter,
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -77,6 +77,13 @@ pub struct OrchestratorMetrics {
     pub cleanup_jobs_processed: Counter<f64>,
     pub cleanup_artifacts_tagged: Counter<f64>,
     pub cleanup_failures_total: Counter<f64>,
+    // Workload metrics
+    pub workload_active_slots: Gauge<f64>,
+    pub workload_busy_seconds_total: Counter<f64>,
+    pub workload_runs_total: Counter<f64>,
+    pub workload_duration_seconds: Histogram<f64>,
+    pub workload_capacity_slots: Gauge<f64>,
+    pub healed_jobs_total: Counter<f64>,
 }
 
 impl Metrics for OrchestratorMetrics {
@@ -436,6 +443,48 @@ impl Metrics for OrchestratorMetrics {
             "errors".to_string(),
         );
 
+        let workload_active_slots = register_gauge_metric_instrument(
+            &orchestrator_meter,
+            "workload_active_slots".to_string(),
+            "Current active workload slots by kind, phase, and class".to_string(),
+            "slots".to_string(),
+        );
+
+        let workload_busy_seconds_total = register_counter_metric_instrument(
+            &orchestrator_meter,
+            "workload_busy_seconds_total".to_string(),
+            "Cumulative busy time spent on orchestrator workloads".to_string(),
+            "s".to_string(),
+        );
+
+        let workload_runs_total = register_counter_metric_instrument(
+            &orchestrator_meter,
+            "workload_runs_total".to_string(),
+            "Total number of orchestrator workload runs".to_string(),
+            "runs".to_string(),
+        );
+
+        let workload_duration_seconds = register_histogram_metric_instrument(
+            &orchestrator_meter,
+            "workload_duration_seconds".to_string(),
+            "Duration of orchestrator workload runs".to_string(),
+            "s".to_string(),
+        );
+
+        let workload_capacity_slots = register_gauge_metric_instrument(
+            &orchestrator_meter,
+            "workload_capacity_slots".to_string(),
+            "Configured local worker slot capacity for queue-backed workloads".to_string(),
+            "slots".to_string(),
+        );
+
+        let healed_jobs_total = register_counter_metric_instrument(
+            &orchestrator_meter,
+            "healed_jobs_total".to_string(),
+            "Total number of stale jobs healed back to Created".to_string(),
+            "jobs".to_string(),
+        );
+
         Self {
             block_gauge,
             successful_job_operations,
@@ -482,6 +531,12 @@ impl Metrics for OrchestratorMetrics {
             cleanup_jobs_processed,
             cleanup_artifacts_tagged,
             cleanup_failures_total,
+            workload_active_slots,
+            workload_busy_seconds_total,
+            workload_runs_total,
+            workload_duration_seconds,
+            workload_capacity_slots,
+            healed_jobs_total,
         }
     }
 }
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -78,7 +78,10 @@ pub struct OrchestratorMetrics {
     pub cleanup_jobs_processed: Counter<f64>,
     pub cleanup_artifacts_tagged: Counter<f64>,
     pub cleanup_failures_total: Counter<f64>,
-    // Workload metrics
+    // Workload metrics.
+    // `workload_active_slots` is an ObservableGauge because the source of truth is
+    // a shared in-memory slot map. Reading that map at scrape time avoids racy
+    // "mutate then record" snapshots that can under-report concurrent work.
     pub workload_active_slots: ObservableGauge<u64>,
     pub workload_busy_seconds_total: Counter<f64>,
     pub workload_capacity_slots: Gauge<f64>,
--- a/orchestrator/src/utils/metrics.rs
+++ b/orchestrator/src/utils/metrics.rs
@@ -81,8 +81,6 @@ pub struct OrchestratorMetrics {
     // Workload metrics
     pub workload_active_slots: ObservableGauge<u64>,
     pub workload_busy_seconds_total: Counter<f64>,
-    pub workload_runs_total: Counter<f64>,
-    pub workload_duration_seconds: Histogram<f64>,
     pub workload_capacity_slots: Gauge<f64>,
     pub healed_jobs_total: Counter<f64>,
 }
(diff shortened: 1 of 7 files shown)


heemankv:
Fixed the verification-timeout workload outcome labeling so timed-out verification runs emit `outcome=error` instead of `success`.

@greptile can you re-review?

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -678,6 +706,10 @@ impl JobHandlerService {
         MetricsRecorder::record_successful_job_operation(1.0, &attributes);
         MetricsRecorder::record_job_response_time(duration.as_secs_f64(), &attributes);
         Self::register_block_gauge(job.job_type, job.internal_id, &attributes, &config).await?;
+        match operation_job_status {
+            Some(JobStatus::VerificationFailed) => workload.finish_error(),
+            _ => workload.finish_success(),
+        }
         Ok(())
     }
```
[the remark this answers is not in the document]
heemankv: Fixed in `ac77e6903`: retryable `VerificationFailed` statuses are now treated as `outcome=error` in the shared verification tail instead of falling through to `finish_success()`.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -1,15 +1,373 @@
...
+            KeyValue::new("work_class", self.work_class.as_str()),
+            KeyValue::new("source_job_type", self.source_job_type.map_or("none", WorkKind::as_str)),
+        ]
+    }
+
+    fn completed_attributes(self, outcome: WorkloadOutcome) -> [KeyValue; 5] {
+        [
+            KeyValue::new("work_kind", self.work_kind.as_str()),
+            KeyValue::new("work_phase", self.work_phase.as_str()),
+            KeyValue::new("work_class", self.work_class.as_str()),
```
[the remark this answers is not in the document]
heemankv: Fixed on the current head in `d5c9da98a`: `WorkloadDescriptor` now serves as the DashMap key directly, so the duplicate `WorkloadKey` struct and `key()` copy path are gone.

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -678,6 +696,7 @@ impl JobHandlerService {
         MetricsRecorder::record_successful_job_operation(1.0, &attributes);
         MetricsRecorder::record_job_response_time(duration.as_secs_f64(), &attributes);
         Self::register_block_gauge(job.job_type, job.internal_id, &attributes, &config).await?;
+        workload.finish_success();
         Ok(())
     }
```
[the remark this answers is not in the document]
heemankv: Fixed in `40bc60953` and preserved on the current head: verification timeouts now branch to `workload.finish_error()` instead of falling through to the success tail.

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -600,6 +607,17 @@ impl JobHandlerService {
...
+                    match move_result {
+                        Ok(()) => {
+                            workload.finish_success();
+                            return Ok(());
+                        }
+                        Err(e) => {
+                            workload.finish_error();
+                            return Err(e);
+                        }
+                    }
```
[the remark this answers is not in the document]
heemankv: Fixed in `ac77e6903` and preserved on the current head: the exhausted `Rejected` branch records `finish_error()` before returning, so failed verifications no longer emit `outcome=success`.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -1,15 +1,373 @@
...
+
+        ORCHESTRATOR_METRICS.workload_busy_seconds_total.add(duration_seconds, &completed_attributes);
+        ORCHESTRATOR_METRICS.workload_runs_total.add(1.0, &completed_attributes);
+        ORCHESTRATOR_METRICS.workload_duration_seconds.record(duration_seconds, &completed_attributes);
+
+        let active_slots = decrement_active_slots(self.descriptor);
+        ORCHESTRATOR_METRICS.workload_active_slots.record(active_slots as f64, &active_attributes);
+    }
+}
+
```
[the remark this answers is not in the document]
heemankv: Fixed on the current head in `d5c9da98a`: `workload_active_slots` is now an observable gauge that reads directly from the active-slot map on scrape, and descriptors are retained at zero after drain so the metric can publish `0` without racy per-event `record()` calls.

heemankv:
Fixed the stale test helper Greptile called out on `ecbe51b28`: `active_slots_for` now keys the `DashMap` by `WorkloadDescriptor` directly and returns `u64`, matching the new slot map type.

@greptile can you re-review?

heemankv (commented):
Self review done, looks good

orchestrator/src/utils/metrics.rs
```diff
@@ -77,6 +78,11 @@ pub struct OrchestratorMetrics {
     pub cleanup_jobs_processed: Counter<f64>,
     pub cleanup_artifacts_tagged: Counter<f64>,
     pub cleanup_failures_total: Counter<f64>,
+    // Workload metrics
+    pub workload_active_slots: ObservableGauge<u64>,
```
Mohiiit, line 85: It would be helpful to have a bit of documentation around this metric block, especially around choices like why `ObservableGauge` is needed here versus a regular `Gauge`.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -1,15 +1,352 @@
...
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Success => "success",
+            Self::Error => "error",
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+struct WorkloadDescriptor {
```
Mohiiit, line 95: I’m not sure I understand the distinction between `work_phase` and `work_class` yet. In the current mappings they seem to move together, so a short comment here might help.

Mohiiit (commented):
Overall this looks good to me. I had a few small comments around metric semantics and one or two control-flow spots that were a bit hard to follow on first read

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -616,6 +623,17 @@ impl JobHandlerService {
                         ),
                     )
                     .await;
+
+                    match move_result {
```
Mohiiit: Do we need the full `match` here? It looks like this could maybe just be `workload.finish_error(); return move_result;`.

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -659,6 +675,16 @@ impl JobHandlerService {
...
+
+                    attributes.push(KeyValue::new("operation_job_status", JobStatus::VerificationTimeout.to_string()));
+
+                    debug!(log_type = "completed", category = "general", function_type = "verify_job", block_no = %internal_id, "General verify job completed for block");
+                    let duration = start.elapsed();
+                    MetricsRecorder::record_successful_job_operation(1.0, &attributes);
+                    MetricsRecorder::record_job_response_time(duration.as_secs_f64(), &attributes);
+                    Self::register_block_gauge(job.job_type, job.internal_id, &attributes, &config).await?;
+                    workload.finish_error();
+                    return Ok(());
```
Mohiiit: This works, but the early return here made the control flow a bit harder to follow since this path no longer goes through the shared footer below. Would it be cleaner to carry the outcome through to the common tail instead?

orchestrator/src/utils/metrics.rs
heemankv: Added a short note here explaining why `workload_active_slots` uses an `ObservableGauge`: the source of truth is the shared in-memory slot map, so reading it during collection avoids stale concurrent snapshots from per-event gauge writes.

orchestrator/src/utils/metrics_recorder.rs
heemankv: I ended up simplifying this further and removed `work_class` entirely. It was speculative structure on top of the current metric model, and `work_phase` already carries the distinction we actually use today.
Thanks for pointing it out!

orchestrator/src/worker/event_handler/service.rs
heemankv: Simplified this branch as suggested. It now records `workload.finish_error()` once and returns `move_result` directly.

orchestrator/src/worker/event_handler/service.rs
heemankv: Refactored this so the timeout path now goes through the shared footer as well. Instead of doing its own local footer and early return, it sets `operation_job_status = VerificationTimeout` and falls through to the common metrics tail, which now treats both `VerificationFailed` and `VerificationTimeout` as workload errors.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -1,15 +1,309 @@
...
+
+    pub fn start_worker_trigger_workload(worker_trigger_type: &WorkerTriggerType) -> WorkloadTracker {
+        WorkloadTracker::start(workload_descriptor_for_worker_trigger(worker_trigger_type))
+    }
+
+    pub fn start_healing_workload(job_type: &JobType) -> WorkloadTracker {
+        WorkloadTracker::start(healing_descriptor(job_type))
+    }
+
+    pub fn record_workload_capacity_for_queue(queue_type: &QueueType, max_slots: usize) {
```
Mohiiit, line 298: Curious about the scope of `workload_capacity_slots`: this currently skips `WorkerTrigger`, `JobHandleFailure`, and the priority queues, while `workload_active_slots` is still emitted for some non-queue-backed workloads. Is the intent for capacity to cover only the processing/verification queues?

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -333,3 +627,62 @@ impl MetricsRecorder {
...
+
+    fn reset_active_workload_state() {
+        ACTIVE_WORKLOAD_SLOTS.clear();
+    }
+
+    fn active_slots_for(descriptor: WorkloadDescriptor) -> u64 {
+        ACTIVE_WORKLOAD_SLOTS.get(&descriptor).map(|entry| *entry).unwrap_or(0)
+    }
+
+    #[test]
```
Mohiiit, line 647: I may be missing some existing coverage, but do we want a behavior-level test around the `verify_job` outcome classification too? This branch now depends on `VerificationFailed` / `VerificationTimeout` finishing as error and the other paths finishing as success, so it feels worth locking that down explicitly.

orchestrator/src/utils/metrics_recorder.rs
```diff
@@ -1,15 +1,309 @@
...
+        Entry::Vacant(entry) => {
+            entry.insert(0);
+        }
+    }
+}
+
+pub fn register_workload_active_slots_observer(meter: &Meter) -> ObservableGauge<u64> {
+    meter
+        .u64_observable_gauge("workload_active_slots")
+        .with_description("Current active workload slots by kind, phase, and class")
```
Mohiiit: Small doc nit: the observable gauge description still says "kind, phase, and class", but I think `work_class` has been removed now.

Mohiiit (commented):
Overall this looks good to me. I had a couple of follow-up questions around the workload metric semantics and coverage.

Also, if I understood correctly, workload_capacity_slots is not an active telemetry collection? it reflects the config directly right? I don't mind to have it in the metrics, just want to draw a clear distinction in my mind about that!

orchestrator/src/utils/metrics_recorder.rs
heemankv: Added focused coverage for this now. I factored the verification workload classification into a small helper and added a unit test that locks down `VerificationFailed` and `VerificationTimeout` as workload errors, while non-failure states stay on the success side.

orchestrator/src/utils/metrics_recorder.rs
heemankv: Fixed this description on the latest head so it now matches the current label set after removing `work_class`.

orchestrator/src/utils/metrics_recorder.rs
heemankv: Yes, that is the intended scope, and I added a short code comment next to the capacity mapping to make that explicit. `workload_capacity_slots` is meant to reflect configured queue-backed processing/verification capacity only. Trigger, failure-handling, priority helper, and maintenance-style workloads still emit active/busy signals for observability, but we are not treating them as part of the sizing denominator in this PR. Also yes: `workload_capacity_slots` is config-derived telemetry, not live occupancy telemetry. It reflects the queue worker concurrency (`max_message_count`) so we can compare configured supply against actual usage.

Mohiiit (approved):

PR #1074, opened 2026-05-22
[codex] Derive Cairo Native cache ABI from dependency version

Mohiiit:
## Summary
- Derives the native cache ABI from the resolved Cairo Native dependency version.
- Keeps the manual schema version for metadata contract changes.
- Ensures Cairo Native dependency bumps invalidate persisted native cache artifacts automatically.

## Validation
- Focused host fingerprint unit coverage passes.
- Formatting and whitespace checks pass.


prkpndy (approved):
greptile's feedback regarding the test looks valid to me. otherwise looks good. would have been good if cairo native exposed a version through code itself, but since that's not there, parsing Cargo.toml looks like a pragmatic approach! gg!

PR #1075, opened 2026-05-22
feat(gateway): add feeder transaction and block lookup endpoints

heemankv:
## Summary

This PR adds the feeder gateway read endpoints currently needed for node compatibility:

- `feeder_gateway/get_transaction`
- `feeder_gateway/get_transaction_status`
- `feeder_gateway/get_block_hash_by_id`
- `feeder_gateway/get_block_id_by_hash`

`feeder_gateway/get_preconfirmed_block` already exists on `madara/main`, so this PR does not reimplement it. The deprecated backlog endpoints are also intentionally left out:

- `feeder_gateway/get_oldest_transaction_age`
- `feeder_gateway/get_number_of_transactions_in_backlog`

## Feeder Transaction Semantics

For `get_transaction` and `get_transaction_status`, this PR deliberately keeps feeder behavior narrower than Starknet JSON-RPC behavior.

The Starknet v0.14.0 pre-release note for feeder `get_transaction_status` says the feeder response finality can only be `ACCEPTED_ON_L2` or `ACCEPTED_ON_L1`, and transactions still in mempool return `Transaction hash not found`. It points readers to RPC 0.9 for the wider `RECEIVED` and preconfirmation status discussion.

Reference:

- https://community.starknet.io/t/sn-0-14-0-pre-release-notes/115618#p-2359352-endpoint-get_transaction_status-18

We also validated this live against integration Sepolia feeder/RPC:

- feeder `get_preconfirmed_block?blockNumber=latest` returned a `PRE_CONFIRMED` block containing transactions with receipts/state diffs
- feeder `get_transaction_status?transactionHash=...` for those same hashes returned `NOT_RECEIVED`
- feeder `get_transaction?transactionHash=...` for those same hashes returned `NOT_RECEIVED`
- RPC `starknet_getTransactionReceipt` for the same hashes returned `finality_status: PRE_CONFIRMED` and `execution_status: SUCCEEDED`

So the feeder endpoints implemented here expose only:

- `NOT_RECEIVED`
- `ACCEPTED_ON_L2`
- `ACCEPTED_ON_L1`

They do **not** expose RPC-layer `CANDIDATE` / `PRE_CONFIRMED`, and they do not expose mempool/preconfirmed transaction payloads through `get_transaction`. Preconfirmed transaction data remains available through `get_preconfirmed_block`; richer preconfirmation status belongs to the RPC layer.

## Block ID Endpoints

`get_block_hash_by_id` and `get_block_id_by_hash` are implemented as the legacy feeder mapping pair:

- `blockId -> blockHash`
- `blockHash -> blockId`

They are intentionally scoped to confirmed data and are tested as a round-trip pair.

## Implementation

- add feeder transaction/status response primitives in `mp-gateway`
- extend the gateway client with the missing transaction and block lookup feeder methods
- expose feeder passthrough hooks from `TransactionLookup` for transaction payloads and statuses
- add gateway server handlers and routes for the missing endpoints
- keep mempool/preconfirmed transaction lookups hidden from feeder `get_transaction` and `get_transaction_status` by returning `NOT_RECEIVED`

## Tests

- gateway server unit coverage for:
  - unknown, mempool, preconfirmed, confirmed, and remote fallback transaction status lookups
  - unknown, mempool, preconfirmed, confirmed, and remote fallback transaction payload lookups
  - block hash/id round-trips and out-of-range handling
- submit-tx unit coverage for:
  - mempool/preconfirmed feeder transaction lookups returning `NOT_RECEIVED`
  - confirmed transaction lookups returning accepted status/payload
  - confirmed-mempool fallback misses returning `NOT_RECEIVED`
- local devnet e2e coverage for:
  - `get_transaction_status`
  - `get_transaction`
  - `get_block_hash_by_id`
  - `get_block_id_by_hash`

## References

- Starknet v0.14.0 feeder `get_transaction_status` note: https://community.starknet.io/t/sn-0-14-0-pre-release-notes/115618#p-2359352-endpoint-get_transaction_status-18
- Starknet version notes: https://docs.starknet.io/learn/cheatsheets/version-notes
- Starknet transaction lifecycle/status docs: https://docs.starknet.io/learn/protocol/transactions
- Historical feeder gateway deprecation post: https://community.starknet.io/t/feeder-gateway-deprecation/100233


madara/crates/client/mempool/src/chain_watcher_task.rs
```diff
@@ -57,11 +57,18 @@ impl<D: MadaraStorageRead + MadaraStorageWrite> Mempool<D> {
...
+                        .executed_transactions()
+                        .nth(tx_index)
+                        .context("Executed transaction should be in block")?
+                        .clone();
                     self.set_transaction_status(
                         tx_hash,
                         Some(TransactionStatus::Preconfirmed(PreConfirmationStatus::Executed {
                             view: preconfirmed.block().clone(),
                             transaction_index: tx_index as u64,
+                            transaction: Arc::new(transaction),
```
prkpndy: **[Medium] This `transaction` is cloned on the block-production hot path but never read.** The only consumers of `PreConfirmationStatus::Executed` are `feeder_status_from_preconfirmed_mempool` / `feeder_transaction_from_preconfirmed_mempool` in `submit_tx/src/lib.rs`, which take `_status` (ignored) and unconditionally return `not_received()`. Nothing in the tree reads `Executed.transaction` (only constructed here + in tests). So every executed-tx status update pays a full `PreconfirmedExecutedTransaction` clone (tx + receipt + state_diff + declared_class) + `Arc` alloc + retained memory for data nobody reads.

Also note line 63: `.executed_transactions().nth(tx_index)` is an O(tx_index) iterator walk, whereas the sibling `Candidate` branch uses `candidate_transactions().get(candidate_index)` (O(1)). Over a block of n executed txs this is O(n²).

Recommend dropping the field + clone entirely (the feeder design hides preconfirmed txs as `NOT_RECEIVED`, so it isn't needed). If it's meant for future use, at least use O(1) indexed access and don't clone until a reader exists.

prkpndy (commented):
Review focused on fatal bugs, efficiency of the new feeder/mempool paths, and test coverage. The trait split (`received_transaction`/`subscribe_new_transactions` → `TransactionLookup`) and the service wiring look correct; findings are concentrated in the genuinely new logic. Inline comments cover the specifics; collecting the smaller items and test gaps here.

### Test coverage gaps

- **Reverted confirmed tx**: no test that a reverted-but-included tx yields `execution_status: REVERTED` + populated `tx_revert_reason` (with `tx_status: ACCEPTED_ON_L2`). The `Reverted` arm of `execution_status` and the revert-reason plumbing are untested on the confirmed path.
- **`ACCEPTED_ON_L1`**: every test uses L2; the `is_on_l1 == true` branch (`accepted_status(true)`) is never exercised.
- **`TransactionLookupSwitch` provider selection** (`L2Sync` → gateway, `BlockProduction` → mempool, neither → `Ok(None)`) is untested — this routing decides local-vs-remote feeder behavior.
- **`get_block_id_by_hash` unknown hash → `BlockNotFound`** is untested (only the round-trip and out-of-range-by-id cases exist).
- The client `get_transaction_and_block_id_endpoints` test hits the **live mainnet feeder** (network-dependent, same pattern as the existing `get_state_update` fixture). Fine, but flag it as integration-only so it isn't treated as a hermetic unit test.

### Smaller notes

- All four new client methods use `.expect("Failed to add URI segment...")` — consistent with existing gateway-client style, so fine; just noting these are `expect`s on request-building paths.
- `tests/js_tests/package.json` pins `viem` `^2.47.6` → `2.47.6` — incidental to this PR's feeder scope (same change appeared in #1067); harmless.

Overall a solid, well-tested feature. The priorities before merge are the dead `Executed.transaction` clone, the duplicated feeder shaping, and the double confirmed-mempool lookup.

madara/crates/client/mempool/src/transaction_status.rs
```diff
@@ -25,7 +27,7 @@ impl TransactionStatus {
 pub enum PreConfirmationStatus {
     Received(Arc<ValidatedTransaction>),
     Candidate { view: Arc<PreconfirmedBlock>, transaction_index: u64, transaction: Arc<ValidatedTransaction> },
-    Executed { view: Arc<PreconfirmedBlock>, transaction_index: u64 },
+    Executed { view: Arc<PreconfirmedBlock>, transaction_index: u64, transaction: Arc<PreconfirmedExecutedTransaction> },
```
prkpndy: The new `transaction: Arc<PreconfirmedExecutedTransaction>` field on `Executed` is currently dead — see the construction site in `chain_watcher_task.rs`; no reader exists. Either remove it or wire a consumer. (The `Candidate` variant's `transaction` field, by contrast, is read.)

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -101,28 +102,167 @@
...
+    }
+
+    let transaction = executed.get_transaction()?;
+    let (_, block_hash, _) = transaction_status_from_block(&executed.block)?;
+    let (execution_status, tx_revert_reason) = execution_status(&transaction.receipt);
+
+    Ok(ProviderTransactionStatus::with_status(fallback_status, Some(execution_status), block_hash, tx_revert_reason))
+}
+
+fn feeder_transaction_from_confirmed_mempool<D: MadaraStorageRead>(
```
prkpndy: **[Low/Med] Double backend lookup + dead params.** The caller (`Mempool::feeder_transaction`) already called `get_transaction_status` — which does a `find_transaction_by_hash` — and learned `Confirmed { block_number, transaction_index, is_on_l1 }`. This fn then does *another* `find_transaction_by_hash` and discards all three passed values (`_block_number`, `_transaction_index`, `_is_on_l1`). That's two DB reads for one logical lookup, plus a TOCTOU window (tx could be reverted between the two reads → second misses → `NOT_RECEIVED`). Collapse to a single lookup and reuse the already-fetched fields, or drop the dead params.

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -101,28 +102,167 @@
...
+    _is_on_l1: bool,
+) -> anyhow::Result<ProviderTransactionResponse> {
+    let Some(executed) = mempool.find_transaction_by_hash(&transaction_hash)? else {
+        return Ok(ProviderTransactionResponse::not_received());
+    };
+
+    feeder_transaction_from_backend_view(&executed)
+}
+
+fn feeder_status_from_preconfirmed_mempool(_status: &PreConfirmationStatus) -> ProviderTransactionStatus {
```
prkpndy: `feeder_status_from_preconfirmed_mempool` / `feeder_transaction_from_preconfirmed_mempool` ignore `_status` and always return `not_received()`. That's the intended design (hide preconfirmed from the feeder), but it means the `transaction` field added to `PreConfirmationStatus::Executed` is never consulted — reinforcing the dead-clone finding in `chain_watcher_task.rs`. A short comment here that preconfirmed/candidate/executed are all deliberately `NOT_RECEIVED` would make the intent obvious.

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -101,28 +102,167 @@
...
+use mp_transactions::{
+    validated::ValidatedTransaction, L1HandlerTransactionResult, L1HandlerTransactionWithFee, TransactionWithHash,
+};
 mod error;
 mod validation;
 
 pub use error::*;
 pub use validation::{TransactionValidator, TransactionValidatorConfig};
 
+fn gateway_executed_transaction(transaction: &mp_block::TransactionWithReceipt) -> GatewayTransaction {
```
prkpndy, line 133: **[Medium] Duplicated feeder-shaping logic.** `gateway_executed_transaction` / `execution_status` / `transaction_status_from_block` here are duplicated in `gateway/server/src/handler.rs` (`gateway_transaction` is byte-for-byte identical; the two `transaction_status_from_block` use *different* predicates — `as_preconfirmed().is_some()` here vs `is_confirmed()`/`is_on_l1()` there). They agree today, but two copies with divergent predicates will drift on edge cases (e.g. a block neither preconfirmed nor accepted-on-L2). Extract one shared helper (e.g. in `mp-gateway`) used by both the server handler and the mempool lookup.

madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -118,6 +121,147 @@ fn parse_add_transaction_request_with_max_body_size(
...
+    if block.is_confirmed() {
+        let block_info = block.as_confirmed().context("Confirmed block view expected")?.get_block_info()?;
+        let status = if block.is_on_l1() { TransactionStatus::AcceptedOnL1 } else { TransactionStatus::AcceptedOnL2 };
+        Ok((status, Some(block_info.block_hash), Some(block_info.header.block_number)))
+    } else {
+        Ok((TransactionStatus::Pending, None, Some(block.block_number())))
+    }
+}
+
+fn gateway_transaction(transaction: &mp_block::TransactionWithReceipt) -> GatewayTransaction {
```
prkpndy: **[Medium] Duplicate of `submit_tx/src/lib.rs`.** `gateway_transaction` here and `gateway_executed_transaction` there are identical; `execution_status` and `transaction_status_from_block` are also duplicated (with a different confirmed-vs-preconfirmed predicate). De-duplicate into a single shared helper so the server-handler path and the mempool-lookup path can't diverge.

madara/crates/client/gateway/server/src/handler.rs
```diff
@@ -118,6 +121,147 @@ fn parse_add_transaction_request_with_max_body_size(
...
+            format!("Block ID should be in the range [0, {}); got: {}.", latest_confirmed + 1, block_id),
+        )
+        .into());
+    }
+
+    let block = backend.block_view_on_confirmed(block_id).ok_or_else(StarknetError::block_not_found)?;
+    Ok(block.get_block_info()?.block_hash)
+}
+
+fn block_id_by_hash_response(block_hash: Felt, backend: &Arc<MadaraBackend>) -> Result<u64, GatewayError> {
```
prkpndy, line 205: **[Low] Not confirmed-scoped, asymmetric with `block_hash_by_id_response`.** `block_hash_by_id_response` explicitly bounds to confirmed data (`latest_confirmed_block_n()` + `block_view_on_confirmed`), but this uses `view_on_latest().find_block_by_hash` with no confirmed guard. In practice preconfirmed blocks have no hash so won't match, but the PR states these endpoints are "intentionally scoped to confirmed data" — add a guard or a comment so the invariant is explicit and can't regress if preconfirmed hashing ever changes.

madara/node/src/submit_tx.rs
```diff
@@ -85,6 +107,26 @@ impl SubmitTransaction for SubmitTransactionSwitch {
...
         }
     }
+
+    async fn feeder_transaction_status(
+        &self,
+        hash: Felt,
+    ) -> Result<Option<mp_gateway::feeder::ProviderTransactionStatus>, SubmitTransactionError> {
+        match self.provider().ok() {
+            Some(provider) => provider.feeder_transaction_status(hash).await,
+            None => Ok(None),
```
prkpndy: **[Low] Downgrades feeder failures to `NOT_RECEIVED`.** When neither `L2Sync` nor `BlockProduction` is on, `self.provider().ok()` is `None` and this returns `Ok(None)`, which the gateway handler turns into `NOT_RECEIVED`. The `TransactionLookup` trait docs explicitly say errors should be propagated "so gateway clients do not silently downgrade transport or upstream failures into `NOT_RECEIVED`." During a transient service-down window this contradicts that intent — consider returning a 503/error instead of `Ok(None)` when no provider is available. (Same for `feeder_transaction` at line 127.)

madara/crates/client/rpc/src/test_utils.rs
```diff
@@ -64,6 +64,11 @@ impl SubmitTransaction for TestTransactionProvider {
     ) -> Result<AddInvokeTransactionResult, SubmitTransactionError> {
         unimplemented!()
     }
+}
+
+#[cfg(test)]
+#[async_trait]
+impl TransactionLookup for TestTransactionProvider {
```
prkpndy, line 71: **[Low]** This `impl TransactionLookup for TestTransactionProvider` is `#[cfg(test)]`-gated, but the sibling `impl SubmitTransaction` (above) is not. If `test_utils` is ever consumed by another crate's integration tests, the `cfg(test)` impl won't be visible there (it only applies to this crate's own tests). Confirm `test_utils` is internal-only, or gate both impls the same way.

madara/crates/client/mempool/src/chain_watcher_task.rs
heemankv: Fixed in 6ea71d723. Dropped the unused `Executed.transaction` payload entirely, so the block-production path no longer does the `executed_transactions().nth(tx_index)` walk, clone, Arc allocation, or retains executed transaction data that feeder never reads.

madara/crates/client/submit_tx/src/lib.rs
heemankv: Fixed in 6ea71d723 by taking the smaller cleanup route: removed the unused `_block_number`, `_transaction_index`, and `_is_on_l1` parameters from `feeder_transaction_from_confirmed_mempool`, so the function signature now matches the lookup it actually performs. I kept the lookup behavior unchanged because the status path and transaction path still need independently shaped feeder responses.

madara/crates/client/mempool/src/transaction_status.rs
heemankv: Fixed in 6ea71d723. Removed the `transaction: Arc<PreconfirmedExecutedTransaction>` field from `PreConfirmationStatus::Executed`; the variant now only carries the preconfirmed view and transaction index needed by the existing status flow.

madara/crates/client/submit_tx/src/lib.rs
heemankv: Fixed in 2aa5b9310. I deduped the feeder shaping into shared `mc-submit-tx` helpers (`feeder_status_from_backend_view` / `feeder_transaction_from_backend_view`) and reused them from both the mempool lookup path and the gateway handler, so the confirmed-vs-preconfirmed mapping now lives in one place.

madara/crates/client/submit_tx/src/lib.rs
heemankv: Addressed across 6ea71d723 and 2aa5b9310. The preconfirmed helper comment is now in place in `submit_tx/src/lib.rs`, and the dead `Executed.transaction` payload it referred to was removed in 6ea71d723.

madara/crates/client/gateway/server/src/handler.rs
heemankv: Fixed in 2aa5b9310. `handler.rs` now calls the shared `mc-submit-tx` feeder shaping helpers instead of carrying its own copy of `gateway_transaction` / `transaction_status_from_block` logic, so the two paths cannot drift.

madara/node/src/submit_tx.rs
heemankv: Fixed in 2aa5b9310. `TransactionLookupSwitch::feeder_transaction_status` and `feeder_transaction` now propagate `self.provider()?` instead of returning `Ok(None)` when neither L2 sync nor block production is active, so gateway callers no longer silently downgrade that state into `NOT_RECEIVED`.

madara/crates/client/rpc/src/test_utils.rs
heemankv: This one is already in the right state on the current head. Both `impl SubmitTransaction for TestTransactionProvider` and `impl TransactionLookup for TestTransactionProvider` are `#[cfg(test)]`-gated now in `rpc/src/test_utils.rs`, so the mismatch noted here is gone.

madara/crates/client/gateway/server/src/handler.rs
heemankv: Fixed in 2aa5b9310. `block_id_by_hash_response` now explicitly bounds the resolved block id against `latest_confirmed_block_n()`. That keeps the endpoint confirmed-scoped even if preconfirmed blocks ever gain hashes later.

madara/crates/primitives/gateway/src/feeder.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct TxFailureReason {
+    pub code: String,
+    pub error_message: String,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ProviderTransactionStatus {
+    pub tx_status: TransactionStatus,
+    pub finality_status: TransactionStatus,
+    pub execution_status: Option<TransactionExecutionStatus>,
```
prkpndy, line 39: `execution_status` here is missing `#[serde(skip_serializing_if = "Option::is_none")]`, unlike every other optional field in this struct (`block_hash`, `tx_failure_reason`, `tx_revert_reason`) **and** unlike `ProviderTransactionResponse::execution_status` just below. As a result a `not_received()` status serializes as `{"tx_status":"NOT_RECEIVED","finality_status":"NOT_RECEIVED","execution_status":null}` — emitting an `execution_status: null` key that the real Starknet feeder omits. Given the stated goal of matching the exact feeder shape, consider adding the same `skip_serializing_if` here for consistency.

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -101,28 +102,170 @@
...
+        TransactionStatus::AcceptedOnL2
+    }
+}
+
+fn transaction_status_from_block<D: MadaraStorageRead>(
+    block: &mc_db::MadaraBlockView<D>,
+) -> anyhow::Result<(TransactionStatus, Option<mp_convert::Felt>, Option<u64>)> {
+    let block_info = block.get_block_info()?;
+    if block.as_preconfirmed().is_some() {
+        Ok((TransactionStatus::Pending, None, Some(block_info.block_number())))
```
prkpndy: This `TransactionStatus::Pending` branch is unreachable: the only two callers of `transaction_status_from_block` (`feeder_status_from_backend_view` / `feeder_transaction_from_backend_view`) both early-return `not_received()` when `block.as_preconfirmed().is_some()` *before* calling this, so `block` is always confirmed here. The dead branch is harmless but misleading — it reads as if preconfirmed maps to `PENDING`, whereas the feeder design deliberately reports preconfirmed as `NOT_RECEIVED`. Consider dropping the preconfirmed arm (or replacing it with a debug-assert/comment).

madara/crates/client/submit_tx/src/lib.rs
```diff
@@ -101,28 +102,170 @@
...
+    }
+
+    let transaction = executed.get_transaction()?;
+    let (status, block_hash, _) = transaction_status_from_block(&executed.block)?;
+    let (execution_status, tx_revert_reason) = execution_status(&transaction.receipt);
+
+    Ok(ProviderTransactionStatus::with_status(status, Some(execution_status), block_hash, tx_revert_reason))
+}
+
+fn feeder_status_from_confirmed_mempool<D: MadaraStorageRead>(
```
prkpndy, line 185: Minor: `get_transaction_status` (called just above in `feeder_transaction_status`) already performs a `find_transaction_by_hash` for the `Confirmed` case, and this helper immediately does a second `find_transaction_by_hash` on the same backend view — a redundant double read plus a small TOCTOU window. Worth noting too that this `Confirmed` path is effectively dead via the only real caller (the gateway handler resolves confirmed txs directly from the backend before falling back to the lookup), so on a sequencer this branch isn't normally hit. Either thread the already-fetched view through, or add a short comment that this exists for non-handler callers.

prkpndy (commented):
Re-reviewed after the latest round — the earlier findings (dead `Executed.transaction` clone, duplicated feeder shaping, the `NOT_RECEIVED` error downgrade, and confirmed-scoping of `get_block_id_by_hash`) all look addressed. Nice. A few smaller things left inline.

One process ask: **CI is currently red** — `lint-rust / cargo-lint` and `test-js / test-js` are both failing. Could you get the pipeline green and also check the **Graphite bot score** before this is merged? (No need to action anything from me on the specifics of those two jobs — just flagging that they need fixing.)

madara/crates/primitives/gateway/src/feeder.rs
heemankv: Fixed in 68057f1b6. `ProviderTransactionStatus::execution_status` now matches the rest of the feeder shape and is omitted entirely when it is `None`; I also added a regression test in `mp-gateway` for the `NOT_RECEIVED` serialization case.

madara/crates/client/submit_tx/src/lib.rs
heemankv: Fixed in 68057f1b6. I removed the dead preconfirmed -> `PENDING` branch from `transaction_status_from_block` and replaced it with a `debug_assert!` documenting that feeder backend shaping must filter preconfirmed blocks earlier.

madara/crates/client/submit_tx/src/lib.rs
heemankv: Addressed in 68057f1b6 with the smaller cleanup route. I kept the extra lookup for now because threading the already-fetched backend view through the `TransactionLookup` fallback would widen this PR more than the remaining value justified, but I added a comment making it explicit that this helper exists for non-handler callers.

prkpndy (approved):

PR #1078, opened 2026-05-22
fix(rpc): align v0.10.2 gateway errors

Mohiiit:
## Summary
- Align JSON-RPC v0.10.2 HTTP behavior with Starknet v0.14.2 for pre-confirmed transaction status, trace errors, storage proof errors, compiled CASM errors, and invalid invoke proof handling.
- Preserve invoke v3 proof_facts through gateway conversion and map invalid proof_facts to INVALID_PROOF.
- Fix gateway enablement and historical hash lookup edge cases used by the RPC/gateway flows.

## Validation
- Formatting check passed.
- RPC unit tests passed.
- Gateway client/server compile check passed.

### DIFF
--- a/madara/crates/client/exec/src/lib.rs
+++ b/madara/crates/client/exec/src/lib.rs
@@ -227,8 +227,12 @@ fn transaction_error_is_entrypoint_not_found(error: &TransactionExecutionError) 
 }
 
 impl Error {
-    pub fn is_entrypoint_not_found(&self) -> bool {
+    pub fn is_call_contract_entrypoint_not_found(&self) -> bool {
         matches!(self, Self::CallContract(error) if transaction_error_is_entrypoint_not_found(&error.err))
+    }
+
+    pub fn is_message_fee_execution_error(&self) -> bool {
+        matches!(self, Self::MessageFeeEstimation(_))
     }
 }
 
--- a/madara/crates/client/exec/src/lib.rs
+++ b/madara/crates/client/exec/src/lib.rs
@@ -150,6 +150,7 @@
 #![allow(clippy::result_large_err)]
 
 use blockifier::{
+    execution::errors::{ConstructorEntryPointExecutionError, EntryPointExecutionError, PreExecutionError},
     state::cached_state::{CommitmentStateDiff, StateMaps},
     transaction::{errors::TransactionExecutionError, objects::TransactionExecutionInfo},
 };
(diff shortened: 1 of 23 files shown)


madara/crates/client/exec/src/lib.rs
```diff
@@ -205,6 +206,32 @@ pub enum Error {
...
+            ConstructorEntryPointExecutionError::ExecutionError { error, .. },
+        ) => is_entrypoint_not_found(error),
+        _ => false,
+    }
+}
+
+impl Error {
+    pub fn is_entrypoint_not_found(&self) -> bool {
+        matches!(self, Self::CallContract(error) if transaction_error_is_entrypoint_not_found(&error.err))
+    }
```
prkpndy, line 232: `is_entrypoint_not_found` only checks the `CallContract` variant, but `mc_exec::Error` has three other variants that also wrap `TransactionExecutionError`: `Reexecution`, `FeeEstimation`, and `MessageFeeEstimation`.

This means when `estimate_fee`, `estimate_message_fee`, or block re-execution (used by trace methods) encounters a missing entrypoint, the `From<mc_exec::Error>` conversion in `rpc/src/errors.rs` falls through to the generic `TxnExecutionError` branch — returning error code **41** instead of the spec-mandated **21** (`ENTRYPOINT_NOT_FOUND`).

Suggested fix:
```rust
pub fn is_entrypoint_not_found(&self) -> bool {
    match self {
        Self::CallContract(e) => transaction_error_is_entrypoint_not_found(&e.err),
        Self::Reexecution(e) => transaction_error_is_entrypoint_not_found(&e.err),
        Self::FeeEstimation(e) => transaction_error_is_entrypoint_not_found(&e.err),
        Self::MessageFeeEstimation(e) => transaction_error_is_entrypoint_not_found(&e.err),
        _ => false,
    }
}
```

madara/crates/client/rpc/src/versions/user/v0_10_2/methods/trace/mod.rs
```diff
@@ -18,6 +18,13 @@ use starknet_types_core::felt::Felt;
 // - traceBlockTransactions now accepts optional trace_flags parameter
 // - Results can include initial_reads when RETURN_INITIAL_READS flag is set
 
+fn validate_trace_block_transactions_block_id(block_id: &BlockId) -> RpcResult<()> {
+    if matches!(block_id, BlockId::Tag(BlockTag::PreConfirmed)) {
+        return Err(StarknetRpcApiError::CallOnPreConfirmed.into());
+    }
```
prkpndy, line 26: Madara rejects `trace_block_transactions` with `BlockId::Tag(PreConfirmed)` using error code **70**, but **Pathfinder accepts pre-confirmed blocks** in `trace_block_transactions` — it processes them if the Starknet version is >= v0.13.1.1. Error code 70 does not exist in Pathfinder at all.

This is a behavioral divergence. If this is mandated by the v0.10.2 spec then Pathfinder is non-compliant, but if not, Madara is being unnecessarily restrictive. Worth confirming against the spec text before merging.

prkpndy (approved):
LGTM

PR #1145, opened 2026-06-11
fix(rpc): structured execution errors + real tx index

Mohiiit:
> Stacked on #1144. Only the last commit is new.

Two spec gaps in error 41/40 `data`, found cross-validating against pathfinder v0.22.5 and juno v0.16.2:
1. `execution_error`/`revert_error` was a flat display string (leaking internal view text) instead of the spec's structured `CONTRACT_EXECUTION_ERROR` — the nested per-frame call chain SDKs parse. Both references emit the structure.
2. `transaction_index` was hardcoded to 0 on the hard-error path, blaming tx 0 in batch estimates/simulations.

```mermaid
flowchart LR
    A[blockifier error stack] --> B[flatten frames]
    B --> C["entrypoint / Cairo1 revert frames → call frames"]
    B --> D["last string frame → leaf"]
    C --> E
    D --> E["fold outermost-first:<br/>{contract_address, class_hash, selector,<br/>error: {…, error: 'panic reason'}}"]
```

The converter mirrors pathfinder's `error_stack_frames_to_json` and juno's `error/stack.rs` exactly. v0.8+ endpoints emit the object on both revert and hard-error paths; v0.7.1 keeps its spec string form.

**Tests:** 184/184 pass. New: batch-of-2 estimate where tx #1 fails → asserts `transaction_index == 1` and the structured error is rooted at the failing account (fails on base); converter unit tests incl. the `ENTRYPOINT_NOT_FOUND` shape.

### DIFF
--- a/madara/crates/client/rpc/src/errors.rs
+++ b/madara/crates/client/rpc/src/errors.rs
@@ -273,6 +273,24 @@ impl From<mc_exec::Error> for StarknetRpcApiError {
     }
 }
 
+impl StarknetRpcApiError {
+    /// The v0.7.1 spec predates structured execution errors: both CONTRACT_ERROR `revert_error`
+    /// and TRANSACTION_EXECUTION_ERROR `execution_error` are flat strings there, whereas the
+    /// shared `From<mc_exec::Error>` impl builds the structured CONTRACT_EXECUTION_ERROR object
+    /// introduced in v0.8. v0.7.1 handlers must convert through this instead of `?`/`From`.
+    pub fn from_exec_error_v0_7(err: mc_exec::Error) -> Self {
+        match &err {
+            mc_exec::Error::CallContract(error) if error.is_entrypoint_not_found() => Self::EntrypointNotFound,
+            mc_exec::Error::CallContract(error) if error.is_contract_not_found() => Self::contract_not_found(),
+            mc_exec::Error::CallContract(_) => Self::ContractError { revert_error: format!("{:#}", err).into() },
+            mc_exec::Error::Reexecution(error) => {
+                Self::TxnExecutionError { tx_index: error.index(), error: format!("{:#}", err).into() }
+            }
+            _ => Self::TxnExecutionError { tx_index: 0, error: format!("{:#}", err).into() },
+        }
+    }
+}
+
 impl From<StarknetRpcApiError> for jsonrpsee::types::ErrorObjectOwned {
     fn from(err: StarknetRpcApiError) -> Self {
         jsonrpsee::types::ErrorObjectOwned::owned((&err).into(), err.to_string(), err.data())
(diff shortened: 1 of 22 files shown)


madara/crates/client/rpc/src/utils/execution_error.rs
```diff
@@ -0,0 +1,231 @@
...
+        .filter_map(|frame| match frame {
+            Frame::Call { contract_address, class_hash, selector } => Some((contract_address, class_hash, selector)),
+            _ => None,
+        })
+        .rev()
+        .fold(json!(leaf), |child, (contract_address, class_hash, selector)| {
+            json!({
+                "contract_address": contract_address,
+                "class_hash": class_hash,
+                "selector": selector,
```
prkpndy: `selector` is always emitted here, so constructor frames serialize as `null`. The Starknet OpenRPC schema types `selector` as `FELT` inside `CONTRACT_EXECUTION_ERROR`, so this payload becomes schema-invalid for constructor failures. Please either omit constructor frames from this shape or encode them in a way that still satisfies the schema.

prkpndy (commented):
This mostly closes the structured execution-error gap, and the real `transaction_index` fix looks correct. I found one remaining compatibility issue in the new serializer: constructor failures now produce a `CONTRACT_EXECUTION_ERROR` frame with `selector: null`, which does not satisfy the Starknet OpenRPC schema (`selector` is a `FELT` there). I left the implementation note inline.

madara/crates/client/rpc/src/errors.rs
```diff
@@ -254,16 +254,57 @@ impl StarknetRpcApiError {
...
+        mc_exec::Error::CallContract(error) => {
+            if error.is_entrypoint_not_found() {
+                E::EntrypointNotFound
+            } else if error.is_contract_not_found() {
+                E::contract_not_found()
+            } else {
+                let revert_error = match payload {
+                    ExecErrorPayload::Structured => crate::utils::contract_execution_error(error.error()),
+                    ExecErrorPayload::FlatString => format!("{:#}", err).into(),
+                };
```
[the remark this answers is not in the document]
Mohiiit: Fixed in `d79c33762`: the v0.7.1 `CallContract` FlatString path now formats the inner `TransactionExecutionError` via `error.error()` instead of the outer `mc_exec::Error`, matching the old `revert_reason()` behavior. I also strengthened `call_reverted_returns_string_revert_error` to assert the response does not contain the `Calling contract ... on top of ...` wrapper. Local validation: `cargo fmt --all -- --check`, `cargo test -p mc-rpc --lib call_reverted_returns_string_revert_error`, and `cargo test -p mc-rpc --lib execution_error`.

madara/crates/client/rpc/src/utils/execution_error.rs
Mohiiit: Fixed and preserved after the rebase onto `main`: constructor/no-selector frames now omit the `selector` key instead of emitting `null`, and the unit test asserts `constructor_frame_omits_selector`. Local validation: `cargo fmt --all -- --check` and `cargo test -p mc-rpc --lib execution_error`.

prkpndy (approved):

PR #1147, opened 2026-06-11
fix(rpc): read-method spec gaps

Mohiiit:
Three read-method divergences from pathfinder v0.22.5 and juno v0.16.2:

```mermaid
flowchart LR
    subgraph before [" before "]
        A3["status of reverted tx"] --> B3["no failure_reason"]
        A4["getEvents(to_block > tip)"] --> B4[⛔ 24]
        A5["latest / l1_accepted, none yet"] --> B5[⛔ 32 NO_BLOCKS]
    end
    subgraph after [" after "]
        C3["✅ failure_reason: '…'"]
        C4[✅ empty page / scan to tip]
        C5[⛔ 24 BLOCK_NOT_FOUND]
    end
    B3 -.-> C3
    B4 -.-> C4
    B5 -.-> C5
```

- `failure_reason` (spec field, "only appears if execution_status is REVERTED") added to `TxnFinalityAndExecutionStatus`.
- `getEvents` numeric range bounds are no longer required to exist — clients paginating by block number at the chain tip got errors where both references succeed; this also lets a bound reference the pre-confirmed block number.
- `NO_BLOCKS` (32) stays only where the spec defines it (`blockNumber`/`blockHashAndNumber`); tag resolution failures are `BLOCK_NOT_FOUND` (24).

Candidate-transaction visibility was dropped from this PR (candidate status is slated for deprecation).

**Tests:** focused mc-rpc regressions pass locally; full PR CI runs after ready-for-review.

### DIFF
--- a/madara/crates/client/rpc/src/block_id.rs
+++ b/madara/crates/client/rpc/src/block_id.rs
@@ -10,6 +10,10 @@ pub trait StateViewResolvable: Sized {
     fn resolve_state_view(&self, starknet: &Starknet) -> Result<MadaraStateView, StarknetRpcApiError>;
 }
 
+pub trait EventRangeBoundResolvable: StateViewResolvable {
+    fn event_range_number(&self) -> Option<u64>;
+}
+
 // v0.7/v0.8 rpc
 
 impl StateViewResolvable for mp_rpc::v0_7_1::BlockId {
@@ -39,6 +43,15 @@ impl StateViewResolvable for mp_rpc::v0_7_1::BlockId {
     }
 }
 
+impl EventRangeBoundResolvable for mp_rpc::v0_7_1::BlockId {
+    fn event_range_number(&self) -> Option<u64> {
+        match self {
+            Self::Number(block_n) => Some(*block_n),
+            _ => None,
+        }
+    }
+}
+
 impl BlockViewResolvable for mp_rpc::v0_7_1::BlockId {
     fn resolve_block_view(&self, starknet: &Starknet) -> Result<MadaraBlockView, StarknetRpcApiError> {
         match self {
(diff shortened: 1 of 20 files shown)


madara/crates/client/rpc/src/block_id.rs
```diff
@@ -152,4 +156,17 @@ impl Starknet {
...
+    /// Resolves a `getEvents` range bound: numeric bounds are raw block numbers that are not
+    /// required to exist — an out-of-range `from_block` yields an empty page and an out-of-range
+    /// `to_block` scans up to the chain tip, like pathfinder and juno (this also lets a bound
+    /// reference the pre-confirmed block number). Hash and tag bounds must resolve through the
+    /// backend. Shared by every RPC version with its own `getEvents` implementation; v0.10.x
+    /// re-export the v0.9 `BlockId`.
+    pub fn resolve_event_range_bound(&self, block_id: mp_rpc::v0_9_0::BlockId) -> Result<u64, StarknetRpcApiError> {
+        match block_id {
+            mp_rpc::v0_9_0::BlockId::Number(block_n) => Ok(block_n),
+            block_id => Ok(self.resolve_view_on(block_id)?.latest_block_n().unwrap_or(0)),
```
prkpndy: `from_block` and `to_block` cannot share the same tag-resolution rule. On an empty chain, `from_block = latest` should still fail with `BLOCK_NOT_FOUND`, but `latest_block_n().unwrap_or(0)` silently turns that missing latest block into block 0, so all three `getEvents` handlers return an empty page instead. Pathfinder keeps lower and upper bound mapping asymmetric for exactly this reason.

prkpndy (commented):
This fixes several real read-method gaps, and the added regression coverage is useful. I found one remaining `getEvents` edge case: lower tag bounds still collapse to block 0 on an empty chain, so `from_block = latest` returns an empty page instead of `BLOCK_NOT_FOUND`. I left the implementation note inline.

madara/crates/client/rpc/src/block_id.rs
Mohiiit: Fixed in the rebased head: `from_block` now uses `resolve_event_from_block_bound` and returns `BLOCK_NOT_FOUND` for empty-chain tag bounds, while `to_block` keeps the empty-page fallback. The focused regression `get_events_from_latest_on_empty_chain_returns_block_not_found` passes for v0.7.1, v0.10.0, and v0.10.2.

prkpndy (approved):

PR #1157, opened 2026-06-17
[codex] Handle pending state update settlement verification

heemankv:
## Summary

- Treat state-update settlement verification as pending when the settlement contract is still reporting the previous batch boundary.
- Only mark the settled batch as completed after verification is actually successful.
- Make Ethereum blob state updates wait for a successful finalized receipt before moving the job to verification.
- Add bounded same-nonce fee-bump replacement handling for pending or underpriced state-update transactions.
- Persist state-update tx nonce and per-attempt tx metadata for operational visibility.
- Increase UpdateState/PriorityProcessing queue visibility based on the configured state-transition processing timeout, and reconcile existing SQS queue attributes during setup.

## Root Cause

StateTransition verification compares the settlement contract's `state_block_number()` to the expected batch end block. On PC testnet, the first verification can see the previous batch end block even though the state update transaction path has just run. The old code treated that previous-boundary read as a hard rejection, and the generic job service surfaced it as `Verification rejected. Max process attempts reached: 1`.

The blob settlement client also accepted a tx hash even when finality returned `None`, and the finality wait did not reject failed receipts. That could move jobs into contract-state verification before a tx was known to have successfully finalized.

For pending Ethereum transactions, the previous processing path waited for the whole state-transition timeout without an explicit same-nonce replacement policy. That made stuck or underpriced transactions depend on queue redelivery/healing behavior instead of being handled directly by the state-update processor.

## Implementation Notes

- Default state-update transaction confirmation window: `MADARA_ORCHESTRATOR_ETHEREUM_TX_CONFIRMATION_TIMEOUT_SECS=300`.
- Default bounded replacements: `MADARA_ORCHESTRATOR_ETHEREUM_MAX_FEE_BUMPS=2`.
- Replacement transactions reuse the original nonce and double the gas multiplier, bounded by `MADARA_ORCHESTRATOR_EIP1559_MAX_GAS_MUL_FACTOR`.
- UpdateState and PriorityProcessing queue visibility is now `max(3 * base_visibility, MADARA_ORCHESTRATOR_STATE_TRANSITION_TIMEOUT_SECONDS + 600)`.

## Validation

- `cargo fmt -p orchestrator -p orchestrator-ethereum-settlement-client -p orchestrator-starknet-settlement-client -p orchestrator-settlement-client-interface`
- `git diff --check`
- `cargo check -p orchestrator-settlement-client-interface`
- `cargo check -p orchestrator-starknet-settlement-client`
- `cargo check -p orchestrator-ethereum-settlement-client`
- `cargo test -p orchestrator-ethereum-settlement-client gas_multiplier_tests --lib`
- `PATH=/Users/heemankverma/Work/Carnot/repos/madara/main/sequencer_venv/bin:$PATH cargo check -p orchestrator`
- `PATH=/Users/heemankverma/Work/Carnot/repos/madara/main/sequencer_venv/bin:$PATH cargo test -p orchestrator queue_control --lib`
- `set -a; source .env.test; set +a; export MADARA_ORCHESTRATOR_ETHEREUM_SETTLEMENT_RPC_URL=http://127.0.0.1:8545; PATH=/Users/heemankverma/Work/Carnot/repos/madara/main/sequencer_venv/bin:$PATH cargo test -p orchestrator state_update_job --lib`


Mohiiit (approved):
Approved. I do not see a blocker for this PR.

PR #1168, opened 2026-06-18
Use direct Atlantic L2 verification for Starknet L3 settlement

heemankv:
## Summary

This updates the Starknet L3 settlement pipeline to use Atlantic's direct `PROOF_VERIFICATION_ON_L2` style flow for Privily/Starkpay, instead of producing a local SNOS proof first and then submitting that proof for a second layout-bridge verification.

## Orchestrator changes

- For `Layer::L3`, `ProofCreation` is skipped. The proving trigger now returns early because proof generation is handled by the Atlantic proof-registration request.
- For `Layer::L3`, `ProofRegistration` jobs are created directly from completed `SnosRun` jobs.
  - The job input is the stored Cairo PIE from SNOS.
  - No `proof.json` or `proof_part2.json` download path is configured.
  - No legacy SNOS fact cross-check is requested from the old fact checker path.
- `ProofRegistration` can now process both existing proof inputs and Cairo PIE inputs.
  - Existing proof input behavior is preserved for the old path.
  - Cairo PIE input submits an Atlantic `CreateJob` request, using the configured Atlantic `result` and `sharp_prover` values. For this Privily path, config should set `result=PROOF_VERIFICATION_ON_L2` and `sharpProver=stwo`.
- L3 `StateTransition` no longer reads `proof.json` / `proof_part2.json` and no longer extracts bootloader public outputs from proof files.
  - It uses the raw stored `snos_output.json` produced by SNOS.
  - L3 state-update metadata now carries only the raw SNOS output path; legacy program/blob artifacts are not passed into settlement.
  - It passes an empty legacy `program_output` vector through the shared settlement trait.
- The Starknet settlement client now calls the new one-argument Piltover ABI: `update_state(snos_output)`.
  - The shared `SettlementClient::update_state_calldata` trait shape is kept unchanged to avoid a wider Ethereum/L2 refactor.
- The Atlantic Starknet proving-layer helper no longer overwrites the configured query result with `PROOF_GENERATION`; it only adds the layout parameter. This allows `PROOF_VERIFICATION_ON_L2` to be configured and respected.

## Expected pipeline for L3

```text
SnosRun
  -> ProofRegistration submits Cairo PIE to Atlantic
  -> Atlantic proves with configured SHARP prover and verifies on L2
  -> DataSubmission remains the existing downstream gate before state update
  -> StateTransition submits raw SNOS output to Piltover
  -> Piltover recomputes the expected Keccak SHARP fact and checks Satellite
```

## Validation

- `cargo check -p orchestrator-atlantic-service -p orchestrator-starknet-settlement-client -p orchestrator`
- `cargo test -p orchestrator-atlantic-service atlantic_status_response_deserializes_l2_bridge_fact_hash_step`
- `cargo test -p orchestrator-starknet-settlement-client --lib`

Notes:

- `cargo fmt --check --manifest-path orchestrator/Cargo.toml` currently reports a pre-existing formatting diff in `orchestrator/src/compression/squash.rs`, outside this PR's changes.
- The full `cargo test -p orchestrator --lib` target currently fails in unrelated DA/batching test fixtures due Starknet type-shape drift before reaching this PR's tests.

### DIFF
--- a/orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
@@ -23,7 +23,9 @@ use tempfile::NamedTempFile;
 use url::Url;
 
 use crate::client::{AtlanticBucketInfo, AtlanticClient, AtlanticJobConfig, AtlanticJobInfo};
-use crate::types::{AtlanticBucketStatus, AtlanticCairoVm, AtlanticQuery, AtlanticQueryStep, AtlanticSharpProver};
+use crate::types::{
+    AtlanticBucketStatus, AtlanticCairoVm, AtlanticChain, AtlanticQuery, AtlanticQueryStep, AtlanticSharpProver,
+};
 
 #[derive(Debug, Clone)]
 pub struct AtlanticValidatedArgs {
@@ -185,7 +187,7 @@ impl ProverClient for AtlanticProverService {
                     AtlanticQueryStatus::Received => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::InProgress => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::Done => {
-                        if matches!(&self.result, AtlanticQueryStep::ProofVerificationOnL2) {
+                        if Self::is_l2_verification_query(&atlantic_query) {
                             if let Err(reason) = Self::ensure_l2_verification_completed(&atlantic_query) {
                                 return Ok(TaskStatus::Failed(reason));
                             }
@@ -444,6 +446,13 @@ impl AtlanticProverService {
 
     fn should_mock_proof(&self) -> bool {
         self.mock_fact_hash
+    }
+
+    fn is_l2_verification_query(atlantic_query: &AtlanticQuery) -> bool {
+        matches!(atlantic_query.chain.as_ref(), Some(AtlanticChain::L2))
+            || matches!(atlantic_query.result.as_ref(), Some(AtlanticQueryStep::ProofVerificationOnL2))
+            || matches!(atlantic_query.step.as_ref(), Some(AtlanticQueryStep::BridgeFactHash))
+            || atlantic_query.steps.iter().any(|step| matches!(step, AtlanticQueryStep::BridgeFactHash))
     }
 
     fn ensure_l2_verification_completed(atlantic_query: &AtlanticQuery) -> Result<(), String> {
--- a/orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
+++ b/orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
@@ -180,10 +180,17 @@ impl ProverClient for AtlanticProverService {
     ) -> Result<TaskStatus, ProverClientError> {
         match task {
             TaskType::Job => {
-                match self.atlantic_client.get_job_status(job_key).await?.atlantic_query.status {
+                let atlantic_query = self.atlantic_client.get_job_status(job_key).await?.atlantic_query;
+                match atlantic_query.status {
                     AtlanticQueryStatus::Received => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::InProgress => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::Done => {
+                        if matches!(&self.result, AtlanticQueryStep::ProofVerificationOnL2) {
+                            if let Err(reason) = Self::ensure_l2_verification_completed(&atlantic_query) {
+                                return Ok(TaskStatus::Failed(reason));
+                            }
+                        }
+
                         if !cross_verify {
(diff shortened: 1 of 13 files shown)


orchestrator/src/worker/event_handler/triggers/proof_registration.rs
```diff
@@ -69,3 +72,62 @@ impl JobTrigger for ProofRegistrationJobTrigger {
...
+                job.internal_id
+            )
+        })?;
+
+        let metadata = JobMetadata {
+            common: CommonMetadata::default(),
+            specific: JobSpecificMetadata::Proving(ProvingMetadata {
+                block_number: snos_metadata.start_block,
+                input_path: Some(ProvingInputType::CairoPie(cairo_pie_path)),
+                ensure_on_chain_registration: None,
```
prkpndy, line 108: Because the new L3 path hardcodes `ensure_on_chain_registration: None`, the downstream verifier now calls `get_task_status(..., cross_verify = false)` and accepts any Atlantic `DONE` status as success. That means we no longer assert that the query actually completed the intended `PROOF_VERIFICATION_ON_L2` flow or reached the expected terminal step before advancing into DA and state transition. Please keep a strict terminal-result check for this path, either by preserving a verification signal in metadata or by validating the returned Atlantic query payload.

prkpndy (changes requested):
I found one blocking issue inline.

Separately, this PR still needs better orchestrator-level coverage for the new L3 path. The added tests cover CLI validation for `PROOF_VERIFICATION_ON_L2` and deserialization of Atlantic's `BRIDGE_FACT_HASH` status, but they do not exercise the core behavior that changed here: direct `SnosRun -> ProofRegistration` job creation, the new Cairo PIE branch in `RegisterProofJobHandler::process_job()`, or the L3 `StateTransition` branch that now settles from raw `snos_output.json`.

Validation I was able to run locally:
- `cargo test -p orchestrator-starknet-settlement-client --lib`
- `cargo test -p orchestrator-atlantic-service atlantic_status_response_deserializes_l2_bridge_fact_hash_step -- --nocapture`

I could not run the broader orchestrator target in this environment because `cargo test -p orchestrator --features testing ...` stops in `apollo_starknet_os_program` without the Cairo 0 `cairo-compile` toolchain.

orchestrator/crates/prover-clients/atlantic-service/src/lib.rs
```diff
@@ -180,10 +180,17 @@ impl ProverClient for AtlanticProverService {
     ) -> Result<TaskStatus, ProverClientError> {
         match task {
             TaskType::Job => {
-                match self.atlantic_client.get_job_status(job_key).await?.atlantic_query.status {
+                let atlantic_query = self.atlantic_client.get_job_status(job_key).await?.atlantic_query;
+                match atlantic_query.status {
                     AtlanticQueryStatus::Received => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::InProgress => Ok(TaskStatus::Processing),
                     AtlanticQueryStatus::Done => {
+                        if matches!(&self.result, AtlanticQueryStep::ProofVerificationOnL2) {
```
prkpndy: This only validates L2-verification completion when the service-wide `self.result` is `PROOF_VERIFICATION_ON_L2`. That fixes the new L3 flow, but it still misses the existing L2 `ProofRegistration` path: `submit_l2_query()` always creates an L2-verification query, yet the Atlantic service for L2 remains configured with `result = PROOF_GENERATION` for the earlier PIE submission step. In that configuration this guard never runs, so `ProofRegistration` on L2 still accepts any Atlantic `DONE` status as success. The expected terminal mode needs to come from the job/query being polled, not from the service's default PIE-submission config.

prkpndy (changes requested):
The latest follow-up commit addresses the earlier L3-specific concern, and I resolved that thread.

I found one remaining blocker inline: the new Atlantic completion check is keyed off the service-wide `self.result`, which fixes the L3 configuration but still leaves the existing L2 `ProofRegistration` path accepting any Atlantic `DONE` status as success. `submit_l2_query()` always creates an L2-verification query, but L2 deployments keep the service configured with `result = PROOF_GENERATION` for the earlier PIE submission step, so the new guard never runs there.

Validation I ran on the current head (`6febe4228b6b71de6b1af0a2c3f5eeff4d0231e4`):
- `cargo test -p orchestrator-atlantic-service atlantic_l2_verification_status -- --nocapture`
- `cargo test -p orchestrator-starknet-settlement-client --lib`

aneesh404 (approved):
Rechecked the latest head. The remaining blocker from my prior review is resolved: the L2 completion validation now keys off the polled Atlantic query rather than the service default, and there is a regression test for that exact path. I also reran `cargo test -p orchestrator-atlantic-service atlantic_l2_verification_status -- --nocapture` on this head, and the targeted status tests passed.

PR #1192, opened 2026-06-19
[codex] Retry transient SNOS failures via SQS

prkpndy:
## Summary
- classify retryable versus fail-fast SNOS failures from typed `generate-pie` errors
- keep fail-fast only for non-retryable SNOS processing failures such as Cairo/OS execution errors
- restore the SQS retry flow for every other job type during processing, and for all processing panics
- point the Madara workspace at the released SNOS tag `v0.14.2-rc.5` from `keep-starknet-strange/snos`
- expand orchestrator tests around SNOS retryability, generic SNOS failures, and panic retry behavior

## Why
Recent fail-fast behavior marked processing failures as `Failed` immediately. That is too aggressive for intermittent SNOS RPC and proof-fetch failures, and it is also too aggressive for non-SNOS jobs, which should continue using the original SQS retry lifecycle until DLQ. The only failures that should still fail fast are non-retryable SNOS execution failures, so alerting can stop the sequencer promptly.

Panics are also not safely classifiable at the service boundary: a `SnosRun` panic could come from transient RPC or transport paths just as easily as from a deterministic internal bug. This PR therefore treats all processing panics as SQS-retryable and reserves fail-fast for positively classified non-retryable SNOS errors.

## Impact
- retryable SNOS processing failures are restored to their original job status and the SQS message is left unacked for redelivery
- all processing panics now restore the original job status and rely on SQS/DLQ retries instead of being marked `Failed` immediately
- all non-SNOS processing failures continue to rely on SQS/DLQ retries instead of being marked `Failed` immediately
- `process_retry_attempt_no` still increments when the job is restored for SQS retry
- deterministic SNOS execution/configuration/proof-shape failures continue to fail fast

## Validation
- `cargo fmt --all --check`
- `source /Users/prakhar/work/karnot/madara/sequencer_venv/bin/activate && cargo clippy -p orchestrator --all-features --tests --no-deps -- -D warnings`
- `source /Users/prakhar/work/karnot/madara/sequencer_venv/bin/activate && cargo test -p orchestrator --lib --no-run`

### DIFF
--- a/orchestrator/src/tests/jobs/mod.rs
+++ b/orchestrator/src/tests/jobs/mod.rs
@@ -32,12 +32,15 @@
 use std::sync::Arc;
 use std::time::Duration;
 
+use generate_pie::error::{BlockProcessingError, PieGenerationError};
 use mockall::predicate::eq;
 use mongodb::bson::doc;
 use rstest::rstest;
+use starknet::providers::ProviderError;
 use tokio::time::sleep;
 
 use crate::core::client::alert::MockAlertClient;
+use crate::error::job::snos::SnosError;
 use crate::tests::common::MessagePayloadType;
 use crate::tests::config::{ConfigType, MockType, TestConfigBuilder};
 use crate::tests::utils::build_job_item;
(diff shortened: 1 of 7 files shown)


orchestrator/src/worker/event_handler/service.rs
```diff
@@ -435,6 +462,52 @@ impl JobHandlerService {
...
+            }
+            Err(db_err) => {
+                error!(job_id = ?job.id, error = ?db_err, "Failed to reset retryable job state for SQS retry");
+                JobError::Other(OtherError::from(format!(
+                    "Failed to reset job state: {}. Original error: {}",
+                    db_err, original_error
+                )))
+            }
+        }
+    }
```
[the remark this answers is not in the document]
prkpndy: This would be a very rare case and we are okay with that. We can manually retry the jobs in this case

orchestrator/src/worker/event_handler/service.rs
```diff
@@ -357,14 +380,36 @@ impl JobHandlerService {
                     .or_else(|| panic.downcast_ref::<&str>().copied())
                     .unwrap_or("Unknown panic message");
 
-                error!(job_id = ?id, panic_msg = %panic_msg, "Job handler panicked during processing, marking as failed");
+                let panic_error =
+                    JobError::Other(OtherError::from(format!("Job handler panicked with message: {}", panic_msg)));
                 let reason =
                     format!("Processing attempt {} panicked: {}", job.metadata.common.process_attempt_no, panic_msg);
+
+                if job.job_type != JobType::SnosRun {
```
heemankv: **Blocking:** This still makes every `SnosRun` panic fail fast, bypassing the retryability split introduced above. At the panic boundary we no longer know whether the underlying cause was deterministic Cairo/OS execution or a transient RPC/transport path that panicked internally. The equivalent `Err(retryable_error)` path restores the job and NACKs for SQS retry, but this path marks the job `Failed` and ACKs it. I think SNOS panics should default to the SQS retry path unless they can be classified as deterministic.

orchestrator/src/tests/jobs/mod.rs
```diff
@@ -626,6 +631,285 @@ async fn process_job_job_handler_returns_error_works() {
...
 }
 
+/// Tests that unclassified SNOS processing errors restore the original job status and rely on SQS/DLQ retries.
+#[rstest]
+#[case::created(JobStatus::Created)]
+#[case::verification_failed(JobStatus::VerificationFailed)]
+#[case::pending_retry(JobStatus::PendingRetry)]
+#[tokio::test]
+#[allow(clippy::await_holding_lock)]
+async fn process_job_unclassified_snos_error_restores_original_status_for_sqs_retry(#[case] initial_status: JobStatus) {
```
heemankv, line 641: **Blocking:** This new expectation conflicts with the existing `process_job_failure_marks_job_as_failed` test later in this file. That older test still uses `SnosRun` + `JobError::Other` and expects `process_job` to return `Ok` and move the job to `Failed`, but the implementation now restores and returns `Err` for that exact shape. Please update/remove the stale test and add explicit SNOS panic coverage for the remaining fail-fast path.

heemankv (changes requested):
Great work on this PR! The move to typed SNOS errors makes the retry policy much clearer, and the added classifier tests cover the important RPC-vs-execution distinction.

A few blockers inline before this is merge-ready:
- `SnosRun` panics still fail fast unconditionally, bypassing the new retryability model.
- The test suite has a stale contradictory expectation for `SnosRun + JobError::Other`.

Also, the PR is currently conflicting with `main` in `Cargo.toml` / `Cargo.lock`, and the description mentions `karnotxyz/snos` while the dependency points at `keep-starknet-strange/snos.git`.

PR #1208, opened 2026-06-30
Guard aggregator input size during batching

Mohiiit:
## Summary
- add a conservative aggregator input size upper bound to aggregator batches
- close aggregator batches before adding a block that would exceed the 800,000 input cap
- validate the formula with the Paradex batch 922 regression case: 866,484

## Validation
- cargo fmt --all
- cargo clippy -p orchestrator --lib --tests -- -D warnings
- cargo test -p orchestrator worker::event_handler::triggers::batching::aggregator::tests --lib
- git pre-push hook completed successfully

## Notes
- Full `cargo test -p orchestrator --lib` was attempted locally but interrupted after many external-service dependent failures and hangs outside this change area.
- Local cargo commands used `CARGO_TARGET_DIR=/Volumes/sandisk_ssd/madara-karnot` from `.zshrc`, `sequencer_venv`, and `BINDGEN_EXTRA_CLANG_ARGS` for macOS SDK headers.


prkpndy (approved):
Everything looks good, just one confirmation: Does aggregator_input_size_upper_bound represent the maximum possible value that we are calculating using the proxy calculation for aggregator input?

It's not 8 lakhs, according to my understanding. Is that correct?

PR #1226, opened 2026-07-06
fix(orchestrator): bound anti-join queries

Mohiiit:
## Summary

Bounds the SNOS and proving anti-join scans to the first unsettled SNOS batch derived from the latest completed state-transition job. This keeps already-settled historical batches/jobs out of the expensive lookup path while preserving the existing fallback when no watermark is available.

Live explain checks before this change showed the bounded shape reducing candidate scans by more than 99% on mock, testnet, and mainnet, with mainnet dropping from about 1.38M candidate batches to about 598 in the sampled path.

## Validation

- Formatting check passed
- Orchestrator clippy passed
- Mock-backed SNOS and proving trigger tests passed
- New DB lower-bound tests compile; local execution needs Mongo at localhost:27017


prkpndy (approved):
lgtm. gg!

Issue #1232, opened 2026-07-14
fix(orchestrator): make verification queue retries idempotent


heemankv:
## Context

This is a follow-up to the pc-mainnet incident investigation around orchestrator SQS verification retries.

Current PR scope should stay limited to:

- reusing cached SQS consumers to avoid per-message AWS client / IRSA credential creation
- preserving full AWS SDK error context in SQS and storage-related failures

This issue tracks the separate design change for making verification retry scheduling idempotent.

Codex thread id for continuation: `019f5fc0-250f-7583-a8fa-3202a049cbdc`

## Problem

The verification worker currently schedules the next verification message before the current SQS message is ACKed.

Current flow:

```text
Receive verification message
        |
verify_job() returns Pending
        |
Send delayed successor verification message
        |
ACK current message
```

If the successor send succeeds but the ACK/delete of the current SQS message fails, the old message can become visible again. Since the queue payload currently only carries the job id, the stale message is indistinguishable from the fresh successor.

That can create this fanout:

```text
Old message becomes visible again
        |
Old message runs verification again
        |
Old message sends another successor
        |
Both chains continue
```

This inflates verification attempts, repeated timeout/failure alerts, and receive/delete mismatch symptoms.

## Agreed Direction

Do not move to a `next_verification_at` DB scheduler for now.

Instead, keep using SQS delay for timing, but make verification queue messages idempotent by adding the expected verification attempt to the message body.

Current message:

```json
{
  "id": "job-id"
}
```

New verification message:

```json
{
  "id": "job-id",
  "verification_attempt": 17
}
```

The verification queue is already specific to verification, so we do not need an additional `action` field for this design.

## Intended Semantics

The queue message should carry the verification attempt it is allowed to execute.

On consuming a verification message:

1. Load the job from DB.
2. Compare `message.verification_attempt` with `job.metadata.common.verification_attempt_no`.
3. If the message attempt is stale, ACK and return success without verifying and without enqueueing a successor.
4. If it matches, run verification.
5. If verification is still pending:
   - increment the DB verification attempt
   - enqueue the next verification message with the incremented `verification_attempt`
   - then ACK the current message via normal worker post-processing

Example stale duplicate handling:

```text
message.verification_attempt = 4
DB verification_attempt_no = 7
=> stale delivery
=> ACK and skip
```

This does not remove the send-before-ACK ordering, but it makes duplicate deliveries harmless.

## Manual / Admin Paths

[...]

PR #1233, opened 2026-07-14
fix(orchestrator): reuse SQS consumers and preserve AWS error context

heemankv:
## Summary

Fixes the orchestrator SQS path involved in the pc-mainnet incident by reducing AWS SDK client/credential churn and preserving the underlying AWS SDK error context in logs/alerts.

Changes included:

- cache and reuse OmniQueue SQS consumers per queue type instead of rebuilding a consumer while wrapping every received message for ACK/NACK
- preserve full AWS SDK error context for SQS receive/ACK/NACK failures
- preserve storage/job error source context so S3 dispatch/service failures are no longer collapsed into a plain string
- retry successful-message ACK once after a short delay before surfacing the ACK failure

## Why

The incident evidence showed a large spike in `AssumeRoleWithWebIdentity` calls from the orchestrator role, close to the SQS receive volume. The old flow received messages with the persistent SQS client, then rebuilt an OmniQueue consumer while wrapping each message for ACK/NACK. That recreated AWS config/client/credential state in the ACK path.

The alerts also hid useful root-cause detail behind plain `dispatch failure` strings because SDK errors were converted with `to_string()` and source chains were dropped.

## Follow-up Required

This PR intentionally does not change verification retry semantics.

A follow-up PR should address idempotent verification retry messages as described in:

- https://github.com/madara-alliance/madara/issues/1232

That follow-up should add `verification_attempt` to verification queue messages so stale duplicate deliveries can be ACKed and ignored instead of producing additional successor verification messages.

## Testing

Not run locally. We intentionally deferred format/lint/checks for one combined validation pass later.

### DIFF
--- a/orchestrator/src/worker/controller/event_worker.rs
+++ b/orchestrator/src/worker/controller/event_worker.rs
@@ -46,15 +46,15 @@ fn format_error_context(error: &impl Error) -> String {
     DisplayErrorContext(error).to_string()
 }
 
-async fn ack_message_with_retry(message: &Delivery) -> Result<(), ConsumptionError> {
-    if let Err(first_error) = message.ack().await {
-        let first_error = format_queue_ack_error(&first_error.0);
+async fn ack_message_with_retry(message: Delivery) -> Result<(), ConsumptionError> {
+    if let Err((first_error, message)) = message.ack().await {
+        let first_error = format_queue_ack_error(&first_error);
         warn!(error = %first_error, "Failed to ACK message, retrying once");
 
         sleep(QUEUE_ACK_RETRY_DELAY).await;
 
-        if let Err(second_error) = message.ack().await {
-            let second_error = format_queue_ack_error(&second_error.0);
+        if let Err((second_error, _message)) = message.ack().await {
+            let second_error = format_queue_ack_error(&second_error);
             error!(
                 first_error = %first_error,
                 error = %second_error,
@@ -207,9 +207,7 @@ impl EventWorker {
         let worker_handler =
             JobHandlerService::get_worker_handler_from_worker_trigger_type(worker_message.worker.clone());
         let workload = MetricsRecorder::start_worker_trigger_workload(&worker_message.worker);
-        worker_handler
-            .run_worker_if_enabled(self.config.clone())
-            .await?;
+        worker_handler.run_worker_if_enabled(self.config.clone()).await.map_err(OtherError::from)?;
         workload.finish_success();
         Ok(())
     }
@@ -313,10 +311,7 @@ impl EventWorker {
                     tracing::error!(worker = ?worker, error = %error_msg, "Failed to handle worker trigger");
                     (
                         format!("Worker {worker:?} handling failed: {error_msg}"),
-                        ConsumptionError::FailedToSpawnWorker {
-                            worker_trigger_type: worker.clone(),
-                            error_msg,
-                        },
+                        ConsumptionError::FailedToSpawnWorker { worker_trigger_type: worker.clone(), error_msg },
                     )
                 }
                 ParsedMessage::JobQueue(msg) => {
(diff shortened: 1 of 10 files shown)


orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -37,6 +40,32 @@ pub struct EventWorker {
 
 const QUEUE_GET_MESSAGE_WAIT_TIMEOUT_SECS: Duration = Duration::from_secs(30);
 const QUEUE_NO_MESSAGE_SLEEP_DURATION: Duration = Duration::from_millis(1000);
+const QUEUE_ACK_RETRY_DELAY: Duration = Duration::from_millis(500);
+
+fn format_error_context(error: &impl Error) -> String {
+    DisplayErrorContext(error).to_string()
+}
```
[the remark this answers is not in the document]
heemankv: Done. Consolidated the duplicate formatter into one generic `format_error_context` helper in `orchestrator/src/error/consumer.rs` and updated the SQS ACK/NACK call sites to use it.

orchestrator/src/worker/controller/event_worker.rs
```diff
@@ -406,7 +420,7 @@ impl EventWorker {
                         // ACK/NACK the priority delivery after processing
                         match &result {
                             Ok(_) => {
-                                if let Err(e) = delivery.ack().await {
+                                if let Err(e) = ack_message_with_retry(delivery).await {
```
prkpndy: can we double check that we are using ack/nack with retry everywhere sqs ack/nack is used?

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -250,13 +251,16 @@ impl PriorityQueueWorker {
             Ok(Some(job)) => job,
             Ok(None) => {
                 warn!("PQ Worker ({}): Job {} not found, ACKing message", self.action, parsed_msg.id);
-                delivery.ack().await.map_err(|e| ConsumptionError::FailedToAcknowledgeMessage(e.0.to_string()))?;
+                delivery
```
prkpndy: should we use ack_with_retry here as well?

orchestrator/src/worker/controller/priority_queue_worker.rs
```diff
@@ -250,13 +251,16 @@ impl PriorityQueueWorker {
...
+                    .ack()
+                    .await
+                    .map_err(|e| ConsumptionError::FailedToAcknowledgeMessage(format_error_context(&e.0)))?;
                 return Ok(());
             }
             Err(e) => {
                 error!("PQ Worker ({}): Database error fetching job {}: {:?}", self.action, parsed_msg.id, e);
                 // NACK to retry later
-                delivery.nack().await.map_err(|e| ConsumptionError::FailedToNackMessage(e.0.to_string()))?;
+                delivery.nack().await.map_err(|e| ConsumptionError::FailedToNackMessage(format_error_context(&e.0)))?;
```
prkpndy: same. nack with retry?

prkpndy (commented):
mostly looks good. let's just make sure that we are using retry for all acks/nacks. also, while you are at it, we can also add retry to aggregator batching s3 push

orchestrator/src/worker/controller/priority_queue_worker.rs
heemankv: Done. Priority queue ACK paths now use `ack_delivery_with_retry`, including malformed-message and job-not-found handling.

orchestrator/src/worker/controller/event_worker.rs
heemankv: Done. I scanned production delivery finalization and centralized all ACK/NACK calls through shared one-time retry helpers in `delivery_retry.rs`. The only remaining direct `ack()`/`nack()` calls are inside those helpers.

orchestrator/src/worker/controller/priority_queue_worker.rs
heemankv: Done. Priority queue NACK paths now use `nack_delivery_with_retry`, including stale-slot cleanup and DB-error retry handling.

prkpndy (approved):

PR #1240, opened 2026-07-21
fix(settlement): retry L1 state update sync on all errors

prkpndy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

A transient L1 RPC failure (timeout, connection reset, 429) that surfaces inside one of the core-contract calls in `get_current_core_contract_state` is wrapped in `EthereumClientError::Contract` (or `StarknetClientError::StateInitialization` on Starknet settlement). `is_recoverable()` classifies these as non-recoverable, so the state update worker returns `Err`, which propagates through `run_sync_worker` and the `ServiceMonitor` and **kills the entire node**. Whether the same network blip crashes the node or is retried depends purely on which RPC call it happens to land on: `get_latest_block_number` failures retry, contract-call failures crash.

## What is the new behavior?

- The state update worker retries **every** sync-session error with the existing capped exponential backoff (1s → 60s). This worker only observes L1 confirmations — an L1 outage degrades finality reporting but is never unsafe — so no error from it should take down the node. `is_recoverable()` now only selects the log level: `warn` for known-transient errors, `error` for unexpected ones, so genuine misconfiguration stays loudly visible without being fatal.
- The backoff resets to base once a session has been healthy for longer than the max delay (previously it escalated across the process lifetime and never reset).
- `EthereumClientError::Contract` and `StarknetClientError::StateInitialization` are reclassified as recoverable, since at runtime they wrap the underlying transport error of contract calls.
- Regression tests: the worker survives an error that is *not* classified recoverable (the exact shape that previously aborted the node), and the classification of the two reclassified variants is pinned.

## Scope notes

- **Gas prices are deliberately not included.** For gas prices we want the node to panic rather than keep producing blocks with stale prices, so the gas price worker's failure behavior is left untouched. (Nodes running with all prices fixed via `--l1-gas-price`/`--blob-gas-price`/`--strk-per-eth` don't run the gas poller at all.)
- Startup behavior is unchanged: client construction still verifies the core contract and fails boot if L1 is unreachable at start. Startup resilience can be addressed separately.
- L1→L2 messaging already has its own infinite reconnect loop and is unaffected.

### DIFF
--- a/madara/crates/client/settlement_client/src/eth/error.rs
+++ b/madara/crates/client/settlement_client/src/eth/error.rs
@@ -46,7 +46,9 @@ impl From<sol_types::Error> for EthereumClientError {
 impl EthereumClientError {
     /// Returns true if the error is recoverable (network/connection issues).
     /// These are transient errors that should be retried with backoff.
+    /// `Contract` is included because contract `.call()` failures wrap the
+    /// underlying transport error, so at runtime they are usually RPC outages.
     pub fn is_recoverable(&self) -> bool {
-        matches!(self, Self::Rpc(_) | Self::EventStream { .. } | Self::NetworkConnection { .. })
+        matches!(self, Self::Rpc(_) | Self::Contract(_) | Self::EventStream { .. } | Self::NetworkConnection { .. })
     }
 }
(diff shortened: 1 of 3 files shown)


heemankv (approved):
Great work on this PR. The retry behavior is scoped to the L1 state-update observer, the retry sleep is cancellation-aware, and gas-price failure behavior remains unchanged.\n\nI reviewed the changed settlement-client paths and did not find any blocking issues. CI is green and the current diff looks safe to merge from my side.

PR #1250, opened 2026-08-31
perf: make Cairo Native execution optimizations configurable

heemankv:
## Summary

- expose one opt-in `exec_hash_cache_enabled` Madara setting for deterministic Starknet API hash memoization and Cairo Native runtime Pedersen memoization
- make all four Starknet API cache capacities independently configurable and add a Cairo Native Pedersen capacity that applies per native execution thread
- reject zero capacities when the hash cache is enabled
- export low-overhead per-kind total-call, hit, miss, and capacity-clear counters through the existing OpenTelemetry pipeline
- add hash-cache activity plus hit- and miss-percentage panels to the Madara overview Grafana dashboard
- retain and document the existing bounded execution read cache, with the benchmarked 512 MiB setting in the sequencer config example
- document the existing sequential Blockifier configuration (`disable_concurrency: true`, `n_workers: 1`) in the chain config example

This branch remains one commit on top of `madara-alliance/madara@daaacf0055fbea1dca67b49aab5a74426efbf49c`.

## Configuration

`configs/sequencer_config.example.yaml` demonstrates:

```yaml
backend_params:
  exec_read_cache_enabled: true
  exec_read_cache_contracts: null
  exec_read_cache_max_memory_mib: 512

  exec_hash_cache_enabled: true
  exec_hash_cache_starknet_keccak_capacity: 8192
  exec_hash_cache_pedersen_pair_capacity: 8192
  exec_hash_cache_pedersen_array_capacity: 2048
  exec_hash_cache_poseidon_array_capacity: 2048
  exec_hash_cache_cairo_native_pedersen_capacity: 8192

cairo_native_params:
  enable_native_execution: true
```

All optimization switches remain disabled by default. The Cairo Native Pedersen capacity is per execution thread because its hot-path cache is thread-local. The read and hash caches memoize deterministic values and do not change execution semantics.

## Metrics and dashboard

Madara registers observable counters backed by relaxed atomics in Starknet API and Cairo Native. Metric collection reads the atomics; hash calls themselves do not invoke OpenTelemetry.

- `exec_hash_cache_calls_call_total`
- `exec_hash_cache_hits_hit_total`
- `exec_hash_cache_misses_miss_total`
- `exec_hash_cache_capacity_clears_clear_total`

Each series has a `kind` label: `starknet_keccak`, `pedersen_pair`, `pedersen_array`, `poseidon_array`, or `cairo_native_pedersen`. The Madara overview dashboard shows total activity plus hit and miss percentages per kind.

## Dependency patches

- Cairo Native: [karnotxyz/cairo_native#1](https://github.com/karnotxyz/cairo_native/pull/1), pinned at `76dec034`
- StarkWare Sequencer / Starknet API: [karnotxyz/sequencer#3](https://github.com/karnotxyz/sequencer/pull/3), pinned at `9a4336dc`

## Validation

- `cargo test pedersen_cache` in Cairo Native: 2 passed
- full Cairo Native Clippy with warnings denied: passed
- `cargo test -p starknet_api hash_cache::tests`: 2 passed
- `cargo check -p mc-exec --locked`: passed
- `cargo check -p madara --locked`: passed after supplying the normal generated Cairo artifact inputs locally
- `cargo metadata --locked --no-deps`: passed
- targeted Rust formatting and `git diff --check`: passed
- sequencer YAML, default JSON, and Grafana dashboard JSON parsing: passed

The focused Madara binary test target remains blocked by an unrelated upstream-main test-only compile error in `madara/node/src/submit_tx.rs`, where it calls the absent `ServiceContext::new_for_testing`. Production checks pass.

## Prior benchmark evidence

The same effective optimization set was validated on the 500-block PC DevNet workload (56,322 transactions): Cairo Native improved from `23.683` to `52.511` normal TPS (`2.217x`) with `56,322 / 56,322` transaction-status parity and zero mismatches. This PR ports that proven configuration to current upstream main; it does not claim a new benchmark run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -340,7 +340,30 @@ time = "0.3"
 zip = "4.3.0"
 
 [patch.crates-io]
+cairo-native = { git = "https://github.com/karnotxyz/cairo_native", rev = "76dec034da93e4f7ef93c5fda83ded540d77fc14" }
 jsonrpsee-core = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 jsonrpsee-types = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 rocksdb = { git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
 librocksdb-sys = { git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
+
+[patch."https://github.com/starkware-libs/sequencer"]
+apollo_compilation_utils = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_compile_to_native = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_compile_to_native_types = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_config = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_infra_utils = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_metrics = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_proc_macros = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_sizeof = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_sizeof_macros = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_starknet_os_program = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+apollo_time = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+blockifier = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+blockifier_test_utils = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+papyrus_common = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+shared_execution_objects = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
+starknet_api = { git = "https://github.com/karnotxyz/sequencer", rev = "9a4336dc1b36943fb77ddfc5becd2370e61d37e1" }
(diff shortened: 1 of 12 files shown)


madara/node/src/cli/backend.rs
```diff
@@ -209,11 +209,57 @@ pub struct BackendParams {
...
+    #[serde(default)]
+    pub exec_hash_cache_enabled: bool,
+
+    /// Maximum Starknet Keccak memoized values.
+    #[clap(
+        env = "MADARA_EXEC_HASH_CACHE_STARKNET_KECCAK_CAPACITY",
+        long,
+        default_value_t = starknet_api::DEFAULT_SN_KECCAK_CACHE_CAPACITY
+    )]
+    pub exec_hash_cache_starknet_keccak_capacity: usize,
```
Mohiiit, line 245: Could we add Serde defaults matching the Clap defaults to all five new capacity fields? Clap’s `default_value_t` is only applied during CLI parsing. When Madara starts with `--config-file`, `main` extracts the selected file directly through Figment, so a valid pre-PR config that omits these keys now fails deserialization even though the cache remains disabled by default.

I think this should also include a regression test that deserializes a config with all five hash-cache capacity keys omitted.

observability/grafana/dashboards/Madara/madara-overview.json
```diff
@@ -8466,6 +8466,179 @@
...
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "mappings": [],
+          "thresholds": {
```
Mohiiit, line 8581: Could we either show only the hit ratio in this stat or add a field override for the miss series? The panel applies the same threshold to both queries, so a miss ratio above 80% is rendered green even though that is the unhealthy outcome. Separate thresholds would keep the dashboard from presenting cache regressions as healthy.

Cargo.toml
```diff
@@ -340,7 +340,30 @@ time = "0.3"
 zip = "4.3.0"
 
 [patch.crates-io]
+cairo-native = { git = "https://github.com/karnotxyz/cairo_native", rev = "76dec034da93e4f7ef93c5fda83ded540d77fc14" }
 jsonrpsee-core = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 jsonrpsee-types = { git = "https://github.com/madara-alliance/jsonrpsee", rev = "52c76441b922b9d1c6b260fcb29d62e94a8ca57c" }
 rocksdb = { git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
 librocksdb-sys = { git = "https://github.com/madara-alliance/rust-rocksdb", branch = "read-options-set-raw-snapshot" }
+
+[patch."https://github.com/starkware-libs/sequencer"]
```
Mohiiit, line 349: Do we need to patch every Blockifier and Sequencer dependency here? Could this broad replacement create dependency or compatibility issues?

madara/crates/client/cairo_native/src/lib.rs
```diff
@@ -33,6 +33,10 @@ pub mod to_blockifier;
 mod test_utils;
 
 // Re-export commonly used types
+pub use cairo_native::{
```
Mohiiit, line 36: Could we re-export only the dependency API Madara currently uses? `set_pedersen_cache_enabled` and `PedersenCacheMetrics` have no callers in this workspace, while the configuration function, metrics snapshot, config type, and default capacity are used. Keeping the facade to those used items avoids creating public API that we may have to preserve later.

Mohiiit (changes requested):
I have a few questions around the configuration and observability surface.

madara/node/src/cli/backend.rs
heemankv: Done! Added Serde defaults matching all five Clap defaults and a Figment regression test that removes every capacity key from the existing config fixture and verifies the legacy config still deserializes with the expected values.

observability/grafana/dashboards/Madara/madara-overview.json
heemankv: Done! Removed the complementary miss series and renamed the panel to Exec Hash Cache Hit Ratio, so the healthy threshold now applies only to the value where higher is better.

Cargo.toml
heemankv: Done! Narrowed the override from the full Sequencer stack to starknet_api plus its four internal path-dependency crates required to keep one coherent source identity. Blockifier and all unrelated Sequencer packages now remain on the upstream tag; the locked production build passes.

madara/crates/client/cairo_native/src/lib.rs
heemankv: Done! Trimmed the Madara facade to the four APIs it actually uses: configuration, metrics snapshot, config type, and default capacity. The unused setter and metrics type re-exports are gone.

Mohiiit (approved):

PR #1251, opened 2026-08-31
feat(sync): add preconfirmed reorg protection

heemankv:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

`MADARA_DISABLE_REORG` only applies to confirmed-block parent-hash divergence. An already-synced preconfirmed block is still replaced when its header changes, its executed transaction count shrinks, or its transaction prefix diverges.
\n
## What is the new behavior?

- Adds the independent `MADARA_DISABLE_REORG_PRECONFIRMED` environment variable and `--disable-reorg-preconfirmed` CLI option.
- When enabled, preserves the local preconfirmed block if the upstream header changes or any previously-synced executed/candidate transaction is removed or replaced.
- Continues to accept forward-only updates, including candidate-to-executed progress and appended transactions.
- Defaults to `false`, leaving existing behavior unchanged unless explicitly enabled.

No.

## Other information

Validation:

- `cargo test -p mc-sync --lib` (34 passed, 1 ignored)
- `cargo test -p madara -F mp-utils/testing full_node_can_disable_preconfirmed_reorgs_independently`
- `cargo check -p madara --bin madara`
- `cargo fmt --all -- --check`

### DIFF
--- a/madara/crates/client/sync/src/tests/gateway_mock.rs
+++ b/madara/crates/client/sync/src/tests/gateway_mock.rs
@@ -198,6 +198,25 @@ impl GatewayMock {
 
     /// Ts is timestamp. We use that to differentiate pending blocks in the tests.
     pub fn mock_block_pending_with_ts(&self, block_number: u64, timestamp: usize) -> Mock<'_> {
+        self.mock_block_pending_with_ts_and_count(block_number, timestamp, 1)
+    }
+
+    pub fn mock_block_pending_with_ts_and_count(
+        &self,
+        block_number: u64,
+        timestamp: usize,
+        transaction_count: usize,
+    ) -> Mock<'_> {
+        self.mock_block_pending_with_ts_and_counts(block_number, timestamp, transaction_count, transaction_count)
+    }
+
+    pub fn mock_block_pending_with_ts_and_counts(
+        &self,
+        block_number: u64,
+        timestamp: usize,
+        transaction_count: usize,
+        executed_count: usize,
+    ) -> Mock<'_> {
         // block alpha-sepolia 171544
         self.mock_server.mock(|when, then| {
             when.method("GET")
@@ -225,8 +244,7 @@ impl GatewayMock {
                 },
                 "timestamp": timestamp,
                 "sequencer_address": "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8",
-                "transactions": [
-                    {
+                "transactions": (0..transaction_count).map(|_| json!({
                         "transaction_hash": "0x6a5a493cf33919e58aa4c75777bffdef97c0e39cac968896d7bee8cc67905a1",
                         "version": "0x1",
                         "max_fee": "0x0",
@@ -250,10 +268,8 @@ impl GatewayMock {
                             "0x1"
                         ],
                         "type": "INVOKE_FUNCTION"
-                    },
-                ],
-                "transaction_receipts": [
-                    {
+                    })).collect::<Vec<_>>(),
+                "transaction_receipts": (0..transaction_count).map(|index| if index < executed_count { json!({
                         "execution_status": "SUCCEEDED",
                         "transaction_index": 3,
                         "transaction_hash": "0x6a5a493cf33919e58aa4c75777bffdef97c0e39cac968896d7bee8cc67905a1",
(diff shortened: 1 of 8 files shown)


madara/crates/client/sync/src/tests/gateway_mock.rs
```diff
@@ -225,8 +244,7 @@ impl GatewayMock {
                 },
                 "timestamp": timestamp,
                 "sequencer_address": "0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8",
-                "transactions": [
-                    {
+                "transactions": (0..transaction_count).map(|_| json!({
                         "transaction_hash": "0x6a5a493cf33919e58aa4c75777bffdef97c0e39cac968896d7bee8cc67905a1",
                         "version": "0x1",
```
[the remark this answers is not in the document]
heemankv: Addressed in 7c22f7a7e. The fixture now emits a distinct hash for every transaction while preserving the original first hash used by the realistic test. The regression also sends a same-length block with a different hash sequence and verifies that the local executed/candidate prefix is retained.

madara/crates/client/sync/src/tests/pipeline.rs
```diff
@@ -176,6 +197,65 @@ async fn test_pending_block_update(mut ctx: TestContext) {
...
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::SyncingTo { target: 1 });
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::Idle);
+    assert_eq!(ctx.service_state_recv.recv().await.unwrap(), ServiceEvent::UpdatedPreconfirmedBlock);
+
+    assert_eq!(preconfirmed_state(&ctx.backend), (1000000000000, 1, 1));
+    drop(task);
+
+    pending_block_mock.delete();
+    let mut grown_block_mock = ctx.gateway_mock.mock_block_pending_with_ts_and_counts(2, 1000000000000, 3, 2);
+    assert!(poll_preconfirmed(&ctx.gateway_mock, &ctx.importer, &ctx.backend, true).await);
```
Mohiiit, line 229: Could we also cover forward growth and candidate replacement with `disable_reorg_preconfirmed = false`? The executed-prefix comparison changes even when protection is disabled, so this would verify that the default path still accepts these updates. Asserting the resulting executed and candidate hashes would also check that appending retains the correct prefix without duplicating transactions.

madara/node/src/cli/l2.rs
```diff
@@ -89,6 +89,12 @@ pub struct L2SyncParams {
     /// operators who want to manually handle chain divergences.
     #[clap(env = "MADARA_DISABLE_REORG", long, default_value_t = false)]
     pub disable_reorg: bool,
+
+    /// Disable replacement of an already-synced preconfirmed block when the upstream block shrinks,
+    /// changes header, or changes its existing transaction prefix. Forward-only updates remain enabled.
```
Mohiiit, line 94: Could we clarify in the flag documentation that confirmed-block import still takes precedence over the preserved preconfirmed view? A regression that rejects a divergent preconfirmed update, then accepts its confirmed block and advances to the next height, would make this boundary explicit and verify that protection cannot stall normal confirmation.

madara/crates/client/sync/src/tests/gateway_mock.rs
```diff
@@ -198,6 +198,64 @@ impl GatewayMock {
...
+    ) -> Mock<'_> {
+        self.mock_block_pending_with_hash_offsets(
+            block_number,
+            timestamp,
+            executed_count,
+            vec![hash_offset; transaction_count],
+        )
+    }
+
+    pub fn mock_block_pending_with_hash_offsets(
```
Mohiiit, line 239: Could we simplify these fixture variants around one helper taking explicit transaction hashes and an executed count, while retaining the existing convenience wrapper? That would make replacement and reordering scenarios readable directly from the supplied hashes, without tracing the wrapper chain or the offset-to-hash calculation.

Mohiiit (commented):
A few non-blocking suggestions to strengthen coverage of the default behavior and confirmation boundary, plus a small fixture simplification.

madara/crates/client/sync/src/tests/pipeline.rs
heemankv: Addressed in #1254: added default-off forward-growth and candidate-only replacement coverage with exact executed/candidate hash assertions, including prefix retention without duplication.

madara/node/src/cli/l2.rs
heemankv: Addressed in #1254: clarified that confirmed imports take precedence, and added a regression that rejects a divergent preconfirmed update, imports the confirmed block at that height, and advances to the next preconfirmed height.

madara/crates/client/sync/src/tests/gateway_mock.rs
heemankv: Addressed in #1254: replaced the offset-based fixture variants with one helper that takes explicit transaction hashes and an executed count, while retaining the count-based convenience wrappers.

PR #1252, opened 2026-09-07
feat(gateway): add negotiated gzip feeder responses

heemankv:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

Feeder-gateway responses are always sent as plain JSON. Madara fullnodes do not advertise gzip support and deserialize response bytes directly, so enabling response compression on the sequencer alone would break existing Madara clients.

Resolves: #NA

## What is the new behavior?

- Madara feeder GET requests advertise `Accept-Encoding: gzip` and centrally decode gzip before parsing successful or structured-error JSON responses.
- The feeder server negotiates gzip for successful, non-empty `feeder_gateway/*` responses when the caller accepts it. Compression remains disabled by default and can be enabled with `--feeder-gateway-gzip-responses` or `MADARA_FEEDER_GATEWAY_GZIP_RESPONSES=true`.
- Compression uses the fast gzip level on blocking workers with four process-wide slots. Saturation and compression failures fall back to identity responses.
- The client independently caps received and decompressed response bodies at 64 MiB, returns a typed overflow error, and applies the existing bounded retry policy to transient decompression failures.
- Gateway telemetry reports bounded route names, selected encoding, logical bytes, transmitted bytes, compression duration, status, and total response time.
- Two cumulative OpenTelemetry counters expose hypothetical uncompressed response-body bytes and actual transmitted response-body bytes. Both have bounded `route` and `encoding` labels and are directly graphable in Grafana/Prometheus.
- Tests cover negotiation, headers, exact decompression, structured errors, invalid/truncated and oversized streams, retry behavior, concurrency saturation, failure fallback, and enabled/disabled two-node sync.

No. The server switch defaults to false, clients continue accepting identity responses, and callers that do not request gzip always receive the existing plain JSON representation.

### Grafana metrics

- `feeder_gateway_response_body_uncompressed_bytes_total`: bytes that would have been sent without gzip.
- `feeder_gateway_response_body_transmitted_bytes_total`: bytes actually sent after optional gzip.

The metrics count response bodies and exclude HTTP/TLS overhead. Compare their byte rates with:

```promql
sum(rate(feeder_gateway_response_body_uncompressed_bytes_total{namespace=~"$namespace", pod=~"$pod"}[$__rate_interval]))
sum(rate(feeder_gateway_response_body_transmitted_bytes_total{namespace=~"$namespace", pod=~"$pod"}[$__rate_interval]))
```

Percentage saved over the selected interval:

```promql
100 * (
  1 -
  sum(increase(feeder_gateway_response_body_transmitted_bytes_total{namespace=~"$namespace", pod=~"$pod"}[$__range]))
  /
  sum(increase(feeder_gateway_response_body_uncompressed_bytes_total{namespace=~"$namespace", pod=~"$pod"}[$__range]))
)
```

### Validation

- `cargo test -p mc-gateway-client -p mc-gateway-server`: client 24 passed / 2 pre-existing ignored; server 23 passed.
- Focused `mc-e2e-tests` negotiated feeder sync: 2 passed (server gzip disabled and enabled).
- `cargo clippy -p mc-gateway-client -p mc-gateway-server --tests -- -D warnings`: passed.
- Metrics follow-up: `cargo test -p mc-gateway-server` (23 passed) and `cargo clippy -p mc-gateway-server --tests -- -D warnings` passed.
- `cargo build --release -p madara`: passed.
- `cargo fmt --all -- --check`, `taplo fmt --check`, and `git diff --check`: passed.
- A new Madara fullnode synced against a preserved pre-change feeder binary.
- `eqlabs/pathfinder:v0.22.2` synced through the gzip-enabled feeder with sampled lag zero.

### Local Devnet benchmark

| Response | Identity | Gzip | Reduction |
| --- | ---: | ---: | ---: |
| `get_block` | 56,261 B | 2,223 B | 96.0% |
| `get_preconfirmed_block` | 56,433 B | 2,190 B | 96.1% |
| `get_state_update` | 1,726 B | 678 B | 60.7% |

A concurrency-four comparison over 3,000 block requests transmitted 168,783,000 bytes with identity and 6,669,000 bytes with gzip, with zero HTTP errors. Mean latency changed from 0.750 ms to 0.839 ms and average sequencer CPU from roughly 12.8% to 14.8% of one core. Block production continued on its two-second cadence and Pathfinder finished at lag zero.


madara/crates/client/gateway/client/src/request_builder.rs
```diff
@@ -175,19 +182,167 @@ where
...
+        .flat_map(|value| value.split(','))
+        .any(|encoding| encoding.trim().eq_ignore_ascii_case("gzip"));
+
+    if !gzip_encoded {
+        return Ok(body);
+    }
+
+    let mut decoder = GzDecoder::new(body.as_ref());
+    let mut decoded = Vec::new();
+    decoder.read_to_end(&mut decoded).map_err(|source| SequencerError::DecompressResponse { source })?;
```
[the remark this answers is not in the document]
heemankv: Done in 60bd5648c. Raw response collection and gzip expansion are now independently capped at 64 MiB. The transport path uses http_body_util::Limited, decompression reads at most limit plus one sentinel byte, and either overflow returns the typed ResponseBodyTooLarge error. A regression test covers both limits.

madara/crates/client/gateway/client/src/methods.rs
```diff
@@ -43,6 +43,7 @@ impl GatewayProvider {
         for attempt in 0..MAX_RETRIES {
             match request_fn().await {
                 Ok(result) => return Ok(result),
+                Err(error @ SequencerError::DecompressResponse { .. }) => return Err(error),
```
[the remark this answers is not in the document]
heemankv: Done in 60bd5648c. I removed the decompression-error fast exit, so transient gzip failures now use the existing bounded five-attempt exponential-backoff policy. The new regression test fails decompression once and verifies recovery on the next attempt.

madara/crates/client/gateway/server/src/service.rs
```diff
@@ -120,3 +161,321 @@ pub async fn start_server(
...
+    gzip_compression_semaphore: Arc<Semaphore>,
+) -> (Response<Full<Bytes>>, ResponseStats) {
+    let uncompressed_bytes = response.body().len() as u64;
+    let should_compress = gzip_enabled
+        && path.starts_with("feeder_gateway/")
+        && response.status().is_success()
+        && uncompressed_bytes > 0
+        && accepts_gzip(request_headers);
+
+    if !should_compress {
```
Mohiiit, line 213: Could we add `Vary: Accept-Encoding` to eligible feeder responses before choosing gzip or identity, including the saturation and compression-failure fallbacks? Currently it is added only after successful compression. A caching proxy can therefore reuse an identity response for subsequent gzip requests, bypassing compression until that entry expires.

madara/crates/client/gateway/client/src/request_builder.rs
```diff
@@ -175,19 +184,207 @@ where
...
+        .filter_map(|value| value.to_str().ok())
+        .flat_map(|value| value.split(','))
+        .any(|encoding| encoding.trim().eq_ignore_ascii_case("gzip"));
+
+    if !gzip_encoded {
+        return Ok(body);
+    }
 
-    Ok(res)
+    let mut decoder = GzDecoder::new(body.as_ref()).take(max_bytes.saturating_add(1) as u64);
```
Mohiiit: Could we use `MultiGzDecoder` here while keeping the existing size limit? `GzDecoder` stops after the first member, so a valid gzip response containing JSON split across multiple members produces incomplete JSON. A small concatenated-member regression test would cover this. The older request decoder uses the same pattern, so this could also be aligned in a follow-up.

madara/crates/client/gateway/server/src/service.rs
```diff
@@ -120,3 +161,321 @@ pub async fn start_server(
...
+        return identity_response(response, 0);
+    }
+
+    let Ok(permit) = gzip_compression_semaphore.try_acquire_owned() else {
+        tracing::warn!(target: "gateway_calls", method = telemetry_route(path), "Feeder gzip concurrency limit reached; returning identity");
+        return identity_response(response, 0);
+    };
+
+    let (parts, body) = response.into_parts();
+    let body = Arc::new(body);
```
Mohiiit: Could we convert the body to `Bytes` here and clone it for the blocking task? That would keep the original available for fallback without `Arc<String>` or the `Arc::try_unwrap`/copy branch. `finish_compression` could then select the body and encoding first and construct the response and statistics once.

Mohiiit (commented):
A few suggestions around HTTP response handling, simplifying body ownership, and making the integration test verify compression explicitly. The multi-member gzip point is a compatibility follow-up; I haven’t observed it with the feeders tested here.

madara/crates/tests/src/transaction_flow.rs
```diff
@@ -304,6 +319,25 @@ fn make_transfer_call(recipient: Felt, amount: u128) -> Vec<Call> {
...
+        .with_feeder_gateway_gzip_responses(gzip_enabled)
+        .run()
+        .await;
+    let RunningTestSetup::TwoNodes { _sequencer: sequencer, mut user_facing } = setup else {
+        unreachable!("full-node setup always starts two nodes")
+    };
+
+    wait_for_next_block(&sequencer.json_rpc()).await;
+    let target = get_latest_block_n(&sequencer.json_rpc()).await;
+    user_facing.wait_for_sync_to(target).await;
```
Mohiiit, line 338: Could the enabled case also request a response from the running feeder with `Accept-Encoding: gzip`, assert `Content-Encoding: gzip`, and decode the body? Currently both cases only check that synchronization succeeds, so they would still pass if the compression flag stopped reaching the server. The disabled case could assert that the same request receives identity.

madara/crates/client/gateway/server/src/service.rs
heemankv: Done in 53fb083eb. Eligible successful, non-empty feeder responses now receive `Vary: Accept-Encoding` before negotiation, so identity negotiation, semaphore saturation, and compression-failure fallbacks all preserve it. The unit cases now assert the header on each fallback path.

madara/crates/client/gateway/client/src/request_builder.rs
heemankv: Done in 53fb083eb. Response decoding now uses `MultiGzDecoder` while retaining the existing compressed and decompressed 64 MiB limits. I added a regression test with one JSON document split across two concatenated gzip members.

madara/crates/client/gateway/server/src/service.rs
heemankv: Done in 53fb083eb. The response body is converted to `Bytes` and cloned for the blocking compression task. `finish_compression` selects the body and encoding first, then constructs the response and statistics once; the fallback reuses the original `Bytes` without `Arc::try_unwrap` or copying.

madara/crates/tests/src/transaction_flow.rs
heemankv: Done in 53fb083eb. Both integration cases now call the running feeder with `Accept-Encoding: gzip`. The enabled case asserts `Content-Encoding: gzip`, manually decodes the wire body, and parses the returned block; the disabled case asserts identity and parses the plain response.

Mohiiit (approved):

